PR #4, opened 2023-01-13
SceneQueryRunner: Pass scene object via scoped vars to enable query interpolation

dprokop:
This PR makes sure the SceneQueryRunner scene object is passed down to `runRequest` function via scoped vars for `template_srv` to interpolate queries when interpolation is handled in the data source implementation.

This can be observed i.e. :
- in http://localhost:3000/scenes/Variables?var-server=AA&var-pod=AAA&var-handler=AAAA&from=now-6h&to=now - legend is now correctly shown.
- in any Prometheus query that uses template vars.

### DIFF
--- a/src/querying/SceneQueryRunner.ts
+++ b/src/querying/SceneQueryRunner.ts
@@ -9,6 +9,7 @@ import {
   DataTransformerConfig,
   PanelData,
   rangeUtil,
+  ScopedVar,
   TimeRange,
   transformDataFrame,
 } from '@grafana/data';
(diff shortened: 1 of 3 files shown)


src/querying/SceneQueryRunner.ts
```diff
@@ -134,14 +137,14 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> {
...
+      scopedVars: sceneObjectScopedVar,
       startTime: Date.now(),
     };
 
     try {
       const ds = await getDataSource(datasource, {
         ...request.scopedVars,
-        __sceneObject: { text: '__sceneObject', value: this },
+        ...sceneObjectScopedVar,
       });
```
torkelo, line 148: think we can just do getDataSource(datasource, request.scopedVars) to reduce unnessary object creation

torkelo (approved):
added small simplification idea

src/querying/SceneQueryRunner.ts
dprokop: Good catch, that's totally unnecessary what I did here! Thanks

PR #6, opened 2023-01-16
SceneVariableSet: Fixes issue with deactivation cleanup

torkelo:
I somehow thought calling unsubscribe was enough but of course not we also need to clear these other sets in deactivate


dprokop (approved):
Tested this, and indeed that's it

PR #7, opened 2023-01-16
Scenes: Clean and simplify EmbeddedScene 

torkelo:
* Rename SceneState to EmbeddedSceneState 
* Simplify EmbeddedScene (remove title)
* Change name of subMenu to controls, make it a simple array of SceneObjects.
* Remove SceneSubMenu component (handled by EmbeddedScene)
* Add SceneControlsSpacer

OSS PR https://github.com/grafana/grafana/pull/61579


dprokop (approved):

src/components/SceneControlsSpacer.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+import { SceneObjectBase } from '../core/SceneObjectBase';
+import { SceneComponentProps } from '../core/types';
+
+export class SceneControlsSpacer extends SceneObjectBase<{}> {
```
dprokop, line 6: Can't we just use SceneFlexLayout as an object passed to the controls array, rather than introducing this spacer?

ivanortegaalba, line 6: Why don't we use `display: flex;` for the container and for the items that want to go to the right uses `align-self: right`?

ivanortegaalba (approved):

PR #11, opened 2023-01-17
Demo app

torkelo:
so we have a two ways to add a demo app to this repo

### Standalone folder (with separate yarn.lock) (this  PR)

Pros
* Skip the complexity of hoisting, lerna and everything related to it 

Cons
* To build, develop against the local version of scenes requires yarn linking
* Need to manually pump the version of example app 

### Workspaces: https://github.com/grafana/scenes/pull/10

* Uses the locally built version of scenes package
* Deal with complexity of lerna, and hosting


dprokop:
Closing in favor of #52

PR #16, opened 2023-01-27
SceneQueryRunner: Run queries when variables changed before re-activation

dprokop:
When working on a demo app, I discovered an issue with queries not being re-run if the variable value has changed before the activation (or re-activation) of the `SceneQueryRunner`.

This can happen i.e. in drill-down views, when variable values are initialized from URL before the SceneQueryRunner is activated. When URL sync happens this way - and it should at least for now ref https://github.com/grafana/scenes/pull/15 - the `SceneVariableValueChangedEvent` listeners are not yet activated, hence the changes are not propagated to `SceneQueryRunner`.

### DIFF
--- a/src/querying/SceneQueryRunner.test.ts
+++ b/src/querying/SceneQueryRunner.test.ts
@@ -14,21 +14,24 @@ import {
 import { SceneTimeRange } from '../core/SceneTimeRange';
 
 import { SceneQueryRunner } from './SceneQueryRunner';
+import { EmbeddedScene } from '../components/EmbeddedScene';
+import { SceneVariableSet } from '../variables/sets/SceneVariableSet';
+import { SceneFlexLayout } from '../components/layout/SceneFlexLayout';
+import { TestVariable } from '../variables/variants/TestVariable';
 
 const getDataSourceMock = jest.fn().mockReturnValue({
   getRef: () => ({ uid: 'test' }),
 });
 
+const result = toDataFrame([
+  [100, 1],
+  [200, 2],
+  [300, 3],
+]);
 const runRequestMock = jest.fn().mockReturnValue(
   of<PanelData>({
     state: LoadingState.Done,
-    series: [
-      toDataFrame([
-        [100, 1],
-        [200, 2],
-        [300, 3],
-      ]),
-    ],
+    series: [result],
     timeRange: getDefaultTimeRange(),
   })
 );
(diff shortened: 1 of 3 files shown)


src/querying/SceneQueryRunner.test.ts
```diff
@@ -88,6 +91,46 @@ describe('SceneQueryRunner', () => {
     });
   });
 
+  describe('when re-activated', () => {
+    describe('variables have changed', () => {
+      it('should run queries', async () => {
```
dprokop, line 96: More like an integration test rather than SceneQueryRunner test, but can't see another way to structure this.

src/querying/SceneQueryRunner.ts
```diff
@@ -69,8 +72,15 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> {
...
 
   private shouldRunQueriesOnActivate() {
-    // If we already have data, no need
+    // Check if variables have changed since last run.
+    // This is relevant in case variable values are applied via URL sync that happened before activation.
+    // In such scenerio no objects are notified about variables change, hence queries would not be run.
     // TODO validate that time range is similar and if not we should run queries again
+    if (!areMapsEqual(this._lastRunVariables, this.getCurrentVariables())) {
+      return true;
+    }
```
torkelo, line 81: This feels a bit wrong. As this is the responsibility of VariableSet, and it should be VariableSewt that somehow figures out that variables have changed before activation and broadcast that change (via _variableDependency) on activate.

src/querying/SceneQueryRunner.ts
```diff
@@ -122,7 +135,15 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> {
   }
 
   private async runWithTimeRange(timeRange: TimeRange) {
+    // Persist values of variables used for last run
+    const variables = sceneGraph.getVariables(this);
+    this._variableDependency.getNames().forEach((name) => {
+      const value = variables.getByName(name)?.getValue();
+      this._lastRunVariables.set(name, value);
+    });
```
torkelo, line 143: For this to work with nested VariableSets it needs to be a recursive lookup, like
https://github.com/grafana/scenes/blob/main/src/variables/interpolation/sceneInterpolator.ts#L55

src/querying/SceneQueryRunner.ts
torkelo: This problem would also be manifest Variables themselves. 

say you have Variable A, and Variable B which depends on A. If we visit the scene again but with URL state that changes A before activation, the B would not get notified of this and re-evaluted

src/querying/SceneQueryRunner.test.ts
```diff
@@ -88,6 +91,46 @@ describe('SceneQueryRunner', () => {
...
+        });
+
+        expect(queryRunner.state.data).toBeUndefined();
+
+        scene.activate();
+        await new Promise((r) => setTimeout(r, 1));
+
+        expect(queryRunner.state.data?.state).toBe(LoadingState.Done);
+        expect(queryRunner.state.data?.series).toEqual([result]);
+        expect(runRequestMock).toHaveBeenCalledTimes(1);
```
torkelo, line 121: this does not actually wait for the variables to complete (By default TestVariable does not complete unless signalUpdateCompleted is called or a delayMs is set)

src/querying/SceneQueryRunner.test.ts
```diff
@@ -88,6 +91,46 @@ describe('SceneQueryRunner', () => {
...
+
+        scene.deactivate();
+        v1.setState({ value: 'a1' });
+        v2.setState({ value: 'b1' });
+        scene.activate();
+        await new Promise((r) => setTimeout(r, 1));
+
+        expect(runRequestMock).toHaveBeenCalledTimes(2);
```
torkelo, line 129: Here again the variables are update process is started but never completed

dprokop:
closing in favor of #19

PR #19, opened 2023-01-31
SceneVariableSet: Improved logic for when to validateAndUpdate variables and notify scene of changes

torkelo:
This tries to solve two major things

* Figure out when we do not need to call validateAndUpdate again after re-activation (when the current value & options are valid)
* Notify the scene of changes when the variable was changed while in-active (Both for other variables, and other scene objects)

### DIFF
--- a/src/variables/sets/SceneVariableSet.ts
+++ b/src/variables/sets/SceneVariableSet.ts
@@ -1,19 +1,28 @@
+import { isEqual } from 'lodash';
 import { Unsubscribable } from 'rxjs';
 
 import { SceneObjectBase } from '../../core/SceneObjectBase';
 import { SceneObject } from '../../core/types';
 import { forEachSceneObjectInState } from '../../core/utils';
-import { SceneVariable, SceneVariables, SceneVariableSetState, SceneVariableValueChangedEvent } from '../types';
+import {
+  SceneVariable,
+  SceneVariables,
+  SceneVariableSetState,
+  SceneVariableValueChangedEvent,
+  VariableValue,
+} from '../types';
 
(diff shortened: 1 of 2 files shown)


src/variables/sets/SceneVariableSet.ts
```diff
@@ -28,8 +37,57 @@ export class SceneVariableSet extends SceneObjectBase<SceneVariableSetState> imp
...
+    );
+
+    this.checkForVariablesThatChangedWhileInactive();
+
+    // Add all variables that need updating to queue
+    for (const variable of this.state.variables) {
+      if (this.variableNeedsUpdate(variable)) {
+        this._variablesToUpdate.add(variable);
+      }
+    }
```
dprokop, line 51: @torkelo I don't think I understand why this loop is necessary. Since we are queuing updates via `checkForVariablesThatChangedWhileInactive`, shouldn't this be sufficient? IIUC it covers pretty much the first activation?

torkelo: @dprokop you can try commenting this out and see the tests that fail. 

This loop and the checks inside variableNeedsUpdate will make it so that 
* We do not add variables that are valid to the variablesToUpdate queue

The check inside checkForVariablesThatChangedWhileInactive checks for CHANGED variables, the check inside variableNeedsUpdate checks for variables that have the same value now as when it de-activated (And we only add valid values to _validValuesWhenDeactivated)

dprokop: OK, this makes sens!

dprokop (approved):

PR #21, opened 2023-02-01
SceneQueryRunner: Wait for variables to complete before issuing query

torkelo:
* [x] Wait with executing queries until VariableSet has activated or finished processing the variable 
 * [x] Detect if variables changed while in-active 
 * [x] Slight change to VariableDependencyLike config so that it now has a function that is always called when the VariableSet has finished processing (even if no variables changed). This is for QueryRunner to be notified when SceneVariableSet is done as it now waits for it. And if it does not get this call, it would not issue queries (when no variable value changed after SceneVariableSet was done processing)

For future PRs to consider
* Need to improve the logic when we need to revalidate a variable when it's value change while in-activation, we always do it now even when it's not need (we could know the new value is valid) 
* Possibly add a new state to LoadingState that is "WaitingForDependencies"  or something like that as all the panels await variables just show "no data" now with no indication of why they are not loading


src/components/SceneCanvasText.tsx
```diff
@@ -32,7 +32,7 @@ export class SceneCanvasText extends SceneObjectBase<SceneCanvasTextState> {
 
     return (
       <div style={style} data-testid={key}>
-        {sceneGraph.interpolate(model, text)}
+        {sceneGraph.interpolate(model, text)} A
```
dprokop, line 35:
```suggestion
        {sceneGraph.interpolate(model, text)}
```

src/core/SceneObjectBase.tsx
```diff
@@ -141,13 +141,13 @@ export abstract class SceneObjectBase<TState extends SceneObjectState = SceneObj
...
-      $data.activate();
-    }
-
     if ($variables && !$variables.isActive) {
       $variables.activate();
     }
+
+    if ($data && !$data.isActive) {
+      $data.activate();
+    }
```
dprokop, line 150: What's the impact of this change? Think it won't have any effect if i.e. variables are top level, and queries are attached to VizPanel?

dprokop: Thinking out loud, maybe we should activate all variables up to the root to make sure that they are all kicked-off before data is activated?

src/querying/SceneQueryRunner.test.ts
```diff
@@ -215,4 +218,120 @@ describe('SceneQueryRunner', () => {
...
+      });
+
+      scene.activate();
+      // should execute query when variable completes update
+      variable.signalUpdateCompleted();
+      await new Promise((r) => setTimeout(r, 1));
+      expect(queryRunner.state.data?.state).toBe(LoadingState.Done);
+    });
+
+    it('Should execute query wven when variable completes but value unchanged', async () => {
```
dprokop, line 315:
```suggestion
    it('Should execute query even when variable completes but value unchanged', async () => {
```

src/variables/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  for (const name of sceneObject.variableDependency.getNames()) {
+    const variable = lookupSceneVariable(name, sceneObject);
+    if (!variable) {
+      continue;
+    }
+
+    const set = variable.parent as SceneVariableSet;
+    return set.isVariableLoadingOrWaitingToUpdate(variable);
```
dprokop, line 40: Think this can prematurely return false. I.e. with A & B variables, when A has loaded, and B is still loading. Think we need to accumulate results for each variable and return result. Or return early if first loading variable is encountered.

torkelo: ah good catch, this needs to only return if this is true

src/querying/SceneQueryRunner.ts
```diff
@@ -45,10 +48,12 @@ export interface DataQueryExtended extends DataQuery {
...
   private _querySub?: Unsubscribable;
   private _containerWidth?: number;
+  private _firstQueryStarted = false;
+  private _variableChangeDetector = new VariableValueChangeDetector(this);
 
-  protected _variableDependency = new VariableDependencyConfig(this, {
+  protected _variableDependency: VariableDependencyConfig<QueryRunnerState> = new VariableDependencyConfig(this, {
     statePaths: ['queries', 'datasource'],
-    onReferencedVariableValueChanged: () => this.runQueries(),
+    onVariableUpdatesCompleted: (variables) => this.onVariableUpdatesCompleted(variables),
```
dprokop, line 56: Oh yes, this makes a lot of sense to wait for all the variables to complete!

src/variables/VariableValueChangeDetector.ts
```diff
@@ -0,0 +1,42 @@
+import { sceneGraph } from '../core/sceneGraph';
+import { SceneObject } from '../core/types';
+import { SceneVariable, VariableValue } from './types';
+import { isVariableValueEqual } from './utils';
+
+/**
+ * Useful for rememmber variable values
+ **/
+export class VariableValueChangeDetector {
```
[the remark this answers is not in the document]
torkelo: I think it could but I also think this problem is going to be limited to only components that want to do state logic when variables changed (so query runners basically). If we see this being used a lot we could build it into SceneObjectBase but not sure it's warranted yet

src/variables/sets/SceneVariableSet.ts
```diff
@@ -104,7 +105,7 @@ export class SceneVariableSet extends SceneObjectBase<SceneVariableSetState> imp
     this._validValuesWhenDeactivated = new Map();
     for (const variable of this.state.variables) {
       // if the current variable is not in queue to update and validate and not being actively updated then the value is ok
-      if (!this._variablesToUpdate.has(variable) || !this._updating.has(variable)) {
+      if (!this._variablesToUpdate.has(variable) && !this._updating.has(variable)) {
         this._validValuesWhenDeactivated.set(variable, variable.getValue());
```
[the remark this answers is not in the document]
torkelo: I am not using VariableValueRecorder in SceneVariableSet

src/variables/VariableValueChangeDetector.ts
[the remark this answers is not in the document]
dprokop: Yup, let's hold on with that

dprokop (approved):
Massive update!

PR #25, opened 2023-02-03
SceneApp: Components for building scene pages in app plugins

dprokop:
Based of https://github.com/grafana/grafana/pull/61746/files#diff-6fa0d2e370daf913de925cfebc01da033392fc381d64012a01fab9b046e7a088

TODO:
- [ ] Docs
- [ ] Tests

### DIFF
--- a/src/components/SceneApp/SceneApp.tsx
+++ b/src/components/SceneApp/SceneApp.tsx
@@ -12,6 +12,7 @@ import { EmbeddedScene } from '../EmbeddedScene';
 const sceneCache = new Map<string, EmbeddedScene>();
 
 export interface SceneAppState extends SceneObjectStatePlain {
+  // Array of SceneAppPage objects that are considered app's top level pages
   pages: SceneAppPage[];
 }
 
@@ -55,13 +56,21 @@ export class SceneApp extends SceneObjectBase<SceneAppState> {
 export interface SceneAppPageState extends SceneObjectStatePlain {
   title: string;
   subTitle?: string;
+  // Use to provide page absolute URL, i.e. /app/overview
   url: string;
+  // Use to provide parametrized page URL, i.e. /app/overview/:clusterId
   routePath?: string;
+  // Whether or not page should be visible in the breadcrumbs path
   hideFromBreadcrumbs?: boolean;
+  // Array of SceneAppPage objects that are used as page tabs displayed on top of the page
   tabs?: SceneAppPage[];
+  // Function that returns a scene object for the page
   getScene: (routeMatch: SceneRouteMatch) => EmbeddedScene;
+  // Array of scenes used for drilldown views
   drilldowns?: SceneAppDrilldownView[];
+  // Function that returns a parent page object, used to create breadcrumbs structure
   getParentPage?: () => SceneAppPage;
+  // Array of query params that will be preserved in breadcrumb and page tab links, i.e. ['from', 'to', 'var-datacenter',...]
   preserveUrlKeys?: string[];
 }
 
@@ -146,8 +155,6 @@ function SceneAppPageRenderer({ model }: SceneComponentProps<SceneAppPage>) {
       {routes}
     </Switch>
   );
-
-  // return ;
 }
 
 interface ScenePageRenderProps {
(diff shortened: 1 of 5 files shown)


src/components/SceneApp/SceneApp.tsx
```diff
@@ -0,0 +1,191 @@
...
+}
+
+export interface SceneAppPageState extends SceneObjectStatePlain {
+  title: string;
+  url: string;
+  routePath?: string;
+  subTitle?: string;
+  hideFromBreadcrumbs?: boolean;
+  tabs?: SceneAppPage[];
+  getScene: (routeMatch: SceneRouteMatch) => EmbeddedScene;
```
dprokop, line 52: @torkelo - I've changed this type to return sth more specific than SceneObject - primarily to expose the API for URLSyncManange init.

src/components/SceneApp/SceneApp.tsx
```diff
@@ -0,0 +1,191 @@
...
+    const routeMatch = useRouteMatch();
+    console.log('routeMatch path', routeMatch.url);
+
+    let scene = sceneCache.get(routeMatch.url);
+    if (!scene) {
+      scene = model.state.getScene(routeMatch);
+      sceneCache.set(routeMatch.url, scene);
+    }
+
+    scene.initUrlSync();
```
dprokop, line 109: Here's the init required prior scene initialisation, ref https://github.com/grafana/scenes/pull/15

src/components/SceneApp/SceneApp.tsx
```diff
@@ -0,0 +1,191 @@
...
+    if (drilldowns) {
+      for (const drilldown of drilldowns) {
+        console.log('registering non-tab drilldown route', drilldown.routePath);
+        routes.push(
+          <Route key={drilldown.routePath} exact={false} path={drilldown.routePath}>
+            <SceneAppDrilldownViewRender drilldown={drilldown} parent={model} />
+          </Route>
+        );
+      }
+    }
```
dprokop, line 98: Added this for scenarios where drilldown is not part of tabbed page.

src/components/SceneApp/SceneApp.tsx
torkelo: makes sense

src/components/SceneApp/SceneApp.tsx
torkelo: not sure this is safe / good place, this is in the render path which can happen multiple times, and it causes state changes which are not allowed during react render. 

We might need to move sceneCache.get, model.state.getScene, and scene.initUrlSync to activate on SceneAppPage (and set scene obj in SceneAppState). And only when we have it in state we render it. Might have to clear it from state on deactivate so we don't render it again on re-active before SceneAppPage activate has a change to call scene.initUrlSync.

src/components/SceneApp/SceneApp.tsx
torkelo: :+1:

src/components/SceneApp/SceneApp.tsx
dprokop: Trickier than expected, will continue next week on this.

src/components/SceneApp/SceneApp.tsx
```diff
@@ -0,0 +1,240 @@
...
+   * We use this flag to make sure the URL sync is enabled before the scene is actually rendered.
+   */
+  const [isInitialized, setIsInitialized] = useState(false);
+  const params = useAppQueryParams();
+  const routeMatch = useRouteMatch();
+
+  let scene = sceneCache.get(routeMatch!.url);
+  if (!scene) {
+    // If we are rendering a tab, we need to get the scene f  rom the tab, otherwise, use page's scene
+    scene = activeTab ? activeTab.state.getScene(routeMatch) : page.state.getScene(routeMatch);
```
dprokop, line 173: To correctly render active tabs, i.e. when they are part of the drill down.

src/components/SceneApp/SceneApp.tsx
```diff
@@ -0,0 +1,240 @@
...
+    // Before rendering scene components, we are making sure the URL sync is enabled for.
+    if (!isInitialized && scene) {
+      scene.initUrlSync();
+      setIsInitialized(true);
+    }
+  }, [isInitialized, scene]);
+
+  if (!isInitialized) {
+    return null;
+  }
```
dprokop, line 187: Scene is now read from cache as deep as the scene page is actually rendered. I don't see a need to get it earlier, as the SceneApp and SceneAppPage are primarily responsible for providing routing.

torkelo (approved):

PR #26, opened 2023-02-03
SceneTimeRange: Some time range model changes 

torkelo:
* Activate time range objects 
* When VizPanel detects a local time range render it's component in the PanelChrome header 
* Remove onIntervalChanged from SceneTimeRangeLike interface


dprokop (approved):
LGTM!

PR #27, opened 2023-02-10
SceneApp: Add test scenarios for routing

dprokop:
This PR adds test scenarios for common routing patterns (nav to scene page, across tabs and drill-down).

Wish we could test the `preserveUrlKeys` but I think we have to save this for an e2e as it requires the Grafana environment to be properly tested.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,6 +91,7 @@
     "prettier": "2.5.1",
     "react": "17.0.2",
     "react-dom": "17.0.2",
+    "react-router-dom": "^5.2.0",
     "rimraf": "^3.0.2",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -91,6 +91,7 @@
     "prettier": "2.5.1",
     "react": "17.0.2",
     "react-dom": "17.0.2",
+    "react-router-dom": "^5.2.0",
```
dprokop, line 94: Had to move (again) to devDeps as otherwise jest is not picking up react-router existence.

torkelo (approved):
Nice tests!!!

PR #32, opened 2023-02-23
Variables: Improve styling and spacing 

torkelo:
Before (not really true before as this is with 

Screenshot from 2023-02-23 15-19-07

After:

Screenshot from 2023-02-23 15-08-23


dprokop (approved):
LGTM

One thought I once had about the variable pickers is to be a little bit more visually verbose about them being actual variables that one can use in queries. I was thinking about using monospace font for the labels:

image

torkelo:
@dprokop it's a cool idea, but variables can also have specific display name (that cannot be used in query), and from a consumer perspective of the variable the fact that these are used in queries should not be that important, just just a filter

PR #33, opened 2023-02-24
SceneComponentWrapper: Wrap in React.memo to eliminate unnessary re-renders

torkelo:
Noticed that using SceneApp with scenes with variables and URL sync enabled causes a lot of unnessary re-renders due to re-renders being triggered due to location changes (due to useLocation or useRouteParams),
these re-renders trickles all the way down.

By wrapping all components in React.memo we make
sure the only thing that triggers re-render is

* State update
* New instance
* isEditing change


torkelo:
@dprokop yea, updating now

PR #34, opened 2023-02-25
SceneReactObject: Utility scene object when you just want to add existing react components to the scene

torkelo:
Adds a utility component that just renders either a component with props or a ReactNode.

### DIFF
--- a/src/components/JustRenderMyReact.tsx
+++ b/src/components/JustRenderMyReact.tsx
@@ -1,0 +1,39 @@
+import React from 'react';
+
+import { SceneObjectBase } from '../core/SceneObjectBase';
+import { SceneComponentProps, SceneLayoutChildState } from '../core/types';
+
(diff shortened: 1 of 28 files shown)


src/components/JustRenderMyReact.tsx
```diff
@@ -0,0 +1,39 @@
...
+   * Alternative to component and props is just to pass a React node
+   */
+  reactNode?: React.ReactNode;
+}
+
+/**
+ * Not a really useful component, just an example of how to create one
+ * @internal
+ */
+export class JustRenderMyReact extends SceneObjectBase<JustRenderMyReactState> {
```
dprokop, line 25: Great we need this, but I think we should name it differently, maybe:

- `SceneReactObject`
- `SceneRenderReact`
- `SceneReactComponent`

dprokop (approved):

PR #36, opened 2023-02-28
Implement auto-refresh for scene time range

dprokop:
Implements SceneRefreshPicker object for picking refresh rate and auto-refreshes.

Closes https://github.com/grafana/grafana/issues/63868


torkelo (approved):
Took this for a spin, works great, we need to update our demos so they add SceneRefreshPicker to the controls (and actions[] for core dashboard scene)

PR #38, opened 2023-03-02
Custom data transformations support 

dprokop:
~This draft proposes a new transformation node that allows custom transformartions(not available in core Grafana's registry).~

This, instead of original approach with a custom scene object, adds support for custom transfromations to the existing transformation APIs. 

Both `SceneDataTransformer` and `SceneQueryRunner` accept either standard Grafana transformation configs or custom transformation operators. Those can easily be mixed and matched to achieve expected results.

It of course introduces some complexity in the code, but at the same time reduces the complexity from the API consumer point of view. And is also a lot more versatile than custom object.

Try demo in https://github.com/grafana/grafana/pull/64044

TODO:
- [x] Add test cases for both SceneQueryRunner and SceneDataTransformer


torkelo (commented):
I think this makes sense, and I think having this feature on SceneQueryRunner would also make some sense.

Or maybe we should make a CompositeDataNode that has an array of data nodes that are processed in order (but not sure how the subscriptions will work), but it would be cool if we could make that work somehow.

src/querying/SceneQueryRunner.ts
```diff
@@ -247,7 +249,57 @@ export const getTransformationsStream: (
...
+            }
+            // @ts-ignore TypeScript has a hard time understanding this construct
+            stream = stream.pipe.apply(stream, operators);
+          } else {
+            stream = stream.pipe(mergeMap((series) => transformDataFrame(t as DataTransformerConfig[], series, ctx)));
+          }
+          return stream;
+        }),
+        map((series) => ({ ...data, series }))
+      );
```
dprokop, line 271: @kaydelaney @torkelo  - any ideas on how we can make this immutable? Mind the `stream` observable mutated in the `mergeMap`.

dprokop:
@torkelo @kaydelaney - updated this PR with the latest changes in core (ref https://github.com/grafana/grafana/pull/64208) and this should be now good to merge.

kaydelaney (approved):
LGTM!

PR #42, opened 2023-03-09
Fix using minified version

bfmatei:
When using a minified version of the library, Terser and/or other uglifiers are tampering with the class names and the constructor name. This is the case for a plugin generate by `@grafana/create-plugin` package.

Checking if an object is instance of class fixes this.

### DIFF
--- a/src/core/sceneGraph.ts
+++ b/src/core/sceneGraph.ts
@@ -6,6 +6,8 @@ import { SceneVariables } from '../variables/types';
 
 import { SceneDataState, SceneEditor, SceneLayoutState, SceneObject, SceneTimeRangeLike } from './types';
 import { lookupVariable } from '../variables/lookupVariable';
+import { SceneFlexLayout } from '../components/layout/SceneFlexLayout';
+import { SceneGridLayout } from '../components/layout/SceneGridLayout';
 
 /**
  * Get the closest node with variables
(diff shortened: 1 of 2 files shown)


dprokop (approved):
Good catch. Should be good to merge, checked if this does not cause any circular deps and it seems not.

torkelo (commented):
I think we cannot use instanceof because there could be different version so scenes package at the same time.

Right @dprokop ?

bfmatei:
@dprokop: I didn't think about circular dependencies hence I ran a test width [madge](https://www.npmjs.com/package/madge) and unfortunately there are some circular deps, one introduced by this PR.

```
1) src/core/types.ts > src/variables/types.ts
2) src/core/SceneObjectBase.tsx > src/core/utils.ts
3) src/components/layout/SceneGridLayout.tsx > src/components/layout/SceneGridRow.tsx
4) src/core/sceneGraph.ts > src/components/layout/SceneGridLayout.tsx > src/components/layout/SceneGridRow.tsx
5) src/components/VizPanel/VizPanel.tsx > src/components/VizPanel/VizPanelRenderer.tsx
6) src/variables/components/VariableValueInput.tsx > src/variables/variants/TextBoxVariable.tsx
7) src/variables/variants/query/QueryVariable.tsx > src/variables/variants/query/createQueryVariableRunner.ts
```

bfmatei:
@torkelo, @dprokop

I pushed another commit that should fix the circular dependency introduced by this PR.

Regarding different package versions, I don't know the library very well. Alternatively to this, I think we can have static string props defining the name of the scene object, similarly to how React handles `displayName` property on components.

src/core/sceneGraph.ts
```diff
@@ -74,7 +76,7 @@ export function getSceneEditor(sceneObject: SceneObject): SceneEditor {
  * Will walk up the scene object graph to the closest $layout scene object
  */
 export function getLayout(scene: SceneObject): SceneObject<SceneLayoutState> {
-  if (scene.constructor.name === 'SceneFlexLayout' || scene.constructor.name === 'SceneGridLayout') {
+  if (scene instanceof SceneFlexLayout || scene instanceof SceneGridLayout) {
```
torkelo, line 79: we should find a better way to find the layout object, maybe if we had a interface for Layouts that had a function, like isDraggingEnabled() we could check for the existence of this function

bfmatei:
Closing in favor of #44 .

PR #43, opened 2023-03-09
QueryEditor: Adds inline query editor scene object

kaydelaney:
- Factored `PanelChromeRenderer` out of `VizPanelRenderer` to be used with  `QueryEditorRenderer`.
- Added generic utility function `getClosest` for finding the closest scene object that matches user-supplied criteria. Refactored existing sceneGraph functions to use this utility.

Leaving this PR as a draft for now in case there's any feedback or I've made any major mistakes 😅


torkelo (commented):
Great job!

torkelo:
Why would we use PanelChromeRenderer when rendering QueryEditor ?

src/core/sceneGraph.ts
```diff
@@ -6,85 +6,77 @@ import { SceneVariables } from '../variables/types';
...
+  let curSceneObject: SceneObject | undefined = sceneObject;
+  let extracted: T | undefined = undefined;
+
+  while (curSceneObject && !extracted) {
+    extracted = extract(curSceneObject);
+    curSceneObject = curSceneObject.parent;
+  }
+
+  return extracted;
+}
```
dprokop, line 22: Nice!

src/core/sceneGraph.ts
```diff
@@ -6,85 +6,77 @@ import { SceneVariables } from '../variables/types';
...
-  if (scene.constructor.name === 'SceneFlexLayout' || scene.constructor.name === 'SceneGridLayout') {
-    return scene as SceneObject<SceneLayoutState>;
-  }
-
-  if (scene.parent) {
-    return getLayout(scene.parent);
+  const layoutSceneObject = getClosest(scene, (s) => (isLayoutSceneObject(s) ? s : undefined));
+  if (layoutSceneObject) {
+    return layoutSceneObject;
   }
```
dprokop, line 68: @kaydelaney - please sync with main, this was recently changed in https://github.com/grafana/scenes/pull/44

src/components/PanelChromeRenderer.tsx
```diff
@@ -0,0 +1,72 @@
...
+  hoverHeader?: boolean;
+}
+
+interface PanelChromeRendererProps {
+  children: React.ReactNode | ((innerWidth: number, innerHeight: number) => React.ReactNode);
+  loadingState?: LoadingState;
+  statusMessage?: string;
+}
+
+export function PanelChromeRenderer({
```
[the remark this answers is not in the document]
dprokop: @kaydelaney - not sure how this change is necessary for QueryEditor? Ref your comment in description: 

> tactored PanelChromeRenderer out of VizPanelRenderer to be used with QueryEditorRenderer.

src/components/QueryEditor/QueryEditorRenderer.tsx
```diff
@@ -0,0 +1,61 @@
...
+    return <div>Datasource has no query editor.</div>;
+  }
+
+  const QueryEditor = loadedDatasource.components.QueryEditor;
+
+  const onChange = (query: DataQuery) => {
+    model.setState({ query });
+
+    if (sceneQueryRunner) {
+      sceneQueryRunner.setState({
```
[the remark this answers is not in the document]
dprokop: Technically we could start simple, and provide the `QueryEditor` scene object with a callback property for `onQueryChange` and allow the consumer to implement the wiring - similar to how we do in https://github.com/grafana/grafana/blob/main/public/app/features/scenes/scenes/demo.tsx#L121 

But I'm not sure about this. I think more and more patterns like this emerge making it super hard in the future to think about serializing scenes.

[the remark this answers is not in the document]
dprokop: Thinking about this more and how potentially this could be used. Maybe the `QueryEditor` should have a local query state as it has right now, and an explicit **Run query** button rendered? Testing this now it appears that changes in the query editor are automatically propagated to `SceneQueryRunner`. While this is fine, and mimics the current Grafana behavior, I think It would be better and follow the changes we've been introducing to query editors in core (Run query button).

src/components/QueryEditor/QueryEditorRenderer.tsx
```diff
@@ -0,0 +1,61 @@
...
+    return <div>Loading data source...</div>;
+  }
+
+  if (!loadedDatasource.components.QueryEditor) {
+    return <div>Datasource has no query editor.</div>;
+  }
+
+  const QueryEditor = loadedDatasource.components.QueryEditor;
+
+  const onChange = (query: DataQuery) => {
```
dprokop, line 31: @kaydelaney to follow some _yet undocumented_  best practices, this should belong to the model, not the renderer.

dprokop:
To summarize what we just discussed:
1. Let's change the implementation slightly so that it allows editing closest's SceneQueryRunner queries.
2. No add/remove button - basically one can only change what's defined in SQR.
3. Add demo in [scenes-app demos](https://github.com/grafana/scenes/tree/main/packages/scenes-app/src/pages/Demos).

kaydelaney:
I've implemented the changes that have been suggested and added a small demo.
One issue I've left unsolved for now is how to handle multiple queries (for now the editor component just deals with one query). Should we just have a scrollable list of query editors, or a tabbed view or something? What do you think @dprokop @torkelo

packages/scenes/src/components/QueryEditor/QueryEditor.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { SceneObjectBase } from '../../core/SceneObjectBase';
+import { SceneLayoutChildState } from '../../core/types';
+
+import { QueryEditorRenderer } from './QueryEditorRenderer';
+import { DataQuery, DataSourceRef } from '@grafana/schema';
+
+import { SceneQueryRunner } from '../../querying/SceneQueryRunner';
+
+export interface QueryEditorState extends SceneLayoutChildState {
+  datasource: string | DataSourceRef;
```
dprokop, line 12: Given the suggested changes I think we should figure out data source based on the SceneQueryRunner rather than provide it explicitly. WDYT?

packages/scenes-app/src/pages/Demos/scenes/queryEditor.tsx
```diff
@@ -0,0 +1,25 @@
...
+        new VizPanel({
+          pluginId: 'timeseries',
+          title: 'Timeseries',
+          placement: {
+            minHeight: 200,
+            minWidth: '40%',
+          },
+        }),
+        new QueryEditor({
+          datasource: 'gdev-testdata',
```
dprokop, line 18: FYI - regardless my other comment - we have a `const DATASOURCE_REF` defined that points to a datasource that is provisioned with the demo app.

dprokop (commented):
> One issue I've left unsolved for now is how to handle multiple queries (for now the editor component just deals with one query).

@kaydelaney I've just tested this, and also looking at the code multiple queries are supported. I'm confused :)

packages/scenes/src/components/QueryEditor/QueryEditor.tsx
kaydelaney: Ah, you're right, will fix

kaydelaney:
> > One issue I've left unsolved for now is how to handle multiple queries (for now the editor component just deals with one query).
> 
> @kaydelaney I've just tested this, and also looking at the code multiple queries are supported. I'm confused :)

Added support after I left that comment 😅

packages/scenes/src/components/QueryEditor/QueryEditor.tsx
```diff
@@ -0,0 +1,36 @@
...
+export class QueryEditor extends SceneObjectBase<QueryEditorState> {
+  public static Component = QueryEditorRenderer;
+
+  public constructor(state?: Partial<SceneLayoutChildState>) {
+    super({ ...state });
+  }
+
+  public activate() {
+    super.activate();
+  }
```
torkelo, line 24: can remove this

packages/scenes/src/components/QueryEditor/QueryEditorRenderer.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const { data } = sceneGraph.getData(model).useState();
+  const sceneQueryRunner = sceneGraph.getSceneQueryRunner(model);
+
+  useEffect(() => {
+    const queryRunnerDatasource = sceneQueryRunner?.state.datasource;
+    getDataSource(queryRunnerDatasource, {})
+      .then((d) => model.setState({ datasource: d }))
+      .catch((err) => model.setState({ datasourceLoadErrorMessage: err }));
+  }, [sceneQueryRunner?.state.datasource, model]);
```
torkelo, line 24: I would move this to the QueryEditor object class, and handle it inside an activation handler.

packages/scenes/src/components/QueryEditor/QueryEditor.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { SceneQueryRunner } from '../../querying/SceneQueryRunner';
+import { SceneLayoutChildState } from '../../core/types';
+import { DataSourceApi } from '@grafana/data';
+
+export interface QueryEditorState extends SceneLayoutChildState {
+  datasource?: DataSourceApi;
+  datasourceLoadErrorMessage?: string;
+}
+
+export class QueryEditor extends SceneObjectBase<QueryEditorState> {
```
torkelo, line 15: not sure what the best name for this is now that it handles multiple queries.

packages/scenes/src/core/sceneGraph.ts
```diff
@@ -6,69 +6,65 @@ import { VariableCustomFormatterFn, SceneVariables } from '../variables/types';
...
 
   throw new Error('No layout found in scene tree');
 }
 
+/**
+ * Will walk up the scene object graph to the closest SceneQueryRunner
+ */
+export function getSceneQueryRunner(sceneObject: SceneObject): SceneQueryRunner | undefined {
+  return getClosest(sceneObject, (s) => (s.state.$data instanceof SceneQueryRunner ? s.state.$data : undefined));
+}
```
torkelo, line 66: This would not find query runners that are wrapped in SceneDataTransformer.

packages/scenes/src/components/QueryEditor/QueryEditorRenderer.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { SceneComponentProps } from '../../core/types';
+
+import { QueryEditor } from './QueryEditor';
+import { getDataSource } from '../../utils/getDataSource';
+
+export function QueryEditorRenderer({ model }: SceneComponentProps<QueryEditor>) {
+  const { datasourceLoadErrorMessage, datasource } = model.useState();
+
+  const { data } = sceneGraph.getData(model).useState();
+  const sceneQueryRunner = sceneGraph.getSceneQueryRunner(model);
```
torkelo, line 17:
```suggestion
  const {queries} = sceneGraph.getSceneQueryRunner(model).useState();
```

torkelo (commented):
I feel pretty stupid for proposing this idea, It was much harder to make a generic/reusable scene object like this than I thought. 

Maybe we should move the whole QueryEditor component to scenes-app, and build a small Explore-like demo app there instead of adding this component to the scenes library. 

Until we have 1-2 clear use cases from plugins it feels like it could just cause confusion. 

thoughts @dprokop @kaydelaney ?

kaydelaney:
Implemented some of the suggested changes @torkelo

packages/scenes/src/components/QueryEditor/QueryEditor.tsx
```diff
@@ -0,0 +1,53 @@
...
+    }
+  };
+
+  public onChange = (sceneQueryRunner: SceneQueryRunner, query: DataQuery) => {
+    const oldQueries = sceneQueryRunner.state.queries;
+    sceneQueryRunner.setState({
+      queries: oldQueries.map((q) => (q.refId === query.refId ? query : q)),
+    });
+
+    sceneQueryRunner.runQueries();
```
torkelo, line 51: not sure we should run queries on change automatically, only if onRunQuery is called.

packages/scenes/src/core/sceneGraph.ts
```diff
@@ -4,69 +4,80 @@ import { DefaultTimeRange, EmptyDataNode, EmptyVariableSet } from '../variables/
...
+    }
+  }
+  
+  if (sceneObject.parent) {
+    return getSceneQueryRunner(sceneObject.parent);
   }
 
-  throw new Error('No layout found in scene tree');
+  return undefined;
 }
```
torkelo, line 81: I would like to wait with adding this to core lib until we know more of how scenes are constructed, and in what scenarios you need to find the closest query runner.

torkelo (commented):
Looks like a good base. I would move QueryEditor and QueryEditorRender 
to a scenes/explore-like demo scene for now where we can have this and over time add things to it maybe to test different scenarios and show how to do more advanced things

packages/scenes/src/core/sceneGraph.ts
kaydelaney: Sounds fine. Move to scenes-app package?

packages/scenes/src/index.ts
```diff
@@ -23,6 +23,8 @@ export { TextBoxVariable } from './variables/variants/TextBoxVariable';
 export { UrlSyncManager } from './services/UrlSyncManager';
 export { SceneObjectUrlSyncConfig } from './services/SceneObjectUrlSyncConfig';
 
+export { getDataSource } from './utils/getDataSource';
```
dprokop, line 26: @kaydelaney let's not expose this util from the scenes ok? It technically has nothing to do with the API so I suggest just resolbing the datasource in the demo implementation.

packages/scenes/src/components/QueryEditor/QueryEditor.tsx
dprokop: Agree with that, think we should have explicit on button in the demo.

dprokop (approved):
Looks good! Some final touches left to be made :)

dprokop:
@kaydelaney I've pushed sync with main and a minir layout fix

packages/scenes-app/src/components/QueryEditor/QueryEditor.tsx
```diff
@@ -0,0 +1,75 @@
+import { SceneObjectBase, SceneLayoutChildState, SceneQueryRunner, getDataSource, SceneDataProvider, SceneObject } from '@grafana/scenes';
+
+
+import { QueryEditorRenderer } from './QueryEditorRenderer';
+import { DataQuery } from '@grafana/schema';
+
+import { DataSourceApi } from '@grafana/data';
+
+export interface QueryEditorState extends SceneLayoutChildState {
```
torkelo, line 9:
```suggestion
export interface QueryEditorState extends SceneObjectState {
```

kaydelaney: Gives an error: "An interface can only extend an object type or intersection of object types with statically known members"

PR #44, opened 2023-03-10
Layout: Introduce interface for layout and make PanelChrome dragging work like it needs to

torkelo:
* Introduce an interface for Layout, use isDraggable function as a test instead of instanceof or constructor name  (not sure if this will break in a minified build?? if that is soo then so much many things would break in core Grafana as we check for function names so many places to check if something implements an interface)
* Fix logic in VizPanelRenderer so it's more correct
* Fix PanelChrome props so that dragClass and dragClassCancel is passed 

I know this is a bit messy, and linked to how the grid works today, but not sure what else we can do right now if we want to support current behaviors and features (in core dashboards). I another interface function idea I had would be getPanelChromeProps , but then we link the Layout interface to PanelChromeProp which is not great, and would not work well if other scene objects want to be draggable


src/components/VizPanel/VizPanelRenderer.tsx
```diff
@@ -31,11 +30,13 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
   const { data } = sceneGraph.getData(model).useState();
   const parentLayout = sceneGraph.getLayout(model);
 
-  // TODO: this should probably be parentLayout.isDraggingEnabled() ? placement?.isDraggable : false
-  // The current logic is not correct, just because parent layout itself is not draggable does not mean children are not
-  const isDraggable = parentLayout.state.placement?.isDraggable ? placement?.isDraggable : false;
-  const dragHandle = <SceneDragHandle layoutKey={parentLayout.state.key!} />;
+  // If parent has enabled dragging and we have not explicitly disabled it then dragging is enabled
+  const isDraggable = parentLayout.isDraggable() ? placement?.isDraggable ?? true : false;
```
torkelo, line 34: does this logic make sense? or should we require each child to explicitly enable dragging? (instead of treating undefined as true)

src/core/types.ts
```diff
@@ -111,7 +111,11 @@ export interface SceneLayoutState extends SceneLayoutChildState {
   children: SceneLayoutChild[];
 }
 
-export type SceneLayout<T extends SceneLayoutState = SceneLayoutState> = SceneObject<T>;
+export interface SceneLayout<T extends SceneLayoutState = SceneLayoutState> extends SceneObject<T> {
+  isDraggable(): boolean;
+  getDragClassName?(): string;
+  getDragCancelClassName?(): string;
+}
```
torkelo, line 118: alt names
* getDragClass
* getDragClassCancel 

shorter and simpler :)

bfmatei: Yeah, I personally would align naming of the props with the methods.

The `name` suffix makes it more clear to people used to React's naming conventions that this needs to be a CSS class rather than a JS class, but I don't have strong opinions on either one.

bfmatei (approved):
Way more elegant solution than in #42. Going to close the other one now.

I left some comments but other than that it looks pretty straight forward to me.

src/components/layout/SceneGridLayout.tsx
```diff
@@ -3,29 +3,53 @@ import ReactGridLayout from 'react-grid-layout';
...
 
   public constructor(state: SceneGridLayoutState) {
     super({
       ...state,
-      placement: {
-        isDraggable: true,
-        ...state.placement,
-      },
+      isDraggable: true,
+      placement: state.placement,
```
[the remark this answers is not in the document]
torkelo: ah, yes your right

bfmatei:
One small suggestion: given that for time being the projects using scenes might be broken due to the issue presented in #42 , should this also release a new version of scenes when merged?

src/components/VizPanel/VizPanelRenderer.tsx
```diff
@@ -31,11 +30,13 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
   const { data } = sceneGraph.getData(model).useState();
   const parentLayout = sceneGraph.getLayout(model);
 
-  // TODO: this should probably be parentLayout.isDraggingEnabled() ? placement?.isDraggable : false
-  // The current logic is not correct, just because parent layout itself is not draggable does not mean children are not
-  const isDraggable = parentLayout.state.placement?.isDraggable ? placement?.isDraggable : false;
-  const dragHandle = <SceneDragHandle layoutKey={parentLayout.state.key!} />;
+  // If parent has enabled dragging and we have not explicitly disabled it then dragging is enabled
+  const isDraggable = parentLayout.isDraggable() && (placement?.isDraggable ?? true);
```
dprokop, line 34: I think this will break with the Flex Layout. Actually, I think you are missing implementing SceneLayout in SceneFlexLayout object in this PR

dprokop: fixed in c48a569e620705115c92ead0806fb4fa08972588

torkelo: @dprokop ah, thanks! forgot about that one :)

dprokop (approved):
@bfmatei @torkelo  - I've just checked and this is making the apps work well when building plugin.

PR #46, opened 2023-03-10
Return value without stringify

idastambuk:
When interpolating variables for json format in [template_srv.ts](https://github.com/grafana/grafana/blob/f8d89eff562dfb7c90b0bb617c639339af30d55c/public/app/features/templating/template_srv.ts#L128), when the variable value is a string, the formatter wraps it in an extra set of unnecessary quotation marks:
Screen Shot 2023-03-09 at 6 27 56 PM

I added a check for typeof variable to prevent this.


torkelo (approved):

dprokop:
Thanks @idastambuk!

PR #48, opened 2023-03-13
SceneVariableSet: Detect state changes and init new variables

torkelo:
Really unsure of the way SceneVariableSet knows which variables to update is going to work long term (record current values).

To make this work I have to record current value on value change (as well as deactive).

### DIFF
--- a/src/variables/sets/SceneVariableSet.ts
+++ b/src/variables/sets/SceneVariableSet.ts
@@ -36,6 +36,9 @@ export class SceneVariableSet extends SceneObjectBase<SceneVariableSetState> imp
       this.subscribeToEvent(SceneVariableValueChangedEvent, (event) => this.handleVariableValueChanged(event.payload))
     );
 
+    // Subscribe to state changes
+    this._subs.add(this.subscribeToState({ next: (state) => this.handleStateChanged(state) }));
+
     this.checkForVariablesThatChangedWhileInactive();
 
     // Add all variables that need updating to queue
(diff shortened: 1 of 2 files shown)


src/variables/sets/SceneVariableSet.ts
```diff
@@ -48,6 +51,29 @@ export class SceneVariableSet extends SceneObjectBase<SceneVariableSetState> imp
...
+   */
+  private handleStateChanged(state: SceneVariableSetState) {
+    const variablesToUpdateCountStart = this._variablesToUpdate.size;
+
+    for (const variable of state.variables) {
+      // If this is a new variable
+      if (
+        !this._variablesToUpdate.has(variable) &&
+        !this._updating.has(variable) &&
+        !this._variableValueRecorder.hasRecordedValue(variable)
```
torkelo, line 65: _variableValueRecorder feels like a fragile solution somehow to know which variables are new, and which variables need updating, but can't come up with anything better right now

dprokop: I agree. And I think more objects in the future might need to inspect previous values for state change reactions. What about exposing previous state too via the state subscription? Sth like https://github.com/grafana/scenes/pull/66

packages/scenes/src/variables/sets/SceneVariableSet.ts
```diff
@@ -48,6 +51,39 @@ export class SceneVariableSet extends SceneObjectBase<SceneVariableSetState> imp
     this.updateNextBatch();
   }
 
+  /**
+   * Look for new variables that need to be initialized
+   */
+  private handleStateChanged = (newState: SceneVariableSetState, oldState: SceneVariableSetState) => {
```
dprokop, line 57: With old/new state this implementation is no so much easier to follow!

dprokop (approved):

PR #50, opened 2023-03-14
QueryRunner: Support triggering processing transformations without re-issung query

torkelo:
This refactors SceneQueryRunner to provide more flexibility around transformations

* To support getting data without transforms applied (not fully implmented in this PR but would be easy to add)
* To support re-process transformations without triggering query
* Support custom transformation scene objects that transform data based on changing variables, or other scene objects they subscribe to (like search box) 
* To support variable use in transformations (When change only triggers transform), not fully implemented in PR

Example use case is to simplify use cases like:
the point of this is component is to simplify use cases covered in https://github.com/grafana/grafana/blob/f7d37a34b2e110d4edb21100535333456da4e8db/public/app/features/scenes/apps/SceneSearchBox.tsx#L53

### DIFF
--- a/src/querying/SceneQueryRunner.test.ts
+++ b/src/querying/SceneQueryRunner.test.ts
@@ -24,7 +24,7 @@ import {
   SceneQueryRunnerDataTransformer,
 } from './transformations';
 import { SceneObjectBase } from '../core/SceneObjectBase';
-import { SceneObjectStatePlain } from '../core/types';
+import { SceneObjectStatePlain, SceneQueryRunnerInterface } from '../core/types';
 
 const getDataSourceMock = jest.fn().mockReturnValue({
   getRef: () => ({ uid: 'test' }),
@@ -361,7 +361,7 @@ describe('SceneQueryRunner', () => {
     });
 
     describe('custom transformer object', () => {
-      it.only('Can re-trigger transformations without issuing new query', async () => {
+      it('Can re-trigger transformations without issuing new query', async () => {
         const customDataTransfomer = new CustomQueryRunnerDataTransformer({ structureRev: 10 });
 
         const queryRunner = new SceneQueryRunner({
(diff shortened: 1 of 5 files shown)


src/querying/SceneQueryRunner.test.ts
```diff
@@ -529,3 +572,21 @@ describe('SceneQueryRunner', () => {
...
+{
+  public updateStateAndRetriggerTransformation() {
+    this.setState({ structureRev: this.state.structureRev + 1 });
+    this.publishEvent(new ReprocessTransformationsEvent());
+  }
+
+  public transform(data: PanelData): Observable<PanelData> {
+    return of(data).pipe(map((d) => ({ ...d, structureRev: this.state.structureRev })));
+  }
+}
```
torkelo, line 592: In a real scenario this would maybe subscribe to control component state (like a searchbox), or implement variable dependency and handle variable value change and then emit ReprocessTransformationsEvent

torkelo: the point of this is component is to simplify use cases covered in https://github.com/grafana/grafana/blob/f7d37a34b2e110d4edb21100535333456da4e8db/public/app/features/scenes/apps/SceneSearchBox.tsx#L53  

So that you don't have to create a full wrapper around SceneQueryRunner  to do this

src/querying/SceneQueryRunner.ts
```diff
@@ -71,6 +79,24 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> {
...
+
+    if (this.state.transformer) {
+      this.state.transformer.activate();
+
+      // Subscribe to transformer wanting to re-process transformations
+      this._subs.add(
+        this.state.transformer.subscribeToEvent(ReprocessTransformationsEvent, () => {
+          this._rawDataSubject.next(this.state.dataPreTransforms!);
+        })
+      );
```
torkelo, line 97: we could potentially change this to be an event emitted on the SceneQueryRunner object directly, or just add a public function "reprocessTransformations" , then the transformer can just call parent.reprocessTransformations();

torkelo: might be simpler

PR #51, opened 2023-03-14
QueryRunner: Alternative more compositional approach to transformations

torkelo:
This is an alternative to https://github.com/grafana/scenes/pull/50 that explores a more compositional approach.

* Separate query runner and transformations so we can re-process transformations without query
* Now SceneQueryRunner is more simple and does not know or care about transformations 

The downside with this is that it's going to be a bit more complex for components that interact with the query runners, they might have to check if is it a normal query runner or a QueryRunnerWithTransforms.

Like updating queries state from panel edit could be more complex, also when adding new transforms it would mean changing and adding this new wrapping query runner. But it does keep the SceneQueryRunner itself simpler.

### DIFF
--- a/src/querying/SceneQueryRunner.test.ts
+++ b/src/querying/SceneQueryRunner.test.ts
@@ -18,14 +18,9 @@ import { SceneVariableSet } from '../variables/sets/SceneVariableSet';
 import { TestVariable } from '../variables/variants/TestVariable';
 import { getCustomTransformOperator } from '../core/SceneDataTransformer.test';
 import { mockTransformationsRegistry } from '../utils/mockTransformationsRegistry';
-import {
-  DefaultQueryRunnerDataTransformer,
-  QueryRunnerWithTransformations,
-  ReprocessTransformationsEvent,
-  SceneQueryRunnerDataTransformer,
-} from './transformations';
+import { QueryRunnerWithTransformations } from './transformations';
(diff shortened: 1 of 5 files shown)


src/querying/SceneQueryRunner.test.ts
```diff
@@ -236,112 +246,162 @@ describe('SceneQueryRunner', () => {
...
     });
 
-    describe('custom transformations', () => {
-      it('applies leading custom transformer', async () => {
-        const queryRunner = new SceneQueryRunner({
-          queries: [{ refId: 'A' }],
-          $timeRange: new SceneTimeRange(),
-          maxDataPoints: 100,
-          // divide by 100, multiply by 2, multiply by 3
+    // describe('custom transformations', () => {
```
torkelo, line 249: these tests are just commented out because this is just draft, have not updated these tests yet

dprokop (approved):
Compared to #50, this approach is more appealing to me. I like how lean those two query runners are, not mixing all concerns in the SceneQueryRunner as in #50. 


> Like updating queries state from panel edit could be more complex, also when adding new transforms it would mean changing and adding this new wrapping query runner. But it does keep the SceneQueryRunner itself simpler.

I'm not sure this is gonna be the case. I think we can easily assume that panel edit's default query runner is the one with transformations support.

PR #52, opened 2023-03-14
Monorepo setup to provide a demo app

dprokop:
Introduces lerna monorepo setup. Consists of:
- @grafana/scenes package (`/packages/scenes`)
- Demo app (`/packages/scenes-app`)

To test different local setup scenarios please follow the guides in:
- Demo app: https://github.com/grafana/scenes/blob/monorepo-setup/packages/scenes-app/README.md
- @grafana/scenes: https://github.com/grafana/scenes/blob/monorepo-setup/packages/scenes/README.md


torkelo:
took this for a spin! works great. 

You have to start a separate watcher for the scene app, it is not built when you run /scripts/dev.sh , so you kind of need to watchers runnning, and if you make a change to scenes core you have to also make a change to app to trigger a rebuild of it. But I think that is fine as start, maybe we can look at ways to have a single watch in the future.

torkelo (approved):

PR #53, opened 2023-03-14
SceneObject: Add add activation handlers that can be registered from the outside

torkelo:
So there is some common patterns emerging where you need to wire up two scene objects, for example maybe you want SearchBox or dropdown component and when it changes state it should update the state of another component or trigger a query (or transformation). This wiring can be done from the outside of scene object by calling functions like subscribeToState or subscribeToEvent but there is no way to do that on activation, or cleanup the wiring on deactivation.

In order to do logic on activate or deactivate you have to create a custom scene object. This feels limiting. This PR explores adding a new function to SceneObject 
named registerActivationHandler. 

These activation handlers are called on activate, and if they return a function that will be called on deactivate (So similar to useEffect). 

I included a test case that demo's the "wire things from the outside" use case. But we could also potentially make activate and deactivate private functions, that cannot be extended by SceneObjects, and have all scene objects use the registerActivationHandler instead in constructor (instead of overriding the activate function and having to remember to call super.activate(). Included a example of that. 

For another demo of a use case look at 
https://github.com/grafana/grafana/blob/dynamic-scene-app-page-experiment/public/app/features/scenes/apps/GrafanaMonitoringApp.tsx#L67 

Where it can hook up an activation handler for SceneAppPage to potentially do something dynamic (add dynamic tabs, and page header controls)


src/core/SceneObjectBase.test.ts
```diff
@@ -148,4 +148,53 @@ describe('SceneObject', () => {
...
+
+      scene.registerActivationHandler(() => {
+        const sub = nestedScene.subscribeToState({
+          next: (state) => scene.setState({ name: state.name }),
+        });
+
+        return () => {
+          sub.unsubscribe();
+          unsubscribed = true;
+        };
```
torkelo, line 172: one annoying thing here is that you do not have access to the internal _subs (rxjs Subscription), so you have to handle unsub yourself. 

We could make subs public, or pass it into the activation handler, but that feels a bit clunky (better to make it public).

dprokop (approved):

PR #54, opened 2023-03-14
Revert "QueryRunner: Alternative more compositional approach to transformations"

torkelo:
Reverts grafana/scenes#51


torkelo:
need some more time with this PR

PR #55, opened 2023-03-14
QueryRunner: Alternative more compositional approach to transformations

torkelo:
This is an alternative to https://github.com/grafana/scenes/pull/50 that explores a more compositional approach.

* Separate query runner and transformations so we can re-process transformations without query
* Now SceneQueryRunner is more simple and does not know or care about transformations 

The downside with this is that it's going to be a bit more complex for components that interact with the query runners, they might have to check if is it a normal query runner or a QueryRunnerWithTransforms.

Like updating queries state from panel edit could be more complex, also when adding new transforms it would mean changing and adding this new wrapping query runner. But it does keep the SceneQueryRunner itself simpler.


src/core/SceneDataTransformer.ts
```diff
@@ -1,67 +1,85 @@
...
+      return this.state.$data;
+    }
+
     if (!this.parent || !this.parent.parent) {
-      return;
+      throw new Error('SceneDataTransformer must either have $data set on it or have a parent.parent with $data');
     }
 
-    const initialData = sceneGraph.getData(this.parent.parent).state.data;
+    return sceneGraph.getData(this.parent.parent);
```
dprokop, line 50: Yes yes yes, this is great.

src/core/SceneDataTransformer.ts
```diff
@@ -1,67 +1,85 @@
...
+  }
 
-    if (initialData) {
-      this.transformData(of(initialData));
+  public setContainerWidth(width: number) {
+    if (this.state.$data && this.state.$data.setContainerWidth) {
+      this.state.$data.setContainerWidth(width);
     }
```
dprokop, line 56: Uhm, think we need to use `getSourceData` here, to support i.e scenario where SQR and SceneDataTransformer are two separate nodes? But this can introduce problems when SQR's data is consumed directly by VizPanel and indirectly via SceneDataTransformer within the same scene.

dprokop (approved):

src/core/SceneDataTransformer.ts
torkelo: @dprokop well, the setContainerWidth call should only be relevant if the query runner is on the same node I think  (IE VizPanel has SceneDataTransformer with an  inner query runner) 

If the query runner is on the parent level, not sure the child VizPanel width should ever impact maxDataPoints, for those scenarios we need something else to set maxDataPoints (manually configured )

torkelo:
@dprokop moved SceneDataTransfomer to querying/  and moved all the transformation related to tests from SceneQueryRunner.tests.ts to SceneDataTransformer

src/core/SceneDataTransformer.ts
dprokop: You are right, VizPanel should only set that for its directly attached $data.

dprokop (approved):

PR #56, opened 2023-03-14
Types: depend on SubscriptionLike rather than Subscription directly 

ryantxu:
In general nicer to have the simple interfaces as core constraints rather than the full class hierarchy.

### DIFF
--- a/src/core/SceneObjectBase.tsx
+++ b/src/core/SceneObjectBase.tsx
@@ -1,5 +1,5 @@
 import { useEffect } from 'react';
-import { Observer, Subject, Subscription, Unsubscribable } from 'rxjs';
+import { Observer, Subject, Subscription, SubscriptionLike, Unsubscribable } from 'rxjs';
 import { v4 as uuidv4 } from 'uuid';
 
 import { BusEvent, BusEventHandler, BusEventType, EventBusSrv } from '@grafana/data';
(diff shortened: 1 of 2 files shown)


src/core/types.ts
```diff
@@ -69,7 +69,7 @@ export interface SceneObject<TState extends SceneObjectState = SceneObjectState>
   readonly urlSync?: SceneObjectUrlSyncHandler<TState>;
 
   /** Subscribe to state changes */
-  subscribeToState(observer?: Partial<Observer<TState>>): Subscription;
+  subscribeToState(observer?: Partial<Observer<TState>>): SubscriptionLike;
```
ryantxu, line 72: Is there a use case for subscribe without an observer?
```suggestion
  subscribeToState(observer: Partial<Observer<TState>>): SubscriptionLike;
```

src/core/SceneObjectBase.tsx
```diff
@@ -82,7 +82,7 @@ export abstract class SceneObjectBase<TState extends SceneObjectState = SceneObj
   /**
    * Subscribe to the scene state subject
    **/
-  public subscribeToState(observerOrNext?: Partial<Observer<TState>>): Subscription {
+  public subscribeToState(observerOrNext?: Partial<Observer<TState>>): SubscriptionLike {
```
ryantxu, line 85: Does "OrNext" mean something?
```suggestion
  public subscribeToState(observer: Partial<Observer<TState>>): SubscriptionLike {
```

torkelo: observerOrNext the next part is from how subscribe method signatures look like insisde rxjs (where they take an observer or a next function), but this function does not take a next function (I wish it did), but can add that later.

torkelo (approved):

PR #60, opened 2023-03-15
SceneApps: Refactor out types 

torkelo:
Just breaks things up a bit and to do that I had to introduce an interface for SceneAppPage (named SceneAppPageLike). 

I have another PR that adds some new some new capabilities to SceneAppPage (to support dynamic tabs, and controls above tabs). 

Doing this PR first will make that later PR easier to review.

This will require some small updates in apps, the getScene function signature has changed for drilldown views (SceneAppPage -> SceneAppPageLike)


src/components/SceneApp/types.ts
```diff
@@ -0,0 +1,57 @@
...
+  getScene: (routeMatch: SceneRouteMatch) => EmbeddedScene;
+  // Array of scenes used for drilldown views
+  drilldowns?: SceneAppDrilldownView[];
+  // Function that returns a parent page object, used to create breadcrumbs structure
+  getParentPage?: () => SceneAppPageLike;
+  // Array of query params that will be preserved in breadcrumb and page tab links, i.e. ['from', 'to', 'var-datacenter',...]
+  preserveUrlKeys?: string[];
+}
+
+export type SceneAppPageLike = SceneObject<SceneAppPageState>;
```
dprokop, line 50: So - thinking about this, I reason this is to avoid circular dependencies when separating the components into different files? Or to provide way to custom app page objects rendering?

dprokop (commented):
Looks good, just one question

dprokop (approved):

src/components/SceneApp/types.ts
torkelo: @dprokop yes, to avoid circular dependencies

dprokop:
@torkelo so this PR is sth we might actually want to hold on merging until next week given it will influence apps.

PR #66, opened 2023-03-15
SceneObjectBase: Expose previous state in state subscription

dprokop:
Changes API of `SceneObject.subscribeToState` to expose previously recorded state.

```ts
subscribeToState(observer?: Partial<Observer<{ previous: TState; current: TState }>>): SubscriptionLike;
```

### DIFF
--- a/packages/scenes/src/core/SceneObjectBase.tsx
+++ b/packages/scenes/src/core/SceneObjectBase.tsx
@@ -1,5 +1,5 @@
 import { useEffect } from 'react';
-import { Observer, Subject, Subscription, SubscriptionLike, Unsubscribable } from 'rxjs';
+import { Observer, Subject, Subscription, SubscriptionLike, Unsubscribable, startWith, pairwise, map } from 'rxjs';
 import { v4 as uuidv4 } from 'uuid';
 
 import { BusEvent, BusEventHandler, BusEventType, EventBusSrv } from '@grafana/data';
@@ -82,8 +82,14 @@ export abstract class SceneObjectBase<TState extends SceneObjectState = SceneObj
   /**
    * Subscribe to the scene state subject
    **/
-  public subscribeToState(observerOrNext?: Partial<Observer<TState>>): SubscriptionLike {
-    return this._subject.subscribe(observerOrNext);
+  public subscribeToState(observerOrNext?: Partial<Observer<{ previous: TState; current: TState }>>): SubscriptionLike {
+    return this._subject
+      .pipe(
+        startWith(this._state),
+        pairwise(),
+        map(([previous, current]) => ({ previous, current }))
+      )
+      .subscribe(observerOrNext);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


torkelo (commented):
if you need the previous state, you can use SceneObjectStateChangedEvent which has this (without the rxjs gymnastics).  I guess if you subscribe to SceneObjectStateChangedEvent you can also get children state changes you have to filter out.

packages/scenes/src/core/SceneObjectBase.tsx
```diff
@@ -82,8 +82,14 @@ export abstract class SceneObjectBase<TState extends SceneObjectState = SceneObj
...
-  public subscribeToState(observerOrNext?: Partial<Observer<TState>>): SubscriptionLike {
-    return this._subject.subscribe(observerOrNext);
+  public subscribeToState(observerOrNext?: Partial<Observer<{ previous: TState; current: TState }>>): SubscriptionLike {
+    return this._subject
+      .pipe(
+        startWith(this._state),
+        pairwise(),
+        map(([previous, current]) => ({ previous, current }))
+      )
+      .subscribe(observerOrNext);
```
torkelo, line 92: this makes every subscription place a bit more complicated, (having to do state.current, deconstruct current).

If previous in some edge cases are needed (and using SceneObjectStateChangedEvent is not feasible) we could expose the state stream directly and then consumers could using this rxjs pairwise code to get a current/previous. 

would like to know the use case / scenario for this. Feels like it would also be easy on the consumer side to remember previous in a local var (but could be more cumbersome than I imagine or needed in more places than I think). so o

dprokop: @torkelo I was thinking primarily about alternative approach to https://github.com/grafana/scenes/pull/48#discussion_r1134143579

torkelo: @dprokop I see, ok. Makes some sense. I mean we can easily remember the previous state in that instance, but I agree this could be useful. Just don't like the rxjs gymnastics or that the callback now is much more complex consisting of an object with current and previous. 

I am sort of inclined to skip the whole rxjs subject thing, and since subscribeToState is not really compatible with rxjs pipelines anyway (would have to be a function that returns Observable for that). we could do something like this instead.

```ts
 subscribeToState<TState>(handler: (newState: TState, prevState: TState>) => void): SubscriptionLike
```

This way it's very similar to React's methods, and you don't really have to care about the prevState if you don't want to

We still return a SubscriptionLike so we can leverage that nice way of unsubbing (and adding many subs to _subs Subscription). but under the hood it's not a Subject and we just have a Set of subscribers. Similar to https://github.com/grafana/scenes/pull/53  (there they are just arrays as there is no way unregister them, but for state subscriptions, it would have to be a Set so we can easily remove them when unsubscribe is called).

dprokop:
Closing in favor of #72

PR #68, opened 2023-03-15
VizPanel: Enable panel menus

dprokop:
cc @yesoreyeram

### DIFF
--- a/packages/scenes-app/src/pages/Demos/Demos.tsx
+++ b/packages/scenes-app/src/pages/Demos/Demos.tsx
@@ -1,6 +1,6 @@
 import { EmbeddedScene, SceneApp, SceneAppPage, SceneFlexLayout, SceneReactObject } from '@grafana/scenes';
 import { Stack } from '@grafana/experimental';
-import * as React from 'react';
+import React, { useMemo } from 'react';
 import { prefixRoute } from '../../utils/utils.routing';
 import { ROUTES } from '../../constants';
 import { getDemoByTitle, getDemos } from './getDemos';
@@ -43,7 +43,7 @@ const getScene = () => {
 };
 
 export const DemoPage = () => {
-  const scene = getScene();
+  const scene = useMemo(() => getScene(), []);
   return <scene.Component model={scene} />;
 };
 
--- a/packages/scenes-app/src/pages/Demos/Demos.tsx
+++ b/packages/scenes-app/src/pages/Demos/Demos.tsx
@@ -1,0 +1,63 @@
+import { EmbeddedScene, SceneApp, SceneAppPage, SceneFlexLayout, SceneReactObject } from '@grafana/scenes';
+import { Stack } from '@grafana/experimental';
+import * as React from 'react';
+import { prefixRoute } from '../../utils/utils.routing';
+import { ROUTES } from '../../constants';
+import { getDemoByTitle, getDemos } from './getDemos';
+import { Card } from '@grafana/ui';
+
+const getScene = () => {
+  return new SceneApp({
+    pages: [
+      new SceneAppPage({
+        title: 'Scene demos',
+        url: prefixRoute(ROUTES.Demos),
+        getScene: () => {
+          return new EmbeddedScene({
+            body: new SceneFlexLayout({
+              direction: 'column',
+              children: [
+                new SceneReactObject({
+                  component: () => <DemosList />,
+                }),
+              ],
+            }),
+          });
+        },
+        drilldowns: [
+          {
+            routePath: `${prefixRoute(ROUTES.Demos)}/:id`,
+            getPage: (match, parent) => {
+              const id = match.params.id;
+              return new SceneAppPage({
+                title: id,
+                url: `${prefixRoute(ROUTES.Demos)}/${id}`,
+                getScene: () => getDemoByTitle(id),
+              });
+            },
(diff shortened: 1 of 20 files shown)


packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -26,6 +27,7 @@ export interface VizPanelState<TOptions = {}, TFieldConfig = {}> extends SceneLa
   pluginVersion?: string;
   displayMode?: 'default' | 'transparent';
   hoverHeader?: boolean;
+  menuItems?: PanelMenuItem[];
```
torkelo, line 30: think we need the option to make this lazy as well (Like how it is in PanelStateWrapper), IE figure out menu items when it opens. That way menu items can be a bit more dynamic and depend on current state (loaded plugin, data)

dprokop:
@torkelo did some tinkering with `addActivationHandler` and it looks good IMO. 

Also - scenes-app now has demos Page, similar to core's demos : http://localhost:3001/a/grafana-scenes-app/demos

packages/scenes/src/components/VizPanel/VizPanelMenu.tsx
```diff
@@ -0,0 +1,58 @@
...
+      items: state.items || [],
+    });
+  }
+
+  // Allows adding menu items dynamically
+  public addItem(item: PanelMenuItem) {
+    this.setState({
+      items: [...this.state.items, item],
+    });
+  }
```
torkelo, line 26: hm... is this going to be common, to add a single item? (instead of updating/setting the full menu)

packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
```diff
@@ -66,6 +67,10 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
   if (model.state.$timeRange) {
     titleItems.push(<model.state.$timeRange.Component model={model.state.$timeRange} />);
   }
+  let panelMenu;
+  if (menu) {
+    panelMenu = <menu.Component model={menu} />;
```
torkelo, line 72:
```suggestion
    panelMenu = () => <menu.Component model={menu} />;
```

I think this should be lazy, so it's only rendered when it's opened

packages/scenes/src/components/VizPanel/VizPanelMenu.tsx
```diff
@@ -0,0 +1,58 @@
...
+  items: PanelMenuItem[];
+}
+
+export class VizPanelMenu extends SceneObjectBase<VizPanelMenuState> {
+  static Component = VizPanelMenuRenderer;
+
+  public constructor(state: Partial<VizPanelMenuState>) {
+    super({
+      ...state,
+      items: state.items || [],
```
torkelo, line 17:
```suggestion
      items: state.items ?? [],
```

nitpick: since items is required (not optional), this code / default setting seems unnecessary

packages/scenes-app/src/pages/Demos/getDemos.ts
```diff
@@ -0,0 +1,27 @@
...
+interface SceneDef {
+  title: string;
+  getScene: () => EmbeddedScene;
+}
+
+export function getDemos(): SceneDef[] {
+  return [{ title: 'Panel menu', getScene: getPanelMenuTest }];
+}
+
+const cache: Record<string, EmbeddedScene> = {};
```
torkelo, line 13: is this really needed, think the cache inside SceneAppPage should handle this

packages/scenes-app/src/pages/Demos/Demos.tsx
```diff
@@ -0,0 +1,63 @@
...
+            },
+          },
+        ],
+      }),
+    ],
+  });
+};
+
+export const DemoPage = () => {
+  const scene = getScene();
```
torkelo, line 46: it's best to keep this outside the react component to create this only once. Or use the same key for SceneApp, or SceneAppPage, creating it here on every render makes every route/url change an unmount/remount

torkelo (approved):

packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
torkelo: @dprokop this is the important bit I think ^ 
that way you can use the VizPanelMenu's activation handler to do things when the menu opens

packages/scenes-app/src/pages/Demos/getDemos.ts
torkelo: I tested this without cache and seems to work still, getPanelMenuTest is only called once

dprokop: good, will push in a bit

packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
torkelo: @dprokop just tested this, it's not really needed. This component is only rendered when menu is shown anyway, so this comment was wrong

dprokop: Precisely, that was done on purpose.

PR #71, opened 2023-03-16
SceneAppPage: Support dynamic pages (changing tabs, title, controls)

torkelo:
* Adds state subscriptions to supports dynamic scene app page states 
* Connects the inner dynamically built scene to the SceneAppPage (initializedScene). This makes the connected (parent) so that the SceneAppPage can have $timeRange and controls that are connected to the EmbeddedScene components. 
* Adds controls to SceneAppPage and a simple component SceneControlWrapper (that adds a label and separator ability). Very experimental as the UX for scene apps feels very much up in the air. We could also add label prop directly to SceneTimeRangePicker.

Screenshot from 2023-03-15 09-14-06

For an example of dynamic tabs and page controls (also relying on activationHandlers, https://github.com/grafana/scenes/pull/53)
https://github.com/grafana/grafana/blob/dynamic-scene-app-page-experiment/public/app/features/scenes/apps/GrafanaMonitoringApp.tsx#L79


packages/scenes/src/components/SceneControlWrapper.tsx
```diff
@@ -0,0 +1,80 @@
+import { css } from '@emotion/css';
+import { GrafanaTheme2 } from '@grafana/data';
+import { useStyles2 } from '@grafana/ui';
+import React from 'react';
+import { SceneObjectBase } from '../core/SceneObjectBase';
+import { isSceneObject, SceneComponentProps, SceneObject, SceneObjectStatePlain } from '../core/types';
+
+export interface SceneControlWrapperState extends SceneObjectStatePlain {
+  label: string;
+  component?: SceneObject;
```
torkelo, line 10: TODO: remove

packages/scenes/src/components/SceneApp/SceneAppPage.tsx
```diff
@@ -60,84 +74,79 @@ function SceneAppPageRenderer({ model }: SceneComponentProps<SceneAppPage>) {
...
-          return page;
-        }}
-      ></Route>
-
-      {/* drilldown routes */}
-      {routes}
-    </Switch>
+    <>
+      <Switch>{routes}</Switch>
+      {currentPageIsRouteMatch && page}
```
dprokop, line 107: Having hard time understanding this. Why is the page rendered outside switch? Shouldn't switch handle rendering it?

dprokop (commented):
All in all, the implementation looks good, yet not very straightforward. I understand this is tricky. I've left one question regarding the routes rendering, I got quite confused by it.

packages/scenes/src/components/SceneApp/SceneAppPage.tsx
torkelo: @dprokop it's not outside a switch really, the page is already wrapped in a Switch and Route (by the parent SceneAppPAge or SceneApp). 

It's just that we did not really know here before if we are on a leaf node that matches the current route or not so we added an Extra Switch + Route here to let react-router handle that. But that is unnessary. 

Its a bit tricky to explain. But think of what happens when a SceneAppPage with tabs is rendered

We go into this condition and render a Switch and Route for each tab. Then react-router will render the correct matching tab page for the current url. So we will go into an other SceneAppPage render (for the tab). In this PR I made is so we can skip an extra Switch + Route when that last SceneAppPage (for the tab) is rendered as we can figure that out by just checking the match url. 
image

torkelo:
```
<>
    <Switch>{routes}</Switch>
    {currentPageIsRouteMatch && page}
<>
 {currentPageIsRouteMatch && page}
```

So the challenge here in this part of the code is knowing "Am I a leaf node that should render my own page" if not then one of the drilldowns added to routes will probably match. So in a sense we actually do this here :

```
if (currentPageIsRouteMatch) {
   return page;
}
return  <Switch>{routes}</Switch>; 
```

Because if the current page is an exact match then the drill-downs will not be needed!

dprokop: Oh yeah, it makes sense now withe the latest changes. Think I had issue getting this right because of the Switch as it technically should not have any effect before your latest change.

dprokop (approved):

PR #72, opened 2023-03-16
SceneObject: subscribeToState signature change and implementation change

torkelo:
Alternative to https://github.com/grafana/scenes/pull/66
Closes #66 

* Add prevState to subscribeToState handler
* Simplify subscribeToState argument to be a handler callback to just receive state (and prevState) as separate args
* Reuse the existing event and eventbus instead of having separate subject

### DIFF
--- a/packages/scenes/src/core/SceneObjectBase.tsx
+++ b/packages/scenes/src/core/SceneObjectBase.tsx
@@ -3,7 +3,13 @@ import { Observer, Subject, Subscription, SubscriptionLike, Unsubscribable } fro
 import { v4 as uuidv4 } from 'uuid';
 
 import { BusEvent, BusEventHandler, BusEventType, EventBusSrv } from '@grafana/data';
-import { SceneObject, SceneComponent, SceneObjectState, SceneObjectUrlSyncHandler } from './types';
+import {
+  SceneObject,
+  SceneComponent,
+  SceneObjectState,
+  SceneObjectUrlSyncHandler,
+  SceneStateChangedHandler,
+} from './types';
 import { useForceUpdate } from '@grafana/ui';
 
 import { SceneComponentWrapper } from './SceneComponentWrapper';
@@ -15,7 +21,6 @@ export abstract class SceneObjectBase<TState extends SceneObjectState = SceneObj
   implements SceneObject<TState>
 {
   private _isActive = false;
-  private _subject = new Subject<TState>();
   private _state: TState;
   private _events = new EventBusSrv();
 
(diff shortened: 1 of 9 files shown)


packages/scenes/src/core/SceneObjectBase.tsx
```diff
@@ -82,8 +86,12 @@ export abstract class SceneObjectBase<TState extends SceneObjectState = SceneObj
...
    * Subscribe to the scene state subject
    **/
-  public subscribeToState(observerOrNext?: Partial<Observer<TState>>): SubscriptionLike {
-    return this._subject.subscribe(observerOrNext);
+  public subscribeToState(handler: SceneStateChangedHandler<TState>): Unsubscribable {
+    return this._events.subscribe(SceneObjectStateChangedEvent, (event) => {
+      if (event.payload.changedObject === this) {
+        handler(event.payload.newState as TState, event.payload.prevState as TState);
+      }
+    });
```
torkelo, line 94: I really like that this simplifies SceneObjectBase implementation (removes rxjs subject) and reuse existing eventbus and the state changed event. 

The only downside is the potential performance impact, where this event bubbles up (hence we need the filter to filter out only events that originate from this instance). So for say a dashboard with a lots of panels and streaming data this  handler is called many times (but with the `changedObject === this` filter to ignore all of them). But if this ever becomes a problem (I sort of doubt it), we can add an option to disable bubbling state change events for some objects (Like SceneQueryRunner)

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -64,10 +64,8 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
     const timeRange = sceneGraph.getTimeRange(this);
 
     this._subs.add(
-      timeRange.subscribeToState({
-        next: (timeRange) => {
-          this.runWithTimeRange(timeRange.value);
-        },
+      timeRange.subscribeToState((timeRange) => {
+        this.runWithTimeRange(timeRange.value);
       })
```
torkelo, line 69: I like the simpler more readable signature for arguments to subscribeToState (just a simple function that get's the new state), instead of the object with a next function

packages/scenes/src/core/types.ts
```diff
@@ -13,6 +13,7 @@ import {
 } from '@grafana/data';
 
 import { SceneVariableDependencyConfigLike, SceneVariables } from '../variables/types';
+import { SceneObjectStateChangedEvent } from './events';
```
dprokop, line 16: Think this import is not used

dprokop (approved):
I think this makes sense. Also, doubt the performance impact will be relevant here. Actually it seems like it's gonna benefit debugging with rx Subject out of the picture.

PR #75, opened 2023-03-16
Typecheck on dev and pr check

dprokop:
Adds https://github.com/fi3ework/rollup-plugin-fork-ts-checker to perform typecheck when developing scenes. Note that this is a plugin that is both alpha and not maintained :D But I haven't found any other way to have reliable typecheck.

Modifies GH action to perform typecheck proior build.


torkelo (approved):

dprokop:
BTW - this PR SHOULD fail first!

PR #76, opened 2023-03-17
SceneQueryRunner: support empty/missing queries

ryantxu:
Text panels do not have queries (and a few others), but still hit the query runner path.

This lets the panels load rather than sitting with a "no data" message.  We may also want to bypass the query runner entirely, but it seems better to complete normally than to not show anything.

With these change https://github.com/grafana/grafana/pull/64936 now shows more panels OK:
image

Still failing with "-- Dashboard --" datasource... but this is progress!

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.ts
@@ -174,12 +174,11 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
 
     // Simple path when no queries exist
     if (!queries?.length) {
-      this._querySub = of({
+      this.onDataReceived({
         state: LoadingState.Done,
         series: [],
         timeRange,
-      }).subscribe(this.onDataReceived);
-      return;
+      });
     }
 
     const request: DataQueryRequest = {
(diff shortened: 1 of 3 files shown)


packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -228,6 +239,15 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
   };
 }
 
+function findFirstDatasource(targets: DataQuery[]): DataSourceRef | undefined {
+  for (const t of targets) {
+    if (t.datasource != null) {
+      return t.datasource;
```
ryantxu, line 245: i'm sure there is some functional 🤷

torkelo (commented):
@ryantxu this is a good change, I added two unit tests as well for these two scenarios. 

Yea been thinking about how to more efficiently deal with panels that don't care about data. as well as how scene app plugins can use VizPanel (or some other abstraction) but with a custom visualization, without registering a visualization plugin. 

https://github.com/grafana/scenes/pull/74

torkelo (approved):

dprokop (approved):
NIce!

PR #80, opened 2023-03-17
Variables: Missing variable props sent to custom variable formatters

torkelo:
* Adds missing props to the object we pass to custom variable formatters


packages/scenes/src/variables/interpolation/sceneInterpolator.ts
```diff
@@ -81,6 +81,8 @@ function formatValue(
     return formatNameOrFn(value, {
       name: variable.state.name,
       type: variable.state.type as VariableType,
+      multi: variable.state.isMulti,
+      includeAll: variable.state.includeAll,
```
dprokop, line 85: 🤦

dprokop (approved):

PR #82, opened 2023-03-17
DataSourceVariable: Fix state property name to be more descriptive

torkelo:
Changes name of the query property to pluginId. The query name is only from the persisted state model, and the reason it's named query there is very old and becuse the data source variable model shared it's model definition with query variable model.

For this new object we should not be use this old name and use a more descriptive name.

like:

* pluginId

or

* dataSourceType

### DIFF
--- a/packages/scenes/src/variables/variants/DataSourceVariable.tsx
+++ b/packages/scenes/src/variables/variants/DataSourceVariable.tsx
@@ -13,7 +13,7 @@ import { MultiValueVariable, MultiValueVariableState, VariableGetOptionsArgs } f
 
 export interface DataSourceVariableState extends MultiValueVariableState {
   /**
-   * Include all data source instance of this plugin type
+   * Include all data source instances with this plugin id
    */
   pluginId: string;
   /**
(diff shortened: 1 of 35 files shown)


dprokop (approved):
Think this requires changelog update too, some people are already using this.

packages/scenes/src/variables/variants/DataSourceVariable.tsx
```diff
@@ -12,7 +12,13 @@ import { VariableValueOption } from '../types';
 import { MultiValueVariable, MultiValueVariableState, VariableGetOptionsArgs } from './MultiValueVariable';
 
 export interface DataSourceVariableState extends MultiValueVariableState {
-  query: string;
+  /**
+   * Include all data source instance of this plugin type
```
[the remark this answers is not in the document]
torkelo: @dprokop yea, maybe it should read "Include all data source instances with this plugin id"

torkelo, line 16:
```suggestion
   * Include all data source instances with this plugin id
```

dprokop: Yes, better. Also - sorry I also misunderstood the change when I added this comment, thought it's about Query variable, and it was not :)

It makes sense to change this tho.

PR #84, opened 2023-03-18
SceneObject: Remove interfaces and props for editing

torkelo:
The editing interfaces and props felt a bit too specific to an unknown future editing mode.

But I still want to support something like that edit mode and support wrapping objects in something like an edit wrapper.

By adding a componentWrapper property to SceneObject I was able to build a similar edit mode with edit wrappers around each of the layout objects.

https://github.com/grafana/grafana/pull/64996

### DIFF
--- a/packages/scenes/src/components/NestedScene.tsx
+++ b/packages/scenes/src/components/NestedScene.tsx
@@ -45,7 +45,7 @@ export class NestedScene extends SceneObjectBase<NestedSceneState> {
   };
 }
 
-export function NestedSceneRenderer({ model, isEditing }: SceneComponentProps<NestedScene>) {
+export function NestedSceneRenderer({ model }: SceneComponentProps<NestedScene>) {
   const { title, isCollapsed, canCollapse, canRemove, body, actions } = model.useState();
   const styles = useStyles2(getStyles);
 
@@ -85,7 +85,7 @@ export function NestedSceneRenderer({ model, isEditing }: SceneComponentProps<Ne
         </Stack>
         <div className={styles.actions}>{toolbarActions}</div>
       </div>
-      {!isCollapsed && <body.Component model={body} isEditing={isEditing} />}
+      {!isCollapsed && <body.Component model={body} />}
     </div>
   );
 }
--- a/packages/scenes/src/core/types.ts
+++ b/packages/scenes/src/core/types.ts
@@ -18,7 +18,6 @@ export interface SceneObjectStatePlain {
   key?: string;
   $timeRange?: SceneTimeRangeLike;
   $data?: SceneDataProvider;
-  $editor?: SceneEditor;
   $variables?: SceneVariables;
 }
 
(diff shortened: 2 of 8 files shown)


packages/scenes/src/core/types.ts
```diff
@@ -68,6 +72,12 @@ export interface SceneObject<TState extends SceneObjectState = SceneObjectState>
   /** This abstraction declares URL sync dependencies of a scene object. **/
   readonly urlSync?: SceneObjectUrlSyncHandler<TState>;
 
+  /**
+   * Component wrapper, can be used to for edit modes, to wrap objects in an edit wrappers. This component wrapper will
+   * be used for all children of this object (unless they specify their own wrapper).
+   **/
+  readonly componentWrapper?: SceneComponentCustomWrapper;
+
```
torkelo, line 80: I am inclined to remove this altogether, I think we can build edit modes instead by walking the object graph and wrapping specific objects in scene object edit wrappers. Having an edit wrapper react component that will be used for all children is problematic as I think some scene objects will be composed of multiple scene objects, but from an editing perspective should be treated as one object. So having a component wrapper that is to be used for all children will be pretty limited (and require overriding)

packages/scenes/src/components/NestedScene.tsx
```diff
@@ -85,7 +85,7 @@ export function NestedSceneRenderer({ model, isEditing }: SceneComponentProps<Ne
         </Stack>
         <div className={styles.actions}>{toolbarActions}</div>
       </div>
-      {!isCollapsed && <body.Component model={body} isEditing={isEditing} />}
```
torkelo, line 88: the point of isEditing to SceneComponentProps was to have a way to trigger re-render without modifying state of every object. But I think we can trigger a re-render of the whole tree by simple wrapping a top level object (say the first layout) in a edit wrapper object. This will cause an unmount & remount (and hence a full tree re-render)

torkelo: Example: 
https://github.com/grafana/grafana/pull/64996/files#diff-3edc88a9fe87847c8b2e9206e1ba45b92c988bc3edb66a70b77a3af4d279c285R58

dprokop: so about this example, don't you think the editing should be handled OUTSIDE of the DashboardScene?

I mean, when editing dashboard scene I can imagine we have a separate class that does sth like this:

```tsx
const scene = xxx;
new SceneEditWrapper({
  scene: xxx
  ...
});
```

And the SceneEditWrapper is responsible for wrapping tree objects by it's own edit components. Rendering standard editors or custom ones (if provided by the object).

torkelo: @dprokop no, as you can see in the example, it's handled inside DashboardScene (it adds the edit wrapper, that will be custom for DashboardScene and add maybe some edit controls at the top, and wrap layout childs in a edit wrapper that allow you to select / resize them etc.

packages/scenes/src/core/types.ts
torkelo: @dprokop i think it's too early to say how editing will work, but I think it's pretty clear that edit UX will not be strictly linked to scene runtime tree, what I mean is the edit components will likely be different from the scene objects that they edit, maybe some scene objects provide an interface to get options registries (like panel and options) but I don't think each scene object will have an "Editor" react component that shows it's edit UI

dprokop: > but I don't think each scene object will have an "Editor" react component that shows it's edit UI

Yes, I agree. But - IMO there's probably gonna be a need for such an optional property of an object, to handle non-trivial edit cases.

dprokop (approved):

PR #86, opened 2023-03-21
SceneInterpolator: Support "__all_variables"

domasx2:
While working on app-o11y traces view I released we need `__all_variables` support for table data links, to presverve selected data sources. I doodles this, does it make sense?


domasx2:
Closing in favor of #98

PR #96, opened 2023-03-22
UrlSyncManager: Improvements and fixes

torkelo:
@domasx2 identified some issues with URL sync when it comes to clearing / restoring state (one of the more complex parts of url sync) 

* Restoring state if the initial state was undefined did not work 
* Coming back to a scene did not remember the initialStates (that is used to restore state when url does not contain the state), this required a change so that we do not new up a new UrlSyncManager every initUrlSync but instead preserve the same instance for as long as that scene exists.
* Also updated the tests to more accurately reflect usage (initUrlSync called, then scene is activated)

### DIFF
--- a/packages/scenes/src/services/UrlSyncManager.test.ts
+++ b/packages/scenes/src/services/UrlSyncManager.test.ts
@@ -5,22 +5,23 @@ import { locationService } from '@grafana/runtime';
 import { SceneFlexLayout } from '../components/layout/SceneFlexLayout';
 import { SceneObjectBase } from '../core/SceneObjectBase';
 import { SceneTimeRange } from '../core/SceneTimeRange';
-import { SceneLayoutChildState, SceneObjectUrlValues } from '../core/types';
+import { SceneLayoutChildState, SceneObject, SceneObjectUrlValues } from '../core/types';
 
 import { SceneObjectUrlSyncConfig } from './SceneObjectUrlSyncConfig';
 import { isUrlValueEqual, UrlSyncManager } from './UrlSyncManager';
 
 interface TestObjectState extends SceneLayoutChildState {
   name: string;
+  optional?: string;
   array?: string[];
   other?: string;
 }
 
 class TestObj extends SceneObjectBase<TestObjectState> {
-  protected _urlSync = new SceneObjectUrlSyncConfig(this, { keys: ['name', 'array'] });
+  protected _urlSync = new SceneObjectUrlSyncConfig(this, { keys: ['name', 'array', 'optional'] });
 
   public getUrlState(state: TestObjectState) {
-    return { name: state.name, array: state.array };
+    return { name: state.name, array: state.array, optional: state.optional };
   }
 
   public updateFromUrl(values: SceneObjectUrlValues) {
@@ -30,6 +31,9 @@ class TestObj extends SceneObjectBase<TestObjectState> {
     if (Array.isArray(values.array)) {
       this.setState({ array: values.array });
     }
+    if (values.hasOwnProperty('optional')) {
+      this.setState({ optional: typeof values.optional === 'string' ? values.optional : undefined });
+    }
   }
 }
 
@@ -37,6 +41,7 @@ describe('UrlSyncManager', () => {
   let urlManager: UrlSyncManager;
   let locationUpdates: Location[] = [];
   let listenUnregister: () => void;
+  let scene: SceneObject;
 
   beforeEach(() => {
     locationUpdates = [];
@@ -46,7 +51,7 @@ describe('UrlSyncManager', () => {
   });
 
   afterEach(() => {
-    urlManager.cleanUp();
+    scene.deactivate();
     locationService.push('/');
     listenUnregister();
   });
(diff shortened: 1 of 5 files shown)


domasx2 (changes requested):
I tried it, some new issues happen:

- On initial load data sources have defaults selected, page works, but data source variables are no longer populated to URL
- Descending into drill down populates url with empty data source variables, data sources become unselected and page breaks
- Bonus, also reproducible on main brach: On initial load data source vars have selected values, but clicking root item in breadcrumbs populates url with empty data source variables and unselects them

https://user-images.githubusercontent.com/847684/226882112-ce5b45db-132d-402a-adf9-7a3d468e9a6c.mp4

torkelo:
@domasx2 thanks for testing more, your drilldown link looks to contain has a `var-prometheus=`  , which is why it's blank on the drilldown view. So it looks like a bug in the link building?

torkelo:
@domasx2 
>  but data source variables are no longer populated to URL
If works for me, if you change a variable does not not get populated to the url? State -> Url only happens when you change the state (ie select a new variable value)

domasx2:
> @domasx2 thanks for testing more, your drilldown link looks to contain a `var-prometheus=` , which is why it's blank on the drilldown view. So it looks like a bug in the link building?

The link we build has no query params, so I think `var-prometheus=` is set by scenes  
image

> If works for me, if you change a variable does not not get populated to the url? State -> Url only happens when you change the state (ie select a new variable value)

torkelo:
@domasx2 ok, if your link is not preserving variables, they will be cleared to their initial state. But think that is a separate issue. 

The problem I am seeing is that variables are initially have no value, then get value (synced to url). But then when you move back to it without them in URL they get get reset to no value. 

Think this is very hard problem, (to know when a blank url state should clear the current state). Need to think more on this

domasx2:
> @domasx2 ok, if your link is not preserving variables, they will be cleared to their initial state. But think that is a separate issue. 

To be clear, the link here preserves query params, it just does not insert `var-prometheus=` and etc if it wasn't present in url to begin with.

**On main branch:** On initial load data  ource variables have a default selected and are synced to the URL. Drilldown link copies query params, it works

**On this branch:** On initial load data source variables have a default selected but the default is not synced to the URL. Thus drilldown link has no query params, drilldown opens with empty data soruce variables and somehow empty values are then synced the URL,

It seems on initialization data source variables select first item from the list if available, but when being re-activated, reset to none if nothing is in URL?

torkelo:
@domasx2 Hm.. the inital values for the variables is en empty string, so this check https://github.com/grafana/scenes/blob/main/packages/scenes/src/variables/variants/MultiValueVariable.ts#L278 does not work (causing them to be cleared when they are not present in URL)

domasx2:
Hm, gonna try explicitly specifying a default data source

torkelo:
@domasx2 yea, going to see how the system can know what what state to preserve and what to clear when value is not present in url, super hard. Also very hard with variables going through quick changes as scenes load causing lots of url syncs

torkelo:
@domasx2 pushed a big update, I had to remove a big feature or UrlSyncManager of being able to restore to an initial value in case the url keys are removed. Now this is passed to updateFromUrl with key values set to null, so it's up to each scene object for how to handle missing url keys, so for time range for example it will just ignore the updateFromUrl if there are not time range keys. But your drilldown will definitely care about this null value. 

This seems to solve all issues. There is an unrelated issue if you have the same variable on multiple scenes, and you select a value on the variable (get's added to URL), then go another app/page (different app), then open the scene app page again, the variable is going to have the same value, but nothing in url state, so when you move to a different scene (via tab or drilldown) then that variable is not going to be part of the url, and that other scene is not going initialize the variable to the same value (because it's not part of url). 

But I think this is a different problem, one we can solve by sharing / caching / restoring state between shared (identical) variables between different scenes. 

@dprokop ^

packages/scenes/src/services/UrlSyncManager.test.ts
```diff
@@ -95,11 +105,11 @@ describe('UrlSyncManager', () => {
...
       expect(obj.state.name).toBe('test2');
 
-      // When relevant key is cleared (say go back)
-      locationService.getHistory().goBack();
+      // When relevant key is cleared
+      locationService.partial({ name: null });
 
       // Should revert to initial state
-      expect(obj.state.name).toBe('test');
+      // expect(obj.state.name).toBe('test');
```
torkelo, line 112: so this is the behavior change, before the UrlSyncManager was able to restore an initial state that it would give the scene object in case the url key was removed. Now this needs to be handled in each scene object

dprokop: think this deserves a note in Changelog with an example snippet what does "handled in each scene object" mean.

dprokop: @torkelo qq - so this basically now needs to be handled using `updateFromUrl`, right?

domasx2 (approved):
re-tested, now works well :+1:

PR #97, opened 2023-03-22
Layout: Create atomic, layout specific objects

dprokop:
This explores an idea of separating `placement` property from all scene objects, and having layout-specifix objects to hold placement configuration. 

First batch of changes is about flex layout. A new `SceneFlexItem` object is introduced. Anything that wants to be rendered via `SceneFlexLayout` needs to be wrapped in this scene object, and it's enforced via SceneFlexLayout types (`children: SceneFlexItem[]`). SceneFlexItem is simply proxying flex css atributes to rendered `div`, so technically it's an ultra dummy object. There is no "magic" in the SceneFlexLayout anymore that configures it's children properties depending on provided placement cnfiguration. It needs to be set up manually, and IMO this is better, as it's faimliar to using browser's web flex configuration, rather than relying on the framework provided assumptions. 

cc @ryantxu 
- [x] SceneFlexLayout 
- [x] SceneGridLayout
- [x] Try figuring out a single interface (using `SceneGridItemLike` for SceneGridLayout and SceneGridRow children property). Currently I'm struggling to get that work nicely with the API.
- [x] Update ts issues in tests.
- [x] Prepare release notes

## Release Notes

The interface of `SceneFlexLayout` and `SceneGridLayout` has changed. These scene objects now accept only dedicated layout item objects as children:
- `SceneFlexItem` for `SceneFlexLayout`
- `SceneGridItem` and `SceneGridRow` for `SceneGridLayout`

`placement` property has been replaced by those layout-specific objects. 

Example
```tsx
// BEFORE
const layout = new SceneFlexLayout({
  direction: 'column',
  children: [
    new VizPanel({
      placement: {
        width: '50%',
        height: '400',
     },
     ...
    })
  ],
  ...
})


// AFTER
const layout = new SceneFlexLayout({
  direction: 'column',
  children: [ 
    new SceneFlexItem({ 
      width: '50%',
      height: '400',
      body: new VizPanel({ ... }),
    }),
  ],
  ...
})

```


packages/scenes/src/components/layout/SceneFlexLayout.tsx
```diff
@@ -3,21 +3,59 @@ import React, { CSSProperties } from 'react';
...
+    flexGrow: rest.flexGrow || 0, // 0 = default?
+    flexShrink: rest.flexShrink, // 1 = default?
+    alignSelf: rest.alignSelf || 'auto',
+    flexBasis: rest.flexBasis || 'auto',
+    width: rest.width || 'auto',
+    height: rest.height || 'auto',
+    minWidth: rest.minWidth || 0,
+    minHeight: rest.minHeight || 0,
+    maxWidth: rest.maxWidth,
+    maxHeight: rest.maxHeight,
```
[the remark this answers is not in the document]
torkelo: @dprokop yea, a flexGrow default feels more natural for scenes

torkelo:
>. There is no "magic" in the SceneFlexLayout anymore that configures it's children properties depending on provided placement cnfiguration. It needs to be set up manually, and IMO this is better, as it's faimliar to using browser's web flex configuration, rather than relying on the framework provided assumptions.

Not sure what you mean, those behaviors where copied from figma and I think are pretty important for ease of use

packages/scenes/src/components/layout/SceneFlexLayout.tsx
```diff
@@ -36,82 +74,24 @@ export class SceneFlexLayout extends SceneObjectBase<SceneFlexLayoutState> imple
...
-    } else {
-      style.alignSelf = ySizing === 'fill' ? 'stretch' : 'flex-start';
-    }
-
-    if (layout.width) {
-      style.width = layout.width;
-    } else {
-      style.flexGrow = xSizing === 'fill' ? 1 : 0;
-    }
-  }
```
torkelo, line 110: Any reason for removing this behavior? it simplifies it a lot

torkelo (commented):
I find it strange that SceneFlexItem has a children array, should it not just wrap a single object? (being an item?). 

Otherwise, it feels like we don't need SceneFlexItem and can just use SceneFlexLayout. 

This makes me a bit doubtful if this is a good idea, becomes so hard to read scene definitions

packages/scenes-app/src/pages/Demos/scenes/flexLayout.tsx
```diff
@@ -0,0 +1,140 @@
...
+                    new VizPanel({
+                      pluginId: 'timeseries',
+                      title: 'Dynamic height and width',
+                      $data: getQueryRunnerWithRandomWalkQuery({}, { maxDataPointsFromWidth: true }),
+                    }),
+                  ],
+                }),
+                new SceneFlexItem({
+                  flexGrow: 1,
+                  children: [
```
torkelo, line 36: feels like having this is be an array breaks a lot of things (spacing between layout objects), and is a bit confusing

packages/scenes/src/components/layout/SceneFlexLayout.tsx
torkelo: Maybe we can simple SceneFlexLayoutItem that has more of a simple figma positioning system, and more advanced raw css flex layout item?

packages/scenes-app/src/pages/Demos/scenes/flexLayout.tsx
dprokop: Yes, I 100% agree

packages/scenes/src/components/layout/SceneFlexLayout.tsx
dprokop: @torkelo I brought this default styling back. Since it's dependant on parent's direction, the renderer looks for parent and ensures SceneFlexItem ain't rendered in a layout other than SceneFlexLayout. Ref https://github.com/grafana/scenes/pull/97/commits/3738ddde21a6290cb6313d488060aeea68400b66#diff-61d9164824772b8e346981c774be75c8cb469ce1a174f4ecc6078908b4b2f9daR46

packages/scenes/src/components/layout/SceneFlexLayout.tsx
```diff
@@ -36,96 +73,80 @@ export class SceneFlexLayout extends SceneObjectBase<SceneFlexLayoutState> imple
...
-    <div style={getItemStyles(direction, placement)}>
-      <item.Component model={item} />
-    </div>
+    <Field label="Direction">
+      <RadioButtonGroup
+        options={options}
+        value={direction}
+        onChange={(value) => model.setState({ direction: value as CSSProperties['flexDirection'] })}
+      />
+    </Field>
```
torkelo, line 108: you can delete this, not used anywhere

packages/scenes/src/core/types.ts
```diff
@@ -25,6 +25,10 @@ export interface SceneLayoutChildState extends SceneObjectStatePlain {
   placement?: SceneLayoutChildOptions;
 }
 
+export interface SceneLayoutItem extends SceneObjectStatePlain {
+  child: SceneObject;
```
torkelo, line 29: torn on child or body is the best naming here (as we use body in a few other places for the "content" object)

torkelo (commented):
Think this looks good, and feels like the right thing to do.

packages/scenes/src/components/layout/SceneFlexLayout.tsx
```diff
@@ -5,19 +5,56 @@ import { Field, RadioButtonGroup } from '@grafana/ui';
...
 } from '../../core/types';
 
-export type FlexLayoutDirection = 'column' | 'row';
-
-interface SceneFlexLayoutState extends SceneLayoutState {
-  direction?: FlexLayoutDirection;
+interface SceneFlexLayoutState extends SceneObjectStatePlain {
+  direction?: CSSProperties['flexDirection'];
   wrap?: CSSProperties['flexWrap'];
+  children: SceneFlexItem[];
```
torkelo, line 17: Maybe this should be SceneFlexItemLike (and interface), to allow custom SceneFlexItems ?

packages/scenes/src/core/types.ts
dprokop: Using `body` now. Think it makes sense to be consitent with `body` for objects that accept single scene object, and `children` for the ones accepting arrays.

torkelo (approved):

PR #98, opened 2023-03-23
Scene: Support for new types of "macro" variables starting with __all_variables

torkelo:
* Builds on https://github.com/grafana/scenes/pull/86  but makes it more general so we can easily support more macro variables similar to this 
* Adds unit tests 

have another an idea for another macro variable that gives you option to specify which variables to include, and another one that instead of checking variables checks url state and gives you option of preserving current all or some current query params. We also need to add the keepTime macro (__url_time_range). 

* New macro abstraction
* Better / more solid concept / abstraction for custom variable value that want to handle formatting themselves 
* Support for custom macros that can take arguments


torkelo:
> Minor inconsistency, possibly be design: variables with default values, that are not included in location query params, are included in the macro variable value
@domasx2 that is because we don't add the default (initial) state to url when the scene loads

packages/scenes/src/variables/macros/AllVariablesMacro.ts
```diff
@@ -0,0 +1,75 @@
...
+ * The sceneInterpolator will detect if getValue returns VariableValueCustom and will skip the normal formatting
+ * This is useful as otherwise we would url encode macros like $__all_variables twice.
+ */
+class SkipFormattingValue implements CustomVariableValue {
+  public constructor(private _value: string) {}
+
+  public formatter(): string {
+    return this._value;
+  }
+}
```
dprokop, line 75: @torkelo  This is implemented twice, here and in URLTimeRangeMacro.

torkelo: @dprokop good catch, think they could drift a bit if we add args to these macros but when they should shared as long as they are identical

PR #99, opened 2023-03-23
ScenesApp: Change to workspace dependency

torkelo:


yarn.lock
```diff
@@ -17398,7 +17398,7 @@ __metadata:
     "@grafana/e2e-selectors": 9.2.1
     "@grafana/eslint-config": 5.0.0
     "@grafana/runtime": canary
-    "@grafana/scenes": ^0.0.28
+    "@grafana/scenes": "workspace:*"
```
dprokop, line 17401: When running `lerna version` whis is not updated. But I think it should be good as we do not publish the app anywhere.

dprokop (approved):

PR #100, opened 2023-03-23
UrlSync: Simplify url sync interface

torkelo:
After the changes in https://github.com/grafana/scenes/pull/96 there is now technically no need for the getUrlState function to pass in state, but it could instead always use current state. The reason for it passing in state was to be able to pass in old state so we could remember the initial state and restore it when url state was removed. 

Burt all of that is no longer needed, so I think it makes sense to simplify this interface and have get just be getUrlState. 

There is a small argument that this could potentially be useful at some point and adding it later could be painful 

## Release Notes

The  SceneObjectUrlSyncHandler interface has changed. The function `getUrlState` no longer takes state as parameter. The implementation needs to use the current scene object state instead.

### DIFF
--- a/packages/scenes/src/components/layout/SceneGridRow.tsx
+++ b/packages/scenes/src/components/layout/SceneGridRow.tsx
@@ -57,8 +57,8 @@ export class SceneGridRow extends SceneObjectBase<SceneGridRowState> {
     this.getGridLayout().toggleRow(this);
   };
 
-  public getUrlState(state: SceneGridRowState) {
-    return { rowc: state.isCollapsed ? '1' : '0' };
+  public getUrlState() {
+    return { rowc: this.state.isCollapsed ? '1' : '0' };
   }
 
   public updateFromUrl(values: SceneObjectUrlValues) {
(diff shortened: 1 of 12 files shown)


dprokop (approved):
Looks good @torkelo.

> There is a small argument that this could potentially be useful at some point and adding it later could be painful

Not sure why this would be useful if it's evaluated against an instance that can access it's own state?

torkelo:
> Not sure why this would be useful if it's evaluated against an instance that can access it's own state?

@dprokop well it could be if we want to get the url state for an old state (ie not the current), that was why the function took in state. But the way url sync works now we will likely never want to know the url state of some past state

dprokop:
> @dprokop well it could be if we want to get the url state for an old state (ie not the current), that was why the function took in state. But the way url sync works now we will likely never want to know the url state of some past state

Yeah, thought about that. Actually, I think this change is very good, feels like adding a possibility to inspect the previous state and react accordingly in the URL is a dangerous path, opening a path to introducing a lot of complexity in URL management.

PR #103, opened 2023-03-23
DataLinks: DataValueMacro for __value.* evaluation

torkelo:
Was so excited how the macros  __all_variables and __url_time_range can now be implemented AND lazily evaluated (instead of how they are ALWAYS evaluated for all data link string interpolations) https://github.com/grafana/grafana/blob/main/packages/grafana-data/src/field/fieldOverrides.ts#L423

Was really excited to see how we could implemented the other data link variables there in a similar lazy way using the same abstraction, starting with the  

* $__value.raw
* $__value.numeric
* $__value.text
* $__value
* $__value.time

All these are also evaluated for every datalink interpolation and passed as scopedVars https://github.com/grafana/grafana/blob/main/packages/grafana-data/src/field/fieldOverrides.ts#L392

Instead this PR propose that we instead pass DataContext via scopedVars 
* field: Field
* valueRowIndex?: number
* timeField?  (optional) 

with this we can implement all the above value "macros" and only evaluate exactly what we need depending on the expression being evaluated! 

This needs a small change in core to set this dataContext (and ignore setting the normal value scopedVars when in a scene context. that should be pretty straight forward)

### DIFF
--- a/packages/scenes/src/variables/macros/DataValueMacro.test.ts
+++ b/packages/scenes/src/variables/macros/DataValueMacro.test.ts
@@ -1,0 +1,43 @@
+import { FieldType, toDataFrame } from '@grafana/data';
+import { DataContextScopedVar } from '../macros/DataValueMacro';
+import { sceneInterpolator } from '../interpolation/sceneInterpolator';
+import { TestScene } from '../TestScene';
+
+describe('DataValueMacro', () => {
+  it('Can use use ${__value.*} interpolation when dataContext exist', () => {
+    const scene = new TestScene({});
+    const data = toDataFrame({
+      name: 'A',
+      fields: [
+        {
+          name: 'number',
+          type: FieldType.number,
+          values: [5, 10],
+          display: (value: number) => {
+            return { text: value.toString(), numeric: value, suffix: '%' };
+          },
+        },
+        {
+          name: 'time',
+          type: FieldType.time,
+          values: [5000, 10000],
+        },
+      ],
+    });
+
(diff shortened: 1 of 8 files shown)


packages/scenes/src/variables/macros/DataValueMacro.test.ts
```diff
@@ -0,0 +1,43 @@
...
+          values: [5000, 10000],
+        },
+      ],
+    });
+
+    const dataContext: DataContextScopedVar = {
+      value: {
+        field: data.fields[0],
+        timeField: data.fields[1],
+        valueRowIndex: 1,
```
torkelo, line 32: maybe a better structure is 
* frame: DataFrame
* fieldIndex?: number
* rowIndex?: number

We can look for the time field frame for `${__value.time}`

packages/scenes/src/variables/types.ts
```diff
@@ -104,6 +104,20 @@ export type VariableCustomFormatterFn = (
...
+ * Used by data link macros
+ */
+export interface DataContextScopedVar {
+  value: {
+    frame: DataFrame;
+    field: Field;
+    valueIndex?: number;
+    calculatedValue?: DisplayValue;
+  };
+}
```
dprokop, line 123: Can we use the type from @grafana/data? Changes in this PR are anyways dependant on latest Grafana version.

dprokop (approved):
Looks great, this nicely proves how versatile the `FormatVariable` interace is.

PR #104, opened 2023-03-23
Fix type import

torkelo:
I am really puzzled why the build on main is green


torkelo:
ah, my bad. I had missed saving part of this file in https://github.com/grafana/scenes/pull/101

PR #111, opened 2023-03-28
Query: support listening to non-parent nodes for data

ryantxu:
Trying to implement the existing "Dashboard Query" pattern we have in dashboards -- we need to find another panel and subscribe to the changes from there.

See https://github.com/grafana/grafana/pull/64936

This is my most naive attempt -- feedback welcome!


torkelo:
While an object like this makes some sense (we need a way to declare what properties are "references" to other scene object but not owned by "this" object) to keep the parent property intact when doing object like this. 

But for dashboard shared queries I think something like this: https://github.com/grafana/grafana/pull/65413 makes better sense. The logic becomes a bit more encapsulated. If we implemented shared query some other way we would need some other object that had the shared query model state and updated the scene accordingly when that changed.

PR #112, opened 2023-03-28
Interpolation: FormatRegistryID is now replaced by VariableFormatID from schema package

ryantxu:
This PR replaces the `FormatRegistryID` with `VariableFormatID` defined in schema package.

Motivation is to get the formats described consistently so we *can* have the same thing in backend macros

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1791,20 +1791,20 @@ __metadata:
   linkType: hard
 
 "@eslint-community/eslint-utils@npm:^4.2.0":
-  version: 4.2.0
-  resolution: "@eslint-community/eslint-utils@npm:4.2.0"
+  version: 4.4.0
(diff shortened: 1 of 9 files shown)


yarn.lock
ryantxu, line 1: This is the result of deleting yarn.lock and then recreating all of it.  I think that is ok/good at this stage in the project :)

torkelo (approved):

torkelo:
@ryantxu 
it's strange that the HTML format ouput changed, found it's because of https://github.com/grafana/grafana/pull/62787/files#r1150058913

yarn.lock
dprokop: Yes, it's fine. Happening all the time (given `canary` dependencies there)

Issue #102, opened 2023-03-23
Geomap panel crashes on rerender


rdubrock:
I'm not 100% sure this is a problem with scenes itself, but I think it might be

To reproduce:

- Add geomap panel to a scene with two tabs
- Click on a tab without the geomap panel
- Click back to the tab with the geomap panel

```
TypeError: can't access property "setActive", this.mouseWheelZoom is undefined
    initControls GeomapPanel.tsx:336
    optionsChanged GeomapPanel.tsx:161
    shouldComponentUpdate GeomapPanel.tsx:106
```    

![geomap_bug](https://user-images.githubusercontent.com/8377044/227284581-dcba58a1-6277-4aad-bd9e-a54156b8f2af.gif)

dprokop:
@rdubrock thanks for reporting this. I have managed to reproduce this easily. 

I think I know where the problem is. It can be perceived either a scenes problem or geomap problem.

---

So what happens when you change tabs back and forth is that we are re-using scenes that are cached. When navigating back to the scene that once has been activated, we are only activating it back and initializing the options that are later on passed down to the particular panel. This happens here: https://github.com/grafana/scenes/blob/main/packages/scenes/src/components/VizPanel/VizPanel.tsx#L86. 

Please note, that the options object is a result of the underlying `lodash.merge` call, which will create a new **options object**.

Now, what happens down the road. In the Geomap panel, there is a `shouldComponentUpdate` called that simply compares current props.options and next props.options for referential equality (https://github.com/grafana/grafana/blob/main/public/app/plugins/panel/geomap/GeomapPanel.tsx#L108) and initializes map controls accordingly. 

The open question we need to discuss tho is if panels should rely on referential equality - and the framework should guarantee that, or rather perform a legitimate deep comparison of the options.

Issue #59, opened 2023-03-15
VizPanel legend onClick doesn't filter series


yesoreyeram:
When I have a timeseries panel in a scene with multiple series, clicking on the legend doesn't filter/highlight the series as we do in regular dashboard. 

In the below example, when clicking `medium3z3` i would like to see only that series instead it does nothing.

<img width="762" alt="image" src="https://user-images.githubusercontent.com/153843/225217601-66b9e3fa-12ed-4b28-bec2-99d826b3db61.png">

torkelo:
ah, good catch! we have to to hook these PanelContext handlers up

PR #113, opened 2023-03-29
VizPanel: Updates to support panel context 

torkelo:
For context, spent a long time on this, trying to find ways to refactor up VizPanel into two scene object one for PanelChrome (ScenePanel) that handles title and the panel chrome, and an inner scene object for visualization. But because the panel chrome is so entwined in the same data/state as the visualization this becomes really messy, both need data and time rnage, and the panel chrome and menu needs visualization context, etc

* Initial panel context 
* Adds support for visibility change
* Adds support for color change
* Adds support for legend sort toggle 
* [x] Tests 

Fixes https://github.com/grafana/scenes/issues/102 
Fixes https://github.com/grafana/scenes/issues/59


packages/scenes/src/components/VizPanel/seriesVisibilityConfigFactory.ts
```diff
@@ -0,0 +1,207 @@
...
+  getFieldDisplayName,
+  isSystemOverrideWithRef,
+  SystemConfigOverrideRule,
+} from '@grafana/data';
+import { SeriesVisibilityChangeMode } from '@grafana/ui';
+
+const displayOverrideRef = 'hideSeriesFrom';
+const isHideSeriesOverride = isSystemOverrideWithRef(displayOverrideRef);
+
+export function seriesVisibilityConfigFactory(
```
torkelo, line 19: his and the colorSeriesConfigFactory is coped from core lib, I don't want to move these to data. and I don't really want to expose them from scenes either. 

Maybe ok to have them duplicated during a period?

packages/scenes-app/src/pages/Home/Home.tsx
```diff
@@ -24,16 +24,25 @@ const getScene = () => {
...
-    {!config.datasources[DATASOURCE_REF.uid] && <Alert title={`Missing ${DATASOURCE_REF.uid} datasource`}>
-      These demos depend on <b>testdata</b> datasource: <code>{JSON.stringify(DATASOURCE_REF)}</code>.  
-      See <a href="https://github.com/grafana/grafana/tree/main/devenv#set-up-your-development-environment">https://github.com/grafana/grafana/tree/main/devenv#set-up-your-development-environment</a> for more details.
-    </Alert>}
+      {!config.datasources[DATASOURCE_REF.uid] && (
+        <Alert title={`Missing ${DATASOURCE_REF.uid} datasource`}>
+          These demos depend on <b>testdata</b> datasource: <code>{JSON.stringify(DATASOURCE_REF)}</code>. See{' '}
+          <a href="https://github.com/grafana/grafana/tree/main/devenv#set-up-your-development-environment">
+            https://github.com/grafana/grafana/tree/main/devenv#set-up-your-development-environment
+          </a>{' '}
```
torkelo, line 40: hm.. thse where not valid prettier formatted, maybe we are missing a "lint" check for prettier

packages/scenes-app/src/pages/Demos/scenes/panelContext.tsx
```diff
@@ -0,0 +1,78 @@
...
+    datasource: {
+      uid: 'gdev-testdata',
+      type: 'testdata',
+    },
+    queries: [
+      {
+        datasource: {
+          type: 'testdata',
+          uid: 'PD8C576611E62080A',
+        },
```
dprokop: Fixed that @torkelo

torkelo: @dprokop that is strange, this is the UID that all checked in gdev dashboards refer to.

dprokop: Yes, it is. My gdev-testdata ui is :
image

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -43,6 +54,9 @@ export class VizPanel<TOptions = {}, TFieldConfig = {}> extends SceneObjectBase<
 
   // Not part of state as this is not serializable
   private _plugin?: PanelPlugin;
+  private _panelContext: PanelContext;
+  private _prevData?: PanelData;
+  private _dataWithFieldConfig?: PanelData;
```
torkelo, line 59: opted to have these here rather than in state as a simple "cache", used by _onSeriesVisibilityChange change callback. Not having this in state simplifies things a lot. For example it makes onFieldConfig change much easier as those can now just trigger a state change (which triggers a render which triggers a call to applyFieldConfig from render). When I tried having data with field config in state it made onFieldConfig much more complex

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -131,4 +173,97 @@ export class VizPanel<TOptions = {}, TFieldConfig = {}> extends SceneObjectBase<
...
+   * Panel context functions
+   */
+  private _onSeriesColorChange = (label: string, color: string) => {
+    this.onFieldConfigChange(changeSeriesColorConfigFactory(label, color, this.state.fieldConfig));
+  };
+
+  private _onSeriesVisibilityChange = (label: string, mode: SeriesVisibilityChangeMode) => {
+    if (!this._dataWithFieldConfig) {
+      return;
+    }
```
dprokop, line 230: Can't figure out why this check happens only in this function?

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -131,4 +173,97 @@ export class VizPanel<TOptions = {}, TFieldConfig = {}> extends SceneObjectBase<
...
+    const plugin = this._plugin!;
+
+    if (!plugin || plugin.meta.skipDataQuery || !rawData) {
+      // TODO setup time range subscription instead
+      return emptyPanelData;
+    }
+
+    const fieldConfigRegistry = plugin.fieldConfigRegistry;
+
+    let structureRev = 0;
```
dprokop, line 190: So for this to work as it is right now, I think we need to move structureRev to instance rather than this function (basically `this._structureRev`). Otherwise structure rev would never change.

packages/scenes/src/components/VizPanel/seriesVisibilityConfigFactory.ts
dprokop: I think it's fine. 

Why don't you want to move it to data?

dprokop (approved):
Works great. Some nit comments I left, but other than that it's looking really good.

packages/scenes/src/components/VizPanel/VizPanel.tsx
dprokop: I see that, and I think it makes sense.

packages/scenes/src/components/VizPanel/seriesVisibilityConfigFactory.ts
torkelo: @dprokop well I see no usage outside scenes (and temporarily in core while we have both).

packages/scenes/src/components/VizPanel/VizPanel.tsx
torkelo: @dprokop yes! you are right. It worked here in a previous implementation as this did subscribe to data here and this was outside the subscribe function. But in this version this needs to be an instance property!

PR #114, opened 2023-03-29
SceneObject: Change how activate works and remove deactivate 

torkelo:
Turns activate into ref counting function, the actual activation function is now internal and private. 

activate will return a "deactivation callback" which when called will decrement the ref count and when 0 the internal deactivation function will be called. 

Beyond solving the problem identified here: https://github.com/grafana/grafana/pull/65413/files/7bc90f100bd8226409db92c2c4565c59d7a811d4#diff-fb19bc02d0a38819ae23ee5819cd89905e383a6276613411a1da8d23fe07fcadR55  

If we can also solve the "parent" problem this would enable sharing the same instance of say a time range or query runner in multiple (not connected by parents)  places in a scene 

* Make activate return a cancellation handler (Deactivate) 
* make deactivate private 

Potential problem:
* The function that activate returns could be called multiple times breaking the ref counting (Solved this problem by preventing/checking if function is called more than once) 

## Release notes 

You can no longer override the activate or deactivate functions in your custom scene objects. Instead, handle activation and deactivation logic like this. 

```ts
 public constructor(state: SceneDataTransformerState) {
    super(state);

    this.addActivationHandler(() => this._onActivate());
  }

  private _onActivate() {
      // activation code
      return () => {
             // Deactivation code 
      }
  }
```

### DIFF
--- a/packages/scenes/src/core/SceneComponentWrapper.tsx
+++ b/packages/scenes/src/core/SceneComponentWrapper.tsx
@@ -5,17 +5,7 @@ import { SceneComponentProps, SceneObject } from './types';
 function SceneComponentWrapperWithoutMemo<T extends SceneObject>({ model, ...otherProps }: SceneComponentProps<T>) {
   const Component = (model as any).constructor['Component'] ?? EmptyRenderer;
 
-  // Handle component activation state state
-  useEffect(() => {
-    if (!model.isActive) {
-      model.activate();
-    }
-    return () => {
-      if (model.isActive) {
-        model.deactivate();
-      }
-    };
-  }, [model]);
+  useEffect(() => model.activate(), [model]);
 
   return <Component {...otherProps} model={model} />;
 }
--- a/packages/scenes/src/core/SceneObjectBase.tsx
+++ b/packages/scenes/src/core/SceneObjectBase.tsx
@@ -173,8 +173,18 @@ export abstract class SceneObjectBase<TState extends SceneObjectState = SceneObj
 
     this._refCount++;
 
+    let called = false;
+
     return () => {
       this._refCount--;
+
+      if (called) {
+        const msg = `SceneObject cancelation handler returned by activate() called a second time`;
+        console.error(msg, this);
+        throw new Error(msg);
+      }
+
+      called = true;
 
       if (this._refCount === 0) {
         this._internalDeactivate();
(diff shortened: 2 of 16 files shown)


packages/scenes/src/core/SceneComponentWrapper.tsx
```diff
@@ -5,17 +5,7 @@ import { SceneComponentProps, SceneObject } from './types';
...
-    if (!model.isActive) {
-      model.activate();
-    }
-    return () => {
-      if (model.isActive) {
-        model.deactivate();
-      }
-    };
-  }, [model]);
+  useEffect(() => model.activate(), [model]);
```
torkelo, line 8: So one of the reasons I opted for an activate function that returns a deactivate function is to support this scenario above.  Otherwise we could have just passed in a sceneObject into activate and have that function add a deactivation handler. sort of like `activateFrom(sceneObject)` , but that does not work for the scenario above. So instead opted for a simpler method of just returning a deactivation handler that the caller needs to call when they are no longer interested in the object (or if its' react the react component get's unmounted)

packages/scenes/src/core/SceneObjectBase.tsx
```diff
@@ -164,27 +166,39 @@ export abstract class SceneObjectBase<TState extends SceneObjectState = SceneObj
...
+    }
+
+    this._refCount++;
+
+    let called = false;
+
+    return () => {
+      this._refCount--;
+
+      if (called) {
```
dprokop, line 181: Nice use of closure

packages/scenes/src/core/SceneObjectBase.tsx
```diff
@@ -139,21 +141,21 @@ export abstract class SceneObjectBase<TState extends SceneObjectState = SceneObj
    * Called by the SceneComponentWrapper when the react component is mounted.
    * Don't override this, instead use addActivationHandler
    */
```
dprokop, line 143: This comment is misplaced (private prevents from this being overriden). Should be in `activate`.

dprokop (approved):
Really like how the activation is shaping now. 

I think this is just missing Release notes with clear indication that `public activate` should not be used and how to achieve the same behavior with `addActivationHandler`. I know we already added a similar note in the PR that introduced activation handlers, but with this PR we are essentially saying `DON'T DO IT`.

torkelo:
@dprokop added some release notes

packages/scenes/src/core/SceneObjectBase.tsx
torkelo: @dprokop moved this to the public activate function

PR #115, opened 2023-03-29
App: Show errors in the landing page if system is not configured properly

ryantxu:
Adding a few warning messages that would have helped me (non-documentation reading 😬 ) figure out why things were not working faster.

image

### DIFF
--- a/packages/scenes-app/provisioning/datasources/default.yaml
+++ b/packages/scenes-app/provisioning/datasources/default.yaml
@@ -1,0 +1,6 @@
+apiVersion: 1
+
+datasources:
+  - name: gdev-testdata
+    isDefault: true
+    type: testdata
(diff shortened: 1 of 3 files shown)


torkelo (commented):
Hm.. it should be using gdev-testdata I think which most grafana developers already have

ryantxu:
> it should be using gdev-testdat

agree... and also want to get back to supporting using *any* testdata instance (when uid is not defined...)

torkelo (approved):

dprokop:
> Hm.. it should be using gdev-testdata I think which most grafana developers already have

Yes, I  used this data source for scene app demo scenario which runs in a container (./scripts/demo.sh), but should have named the provisioned test datasource gdev-testdata to work simlessly with core. I've brought back correct provisioning file in [2e24cbe](https://github.com/grafana/scenes/pull/115/commits/2e24cbe75e9ec5b10e674cce300fb93a66e4f977)

PR #117, opened 2023-03-31
VizPanel: Support for passing a custom PanelPlugin

torkelo:


packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -31,7 +31,7 @@ import { changeSeriesColorConfigFactory } from './colorSeriesConfigFactory';
 export interface VizPanelState<TOptions = {}, TFieldConfig = {}> extends SceneObjectStatePlain {
   title: string;
   description?: string;
-  pluginId: string;
+  pluginId: string | PanelPlugin;
```
torkelo, line 34: rename to plugin?

dprokop: yes! 

Thinking if it wouldn't make sense to actually have CustomViz object instead of mixing these two here. But this means code duplication. Renderer could be re used tho.

torkelo: I was looking into that, belive me trying to make break up viz into more composable parts has been super hard. 

But with this, the only difference between pluginId and panel plugin is just an additional step of loading the plugin (when it's a string id), all the rest is the same. 

Maybe if they inherited the same base class?

dprokop: You know what, the more I think about it the more skeptical I am about the separation. I mean - the VizPanel code is not impacted that much by this change, nor it's more complex. I think it's a good first step and would drop the separation idea until we have a clear insight into this being needed.

The only thing I would change is the `pluginId` -> `plugin` rename for now.

PR #119, opened 2023-04-01
Behaviors: Add state and runtime behavior to any scene object

torkelo:
Explores a way to do conditional panels and a way to add new behaviors and state to other scene objects 

* Adds $behaviors array to SceneObjectStatePlain  
* All behaviors are activated when the parent scene object is activated (exactly like $data, $variables, etc) 
* Adds isHidden to SceneLayoutItemState, but really only SceneFlexItem supports it (much trickier to support for grid) *

Adds two basic proof of concepts 
* HiddenWhenNoDataBehavior
* HiddenForTimeRangeBehavior 
* ShowBasedOnConditionBehavior (hooked up to a radio button toggle) 
* Adds a demo scene 

Beyond behaviors like above I think a behavior could also be used to do things like changing a layout based on some drill down URL parameter (So implement url sync).  They can also be used to add some custom state to an object (But maybe there should be another way to do that, but this at least makes it possible)

## Release notes

You can now augment any scene object with runtime state & behavior using the new `$behaviors` state key. Behaviors are implemented as SceneObjects that are activated when their parent is activated or as pure functions that get called when the SceneObject they are attached to get's activated. 

With behaviors you can easily implement conditional display of panels using the new `isHidden` property on SceneFlexItem. and other dynamic layout behaviors. View the [behaviors demo](https://github.com/grafana/scenes/blob/main/packages/scenes-app/src/demos/behaviorsDemo.tsx) for some examples.


packages/scenes/src/behaviors/ShowBasedOnConditionBehavior.ts
```diff
@@ -0,0 +1,39 @@
+import { SceneObject, SceneObjectStatePlain } from '@grafana/scenes';
+import { HiddenLayoutItemBehavior } from './HiddenLayoutItemBehavior';
+
+export interface ShowBasedOnConditionBehaviorState extends SceneObjectStatePlain {
+  getCondition: () => ShowBasedCondition;
```
torkelo, line 5: the condition is accessed via a function to work around the problem of having references to other scene objects in state (without overwriting parent). Think we need a solution for this problem

packages/scenes-app/src/pages/Demos/scenes/behaviors.tsx
```diff
@@ -0,0 +1,85 @@
...
+          body: newTimeSeriesPanel({ title: 'Hidden when no time series' }),
+        }),
+      ],
+    }),
+  });
+}
+
+function getTextPanelToggle(toggle: SceneRadioToggle) {
+  return () => ({ references: [toggle], condition: () => toggle.state.value === 'visible' });
+}
```
torkelo, line 85: Another way to do this condition would be if references were only scene object keys, and the condition would get passed these objects. The benefit of this would be that the scene could be declared inline and the radio toggle variable would not have be declared before the rest of the scene 

```suggestion
function getTextPanelToggle() {
  return { references: ["toggle-key"], condition: (toggle: SceneRadioToggle) => toggle.state.value === 'visible' };
}
```

ryantxu:
I would expect that we implement things like "hidden" by adding and removing things from the scene graph -- i get a bit nervous if we have trees including things that should not really be active.

torkelo:
> I would expect that we implement things like "hidden" by adding and removing things from the scene graph -- i get a bit nervous if we have trees including things that should not really be active.

@ryantxu 

That does not really compute in my mind.  And I think your misunderstand the logic here. a SceneFlexItem that is hidden (via state isHidden) never renders its body, so the VizPanel or Layout inside it will never be actived. 

* If you want to hide something in response to data, the data (query runner) needs to be attached to something in the scene that is active! 
* If we want to do declarative conditional logic I don't see any other way to attach that conditional logic to some parts of the scene (and those parts need to be active)

Your are right that we can modify children's layout and add new conditional parts to the layout, but that is fully possible with this behaviors model as well.  Super easy to create a behavior that can add a new child or removes a child. The behaviors can do anything. The example behaviors added in this PR will likely be moved to the demo app as I don't think they are solid enough to add to the library yet (until we have clear use cases), but are just examples of what you can build with behaviors. And they do simplify otherwise much more complex logic of creating scene objects, adding activation handlers, subscribing to another scene object, and modifying scene state. 

They are a first stab at trying to make dynamic behaviors declarative but are just early pocs that showcase what behaviors can do.

torkelo:
Pushed an update that moved all the "POC" behaviors to scene-app , also added a new type of behavior SceneStatelessBehavior  (so behaviors can either be SceneObjects or SceneStatelessBehavior)

SceneStatelessBehavior are pure functions that will be called when object is activated and return a cancellation handler. 
```ts
export type SceneStatelessBehavior = (sceneObject: SceneObject) => CancelActivationHandler;
```

with this you can define simple behaviors that don't require any state on their own as simple functions. 

Example:

```
body: new VizPanel({
   $behaviors: [logEventsBehavior],
    ....
}),

function logEventsBehavior(sceneObject: SceneObject) {
  console.log(`[SceneObjectEvent]: ${sceneObject.constructor?.name} ${sceneObject.state.key} activated!`);

  sceneObject.subscribeToState((state) => {
    console.log(`[SceneObjectEvent]: ${sceneObject.constructor?.name} ${sceneObject.state.key} state changed!`, state);
  });

  return () => {
    console.log(`[SceneObjectEvent]: ${sceneObject.constructor?.name} ${sceneObject.state.key} deactivated!`);
  };
}

```

packages/scenes/src/components/layout/SceneFlexLayout.tsx
```diff
@@ -68,10 +72,16 @@ interface SceneFlexItemPlacement {
...
+  /**
+   * True when the item should rendered but not visible.
+   * Useful for conditional display of layout items
+   */
+  isHidden?: boolean;
 }
 
 interface SceneFlexItemState extends SceneFlexItemPlacement, SceneObjectState {
   body: SceneObject | undefined;
+  isHidden?: boolean;
```
dprokop, line 84: Thinking out loud... I would consider moving this to `SceneObjectState` and maybe handle visibility in the `SceneComponentWrapper`? I guess the problematic thing about this may be that activation happens there in the `useEffect` hook. 

I wish we had sth like an internal object state so that such property wouldn't become available as aa public api in the constructor.

packages/scenes-app/src/demos/behaviors/ShowBasedOnConditionBehavior.ts
```diff
@@ -0,0 +1,45 @@
+import { sceneGraph, SceneObject, SceneObjectState } from '@grafana/scenes';
+import { HiddenLayoutItemBehavior } from './HiddenLayoutItemBehavior';
+
+export interface ShowBasedOnConditionBehaviorState extends SceneObjectState {
+  references: string[];
```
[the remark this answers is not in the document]
torkelo: @dprokop well there are problems with that 
* Having the reference be a string makes be serializable (I know the condition is not at this point , we would need some kind of scripting language for the condition to be serializable) 
* Makes it possible to declare the scene in one expression without having to break it up into separate variable declarations 
* And it works around the problem of we do currently not support having SceneObjects are direct references (as it would change parent)

So the ShowBasedOnConditionBehavior is really just an exploration of how it could be done, I am not saying it's how we should do it. This is all part of the demo app

packages/scenes/src/components/layout/SceneFlexLayout.tsx
torkelo: @dprokop yea, had the same thought of moving this to SceneObjectState & SceneComponentWrapper. With https://github.com/grafana/scenes/pull/148  merged I can give it a go 

I don't think hidden should be hidden/internal state though (needs to be settable from the outside)

dprokop: > but then we need to subscribe to model state from the SceneComponentWrapper and this doubles the renders of the real SceneComponent.

RIght, this is logical. Then I think this is a good first approach with the item's `isHidden` state

packages/scenes-app/src/demos/behaviors/HiddenLayoutItemBehavior.ts
```diff
@@ -0,0 +1,40 @@
...
+    const parentLayoutItem = getClosestLayoutItem(this);
+
+    if (parentLayoutItem.state.isHidden) {
+      parentLayoutItem.setState({ isHidden: false });
+    }
+  }
+}
+
+function getClosestLayoutItem(obj: SceneObject): SceneFlexItem {
+  if (obj instanceof SceneFlexItem) {
```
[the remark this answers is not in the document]
torkelo: fixed this

dprokop (approved):

PR #120, opened 2023-04-01
LayoutTypes: Cleanup old types that are no longer needed 

torkelo:
* Simplify types now that the union type SceneObjectState is no longer needed 
* Removes SceneLayoutChildState

### DIFF
--- a/packages/scenes/src/core/types.ts
+++ b/packages/scenes/src/core/types.ts
@@ -21,15 +21,9 @@ export interface SceneObjectStatePlain {
   $variables?: SceneVariables;
 }
 
-export interface SceneLayoutChildState extends SceneObjectStatePlain {
-  placement?: SceneLayoutChildOptions;
-}
-
 export interface SceneLayoutItemState extends SceneObjectStatePlain {
   body: SceneObject | undefined;
 }
-
-export type SceneObjectState = SceneObjectStatePlain | SceneLayoutState | SceneLayoutChildState;
 
 export interface SceneLayoutChildOptions {
   width?: number | string;
@@ -60,7 +54,7 @@ export interface SceneDataState extends SceneObjectStatePlain {
   data?: PanelData;
 }
 
-export interface SceneObject<TState extends SceneObjectState = SceneObjectState> {
+export interface SceneObject<TState extends SceneObjectStatePlain = SceneObjectStatePlain> {
   /** The current state */
   readonly state: TState;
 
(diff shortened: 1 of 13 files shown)


packages/scenes/src/core/types.ts
```diff
@@ -60,7 +54,7 @@ export interface SceneDataState extends SceneObjectStatePlain {
   data?: PanelData;
 }
 
-export interface SceneObject<TState extends SceneObjectState = SceneObjectState> {
+export interface SceneObject<TState extends SceneObjectStatePlain = SceneObjectStatePlain> {
```
torkelo, line 57: think we should probably take the time to rename SceneObjectStatePlain to SceneObjectState or SceneObjectStateBase

packages/scenes/src/core/types.ts
```diff
@@ -21,16 +21,10 @@ export interface SceneObjectStatePlain {
...
 
-export interface SceneLayoutChildState extends SceneObjectStatePlain {
-  placement?: SceneLayoutChildOptions;
-}
-
 export interface SceneLayoutItemState extends SceneObjectStatePlain {
   body: SceneObject | undefined;
 }
 
-export type SceneObjectState = SceneObjectStatePlain | SceneLayoutState | SceneLayoutChildState;
```
torkelo, line 32: this is the main thing that I don't think we need anymore (this union type was only used from NestedScene in a bit of code that no longer works (where it tries to remove itself from it's parent children array, but it's parent never has a children array in the current layouts)

dprokop: > this union type was only used from NestedScene in a bit of code that no longer works (where it tries to remove itself from it's parent children array, but it's parent never has a children array in the current layouts)

It does work, added in https://github.com/grafana/scenes/pull/97/commits/fa5b1412e9abbd5d75dcdfa4828813882880f072#diff-a02fdda4e1d01bbcde83f440d8f484157069ab91780e41555f9d4f7d76a467e0 and the updated in https://github.com/grafana/scenes/pull/97/commits/f1eabcaa71580dd18b92de6ef5cd5d2de001cfd4#diff-a02fdda4e1d01bbcde83f440d8f484157069ab91780e41555f9d4f7d76a467e0

packages/scenes/src/core/types.ts
dprokop: Agree

dprokop (approved):

torkelo:
@dprokop will rename SceneObjectStatePlain to SceneObjectState in another PR (as it will touch so many files, and warrants a release notes entry)

PR #123, opened 2023-04-04
ValueMacro: Fixes so __value works for rowIndex 0

torkelo:
Stupid javascript or stupid developer (me)? Answer: both!


dprokop (approved):
It happens ... ;)

Issue #29, opened 2023-02-23
QueryVariable: Refresh on time range change does not wait for dependencies 


torkelo:
Variable that should refresh on time range change does not wait for running dependencies https://github.com/grafana/scenes/blob/main/src/variables/variants/query/QueryVariable.tsx#L72  

So if you have two variables A and B, and B depends on A and both have refresh on time range change both will trigger an update at the same time. Which could make A trigger twice if the update causes a value change in A.

PR #124, opened 2023-04-04
SceneVariableSet: Refresh variables that depend on time range

dprokop:
Closes https://github.com/grafana/grafana/issues/65832
Closes https://github.com/grafana/scenes/issues/29

First approach to refreshing variables that are dependant on time range.

1. Removed time range dependency from QueryVariable
2. Implemented logic in SceneVariableSet that will add all time range dependant variables to the update queue. My understanding is that doing this and calling `_updateNextBatch` is sufficient to trigger the chain of variable updates.

### DIFF
--- a/packages/scenes-app/src/pages/Demos/scenes/variables.tsx
+++ b/packages/scenes-app/src/pages/Demos/scenes/variables.tsx
@@ -15,7 +15,6 @@ import {
   NestedScene,
 } from '@grafana/scenes';
 import { getQueryRunnerWithRandomWalkQuery } from '../utils';
-// import { getQueryRunnerWithRandomWalkQuery } from '../utils';
 
 export function getVariablesDemo(): EmbeddedScene {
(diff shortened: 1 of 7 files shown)


packages/scenes-app/src/pages/Demos/scenes/variables.tsx
```diff
@@ -0,0 +1,136 @@
...
+  EmbeddedScene,
+  SceneControlsSpacer,
+  SceneRefreshPicker,
+  SceneFlexItem,
+  VizPanel,
+  SceneCanvasText,
+  NestedScene,
+} from '@grafana/scenes';
+import { getQueryRunnerWithRandomWalkQuery } from '../utils';
+// import { getQueryRunnerWithRandomWalkQuery } from '../utils';
```
dprokop, line 18:
```suggestion
```

torkelo (approved):
Think this looks good

PR #125, opened 2023-04-04
scene-app: Refactor to use SceneAppPage for demos

torkelo:
This refactors and moves things around quite a bit in the scenes-app App 

* Makes all demos use SceneAppPage , this makes it possible to create demos with subtitle, tabs, and drilldowns. Also makes it possible to to make things in demos that https://github.com/grafana/scenes/pull/71 will enable. 
* Makes demos the default route.
* Make the with drilldown demo page into a normal demo like the others
* Removes the "home" page and the "with tabs" page. The drilldown demo has tabs that scenario is kind of covered there. And I expect other demos will likely have tabs. The Home demo has a custom query variable. But I expect we will add variables demos (test cases really)  that will cover variables.


dprokop:
@torkelo pushed an update adding refresh pickers and isOnCanvas to controls

torkelo:
@dprokop I think further refactors are needed to reduce duplication a bit between demos, but wanted to do that in a separate PR

PR #127, opened 2023-04-04
sceneGraph: findObject

torkelo:
Implements a simple findObject function that can find any object in the tree (from any starting point) 

Also makes forEachChild support a "break" ability by returning true


dprokop (approved):

packages/scenes/src/core/SceneObjectBase.tsx
```diff
@@ -241,16 +241,20 @@ export abstract class SceneObjectBase<TState extends SceneObjectState = SceneObj
    * Loop through state and call callback for each direct child scene object.
    * Checks 1 level deep properties and arrays. So a scene object hidden in a nested plain object will not be detected.
    */
-  public forEachChild(callback: (child: SceneObjectBase) => void) {
+  public forEachChild(callback: (child: SceneObjectBase) => boolean | undefined) {
```
ryantxu, line 244: this looks like we are now combining the `Array.forEach()` semantics with `Array.find(...)` does it make more sense to have two function?

torkelo: @ryantxu I can't implement this search algorithm using find, as find cannot be recursive, the check only returns true or false, so it cannot be used to find something deep in the tree. At least I can't make it work, but might be missing something 

image

torkelo: This will not either work, as it will not return the deep hit 
image

torkelo: This was with this implementation, see branch here: https://github.com/grafana/scenes/compare/findObject...scene-object-find-child?expand=1

image

torkelo: > this looks like we are now combining the Array.forEach() semantics with Array.find(...) 

So in the end, not really. They are still fundamentally different. Just adding a break ability to forEach. I know it's not a super common behavior, so torn on adding it. But it looks pretty useful.

PR #129, opened 2023-04-05
SceneGridLayout: Support lazy rendering of items out of view

### LINKED ISSUE #121: SceneGridLayout: Support lazy rendering of items out of view
We need something like https://github.com/grafana/grafana/blob/main/public/app/features/dashboard/dashgrid/LazyLoader.tsx  

Would be nice to support it for SceneFlexLayout as well

kaydelaney:
Closes #121

Don't know if there's a better way to get the "real" dimensions of a grid item other than what I've done here, but it seems to work okay.
Basically just copied the LazyLoader component from grafana main - could maybe factor it out to grafana-ui rather than having it in both places?


dprokop (approved):
Works great as far as I can test. Agree with @torkelo that we should add some demo scene for testing the behavior.

packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
```diff
@@ -14,6 +14,7 @@ interface SceneGridLayoutState extends SceneObjectState {
    * Turn on or off dragging for all items. Indiviadual items can still disabled via isDraggable property
    **/
   isDraggable?: boolean;
+  lazy?: boolean;
```
torkelo, line 17:
```suggestion
  isLazy?: boolean;
```

just to keep same naming as isDraggable

PR #132, opened 2023-04-05
SceneApp: Share defaults between demos

torkelo:
An attempt at sharing more of the "controls" (and time range) between demos


dprokop (approved):
nice

PR #133, opened 2023-04-05
PackageJson: Simple scripts to run app dev from root

torkelo:
After having been frustrated with having to cd into packages/scenes-app and running yarn dev there, and then not being able to run root yarn commands etc. I think
having these root scripts makes sense.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,9 @@
     "precommit": "yarn run lint-staged",
     "prepare": "husky install",
     "packages:publish": "lerna exec --no-private -- npm publish",
-    "test:scenes": "lerna run test --scope '@grafana/scenes' --"
+    "test:lib": "lerna run test --scope '@grafana/scenes' --",
+    "dev:lib": "lerna run dev --scope '@grafana/scenes' --",
+    "dev:app": "lerna run dev --scope 'scenes-app' --"
   },
   "resolutions": {
     "@types/react": "17.0.42"


package.json
```diff
@@ -29,7 +29,9 @@
     "precommit": "yarn run lint-staged",
     "prepare": "husky install",
     "packages:publish": "lerna exec --no-private -- npm publish",
-    "test:scenes": "lerna run test --scope '@grafana/scenes' --"
+    "test:lib": "lerna run test --scope '@grafana/scenes' --",
+    "dev:lib": "lerna run dev --scope '@grafana/scenes' --",
+    "dev:app": "lerna run dev --scope 'scenes-app' --"
```
dprokop, line 34: There are also `./scripts/dev.sh` that links to local grafana and run `yarn dev` again scenes, and `./scripts.demo` that will start the scenes-app in container.

I most often have both of them running.

But I guess you most often develop against local grafana with scenes-app there?

dprokop (approved):

package.json
torkelo: @dprokop yes, I develop against local grafana with scenes-app installed (having multiple grafanas locally is frustrating as they log you out all the time)

dprokop: That's right :) I use Chrome Canary for the demo app. I kind of like that the demo app has LiveReload :)

PR #134, opened 2023-04-06
Revert "Docs: Setup Docusaurus (#128)"

dprokop:
This reverts commit fb6d6b5b09ed8e6a133c9b6204a73b237278481d.

Reverting this from main for now - seems like I've introduced typescript regression with this, and haven't yett figured it out.


dprokop:
Closing, seems like a flakiness on my machine.

PR #135, opened 2023-04-06
FlexLayout: Allow SceneFlexLayout to be child of another flex layout

dprokop:
This allows SceneFlexLayout to be used as a child of another SceneFlexLayout, without the need for wrapping it with SceneFlexItem. Tried to make SceneFlexLayout to implement SceneFlexItemLike interface, but it's quite tricky since the primarily because the `body` and `children` differentiation.

### DIFF
--- a/packages/scenes/src/components/layout/SceneFlexLayout.tsx
+++ b/packages/scenes/src/components/layout/SceneFlexLayout.tsx
@@ -1,20 +1,14 @@
 import React, { CSSProperties } from 'react';
 
 import { SceneObjectBase } from '../../core/SceneObjectBase';
-import {
-  SceneComponentProps,
-  SceneLayout,
-  SceneObjectState,
-  SceneObject,
-  SceneLayoutItemState,
-} from '../../core/types';
+import { SceneComponentProps, SceneLayout, SceneObjectState, SceneObject } from '../../core/types';
 
-export interface SceneFlexItemLike extends SceneObject<SceneFlexItemState> {}
+export interface SceneFlexItemLike extends SceneObject<SceneFlexItemSizeState> {}
 
-interface SceneFlexLayoutState extends SceneObjectState {
+interface SceneFlexLayoutState extends SceneObjectState, SceneFlexItemSizeState {
   direction?: CSSProperties['flexDirection'];
   wrap?: CSSProperties['flexWrap'];
-  children: Array<SceneFlexItemLike | SceneFlexLayout>;
+  children: SceneFlexItemLike[];
 }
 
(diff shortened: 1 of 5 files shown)


packages/scenes/src/components/layout/SceneFlexLayout.tsx
```diff
@@ -45,9 +45,12 @@ function SceneFlexLayoutRenderer({ model }: SceneComponentProps<SceneFlexLayout>
 
   return (
     <div style={style}>
-      {children.map((item) => (
-        <item.Component key={item.state.key} model={item} />
-      ))}
+      {children.map((item) => {
+        if (isSceneFlexLayout(item)) {
```
dprokop, line 49: This is awkward, but don't see any other way (apart from tsignore) to make TS happy

torkelo: @dprokop I am a bit confused as well, this should not be needed as this component does not care about children or body.

I was able to make this work here, just need to simplify SceneFlexItemLikeState so that it does not include body 
https://github.com/grafana/scenes/compare/flex-layout-children...flex-item-like-state?expand=1

torkelo: tempted to remove SceneLayoutItemState as it just adds a bit of confusing (just specifies body, but can have this as a convention instead for SceneFlexItem and SceneGridItem )

packages/scenes/src/components/layout/SceneFlexLayout.tsx
```diff
@@ -45,9 +45,12 @@ function SceneFlexLayoutRenderer({ model }: SceneComponentProps<SceneFlexLayout>
 
   return (
     <div style={style}>
```
torkelo, line 47: this is missing getFlexItemItemStyles(parent.state.direction || 'row', model);

If we are to use a SceneFlexLayout as an item it needs to behave as one as well

dprokop:
Thanks for having a look @torkelo. Did plenty of changes (SceneLayoutItemState removal included) so that it's I think slightly better now :)

packages/scenes/src/components/layout/SceneFlexLayout.tsx
```diff
@@ -43,16 +40,23 @@ function SceneFlexLayoutRenderer({ model }: SceneComponentProps<SceneFlexLayout>
     minHeight: 0,
   };
 
+  if (parent && isSceneFlexLayout(parent)) {
```
torkelo, line 43: I think you can just use instanceof SceneFlexLayout here

torkelo (approved):

PR #136, opened 2023-04-07
Docs: Getting started and core concepts

dprokop:
This adds content for:

- Getting started
- Core concepts of Scenes


docusaurus/docs/core-concepts.md
```diff
@@ -0,0 +1,236 @@
...
+```tsx
+function CounterRenderer({ model }: SceneComponentProps<Counter>) {
+  const { count } = model.useState();
+  const onIncrement = () => model.setState({ count: count + 1 });
+
+  // ...
+}
+```
+
+Or in the scene object implementation:
```
torkelo, line 96:
```suggestion
Or in the scene object class:
```

torkelo (approved):
Looks good!

PR #137, opened 2023-04-09
SceneApp: Correctly build demo pages with getParentPage 

torkelo:
I used a bit of a hack before to connect the demo page it's parent by doing this

```
 getPage: (routeMatch, parent: SceneAppPageLike) => {
            demo.setState({ getParentPage: () => parent });
``` 

Setting state inside getPage is not allowed as this is called in the React render path (and was a bit hacky either way). So this refactors the demos to be generated inside the getPage instead. To not duplicate the demo title and url I changed the signature a bit so that the demo getPage call get's passed in SceneAppState which already includes the title & url and getParentPage 

* Also adds a cache for drilldown pages 
* Will need to be reworked a bit, have a follow-up PR for this that refactors SceneAppPage a bit to add support for fallback routes

### DIFF
--- a/packages/scenes-app/src/demos/dynamicPage.tsx
+++ b/packages/scenes-app/src/demos/dynamicPage.tsx
@@ -14,7 +14,6 @@ import { getQueryRunnerWithRandomWalkQuery } from './utils';
 
 export function getDynamicPageDemo(defaults: SceneAppPageState): SceneAppPage {
   const defaultTabs = [getSceneAppPage('/tab1', 'Temperature')];
-  console.log('get dynamic page demo');
 
   const page = new SceneAppPage({
     ...defaults,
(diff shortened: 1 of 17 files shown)


packages/scenes-app/src/demos/dynamicPage.tsx
```diff
@@ -7,17 +7,18 @@ import {
...
   SceneRefreshPicker,
+  SceneAppPageState,
 } from '@grafana/scenes';
 import { demoUrl } from '../utils/utils.routing';
 import { getQueryRunnerWithRandomWalkQuery } from './utils';
 
-export function getDynamicPageDemo(): SceneAppPage {
+export function getDynamicPageDemo(defaults: SceneAppPageState): SceneAppPage {
   const defaultTabs = [getSceneAppPage('/tab1', 'Temperature')];
+  console.log('get dynamic page demo');
```
dprokop, line 17:
```suggestion

```

dprokop (approved):

PR #139, opened 2023-04-09
SceneObject: findChild & sceneGraph.findObject 

torkelo:
- sceneGraph: findObject
- Fixing ts issue
- SceneObject: findChild and sceneGraph.findObject


packages/scenes/src/core/sceneGraph.ts
```diff
@@ -94,6 +94,50 @@ export function hasVariableDependencyInLoadingState(sceneObject: SceneObject) {
...
+  check: (obj: SceneObject) => boolean,
+  alreadySearchedChild?: SceneObject
+): SceneObject | null {
+  if (check(scene)) {
+    return scene;
+  }
+
+  let found: SceneObject | null = null;
+
+  scene.findChild((child) => {
```
torkelo, line 108: this is the awkward bit, I can't actually use the return value of findChild here, but have to use as a sort of loop that I can break out of.

torkelo:
think I don't like this, will stick with https://github.com/grafana/scenes/pull/127 but remove break ability for now

PR #141, opened 2023-04-11
Flex layout item parent direction

torkelo:
This builds on https://github.com/grafana/scenes/pull/135  (now merged to main). 

In that PR we don't subscribe to the parent state so the layout of children will not be correctly changed if you where to change direction (via setState).  To avoid having to subscribe to parent state in the SceneFlexLayout and in SceneFlexItem this PR changes the rendering to pass the direction down from parent to child via normal react prop.


packages/scenes/src/components/layout/SceneFlexLayout.tsx
```diff
@@ -73,21 +74,24 @@ interface SceneFlexItemState extends SceneFlexItemPlacement, SceneObjectState {
   body: SceneObject | undefined;
 }
 
+interface SceneFlexItemRenderProps<T> extends SceneComponentProps<T> {
+  parentDirection?: CSSProperties['flexDirection'];
```
dprokop, line 78: Will this save us any re-renders? I would rather expect reading parent direction using `model.parent.useState()`.

torkelo: @dprokop Well sort of, it eliminates having to subscribe to parent state from child. making the update flow a bit easier and simpler (update direction will trigger re-render and then re-render of the child as it passes it to the child). 

Without this every child needs to subscribe to it's parent, and it becomes a bit messy with conditional logic for SceneFlexLayout that might have parent or not to subscribe to

dprokop: Right, I see. Think this makes sens, just felt sort of unnatural at first sight given we usually don't pass state to the children via props. But I think it's fine.

torkelo: @dprokop yea, agree it's not very natural (from the a scene) but it really comes in handy for some edge cases. We use the same thing to pass routeProps between SceneAppPage and their inner tabs and drilldowns.

dprokop: > We use the same thing to pass routeProps between SceneAppPage and their inner tabs and drilldowns.

haha, yeah I know, that was the only place where I could find this pattern :)

PR #142, opened 2023-04-11
SceneAppPage: Refactorings and adding default fallback routes 

torkelo:
* Refactors and clean-up logic 
* Refactors tests so we can test logic in different "it"  
* Move EmbeddedScene cache to SceneAppPAge
* Move generated drilldown pages cache to SceneAppPage 
* Adds default fallback routes for drilldowns (Simple static 404 page for now), need to make this configurable in the future. Also need to investigate how we can have apps show the default 404 page 

Holding some further refactors for follow-up PR 
* Refactor scenes-app routing a bit to go directly to a SceneApp 
* Add support for fallback & custom (ie additional routes) to SceneApp. Useful for 404 and redirect handling.


packages/scenes/src/components/SceneApp/SceneAppPageView.tsx
```diff
@@ -4,60 +4,65 @@ import React, { useEffect } from 'react';
...
-import { EmbeddedScene } from '../EmbeddedScene';
 import { SceneAppPage } from './SceneAppPage';
-import { SceneAppDrilldownView, SceneAppPageLike, SceneRouteMatch } from './types';
+import { SceneAppDrilldownView, SceneAppPageLike } from './types';
 import { getLinkUrlWithAppUrlState, renderSceneComponentWithRouteProps, useAppQueryParams } from './utils';
 
 export interface Props {
   page: SceneAppPageLike;
-  activeTab?: SceneAppPageLike;
+  //   activeTab?: SceneAppPageLike;
```
dprokop, line 13:
```suggestion
```

packages/scenes/src/components/SceneApp/SceneAppPageView.tsx
```diff
@@ -4,60 +4,65 @@ import React, { useEffect } from 'react';
...
   routeProps: RouteComponentProps;
 }
 
-export function SceneAppPageView({ page, activeTab, routeProps }: Props) {
-  const pageState = page.useState();
+export function SceneAppPageView({ page, routeProps }: Props) {
+  const containerPage = getParentPageIfTab(page);
```
dprokop, line 18: the `containerPage` is quite misleading given it will return `page` if it's not a tab. Maybe we should name it juest `page` instead, and refer to the page provided via props as `props.page`?

dprokop (approved):

packages/scenes/src/components/SceneApp/SceneAppPageView.tsx
torkelo: @dprokop but that becomes even more confusing having two page variables in the same function with no help on what is what. 

I struggled naming this but could not find anything better. 

Maybe name the page passed via props as "innerPage", and ` const outerPage = getParentPageIfTab(innerPage);`

I think it's confusing no matter what, as the variable means different things depending on if it's a tabbed page or not.

torkelo: I tested this but it becomes really unclear when to use props.page vs page  (getScene should use props.page for example)
image

PR #148, opened 2023-04-17
SceneObjectBase: Activate parents before children

torkelo:
* Wait with rendering scene objects' real component until after activation is completed 

Enables
* Activation logic can now happen before the first rendering of the SceneObject Component 
* Behaviors can run before children are activated (and set isHidden for example to prevent children from activating/rendering)

### DIFF
--- a/packages/scenes/src/core/SceneComponentWrapper.tsx
+++ b/packages/scenes/src/core/SceneComponentWrapper.tsx
@@ -4,8 +4,19 @@ import { SceneComponentProps, SceneObject } from './types';
 
 function SceneComponentWrapperWithoutMemo<T extends SceneObject>({ model, ...otherProps }: SceneComponentProps<T>) {
   const Component = (model as any).constructor['Component'] ?? EmptyRenderer;
+  const [activated, setActivated] = React.useState(false);
 
-  useEffect(() => model.activate(), [model]);
+  useEffect(() => {
+    setActivated(true);
+    return model.activate();
+  }, [model]);
+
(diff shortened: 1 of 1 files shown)


dprokop (approved):
Ah, it worked!

packages/scenes/src/core/SceneComponentWrapper.tsx
```diff
@@ -4,8 +4,19 @@ import { SceneComponentProps, SceneObject } from './types';
...
+    setActivated(true);
+    return model.activate();
+  }, [model]);
+
+  // By not rendering the component until the model is actiavted we make sure that parent models get activated before child models
+  // Otherwise child models would be activated before parents as that is the order of React mount effects.
+  // This also enables static logic to happen inside activate that can change state before the first render.
+  if (!activated) {
+    return null;
+  }
```
torkelo, line 19: @dprokop question is, should we move isActive to SceneObjectState? maybe under property name $isActive 

The reason it's not I think was because I did not want activation to trigger an unnecessary re-render but with this logic the SceneObject Component is not rendered until after activation so that reason no longer exist.

torkelo: but no I don't like that, then we need to subscribe to the scene object state here in the SceneComponentWrapper (and it would re-render on every state change).

dprokop: I'm on a fence, I think I would keep it as it is (React local state). Moving it to SceneObjectState makes it available in every constructor of a SceneObject. Which with `isActive` is pointless. Literally, 2 min ago I left this comment:

image

PR #153, opened 2023-04-18
SceneFlexLayout: Add support for lazy-loading

kaydelaney:
Leaving this as a draft because I have the feeling there are some edge-cases where this breaks down, but it's at least a start.

### DIFF
--- a/packages/scenes-app/src/demos/lazyLoadedFlex.tsx
+++ b/packages/scenes-app/src/demos/lazyLoadedFlex.tsx
@@ -1,4 +1,3 @@
-import React, { useMemo } from 'react';
 import {
   EmbeddedScene,
   SceneAppPage,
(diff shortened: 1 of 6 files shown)


bfmatei:
Ah, this is something I wanted to point out in one of the feedback threads I posted. Given a multi-page, multi-tab, and a drill-down, the bundle went to ~ 500kb. But we don't need to load everything until the user accesses that area, so lazy loading is more than welcomed.

packages/scenes/src/components/layout/SceneFlexLayout.tsx
```diff
@@ -86,13 +91,14 @@ interface SceneFlexItemState extends SceneFlexItemPlacement, SceneObjectState {
 
 interface SceneFlexItemRenderProps<T> extends SceneComponentProps<T> {
   parentDirection?: CSSProperties['flexDirection'];
+  isLazy?: boolean;
```
dprokop, line 94: Think we can make FlexLayout lazy by default. With the latest changes from @torkelo in https://github.com/grafana/scenes/pull/156 the parent state is provided via `parentState` property, so you don't have to include `isLazy` in the item state.

packages/scenes-app/src/demos/lazyLoadedFlex.tsx
```diff
@@ -0,0 +1,124 @@
...
+  SceneAppPageState,
+  SceneCanvasText,
+  SceneFlexItem,
+  SceneFlexLayout,
+  VizPanel,
+} from '@grafana/scenes';
+import { getQueryRunnerWithRandomWalkQuery, getEmbeddedSceneDefaults } from './utils';
+import { DebugItem } from './flexLayout';
+
+export function getLazyFlexDemo(defaults: SceneAppPageState) {
```
dprokop, line 13: I couldn't get this demo show the lazy loading. So i've tried panel repeater demo with ridiculous 100 repeats and it seems like that demo doesn't render the timeseries viz anymore.

PR #154, opened 2023-04-18
VizPanel: Support runtime registered panel plugins  

torkelo:
This PR makes it possible for scene app plugins to include custom visualizations without those visualizations having to be defined as standalone panel plugins. 

This can in a way already be accomplished via a custom SceneObject but requires plugins using PanelChrome directly and there is quire a bit of complexity in using that and setting the correct loading state and props for PanelChrome. 

Only problem with this implementation that I can think of right now
* The id of the runtime registered panel plugin could conflict with other plugins. This is not a problem right now as each scene app plugin get's it's own version of scenes so they get their own version of the runtimePanelPlugin registry. But if we where to switch to federated modules in the future then this could become a problem.

## Release notes 

It's now easier for scene app plugins to include custom visualizations in their scenes. By registering a runtime-defined PanelPlugin via the function `registerRuntimePanelPlugin` a scene app plugin can now take advantage of the full VizPanel logic that creates a PanelChrome with panel title, description, loading state, menu as well as the logic for applying panel option defaults and applying field config.

### DIFF
--- a/packages/scenes/src/components/VizPanel/registerRuntimePanelPlugin.ts
+++ b/packages/scenes/src/components/VizPanel/registerRuntimePanelPlugin.ts
@@ -4,11 +4,22 @@ import { getPluginImportUtils } from '@grafana/runtime';
 export const runtimePanelPlugins = new Map<string, PanelPlugin>();
 
 export interface RuntimePanelPluginOptions {
+  /**
+   * Please specify a pluginId that is unlikely to collide with other plugins.
+   */
   pluginId: string;
   plugin: PanelPlugin;
 }
 
+/**
+ * Provides a way to register runtime panel plugins.
+ * Please use a pluginId that is unlikely to collide with other plugins.
+ */
 export function registerRuntimePanelPlugin({ pluginId, plugin }: RuntimePanelPluginOptions) {
+  if (runtimePanelPlugins.has(pluginId)) {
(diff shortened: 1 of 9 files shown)


packages/scenes/src/components/VizPanel/registerRuntimePanelPlugin.ts
```diff
@@ -0,0 +1,41 @@
...
+        large: '',
+        small: '',
+      },
+      screenshots: [],
+      updated: '',
+      version: '',
+    } as PluginMetaInfo,
+  };
+
+  runtimePanelPlugins.set(pluginId, plugin);
```
dprokop, line 34: To address your concern about name conflicts with already registered plugins, we may check if there is a plugin with such name already regitered and throw an exception accordingly?

dprokop: Or alternatively, when registering a new runtime plugin, we can prefix it's plugin id : `__runtime_${pluginId}`. And use the same prefixed name when loading it, making the installed plugins take priority: 

```ts
return getPanelPluginFromCache(pluginId) ?? runtimePanelPlugins.get(`__runtime_${pluginId}`)
```

This way I think we can work around the potential name conflicts.

dprokop (approved):

dprokop (approved):

PR #156, opened 2023-04-19
FlexLayout: Responsive breakpoints

torkelo:
Just started exploring this, opening a draft so no one else starts on it and get some early feedback.

I noticed that this is actually pretty tricky as all constraints you put change when you change flex direction. Say you have a row of stat panels and the row has a maxHeight of 150px, if we have a responsive rule to change the direction of that row to column then all those panels would be also constrained to 150px. So every responsive rule would also have to change / remove the height (and potentially width constraints).

I am trying to make SceneFlexLayout responsive by default as well, in order for this to not require too much work for every scene. What I think is required is minHeight constraints (either on row or visualization level).

* [x] Automatically change direction to column when media query match is lower than breakpoint md, or manual direction is set in screenSmall prop 
* [x] Automatically clear maxHeight constraints for media query (so that rows that become column layouts don't have each item constrained to the same height as the row layout had)
* [x] Automatically clear maxWidth constraints for meda query (unless specified in screenSmall)
* [x] Automatically set minHeight on children if minHeight set on layout level. This makes it so you do not have to set minHeight on all children and it's enough to set it on the parent layout (row). Why is this needed? Well, otherwise you ALWAYS have to set minHeight on basically all children. Because when the responsive rule kicks in and a row layout is changed to a column the minHeight constraint needs to be on the child level, not the parent level. 

[Responsive-layout---Demos---Scenes-Test-App---Apps---Grafana.webm](https://user-images.githubusercontent.com/10999/233094376-fd972afd-c525-4f29-893d-04d873c230c7.webm)

### DIFF
--- a/docusaurus/docs/scene-layout.md
+++ b/docusaurus/docs/scene-layout.md
@@ -67,8 +67,8 @@ Both `ScenFlexLayout` and `SceneFlexItem` object types accept the following conf
   md?: SceneFlexItemPlacement;
 ```
 
-We really recommend setting a minHeight on all children of layout that use `column` direction. This will make sure that they don't get squashed to much on smaller screens. If you set minHeight or height on a SceneFlexLayout you do not need
-to set it on each child as they will inherit this constraint.
+We really recommend setting a minHeight on all children of layout that use `column` direction. This will make sure that they don't get squashed too much on smaller screens. If you set minHeight or height on a SceneFlexLayout you do not need
+to set it on each child as they will inherit these constraints.
 
 ### Step 4. Add panels to flex layout items
 
@@ -144,7 +144,7 @@ For `SceneFlexItems` that contain a `VizPanel`, it's usually a good idea to set 
 
 ### Responsive flex layouts
 
-By default SceneFlexLayout has some responsive behaviors for smaller screens. These kick in for screens that match the media query of theme.breakpoints.down('md').
+By default SceneFlexLayout has some responsive behaviors for smaller screens. These kick in for screens that match the media query of Grafana's theme.breakpoints.down('md').
 
 * SceneFlexLayout direction will change from row to column.
 * SceneFlexLayout maxWidth, maxHeight, height or width constraints are removed.
--- a/docusaurus/docs/scene-layout.md
+++ b/docusaurus/docs/scene-layout.md
@@ -45,7 +45,7 @@ Create scene with two, equally sized layout items in a column:
 const myScene = new EmbeddedScene({
   body: new SceneFlexLayout({
     direction: 'column',
-    children: [new SceneFlexItem({}), new SceneFlexItem({})],
+    children: [new SceneFlexItem({ minHeight: 200 }), new SceneFlexItem({ minHeight: 300})],
   }),
 });
 ```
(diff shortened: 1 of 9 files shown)


packages/scenes/src/components/layout/SceneFlexLayout.tsx
```diff
@@ -7,10 +9,23 @@ export interface SceneFlexItemStateLike extends SceneFlexItemPlacement, SceneObj
...
+
 interface SceneFlexLayoutState extends SceneObjectState, SceneFlexItemPlacement {
   direction?: CSSProperties['flexDirection'];
   wrap?: CSSProperties['flexWrap'];
   children: SceneFlexItemLike[];
+  /**
+   * Set direction for smaller screens. This defaults to column.
+   * This equals media query theme.breakpoints.down('md')
+   */
+  screenSmall?: SceneFlexItemResponsiveProps;
```
dprokop, line 28: An alternative idea to this. We may possibly use `SceneFlexItemPlacement` this way:

```ts
interface SceneFlexLayoutState extends SceneObjectState {
  wrap?: CSSProperties['flexWrap'];
  children: SceneFlexItemLike[];
  placement: {
   default: SceneFlexItemPlacement & { direction?: CSSProperties['flexDirection'] }, // this is default and required?
   sm?: SceneFlexItemPlacement & { direction?: CSSProperties['flexDirection'] },
  }
}
```


A similar change would be required for SceneFlexItem. This is a breaking change, but unifies how responsive breakpoints can be configured. And sets us for the future config or more than two breakpoints if needed. WDYT?

torkelo: @dprokop yes, but the definitions become so nested, and with this approach you will kind of rarely need to specify breakpoint overrides. Because the automatic responsive rules make the layouts responsive by default, and only in edge cases you will need to override the defaults (to preserve row layout for example). I don't think we need to move the default constraints to a sub-property 

 ```ts
new SceneFlexLayout({
              direction: 'row',
              placement: {
                default: {
                  maxHeight: 150,
                  minHeight: 150,
                },
                sm: {
                  direction: 'row'
                }
              },
```

vs 

```ts
new SceneFlexLayout({
              direction: 'row',
              maxHeight: 150,
              minHeight: 150,              
              sm: {
                direction: 'row',
              },
```

torkelo:
@dprokop @domasx2 @bfmatei I updated the implementation a bit, works really well. But could need some more eyes / testing

bfmatei:
Hey @torkelo!

From my experience up until now, `minHeight` is a property that should be set on all rows for proper alignment and placement of the panels. Hence, I don't see a major drawback to the `minHeight` constraint.

I will switch to this branch locally and play w/ it on our plugin to see the behavior. Expect a video pretty soon.

torkelo:
@bfmatei Yes. that be great. 

So the logic is now that children will by default use the height or minHeight of their parent layout (unless specified on the child).

dprokop (approved):
Works great, awesome you managed to have the default responsive behavior! One and only thing that I find missing is the docs for responsive behavior in https://github.com/grafana/scenes/blob/flex-layout-breakpoints/docusaurus/docs/scene-layout.md#L8

torkelo:
@dprokop agree, Will add docs today and get this merged

torkelo:
@dprokop pushed some docs, can you check if they look ok / understandable?

docusaurus/docs/scene-layout.md
```diff
@@ -137,6 +142,32 @@ The above example will render two panels, a Timeseries and a Table panel.
 For `SceneFlexItems` that contain a `VizPanel`, it's usually a good idea to set `minHeight` or `minWidth` constraints so they don't get squashed too small by limited screen space.
 :::
 
+### Responsive flex layouts
+
+By default SceneFlexLayout has some responsive behaviors for smaller screens. These kick in for screens that match the media query of theme.breakpoints.down('md').
```
dprokop, line 147:
```suggestion
By default SceneFlexLayout has some responsive behaviors for smaller screens. These kick in for screens that match the media query of Grafana's `theme.breakpoints.down('md')`.
```

dprokop (approved):
LGTM!

PR #159, opened 2023-04-19
Querying: Support runtime registered data source

torkelo:
This is another attempt at https://github.com/grafana/scenes/pull/45 but instead solving it by allowing scene library consumers to register an instance of DataSourceApi. 

Trying to make it possible to issue custom API call queries that result in DataFrames but via the SceneQueryRunner 

Advantages to this over a customQueryHandler passed to SceneQueryRunner: 

* [x] You can now use mixed data source queries to merge data from Prometheus and to your custom data source  (without adding customQueryHandler to the DataQuery model or creating a new CompositeQueryRunner) 
* [x] Docs 
* [x] Demo scene


packages/scenes-app/src/demos/flexLayout.tsx
```diff
@@ -25,10 +26,11 @@ export function getFlexLayoutTest(defaults: SceneAppPageState) {
               children: [
                 new SceneFlexItem({
                   minWidth: '70%',
-                  body: PanelBuilders.timeseries()
-                    .setTitle('Dynamic height and width')
-                    .setData(getQueryRunnerWithRandomWalkQuery({}, { maxDataPointsFromWidth: true }))
-                    .build(),
+                  body: new VizPanel({
```
bfmatei, line 29: Out of curiosity, what's the reason for changing from builder to declarative here?

bfmatei (commented):
This is pure awesomeness, ty @torkelo. I'll leave some comments here and there.

packages/scenes-app/src/demos/flexLayout.tsx
torkelo: did not really need it here, was not setting any visualization settings

dprokop (approved):
Looks good - apart from that one test failing! 

Adding @imatwawana for docs review.

bfmatei (approved):
Looks fine from my pov as well.

PR #160, opened 2023-04-21
SceneObjectBase: Force render when there are pending state chages before any state subscription

dprokop:
This tries to address a weird timing issue that can happen when an async action in the model that updates model state happens before state subscription is created in renderer.

For exampe in VizPanel, plugin can load quicker than the VizPanelRenderer renders and sets subscription to the state. So technically there was a state change, that will be undetected by VizPanel. 

In this PR I'm trying to solve this with a simple counter in SceneObjectBase that counts state updates (pending state updates) until a state subscription is created. When such subscription is added, and we have indicated some pending changes, the component is force updated and pending changes counter cancelled.


dprokop:
closing in favor of https://github.com/grafana/scenes/pull/161

PR #161, opened 2023-04-21
SceneObjectBase: Fixes issue with useState subscription misses state change that happens between frist render and useEffect 

torkelo:


packages/scenes/src/core/SceneObjectBase.tsx
```diff
@@ -278,12 +277,19 @@ export abstract class SceneObjectBase<TState extends SceneObjectState = SceneObj
...
 function useSceneObjectState<TState extends SceneObjectState>(model: SceneObjectBase<TState>): TState {
-  const forceUpdate = useForceUpdate();
+  const [_, setState] = useState<TState>(model.state);
+  const stateAtFirstRender = model.state;
 
   useEffect(() => {
-    const s = model.subscribeToState(forceUpdate);
+    const s = model.subscribeToState(setState);
+
+    if (model.state !== stateAtFirstRender) {
```
dprokop, line 286:
```suggestion
    // Re-render component if the state changed before state subscription (model.subscribeToState) has been called.
    if (model.state !== stateAtFirstRender) {
```

dprokop (approved):

packages/scenes/src/core/SceneObjectBase.tsx
dprokop: Wait this comment is not 100% true:D

dprokop: Should be better now

torkelo: @dprokop what about this comment 
// Re-render component if the state changed between first render and useEffect (mount)

dprokop: Great

PR #162, opened 2023-04-21
Scenes App: Fix scenes dependency

matyax:
I followed the demo-app instructions and stumbed upon a bunch of errors telling ` Error: Can't resolve '@grafana/scenes' in '/Users/matyax/Apps/scenes/packages/scenes-app/src/pages'`.

Changing the dependency version fixed them and I was able to see the demo app.

### DIFF
--- a/packages/scenes-app/package.json
+++ b/packages/scenes-app/package.json
@@ -62,7 +62,7 @@
     "@emotion/css": "^11.1.3",
     "@grafana/data": "canary",
     "@grafana/runtime": "canary",
-    "@grafana/scenes": "workspace:*",
+    "@grafana/scenes": "canary",
     "@grafana/schema": "canary",
     "@grafana/ui": "canary",
     "@types/lodash": "latest",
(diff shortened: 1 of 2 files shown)


torkelo:
@matyax strange, where are you getting these error? the dependency is correctly set to workspace (ie it's using the version in the your local repo)

torkelo:
@matyax try yarn install from repo root, then `yarn dev:app` to build the demo app (need to build the lib first with `yarn dev:lib` )

packages/scenes-app/package.json
```diff
@@ -62,7 +62,7 @@
     "@emotion/css": "^11.1.3",
     "@grafana/data": "canary",
     "@grafana/runtime": "canary",
-    "@grafana/scenes": "workspace:*",
```
torkelo, line 65: workspace is correct

matyax:
@torkelo With a fresh clone of the repo. Followed the instructions and the `scenes-app` could not resolve the scenes dependency with workspace. Was  I missing anything?

matyax:
The problem was my `GRAFANA_PATH` variable.

torkelo:
@matyax no I think your right the demo.sh script is missing building the library

PR #163, opened 2023-04-24
Scenes app: add script step to build scenes

matyax:
Follow up to a previous PR https://github.com/grafana/scenes/pull/162 where there was a missing script step to build the Scenes library.

Adding it now.


dprokop (approved):
thanks @matyax, this works!

PR #165, opened 2023-04-27
Docs: Activation handlers

dprokop:


dprokop:
>Looks good. Wonder if we should also have an example of activation handlers from inside the a custom scene object (in the constructor)?
@torkelo it's already there https://github.com/grafana/scenes/blob/main/docusaurus/docs/advanced-data.md#use-data-in-custom-object-class :)

torkelo:
@dprokop ah, I missed it :)

PR #167, opened 2023-04-28
Add support for timezones

dprokop:
This brings support for time zones configuration in a scene.

1. `SceneTimeRange` and `SceneTimePicker` now handle time zone changes.
2. There is a new object called `SceneTimeZoneOverride`. It implements `SceneTimeRangeLike` interface so that it can be used.... as `SceneTimeRange` object:

```ts
const scene = new EmbeddedScene({
  ...
  $timeRange: new SceneTimeRange({ timeZone: 'YYY' }),
  body: new SceneFlexLayout({
    children: [
      new SceneFlexItem({
        // This panel will use global time range and time zone
        body: new VizPanel({
          ...
        }),
      }),
      new SceneFlexItem({
        // This panel will use GLOBAL time range but with locally specified time zone
        body: new VizPanel({
          $timeRange: new SceneTimeZoneOverride({ timeZone: 'XXX'}),
        }),
      }),
      new SceneFlexItem({
        // This panel will use LOCAL time range but GLOBAL time zone YYY
        body: new VizPanel({
          $timeRange: new SceneTimeRange({  from: ..., to: ...}),
        }),
      })

  }),
```

 

TODO:
- [x] Make `SceneQueryRunner` use correct time range rather than fixed 'browser' time range
- [x] Add tests

## Release notes

You can now use multiple time zones in Scene. `SceneTimeRange` and `SceneTimePicker` respect time zone settings. Additionally, a new object was added, `SceneTimeZoneOverride`. It can be used to override the time zone provided by a time range object higher in the scene hierarchy. Objects within `SceneTimeZoneOverride` scope will use the closest `SceneTimeRange` range, but a locally specified time zone.

Example: 

```ts
const scene = new EmbeddedScene({
  $timeRange: new SceneTimeRange({ from: 'now-6h', to: 'now', timeZone: 'browser'}),
  children: [
    // Will use global time range and time zone
    new VizPanel({
      $data: new SceneQueryRunner({ ... }),
      ...
    }),
    // Will use global time range and locally specified time zone
    new VizPanel({ 
      $timeRange: new SceneTimeZoneOverride({ timeZone: 'America/New_York' }),
      $data: new SceneQueryRunner({ ... }),
      ...
    }),
  ],
  ...
})
```

### DIFF
--- a/packages/scenes/src/core/SceneTimeRange.tsx
+++ b/packages/scenes/src/core/SceneTimeRange.tsx
@@ -49,6 +49,10 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
     });
   }
 
+  public getTimeZone(): TimeZone {
+    return this.state.timeZone || getTimeZone();
+  }
+
   public onTimeRangeChange = (timeRange: TimeRange) => {
     const update: Partial<SceneTimeRangeState> = {};
 
@@ -64,7 +68,7 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
       update.to = timeRange.raw.to.toISOString();
     }
 
-    update.value = evaluateTimeRange(update.from, update.to, this.state.timeZone);
+    update.value = evaluateTimeRange(update.from, update.to, this.getTimeZone());
     this.setState(update);
   };
 
@@ -73,7 +77,7 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
   };
 
   public onRefresh = () => {
-    this.setState({ value: evaluateTimeRange(this.state.from, this.state.to, this.state.timeZone) });
+    this.setState({ value: evaluateTimeRange(this.state.from, this.state.to, this.getTimeZone()) });
   };
 
   public getUrlState() {
@@ -98,7 +102,7 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
       update.to = to;
     }
 
-    update.value = evaluateTimeRange(update.from ?? this.state.from, update.to ?? this.state.to, this.state.timeZone);
+    update.value = evaluateTimeRange(update.from ?? this.state.from, update.to ?? this.state.to, this.getTimeZone());
     this.setState(update);
   }
 }
--- a/packages/scenes/src/core/SceneTimeRange.tsx
+++ b/packages/scenes/src/core/SceneTimeRange.tsx
@@ -2,6 +2,7 @@ import { dateMath, getTimeZone, TimeRange, toUtc } from '@grafana/data';
 import { TimeZone } from '@grafana/schema';
 
 import { SceneObjectUrlSyncConfig } from '../services/SceneObjectUrlSyncConfig';
+import { sceneGraph } from './sceneGraph';
 
 import { SceneObjectBase } from './SceneObjectBase';
 import { SceneTimeRangeLike, SceneTimeRangeState, SceneObjectUrlValues, SceneObjectUrlValue } from './types';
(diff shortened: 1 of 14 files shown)


dprokop:
@torkelo  - I've changed the implementation a bit. Now it does not use the proxy anymore, just keeps the local state read from the closes time range on activation. Also, added support for the discussed case when you have a global time zone privided via SceneTimeRange, and some locally defined range without time zone specified. This required to make the `timeZone` property optional in the SceneTimeRangeState. 

There is a demo that showcases 3 different scnarios:
1. Using global time range and TZ (what we have now)
2. Using local time range, but with time zone not specified (will read from the closes SceneTimeRange that has TZ defined)
3. Using global time range, and locally overwriten TZ (via `SceneTimeRangeOverride` object, which is a `SceneTimeRangeLike`, but only allows for TZ config).

packages/scenes/src/core/SceneTimeRange.tsx
```diff
@@ -11,9 +13,44 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
...
+    }
+  };
+
+  private getTimeZoneSource() {
+    return sceneGraph.getClosest<SceneTimeRangeLike>(this.parent!.parent!, (o) => {
+      if (o.state.$timeRange && o.state.$timeRange.state.timeZone) {
+        return o.state.$timeRange;
+      }
+      return undefined;
+    });
```
dprokop, line 49: OK, this is causing the tests to fail, basically importing sceneGraph introduces a circular dependency.

packages/scenes/src/core/SceneTimeRange.tsx
```diff
@@ -1,19 +1,56 @@
...
+
+    this.addActivationHandler(this._onActivate);
+  }
+
+  private _onActivate = () => {
+    // When SceneTimeRange has no time zone provided, find closest source of time zone and subscribe to it
+    if (!this.state.timeZone) {
+      const timeZoneSource = this.getTimeZoneSource();
+      if (timeZoneSource) {
+        this.setState({ timeZone: timeZoneSource ? timeZoneSource.state.timeZone : getTimeZone() });
```
torkelo, line 28: timeZoneSoiurce must always have a value here as we are inside an `if(timeZoneSource)`  so this `?` statement looks unnessary

packages/scenes/src/core/SceneTimeRange.tsx
```diff
@@ -1,19 +1,56 @@
...
+        this._subs.add(
+          timeZoneSource.subscribeToState((n, p) => {
+            if (n.timeZone !== p.timeZone) {
+              this.setState({ timeZone: n.timeZone });
+            }
+          })
```
torkelo, line 34: The next time this time range is activated we will already have a timeZone in state, and this subscription will not be setup. 

Why do we store this parent timezone on this level? We only need to know that it changed, to re-evaluate our time range. I do not think we need to store it on every level.

packages/scenes/src/core/SceneTimeRange.tsx
```diff
@@ -1,19 +1,56 @@
...
+    return getClosest<SceneTimeRangeLike>(this.parent!.parent!, (o) => {
+      if (o.state.$timeRange && o.state.$timeRange.state.timeZone) {
+        return o.state.$timeRange;
+      }
+      return undefined;
+    });
+  }
+
+  public getTimeZone(): TimeZone {
+    return this.state.timeZone || getTimeZone();
```
torkelo, line 53: should this not walk up the sceneGraph to get the closest timezone if not present here?

packages/scenes/src/core/SceneTimeRange.tsx
dprokop: Good point @torkelo . I was thinking about having an _internal flag_ to know whether or not time zone was set via constructor. This flag would be set/reset on activation/deactivation Let me see how this will look when implemented.

torkelo: @dprokop I would explore not storing it

dprokop: yes, that was a dummy assumption from me to store it, will just read from closed time range and re-evaluate value when needed. Update coming shortly.

packages/scenes/src/core/SceneTimeRange.tsx
dprokop: This is built on assumption that we store time zone in the local state. Other comments this will change :)

torkelo (approved):
looks good!

PR #171, opened 2023-05-04
SceneAppPage: Add support for custom title

torkelo:
Why: 
App observability app needs to show multiple images after title. 

I am torn on how best to expose it
 * Just use the plain renderTitle that Page/PluginPage expose 
 * Use `React.ComponentType<{page: SceneAppPage}>`  component (this way you can define a title component that can be passed the page model).
 * Or titleObject, a full scene object handling title rendering.

### DIFF
--- a/packages/scenes/src/components/SceneApp/types.ts
+++ b/packages/scenes/src/components/SceneApp/types.ts
@@ -25,6 +25,11 @@ export interface SceneAppPageState extends SceneObjectState {
   title: string;
   /** Page subTitle */
   subTitle?: string;
+  /**
+   * Customize title rendering.
+   * Please return an unstyled h1 tag here + any additional elements you need.
+   **/
+  renderTitle?: (title: string) => React.ReactNode;
   /** For an image before title */
   titleImg?: string;
   /** For an icon before title */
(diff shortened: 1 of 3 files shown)


bfmatei (approved):
LGTM!

packages/scenes/src/components/SceneApp/types.ts
```diff
@@ -25,6 +25,11 @@ export interface SceneAppPageState extends SceneObjectState {
   title: string;
   /** Page subTitle */
   subTitle?: string;
+  /**
+   * Customize title rendering.
+   * Please return an unstyled h1 tag here + any additional elements you need.
+   **/
+  renderTitle?: (title: string) => React.ReactNode;
```
dprokop, line 32: Maybe `title: string | (title: string) => React.ReactNode;` ?

torkelo: @dprokop we need the title as string always for the breadcrumbs

dprokop (approved):

packages/scenes/src/components/SceneApp/types.ts
torkelo: yea, another way to type if would be `renderTitle: React.ComponentType<{title: string}`

PR #174, opened 2023-05-07
VizPanel: Support adding header actions to top right corner of PanelChrome 

torkelo:
Want to polish up the Grafana monitoring scene to use for demos during GrafanaCon, so needed actions.

Think this is a pretty straight forward way to expose it.

Allow either react nodes or a scene object.

React node should be useful for links to other places while a scene object
could be more useful if it's a button that changes state on current scene.

### DIFF
--- a/packages/scenes-app/src/demos/withDrilldown/panels.tsx
+++ b/packages/scenes-app/src/demos/withDrilldown/panels.tsx
@@ -1,4 +1,4 @@
-import { ReducerID, ThresholdsMode } from '@grafana/data';
+import { getFrameDisplayName, ReducerID } from '@grafana/data';
 import {
   SceneByFrameRepeater,
   SceneDataNode,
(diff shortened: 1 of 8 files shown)


packages/scenes-app/src/demos/withDrilldown/panels.tsx
```diff
@@ -124,40 +107,60 @@ export function getRoomsTemperatureTable() {
...
+                headerActions: (
+                  <a
+                    className="external-link"
+                    href={`${demoUrl('with-drilldowns')}/room/${encodeURIComponent(
+                      getFrameDisplayName(frame)
+                    )}/temperature`}
+                  >
+                    <Icon name="arrow-right" />
+                    Go to room
+                  </a>
```
dprokop, line 138: More of a UX thing. I completely missed the action the first time I wanted to test this demo out. Maybe a button would be a better choice here. Just to consider in the Grafana Monitoring app you will update. In this PR it is ok.

dprokop (approved):

dprokop:
@torkelo do you think we should add this to `Visualizations` docs?

torkelo:
@dprokop yes

PR #178, opened 2023-05-10
Docs: Improve headings

imatwawana:
Changed **Getting started > Getting started** to **Getting started > Setup** since it's bad practice to have duplicated headings. Also updated the page title to **Set up Scenes** to align with the new sidebar naming.

Changed **Learn Scenes** to **Scene objects** to make it more explicit about what kind of information will be in that section. All of the docs are aimed at helping you learn Scenes and it doesn't give you a clear idea of what content will be in that section.


dprokop (approved):

dprokop:
Thanks @imatwawana!

PR #189, opened 2023-05-15
Docs: Add advanced variables usage docs

dprokop:


docusaurus/docs/advanced-variables.md
```diff
@@ -3,4 +3,218 @@ id: advanced-variables
...
+- `onReferencedVariableValueChanged` - Configure a callback that will be executed when variable(s) that the object depends on changed.
+
+### Step 5. Add callback to handle referenced variables changes
+
+Introduce a new property `interpolatedText` to `TextInterpolatorState` that will contain the provided text, but with variables interpolated:
+
+```tsx
+interface TextInterpolatorState extends SceneObjectState {
+  text: string;
+  interpolatedText?: string;
```
torkelo, line 138: Do we have a more simple example of just using sceneGraph.interpolate in the render (component)? I don't think 99% of use cases of variables will need to store them in the state, only interpolate on rendering / usage.

docusaurus/docs/advanced-variables.md
```diff
@@ -3,4 +3,218 @@ id: advanced-variables
...
+
+Modify the `TextInterpolator` code to store interpolated value of the provided text. Use `sceneGraph.interpolate` helper to replace variables used in `text` with the correct value.
+
+```tsx
+class TextInterpolator extends SceneObjectBase<TextInterpolatorState> {
+  static Component = TextInterpolatorRenderer;
+
+  protected _variableDependency = new VariableDependencyConfig(this, {
+    statePaths: ['text'],
+    onReferencedVariableValueChanged: () => this.onVariableChange(),
```
torkelo, line 166: There is also the default behavior where if not onReferencedVariableValueChanged is passed the component will just be re-rendered automatically whenever a variable dependency changed, no need to store an d update any interpolated state.

torkelo (commented):
Looks great! but also a bit more complicated than what I hope common variable uses will be. 

I would start with a simple example of just using sceneGraph.interpolate while rendering, and relying on the default implementation of onReferencedVariableValueChanged

PR #190, opened 2023-05-16
SceneQueryRunner: Initial data state to avoid unnecesary No data messages

dprokop:
This PR adds a default initial data state to SceneQueryRunner to avoid unnecessary No data messages in VizPanel caused by the emptyPanelData (which has `LoadingState.Done` set).

<details>
  <summary>📦 Published PR as canary version: <code>0.8.0--canary.190.5014221438.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@0.8.0--canary.190.5014221438.0
  # or 
  yarn add @grafana/scenes@0.8.0--canary.190.5014221438.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/core/types.ts
+++ b/packages/scenes/src/core/types.ts
@@ -176,7 +176,7 @@ export type DeepPartial<T> = {
 
 export interface SceneDataProvider extends SceneObject<SceneDataState> {
   setContainerWidth?: (width: number) => void;
-  isReadyToRender: () => boolean;
+  isDataReadyToDisplay?: () => boolean;
 }
 
 export type SceneStatelessBehavior = (sceneObject: SceneObject) => CancelActivationHandler | void;
--- a/packages/scenes/src/core/types.ts
+++ b/packages/scenes/src/core/types.ts
@@ -176,6 +176,7 @@ export type DeepPartial<T> = {
 
 export interface SceneDataProvider extends SceneObject<SceneDataState> {
   setContainerWidth?: (width: number) => void;
+  isReadyToRender: () => boolean;
 }
 
 export type SceneStatelessBehavior = (sceneObject: SceneObject) => CancelActivationHandler | void;
--- a/packages/scenes/src/querying/SceneQueryRunner.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.ts
@@ -1,9 +1,17 @@
 import { cloneDeep } from 'lodash';
 import { Unsubscribable } from 'rxjs';
 
-import { DataQuery, DataSourceRef } from '@grafana/schema';
+import { DataQuery, DataSourceRef, LoadingState } from '@grafana/schema';
 
-import { CoreApp, DataQueryRequest, PanelData, preProcessPanelData, rangeUtil, ScopedVar } from '@grafana/data';
+import {
+  CoreApp,
+  DataQueryRequest,
+  getDefaultTimeRange,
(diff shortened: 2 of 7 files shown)


torkelo (commented):
I think we should maybe take the time and change SceneDataState to not have an optional data prop? but it's trickt, needs custom constructors then to make it optional when creating a SceneQueryRunner and SceneDataTransformer. Maybe best tackled in a separate PR. 

So for this change, what happens if the SceneQueryRunner is wrapped in a SceneDataTransformer? Do we need to make a similar change in SceneDataTransformer?

dprokop:
> So for this change, what happens if the SceneQueryRunner is wrapped in a SceneDataTransformer? Do we need to make a similar change in SceneDataTransformer?

We don't as transformer is using the data provided from runner state.

But need to polish this PR a bit as refreshes now cause viz to disappear (because of the state-dependant rendering of the plugin component). Will update soon

dprokop:
> I think we should maybe take the time and change SceneDataState to not have an optional data prop? but it's trickt, needs custom constructors then to make it optional when creating a SceneQueryRunner and SceneDataTransformer. Maybe best tackled in a separate PR.

@torkel yes, i'll do it separately. In the meantime, I've pushed some updates, wonder what you think about those. It's quite tricky to solve it relying only on the loading state, to solve the issue of refresh hiding the viz. But i think this approch now is good.

rdubrock:
> as refreshes now cause viz to disappear

I did notice the flicker on refresh, but it's definitely an improvement!

rdubrock (approved):

packages/scenes/src/core/types.ts
```diff
@@ -176,6 +176,7 @@ export type DeepPartial<T> = {
 
 export interface SceneDataProvider extends SceneObject<SceneDataState> {
   setContainerWidth?: (width: number) => void;
+  isReadyToRender: () => boolean;
```
torkelo, line 179: hmm...feels likethe wrong terminology to use (from this components point of view). 

DataProvider has no concept of rendering.

dprokop: - `isDataReadyForDisplay`
- `isDataReady`
 
maybe? I think i'm gonna also make this optional.

leeoniya: plan is to roll it out the capabilities to all TSDBs. working on it with @grafana/observability-logs  and @grafana/observability-metrics. (pinged you on slack)

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -224,6 +246,14 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
 
   private onDataReceived = (data: PanelData) => {
     const preProcessedData = preProcessPanelData(data, this.state.data);
+    if (
+      !this._hasFetchedData &&
+      preProcessedData.state !== LoadingState.NotStarted &&
+      preProcessedData.state !== LoadingState.Loading
```
torkelo, line 252: the not started check feels unnessary here :)  feels like something is wrong if that happens

torkelo (approved):

packages/scenes/src/querying/SceneQueryRunner.ts
dprokop: Yeah, I've cleaned up the code a bit, it ain't there anymore.

Issue #180, opened 2023-05-11
DrilldownLinks: Make it easier to specify links that include current URL state 


torkelo:
When creating a datalink you currently have to do this:

### Problem no built-in helper function for drilldown page links 
```ts
 value: [
      {
        title: 'Go to handler drilldown view',
        onBuildUrl: () => {
          const params = locationService.getSearchObject();
          return getLinkUrlWithAppUrlState(
            '/scenes/grafana-monitoring/handlers/${__value.text:percentencode}',
            params
          );
        },
      },
    ],
```

But we do not expose a util like getLinkUrlWithAppUrlState from scenes. 

### Problem no way to to drilldown links that just add a query parameter to current scene url 

If you have a drilldown link that should just ADD a query parameter then you can't use the above approach. Because if you want to use `${__value.text:percentencode}` in in query parameter you need to interpolate it this function before passing it to getLinkUrlWithAppUrlState (otherwise the template variable expression will be escaped before being interpolated). But the onBuildUrl does not get any interpolation function passed to it. the interpolation happens after it is called. which is too late if we are to build the url with the additional url paramter using `urlUtil.renderUrl(path, params);`  (which is what getLinkUrlWithAppUrlState is using). 

### Links in VizPanel header actions 

Let's say we have a link in a VizPanel (that goes to a differnet scene or just wants to add a query parameter to current scene)

```tsx
body: new VizPanel({
  pluginId: 'timeseries',
  headerActions: (
    <LinkButton
      fill="text"
      size="sm"
      icon="arrow-right"
      href={"/plugin-id/scene?query}
    >
      Details
    </LinkButton>
  ),
  title: getFrameDisplayName(frame),
  options: {
    legend: { showLegend: false },
  },
}),
```

How is that link to be defined in a away to that adds the current URL state (or the state you wish to preserve if the link goes to another scene). Do we need a wrapper scene object around LinkButton that can update link on URL location change?

dprokop:
> If you have a drilldown link that should just ADD a query parameter then you can't use the above approach. Because if you want to use ${__value.text:percentencode} in in query parameter you need to interpolate it this function before passing it to getLinkUrlWithAppUrlState (otherwise the template variable expression will be escaped before being interpolated). But the onBuildUrl does not get any interpolation function passed to it. the interpolation happens after it is called. which is too late if we are to build the url with the additional url paramter using urlUtil.renderUrl(path, params); (which is what getLinkUrlWithAppUrlState is using).

[...]

PR #193, opened 2023-05-18
Scene utils: Expose helper for building drilldown links

dprokop:
- Exposes `getLinkUrlWithAppUrlState` under `sceneUtils` object.
- Changes `getLinkUrlWithAppUrlState` not to require `params` as an argument. This now happens under the hood using runtime's `locationService`
- Moves `registerRuntimePanelPlugin` to `sceneUtils`

Ref https://github.com/grafana/scenes/issues/180

<details>
  <summary>📦 Published PR as canary version: <code>0.9.0--canary.193.5042913418.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@0.9.0--canary.193.5042913418.0
  # or 
  yarn add @grafana/scenes@0.9.0--canary.193.5042913418.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/SceneApp/utils.ts
+++ b/packages/scenes/src/components/SceneApp/utils.ts
@@ -15,7 +15,7 @@ export function useAppQueryParams(): UrlQueryMap {
  * @param preserveParams Query params to preserve
  * @returns Url with query params
  */
-export function getLinkUrlWithAppUrlState(path: string, preserveParams?: string[]): string {
+export function getLinkWithAppState(path: string, preserveParams?: string[]): string {
   // make a copy of params as the renderUrl function mutates the object
   const paramsCopy = { ...locationService.getSearchObject() };
 
--- a/packages/scenes/src/components/SceneApp/utils.ts
+++ b/packages/scenes/src/components/SceneApp/utils.ts
@@ -1,18 +1,23 @@
-import { RouteComponentProps, useLocation } from 'react-router-dom';
-
-import { UrlQueryMap, urlUtil } from '@grafana/data';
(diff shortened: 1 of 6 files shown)


packages/scenes/src/components/SceneApp/utils.ts
```diff
@@ -1,18 +1,23 @@
...
 }
 
-export function getLinkUrlWithAppUrlState(path: string, params: UrlQueryMap, preserveParams?: string[]): string {
+/**
+ *
+ * @param path Url to append query params to
+ * @param preserveParams Query params to preserve
+ * @returns Url with query params
+ */
+export function getLinkUrlWithAppUrlState(path: string, preserveParams?: string[]): string {
```
torkelo, line 18: Think we can shorten this name to something like getUrlWithPreservedState 

Having the word Link and the word Url twice feels unnecessary:)

torkelo (approved):

packages/scenes/src/components/SceneApp/utils.ts
dprokop: Thanks @torkelo, renamed to `getLinkWithAppState`

torkelo: @dprokop sounds good. Although I would prefer  `GetUrlWithAppState` as Link word would be sure what the response would be, a link model? a link react element/scene object? but with getUrlXX I would be more sure it's a function that returns a string that can be used in href/url props

dprokop: yeah, that's a fair point, will update shortly

PR #197, opened 2023-05-23
Macros: Add from and to macro

torkelo:
Closes #195

Adds macros for `$__from' and '$__to'

PR #198, opened 2023-05-23
Refactor sceneGraph structure to avoid circular dependencies

dprokop:
Based on https://github.com/grafana/scenes/pull/197 

Wanted to get rid of duplicate implementation of `getTimeRange` in  https://github.com/grafana/scenes/blob/from-to-macros/packages/scenes/src/variables/macros/timeMacros.ts#L61 but this resulted in a circular dependency.

This PR slightly changes the structure of the `sceneGraph` utils so that we can refactor them in the future if more issues like this appear. Currently, the `getTimeRange` util is extracted to a separate file making it possible to re-use directly in code that uses it (`timeMacros`). 

The `sceneGraph` API remains unchanged.


torkelo (approved):

torkelo:
Looks good!

PR #199, opened 2023-05-23
Macros: Url macro 

### LINKED ISSUE #180: DrilldownLinks: Make it easier to specify links that include current URL state

torkelo:
Closes #180

* Current full url `${__url}`
* Current url only path `${__url.path}` 
* Current url only state `${__url.state}`
* Current url only state without query parameter `${__url.state:exclude:var-handler}`
* Current url only state include only specified query parameters `${__url.state:include:from,to,var-instance}`

there is also an option to not use fieldPath and let these be separate macros __url, __url_state, __url_path 

TODO
* Review feedback
* Update docs

Any missing variants / capabilities? 

I think this could be useful

* Current parent path `${__url.parent}`  (to link to a sibling url)


packages/scenes/src/variables/macros/urlMacros.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  public constructor(name: string, _: SceneObject) {
+    this.state = { name: name, type: 'url_maco' };
+  }
+
+  public getValue(fieldPath?: string) {
+    const location = locationService.getLocation();
+
+    switch (fieldPath ?? '') {
+      case 'state':
```
dprokop, line 17: I think I would name this either `params` or `query`. `state` is very ambiguous, one may claim that URL state includes both path and params.

dprokop (approved):
@torkel - I've pushed merge commit which pretty much sums up my review. This looks great!

I'm not sure about `__url.parent`, I would wait with adding it until we see if there's a need for that.

torkelo:
@dprokop pushed some docs for macros , please review when you have time

dprokop (approved):

PR #205, opened 2023-05-24
GettingStarted: Slightly change wording

kaydelaney:
Feel like "Alternatively" fits better than "Conversely" in this case.


dprokop (approved):

imatwawana (approved):
Yep, makes sense to me. Good call!

PR #206, opened 2023-05-24
Core Concepts: Small docs changes

kaydelaney:
Just a few small changes to the Core Concepts docs.

### DIFF
--- a/docusaurus/docs/core-concepts.md
+++ b/docusaurus/docs/core-concepts.md
@@ -11,13 +11,13 @@ A scene is a collection of objects, called _scene objects_. These objects repres
 
 ![Scene objects tree](../docs/assets/sceneTree.png)
 
-Scenes allow you to group and nest object. Things like data, time ranges, or variables can be added to any object in the tree, making them accessible to that object and all descendant objects. Because of this, scenes allow you to create dashboards that have multiple time ranges, queries that can be shared and transformed, or nested variables.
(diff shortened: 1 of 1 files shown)


docusaurus/docs/core-concepts.md
```diff
@@ -72,7 +72,7 @@ function CounterRenderer({ model }: SceneComponentProps<Counter>) {
 
 ## State
 
-A scene object can have a state. The shape of the object's state is expressed through an interface that _must_ extend the `SceneObjectState` interface:
+Scene objects can have associated state. The shape of an object's state is expressed through an interface that _must_ extend the `SceneObjectState` interface:
```
imatwawana, line 75:
```suggestion
Scene objects can have an associated state. The shape of an object's state is expressed through an interface that _must_ extend the `SceneObjectState` interface:
```

imatwawana (approved):
Thanks for all the catches! I made one teeny fix, but otherwise LGTM!

docusaurus/docs/core-concepts.md
imatwawana: Could we say "data" explicitly here? So:
```suggestion
Scene objects can have associated state data. The shape of an object's state is expressed through an interface that _must_ extend the `SceneObjectState` interface:
```

If that still doesn't really convey the meaning or nuance of what you're going for, then go with your original edit. I defer to you on this because I don't want to change the meaning.

kaydelaney: I think it's usage is common enough, especially in the React world that the `data` is implicit (see https://react.dev/learn/choosing-the-state-structure for example).
I think I'll stick with the original edit for now, if that's okay

PR #219, opened 2023-05-26
Behaviors: Variable changed

dprokop:
A simple behavior to perform effect when variable changes. May be used for i.e. app settings persistence that depend on variable value (https://github.com/grafana/grafana-k8s-plugin/pull/561#discussion_r1206696676).

WDYT @torkelo ?

<details>
  <summary>📦 Published PR as canary version: <code>0.13.0--canary.219.5120943500.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@0.13.0--canary.219.5120943500.0
  # or 
  yarn add @grafana/scenes@0.13.0--canary.219.5120943500.0
  ```
</details>


packages/scenes-app/src/demos/behaviors/VariableEffectBehavior.ts
```diff
@@ -0,0 +1,44 @@
...
+  });
+
+  private _onVariableChanged(changedVariables: Set<SceneVariable>): void {
+    const effect = this.state.effect;
+
+    for (const variable of changedVariables) {
+      if (this.state.variables.includes(variable.state.name)) {
+        if (this._runningEffects.has(variable.state.name)) {
+          this._runningEffects.get(variable.state.name)!();
+          this._runningEffects.delete(variable.state.name);
```
torkelo, line 35: not sure I understand the point of the cancellation effect (that triggers on the second change?)

torkelo (commented):
I assume this is to create behaviors that can change the scene in different ways depending on the value of a variable. 

I can't really groc the cancellation thing. and the "runningEffects" map, in what way are the effect running? or is just to cancel the previous effect when a new instance of the effect is executed?

dprokop:
> I can't really groc the cancellation thing. and the "runningEffects" map, in what way are the effect running? or is just to cancel the previous effect when a new instance of the effect is executed?


Exactly. Say when you have an API request that u want to cancel if another is requested.

packages/scenes-app/src/demos/behaviors/VariableEffectBehavior.ts
```diff
@@ -0,0 +1,44 @@
...
+  SceneVariableState,
+  VariableDependencyConfig,
+} from '@grafana/scenes';
+
+interface VariableEffectBehaviorState extends SceneObjectState {
+  // The names of the variables to effect when they change
+  variables: string[];
+  // The effect to run when a variable changes. Optionally return a cancel function
+  effect: (variableState: SceneVariableState) => (() => void) | void;
+}
```
torkelo, line 14:
```suggestion
interface VariableChangedActionBehaviorState extends SceneObjectState {
  // The names of the variable to subscribe to changes to 
  variableName: string;
  // The handler to run when a variable changes. Return a cancellation function if you do anything async like issue a query
  onChange: (variable: SceneVariable) => (() => void) | void;
} 
```

Wonder if it makes more sense to limit this to a single variable, and the effect handler to take the full SceneVariable instance? That way it could potentially be made generic as well (ie typed SceneVariable). 

The array of variables becomes a bit strange behavior-wise, if multiple variables are specified, will the effect trigger for each if many are changed in one variable change transaction?

dprokop: Yes, I think this is better to handle a single variable effect!

dprokop:
@torkelo pushed updates and tests. One important thing - this new behavior is exposed via the `behaviors` namespace. I haven't noticed it before but all behaviors are currently in the scenes-app repo. Moved the `VariablesChangedBehavior` to `scenes`.

packages/scenes-app/src/demos/variables.tsx
```diff
@@ -66,6 +67,29 @@ export function getVariablesDemo(defaults: SceneAppPageState) {
         }),
         body: new SceneFlexLayout({
           direction: 'row',
+          $behaviors: [
+            new behaviors.VariableChangedBehavior({
```
torkelo, line 71: we don't export anything else like this (except maybe sceneGraph). I like it in the sense it makes it easy to import behaviors and get auto-complete for all types of behaviors we export, the only ugly bit is if you want to remove the namespace, or if you cannot import the behavior directly. 

If we where to adopt this, should we do it for Variables as well?

packages/scenes/src/behaviors/VariableChangedBehavior.ts
```diff
@@ -0,0 +1,42 @@
...
+  variableName: string;
+  // The handler to run when a variable changes. Return a cancellation function if you do anything async like issue a query.
+  onChange: (variable: SceneVariable) => (() => void) | void;
+}
+
+/**
+ * This behavior will run an effect function when specified variables change
+ */
+
+export class VariableChangedBehavior extends SceneObjectBase<VariableChangedBehaviorState> {
```
torkelo, line 17: the naming of this behavior feels off, a behavior should be a behavior we can attribute/give to a scene object, examples comparison to:

* HiddenWhenNoDataBehavior
* ShowBasedOnConditionBehavior 

given that, here are some ideas

* RespondToVariableChangedBehavior 
* TakeActionOnVariableChangedBehavior 
* ReactToVariableChangedBehavior 
* PerformWhenVariableChangedBehavior 
* OnVariableChangedBehavior 
* OnVariableValueChangedBehavior (probably unnecessary to add the word value, but usually never hurts to be explicit) 
* ActWhenVariableValueChangedBehavior 
* ActWhenVariableChangedBehavior

@dprokop thoughts, any favorites? 

ActWhenVariableChangedBehavior sounds pretty good 

```ts
  body: new SceneFlexLayout({
          direction: 'row',
          $behaviors: [
            new behaviors.ActWhenVariableChangedBehavior({ variableName: 'ds', onChange: handleDSChange })
```

packages/scenes-app/src/demos/variables.tsx
dprokop: @torkelo - yeah, this is for a discussion. I figured that exporting yet-another-object makes it probably trickier to reason about what's it for. 

I think we can consider this for variables (another PR). 

But this leads me to thinking - is the `Behavior` suffix necessary if it's namespaced? I would suggest doing namespaced export like this: `bahaviors.ActWhenVariableChanged` 

> or if you cannot import the behavior directly.

What do you mean?

packages/scenes/src/behaviors/VariableChangedBehavior.ts
dprokop: I like `OnVariableChangedBehavior` and `ActWhenVariableChangedBehavior`, with a preference to the former, feels more natural.

packages/scenes-app/src/demos/variables.tsx
torkelo: > or if you cannot import the behavior directly.

@dprokop 
I mean you have to import behaviors and not the specific behavior you want. So forced into using the namespace always. 

But your point of not needing the suffix is a good one, so the expression just really just 1 extra character with the namespace vs without. 

Let's try this namespace approach :)

torkelo (commented):
let's go with:

* behaviors.ActWhenVariableChanged

torkelo (approved):

Issue #207, opened 2023-05-25
Queries: Provide a way for query cancellation


dprokop:
In core Grafana dashboards there is a way to cancel a running query:
1. Query variable query shows a spinner next to the dropdown that when clicked cancels the query:

https://github.com/grafana/scenes/assets/2376619/608c9abb-0022-413e-a5c8-ab9e106271a8

2. Panel shows an explicit Cancel query button when query is taking longer than 2s.

https://github.com/grafana/scenes/assets/2376619/6db7f156-401c-4424-bd8b-b9ee6acdfaf6

We need to introduce a cancellation mechanism for variable and panel queries in scenes to support the above use cases.

PR #220, opened 2023-05-31
VizPanel: Allow queries to be cancelled

kaydelaney:
Not sure this is the right approach (specifically with how I'm getting the closest SceneQueryRunner), but it seems to work.

Part of #207
Will tackle query variable cancellation in another PR

<details>
  <summary>📦 Published PR as canary version: <code>0.17.0--canary.220.5309552119.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@0.17.0--canary.220.5309552119.0
  # or 
  yarn add @grafana/scenes@0.17.0--canary.220.5309552119.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
@@ -9,8 +9,6 @@ import { sceneGraph } from '../../core/sceneGraph';
 import { isSceneObject, SceneComponentProps } from '../../core/types';
 
 import { VizPanel } from './VizPanel';
-import { SceneQueryRunner } from '../../querying/SceneQueryRunner';
-import { SceneDataTransformer } from '../../querying/SceneDataTransformer';
 
 export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
   const {
@@ -88,7 +86,6 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
   // Data is always returned. For non-data panels, empty PanelData is returned.
   const data = dataWithFieldConfig!;
   const isReadyToRender = dataObject.isDataReadyToDisplay ? dataObject.isDataReadyToDisplay() : true;
-  const canCancel = $data instanceof SceneQueryRunner || ($data instanceof SceneDataTransformer && $data.state.$data instanceof SceneQueryRunner);
 
   return (
     <div ref={ref as RefCallback<HTMLDivElement>} style={{ position: 'absolute', width: '100%', height: '100%' }}>
@@ -108,7 +105,7 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
           dragClassCancel={dragClassCancel}
           padding={plugin.noPadding ? 'none' : 'md'}
           menu={panelMenu}
-          onCancelQuery={canCancel ? model.onCancelQuery : undefined}
+          onCancelQuery={model.onCancelQuery}
         >
           {(innerWidth, innerHeight) => (
             <>
--- a/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
@@ -9,8 +9,6 @@ import { sceneGraph } from '../../core/sceneGraph';
 import { isSceneObject, SceneComponentProps } from '../../core/types';
 
 import { VizPanel } from './VizPanel';
-import { getClosest } from '../../core/sceneGraph/utils';
-import { SceneQueryRunner } from '../../querying/SceneQueryRunner';
 
 export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
   const {
@@ -37,8 +35,6 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
   const dataObject = sceneGraph.getData(model);
   const rawData = dataObject.useState();
   const dataWithFieldConfig = model.applyFieldConfig(rawData.data!);
-  // maybe move this logic to VizPanel itself?
-  const queryRunner = getClosest(model, (s) => (s.state.$data instanceof SceneQueryRunner) ? s.state.$data : undefined);
 
   // Interpolate title
   const titleInterpolated = model.interpolate(title, undefined, 'text');
@@ -109,7 +105,7 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
           dragClassCancel={dragClassCancel}
           padding={plugin.noPadding ? 'none' : 'md'}
           menu={panelMenu}
-          onCancelQuery={() => queryRunner?.cancelQuery()}
+          onCancelQuery={model.onCancelQuery}
         >
           {(innerWidth, innerHeight) => (
             <>
--- a/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
@@ -9,6 +9,8 @@ import { sceneGraph } from '../../core/sceneGraph';
 import { isSceneObject, SceneComponentProps } from '../../core/types';
 
 import { VizPanel } from './VizPanel';
+import { SceneQueryRunner } from '../../querying/SceneQueryRunner';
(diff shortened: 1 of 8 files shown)


packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
```diff
@@ -35,6 +37,8 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
   const dataObject = sceneGraph.getData(model);
   const rawData = dataObject.useState();
   const dataWithFieldConfig = model.applyFieldConfig(rawData.data!);
+  // maybe move this logic to VizPanel itself?
+  const queryRunner = getClosest(model, (s) => (s.state.$data instanceof SceneQueryRunner) ? s.state.$data : undefined);
```
dprokop, line 41: Moving this to VizPanel end exposing `model.onCancelQuery` sounds like an optimal solution.

Another thing is how you retrieve the query runner. You should do that by using `sceneGraph.getData(...)` which will return an object implementing `SceneDataProvider` interface. This interface is currently implemented by three types of objects:
- `SceneQueryRunner`
- `SceneDataTransformer`
- `SceneDataNode`

I think we may want to extend the `SceneDataProvider` interface with a method that implements the cancellation. Thinking about this, especially for the reason that `SceneDataTransformer` can wrap `SceneQueryRunner` :

```
new SceneDataTransformer({
  $data: new SceneQeuryRunner(...),
  ...
});
```

So the cancellation can happen on SQR inderectly via SceneDataTransformer implementation.


The cancellation does not make sense for `SceneDataNode` tho, as this only provides static data to siblings.

dprokop:
Would appreciate a demo scene in scene-app for testing this behavior :)

kaydelaney:
> Would appreciate a demo scene in scene-app for testing this behavior :)

I think the lazy loading demo should work for testing (the queries take at least 5 seconds)

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -153,6 +153,13 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
     return this.state.maxDataPoints ?? this._containerWidth ?? 500;
   }
 
+  public cancel() {
```
dprokop, line 156: Think we are missing implementation of `cancel` method for `SceneDataTransformer`

dprokop (changes requested):

packages/scenes/src/querying/SceneQueryRunner.ts
kaydelaney: Added. Haven't dealt with SceneDataTransformer before so let me know if the cancel method I added makes sense

packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -87,6 +87,10 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
     this.transform(this.getSourceData().state.data);
   }
 
+  public cancel() {
+    this._transformSub?.unsubscribe();
```
dprokop, line 91: @kaydelaney I don't think this is sufficient and also doesn't seem to work. What we need to unsubscribe from is the data subscription which we are setting in the activation handler and which is then subject of the transformation.

But, there are two paths here:
1. `SceneTransformationNode` that does not have `$data: new SceneQueryRunner(...)` defined and transforms the closest data provider.
2. `SceneTransformationNode` that has it's on SceneQueryRunner.

In 1) the source SQR can be shared between multiple panels and IMO the cancellation should only happen in the context of the current transformation node to avoid cancellation in other panels which I think can be confusing. I think we should not allow cancellation in such a scenario at all, WDYT?

In 2) the approach is similar to SQR cancellation - so that just cancels the closes SQR (available in transformation node via `this.getSourceData` method)

kaydelaney: Ah, I see, thanks for the context. I think I agree that we shouldn't allow cancellation in the first scenario

kaydelaney: btw, guessing SceneTransformationNode is a mistype and you meant SceneDataTransformer?

kaydelaney:
Not sure what the best approach is for checking if a SQR is shared between multiple panels

packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -87,6 +88,14 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
     this.transform(this.getSourceData().state.data);
   }
 
+  public cancel() {
+    if (this.state.$data instanceof SceneQueryRunner) {
+      try {
+        this.getSourceData().cancel?.();
+      } catch {}
```
torkelo, line 95: why the try catch?

kaydelaney: I had it because getSourceData can throw, but I guess there's actually no need to wrap it

packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
```diff
@@ -86,6 +88,7 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
   // Data is always returned. For non-data panels, empty PanelData is returned.
   const data = dataWithFieldConfig!;
   const isReadyToRender = dataObject.isDataReadyToDisplay ? dataObject.isDataReadyToDisplay() : true;
+  const canCancel = $data instanceof SceneQueryRunner || ($data instanceof SceneDataTransformer && $data.state.$data instanceof SceneQueryRunner);
```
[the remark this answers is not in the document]
torkelo: @kaydelaney I would just check if $data exists, I don't think we need any of these instance of checks

packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -87,6 +88,12 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
     this.transform(this.getSourceData().state.data);
   }
 
+  public cancelQuery() {
+    if (this.state.$data instanceof SceneQueryRunner) {
+        this.getSourceData().cancelQuery?.();
+    }
```
torkelo, line 94: we do not need this instanceof check, cancel is part of the SceneDataProvider interface

torkelo, line 94:
```suggestion
    if (this.state.$data) {
        this.state.$data.cancelQuery();
    }
```

packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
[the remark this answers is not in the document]
kaydelaney: So, the instance checks are there for the case where $data is a SceneDataTransformer that doesn't have its "own" SceneQueryRunner. Since in this case the SceneQueryRunner could be shared by multiple panels, we didn't want to allow for cancellation as that would cancel other panels, which could be kind of weird. What do you think?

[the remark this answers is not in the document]
torkelo: I think that is fine (until we know otherwise) that it would also cancel results for other panels. 

if it becomes an issue we should probably add a "canCancel" or something to SceneDataProvider interface. These instance of checks make the framework less extensible so need to limit where we use them :)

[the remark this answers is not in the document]
dprokop: @kaydelaney I've pushed minor update to bring this behavior back.

packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -87,6 +88,12 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
     this.transform(this.getSourceData().state.data);
   }
 
+  public cancelQuery() {
+    if (this.getSourceData() instanceof SceneQueryRunner) {
```
torkelo, line 92: Why the instance of check here?

dprokop: We don't need it, just removed it

PR #222, opened 2023-05-31
TimeRangePicker: Default to the "isOnCanvas" true look

torkelo:
Seeing scene apps with this:
image
 

I think we should default to isOnCanvas true behavior, only core dashboards will have these in the nav toolbar (and not for long), all other use cases should use the canvas style


domasx2 (approved):
:+1:

PR #225, opened 2023-06-07
PanelBuilders: Typed API for VizPanel creation

dprokop:
This PR introduces an API for building VizPanels using core visualizations easily. It supports code support for core panel types for Options and FieldConfig. 

We need to merge https://github.com/grafana/grafana/pull/69678 first before proceeding with this ✅ 

TODO:
- [x] Add API methods for remaining core panels
- [x] Add default values handling (provided via @grafana/schema)
- [x] Add version to specific builders (waits for https://github.com/grafana/grafana/pull/69781/)
- [x] Add documentation

<details>
  <summary>📦 Published PR as canary version: <code>0.16.0--canary.225.5291385533.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@0.16.0--canary.225.5291385533.0
  # or 
  yarn add @grafana/scenes@0.16.0--canary.225.5291385533.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/core/PanelBuilders/VizPanelBuilder.test.ts
+++ b/packages/scenes/src/core/PanelBuilders/VizPanelBuilder.test.ts
@@ -97,12 +97,7 @@ describe('VizPanelBuilder', () => {
 
     it('allows options configuration', () => {
       const builder = getTestBuilder();
-      builder.setOptions({
-        complex: {
-          a: 2,
-        },
-        numeric: 2,
-      });
+      builder.setOption('complex', { a: 2 }).setOption('numeric', 2);
 
       expect(builder.build().state.options.complex!.a).toEqual(2);
       expect(builder.build().state.options.complex!.b).toEqual('text');
--- a/packages/scenes/src/core/PanelBuilders/VizPanelBuilder.test.ts
+++ b/packages/scenes/src/core/PanelBuilders/VizPanelBuilder.test.ts
@@ -1,0 +1,179 @@
+import { VizPanelBuilder } from './VizPanelBuilder';
+
+interface OptionsTest {
+  numeric: number;
+  text: string;
+  complex: {
+    a: number;
+    b: string;
+  };
+}
+
+interface FieldConfigTest {
+  numeric: number;
+  text: string;
+  complex: {
+    a: number;
+    b: string;
+  };
+}
+
+const createDefaultOptions = (): OptionsTest => ({
+  numeric: 1,
+  text: 'text',
+  complex: {
+    a: 1,
+    b: 'text',
+  },
+});
+
+const createDefaultFieldConfig = (): FieldConfigTest => ({
+  numeric: 1,
+  text: 'text',
+  complex: {
+    a: 1,
+    b: 'text',
+  },
+});
+
+const getTestBuilder = () =>
+  new VizPanelBuilder<OptionsTest, FieldConfigTest>(
+    'testplugin',
+    '1.0.0',
+    createDefaultOptions,
+    createDefaultFieldConfig
+  );
+
+const configurablePropertiesTest: Array<[string, string | boolean]> = [
+  ['title', 'My panel title'],
+  ['description', 'My panel description'],
+  ['displayMode', 'transparent'],
+  ['hoverHeader', true],
+  ['isDraggable', true],
+  ['isResizable', true],
+];
+describe('VizPanelBuilder', () => {
+  it.each(configurablePropertiesTest)('should configure %s', (property, expected) => {
+    const methodName = `set${property.charAt(0).toUpperCase()}${property.slice(1)}`;
+    const builder = getTestBuilder();
+
+    // @ts-ignore
+    builder[methodName](expected);
+
+    // @ts-ignore
+    expect(builder.build().state[property]).toEqual(expected);
+    expect(builder.build().state.pluginId).toEqual('testplugin');
+    expect(builder.build().state.pluginVersion).toEqual('1.0.0');
+  });
+
+  describe('options and field config', () => {
+    it('provides default options', () => {
+      const builder = getTestBuilder();
+      expect(builder.build().state.options).toEqual(createDefaultOptions());
+    });
+
+    it('provides default field config', () => {
+      const builder = getTestBuilder();
+      expect(builder.build().state.fieldConfig).toEqual({
+        defaults: {
+          custom: createDefaultFieldConfig(),
+        },
+        overrides: [],
+      });
+    });
+
+    it('allows options configuration', () => {
+      const builder = getTestBuilder();
+      builder.setOptions({
+        complex: {
+          a: 2,
+        },
+        numeric: 2,
+      });
+
+      expect(builder.build().state.options.complex!.a).toEqual(2);
+      expect(builder.build().state.options.complex!.b).toEqual('text');
+      expect(builder.build().state.options.numeric).toEqual(2);
+      expect(builder.build().state.options.text).toEqual('text');
+    });
+
+    it('allows standard field config configuration', () => {
+      const builder = getTestBuilder();
+      builder.setStandardConfig({
+        decimals: 2,
+        unit: 'ms',
+        displayName: 'testDisplayName',
+      });
+
+      expect(builder.build().state.fieldConfig!.defaults.decimals).toEqual(2);
+      expect(builder.build().state.fieldConfig.defaults.unit).toEqual('ms');
+      expect(builder.build().state.fieldConfig.defaults.displayName).toEqual('testDisplayName');
+    });
+
+    it('allows field config configuration', () => {
+      const builder = getTestBuilder();
+      builder.setFieldConfig({
+        complex: {
+          a: 2,
+        },
+        numeric: 2,
+      });
+
+      expect(builder.build().state.fieldConfig!.defaults.custom!.complex.a).toEqual(2);
+      expect(builder.build().state.fieldConfig.defaults.custom!.complex.b).toEqual('text');
+      expect(builder.build().state.fieldConfig.defaults.custom?.numeric).toEqual(2);
+      expect(builder.build().state.fieldConfig.defaults.custom?.text).toEqual('text');
+    });
+  });
+
+  describe('overrides', () => {
+    it('allows overrides configuration', () => {
+      const builder = getTestBuilder();
+      builder.setFieldConfigOverrides((b) =>
+        b
+          .matchFieldsWithName('fieldName')
+          .override('complex', { a: 2, b: 'text' })
+          .matchFieldsByQuery('A')
+          .override('numeric', 2)
+          .override('decimals', 2)
+      );
+
+      expect(builder.build().state.fieldConfig.overrides).toHaveLength(2);
+      expect(builder.build().state.fieldConfig.overrides).toMatchInlineSnapshot(`
+        [
+          {
+            "matcher": {
+              "id": "byName",
+              "options": "fieldName",
+            },
+            "properties": [
+              {
+                "id": "custom.complex",
+                "value": {
+                  "a": 2,
+                  "b": "text",
+                },
+              },
+            ],
+          },
+          {
+            "matcher": {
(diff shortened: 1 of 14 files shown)


packages/scenes/src/core/PanelBuilders/VizPanelBuilder.test.ts
```diff
@@ -0,0 +1,179 @@
...
+  describe('overrides', () => {
+    it('allows overrides configuration', () => {
+      const builder = getTestBuilder();
+      builder.setFieldConfigOverrides((b) =>
+        b
+          .matchFieldsWithName('fieldName')
+          .override('complex', { a: 2, b: 'text' })
+          .matchFieldsByQuery('A')
+          .override('numeric', 2)
+          .override('decimals', 2)
```
dprokop, line 138: @torkelo this overrides builder API proposal assumes that all `override` calls after `match*` call will create an override for that matcher. This test shows how it goes.

packages/scenes/src/core/PanelBuilders/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  Options as TimeSeriesOptions,
+  FieldConfig as TimeSeriesFieldConfig,
+} from '@grafana/schema/dist/esm/raw/composable/timeseries/panelcfg/x/TimeSeriesPanelCfg_types.gen';
+import { Options as TableOptions } from '@grafana/schema/dist/esm/raw/composable/table/panelcfg/x/TablePanelCfg_types.gen';
+import { Options as BarGaugeOptions } from '@grafana/schema/dist/esm/raw/composable/bargauge/panelcfg/x/BarGaugePanelCfg_types.gen';
+import { VizPanelBuilder } from './VizPanelBuilder';
+
+export const PanelBuilders = {
+  timeseries() {
+    return new VizPanelBuilder<TimeSeriesOptions, TimeSeriesFieldConfig>('timeseries', '1.0.0');
```
dprokop, line 12: Version... think we need to expose `setVersion` method rather than relying on setting version here. It's up to consumer to define which version they want to build panel with.

dprokop: Or, I think we should fix the version here to the one that the currently used schema has. Thoughts?

packages/scenes/src/core/PanelBuilders/VizPanelBuilder.test.ts
torkelo: @dprokop how does ` .override('numeric', 2)`  know it's a custom property and not a standard one?

packages/scenes/src/core/PanelBuilders/VizPanelBuilder.ts
```diff
@@ -0,0 +1,231 @@
...
+        id,
+        options,
+      },
+      properties: [],
+    });
+    return this;
+  }
+
+  public override<T extends TFieldConfig & StandardFieldConfig, K extends keyof T>(id: K, value: T[K]): this {
+    const _id = this.isStandardFieldConfigId(String(id)) ? String(id) : `custom.${String(id)}`;
```
torkelo, line 218: ah, I see it's controlled here. Clever :)

torkelo: there is a problem though. TFieldConfig and StandardFieldConfig can contain the same property name. So would be impossible to set a custom field config with the property name "color" or any of the standard property names.

packages/scenes/src/core/PanelBuilders/index.ts
torkelo: @dprokop yea, that could work. But what version is that? 10.0.0? when do we increment it?

torkelo (approved):
Looks great. Some concern conflicts between custom and standard field config properties. Maybe we can add two separate functions? One to set custom props and one to set standard?

packages/scenes/src/core/PanelBuilders/VizPanelBuilder.ts
dprokop: Good point, missed that completely. Revisiting this today, will address :)

packages/scenes/src/core/PanelBuilders/index.ts
torkelo: @dprokop this PR looks related to this  https://github.com/grafana/grafana/pull/69781

dprokop: Ah yes, if the plugin version is exposed via schema, then we have the available given the schema version being used in Scenes. Think this should be a good first step for handling the initial version.

packages/scenes/src/core/PanelBuilders/VizPanelBuilder.ts
dprokop: @torkelo I changed this a bit. There are now standard properties-specific methods available for setting and overriding field config:

```ts
builder.setUnit('ms')....
```

```ts
builder.setFieldConfigOverrides(b => b.match(...).overrideUnit(...)
```

For custom field config overrides the `b.override('propName', value)` works (it dosen't support std field config, only custom). 

I think this API is quite nice now.

packages/scenes/src/core/PanelBuilders/VizPanelBuilder.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  public setFieldConfig(fieldConfig: DeepPartial<TFieldConfig>): this {
+    this._state.fieldConfig.defaults = {
+      ...this._state.fieldConfig.defaults,
+      custom: merge(this._state.fieldConfig.defaults.custom, fieldConfig),
+    };
+    return this;
+  }
+
+  public setFieldConfigOverrides(builder: (b: FieldConfigOverridesBuilder<TFieldConfig>) => void): this {
```
dprokop, line 146: Thinking about changing the name here to `setOverrides`, thoughts? Field config doesn't provide much value, in the UI we call field config overrides just Overrides.

torkelo: @dprokop yes, agree :+1:

packages/scenes/src/core/PanelBuilders/VizPanelBuilder.test.ts
```diff
@@ -0,0 +1,391 @@
...
+          "complex": {
+            "a": 2,
+            "b": "text",
+          },
+          "numeric": 2,
+          "text": "text",
+        }
+      `);
+    });
+  });
```
torkelo, line 213: feels like there is a mismatch here between using the plain VizPanel where fieldConfig property is very different from setFieldConfig. One would think based on the naming that they would accept the same thing. It also mismatch between setFieldConfigOverrides which starts a builder that accepts both standard and custom field options. 

Maybe a function name like setCustomFieldConfig would make it more clear?

packages/scenes/src/core/PanelBuilders/VizPanelBuilder.test.ts
```diff
@@ -0,0 +1,391 @@
...
+          .matchFieldsWithName('fieldName')
+          .overrideColor({ mode: 'thresholds' })
+          .overrideDecimals(2)
+          .overrideDisplayName('testDisplayName')
+          .overrideFilterable(true)
+          .overrideMax(100)
+          .overrideMin(1)
+          .overrideNoValue('no value')
+          .overrideUnit('ms')
+          .overrideLinks([
```
torkelo, line 229: wonder if prefixing these with `set` instead of override here would make it cleaner (similar to how the default functions are defined).  That these are overrides is kind of implied already with setFieldConfigOverrides.  But not sure, might be good to be overly clear that these are overrides

packages/scenes/src/core/PanelBuilders/VizPanelBuilder.test.ts
```diff
@@ -0,0 +1,391 @@
...
+            },
+          },
+        ])
+        .setThresholds({
+          mode: ThresholdsMode.Absolute,
+          steps: [
+            { color: 'green', value: 0 },
+            { color: 'red', value: 80 },
+          ],
+        })
```
torkelo, line 128: should we help / enforce that the first should be -Infinity ?

packages/scenes/src/core/PanelBuilders/VizPanelBuilder.test.ts
```diff
@@ -0,0 +1,391 @@
...
+          "text": "text",
+        }
+      `);
+    });
+  });
+
+  describe('overrides', () => {
+    it('allows standard overrides configuration', () => {
+      const builder = getTestBuilder();
+      builder.setFieldConfigOverrides((b) =>
```
torkelo, line 218: did you plan to rename this to setOverrides?

dprokop: Yes @torkelo, but didn't work during the weekend ;) Will update shortly

dprokop:
> @dprokop how does  .override('numeric', 2) know it's a custom property and not a standard one?

@torkelo because override is typed with `TFieldConfig` - which represents panel-specific field config (https://github.com/grafana/scenes/blob/panel-builders-api/packages/scenes/src/core/PanelBuilders/FieldConfigOverridesBuilder.ts#L79). It only accepts keys/values from this interface, and this interface describes `.custom` shape only.

packages/scenes/src/core/PanelBuilders/VizPanelBuilder.test.ts
dprokop: > feels like there is a mismatch here between using the plain VizPanel where fieldConfig property is very different from setFieldConfig.


@torkelo - I'm bringing back VizPanel types for options and field config (DeepPartial) so that we suddenly do not break every VizPanel when this is merged.

IMO there is no point in using the builder pattern for setting field config (not overrides!). It's a plain simple object merge, while overrides require a specific structure to be built (matcher + properties). 

In the builder and overrides builder we have `setXXX`/`overrideXXX` methods for standard config and `override`. Agree there is a mismatch in the viz builder where we have `setXXX` for standard config and `setFieldConfig` for custom config. 

What I think can work is alignment on `override` and `setFieldConfig`. I suggest skipping `setFieldConfig` and instead using `setConfigProperty(key, value)` and then renaming `override` to `overrideConfigProperty(key, value)`. This will provide method equivalents between the two builders.

This is now pushed with [858d62d](https://github.com/grafana/scenes/pull/225/commits/858d62d1146d34f79847a97ed729422707aaa482)

packages/scenes/src/core/PanelBuilders/VizPanelBuilder.test.ts
dprokop: Wanted to be explicit about the meaning of those and I don't think `set...` is a good option here. We could align with edit UI and have `addOverride...`, but I think it's an overkill given override is a verb already.

packages/scenes/src/core/PanelBuilders/VizPanelBuilder.test.ts
```diff
@@ -0,0 +1,388 @@
...
+          "unit": "ms",
+        }
+      `);
+    });
+
+    it('allows field config configuration', () => {
+      const builder = getTestBuilder();
+
+      builder.setConfigProperty('complex', { a: 2 });
+      builder.setConfigProperty('numeric', 2);
```
torkelo, line 197: I think naming this function `setCustomFieldConfig` could be more clear, the thing that feels ambiguous with the name `setConfigProperty` is that it's too simiar to setOptions (config & options). think we need the word field to make it clear where this option ends up and maybe even the word custom (if they look at the data model).

torkelo (approved):
Looks good! setConfigProperty naming is the only thing that stands out as potentially ambiguous, what config? panel option?

packages/scenes/src/core/PanelBuilders/VizPanelBuilder.test.ts
dprokop: agree, changed that

torkelo (approved):
looks great

Issue #176, opened 2023-05-09
LayoutObject: SceneSplitter 


torkelo:
Something that enables you to drag and resize a scene split vertically or horizontally 

to be used for views like this: 
![image](https://github.com/grafana/scenes/assets/10999/a659b287-9891-4991-a635-76a8edc84c41)

kaydelaney:
Quite keen to work on this! I did some work a while ago on the current component (not sure if it ever got merged though...)

torkelo:
@kaydelaney feel free, I was not going to do it anytime soon :)

PR #229, opened 2023-06-09
SplitLayout: Add Splitter and SplitLayout

kaydelaney:
Adds a SplitLayout scene object.
I wanted to use the opportunity to write an accessible and robust Splitter component (possibly one we can add to grafana-ui). To this end it supports keyboard controls, shortcuts, aria values, and maintains proportions if a screen is resized. Min/max dimensions can also be set on splitter panes.

I think there's still room for refactoring and improvement, but it's a start.

Solves #176

### DIFF
--- a/packages/scenes-app/src/demos/split.tsx
+++ b/packages/scenes-app/src/demos/split.tsx
@@ -1,0 +1,44 @@
+import {
+  EmbeddedScene,
+  SceneAppPage,
+  SceneAppPageState,
+  SceneCanvasText,
+  SplitLayout,
+  VizPanel,
+} from '@grafana/scenes';
+import { getQueryRunnerWithRandomWalkQuery, getEmbeddedSceneDefaults } from './utils';
+
+export function getSplitTest(defaults: SceneAppPageState) {
+  return new SceneAppPage({
+    ...defaults,
+    subTitle: 'A simple demo of different flex layout options',
+    getScene: () => {
+      return new EmbeddedScene({
+        ...getEmbeddedSceneDefaults(),
+        key: 'Flex layout embedded scene',
+        body: new SplitLayout({
+          direction: 'row',
+          primary: new VizPanel({
+            pluginId: 'timeseries',
+            title: 'Dynamic height and width',
+            $data: getQueryRunnerWithRandomWalkQuery({}, { maxDataPointsFromWidth: true }),
+          }),
+          secondary: new SplitLayout({
+            ySizing: 'content',
+            direction: 'column',
+            primary: new SceneCanvasText({
+              text: 'Size to content',
+              fontSize: 20,
+              align: 'center',
+            }),
+            secondary: new SceneCanvasText({
+              text: 'Blah blah',
+              fontSize: 30,
+              align: 'center',
+            }),
+          }),
+        }),
+      });
+    },
+  });
+}
--- a/packages/scenes-app/src/demos/split.tsx
+++ b/packages/scenes-app/src/demos/split.tsx
@@ -5,6 +5,7 @@ import {
   SceneAppPageState,
   SceneCanvasText,
   SceneDataTransformer,
+  SceneFlexItem,
   SceneQueryRunner,
   SplitLayout,
   VizPanel,
@@ -62,10 +63,13 @@ const roomsTemperatureQuery = {
 };
 
 const getDynamicSplitScene = () => {
-  const defaultSecondary = new SceneCanvasText({
-    text: 'Select room to see details',
-    fontSize: 20,
-    align: 'center',
+  const defaultSecondary = new SceneFlexItem({
+    minWidth: 500,
+    body: new SceneCanvasText({
+      text: 'Select room to see details',
+      fontSize: 20,
+      align: 'center',
+    }),
   });
   const runner = new SceneQueryRunner({
     datasource: DATASOURCE_REF,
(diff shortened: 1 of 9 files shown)


domasx2:
Nice!

packages/scenes/src/components/layout/Splitter.tsx
```diff
@@ -0,0 +1,531 @@
...
+      direction: this.state.direction === 'row' ? 'column' : 'row',
+    });
+  }
+
+  public isDraggable(): boolean {
+    return false;
+  }
+}
+
+function SceneFlexLayoutRenderer({ model, parentState }: SceneFlexItemRenderProps<SplitLayout>) {
```
torkelo, line 35:
```suggestion
function SplitLayoutRenderer({ model, parentState }: SceneFlexItemRenderProps<SplitLayout>) {
```

packages/scenes/src/components/layout/Splitter.tsx
```diff
@@ -0,0 +1,531 @@
...
+import { clamp, throttle } from 'lodash';
+import React, { ComponentType, CSSProperties, useCallback, useLayoutEffect, useRef, useState } from 'react';
+
+import { SceneObjectBase } from '../../core/SceneObjectBase';
+import { SceneComponentProps, SceneObjectState, SceneObject } from '../../core/types';
+import { useUniqueId } from './grid/LazyLoader';
+
+export interface SceneFlexItemStateLike extends SceneFlexItemPlacement, SceneObjectState {}
+
+export interface SceneFlexItemLike extends SceneObject<SceneFlexItemStateLike> {}
```
torkelo, line 13: can we import these from SceneFlexLayout?

packages/scenes/src/components/layout/Splitter.tsx
```diff
@@ -0,0 +1,531 @@
...
+  return (
+
+    <Splitter direction={direction}>
+        <Prim key={primary.state.key} model={primary} parentState={model.state} />
+        <Sec key={secondary.state.key} model={secondary} parentState={model.state} />
+    </Splitter>
+  );
+}
+
+export interface SceneFlexItemPlacement {
```
torkelo, line 53: import this so don't duplicate this type :)

torkelo (commented):
This is looking awesome! great job @kaydelaney! 

 Some minor issues with duplicated type that I think can be imported without circular references (if not we will just have to move them somewhere). 

One thing missing is for the view to remember the split state when you come back to the view. I don't think we need to remember state in URL but it would be good to remember size changes in scene object state. 

Then consumers of the object could potentially add new behaviors (or via built in options) to persist state in local storage with some key (but can be added in follow-up PRs).

packages/scenes/src/components/layout/Splitter.tsx
dprokop: +1

dprokop (commented):
@kaydelaney this is looking great! I've added a dynamic split demo and I found some things we can improve:

1. Figure out a way to handle split pane size constraints (min width / height depending on direction). Now one can collapse the pane completely.
2. Figure out a way for the pane to be optional. I think this will be especially useful for the dynamic split scenario when the pane appears based on an interaction (i.e. as in the demo I have added). As you can see in the demo there is no way to set the `secondary` pane as undefined which forced me to provide it with a placeholder. I could achieve this by replacing the entire layout on table row click, but this would be a lot more code. See the recording for the demo behavior:


https://github.com/grafana/scenes/assets/2376619/3675fac6-9e40-4519-bb39-f6cde551834f

dprokop:
Btw - split layout demos are organised in tabs if you struggle to find the dynamic split demo .

packages/scenes-app/src/demos/split.tsx
```diff
@@ -0,0 +1,208 @@
...
+          direction: 'row',
+          primary: new VizPanel({
+            pluginId: 'timeseries',
+            title: 'Dynamic height and width',
+            $data: getQueryRunnerWithRandomWalkQuery({}, { maxDataPointsFromWidth: true }),
+          }),
+          secondary: new SplitLayout({
+            ySizing: 'content',
+            direction: 'column',
+            primary: new SceneCanvasText({
```
dprokop, line 38: Hmm, SceneCanvasText is not a SceneFlexItemLike, not sure why TS ain't complaining here?

dprokop: Ah I see, `SceneFlexItemLike` is all optional props.

dprokop:
@kaydelaney regarding my comment

> Figure out a way to handle split pane size constraints (min width / height depending on direction). Now one can collapse the pane completely.

I think I used it in the wrong way, hence I couldn't get constraints working. I've pushed an update to the demo with the constraints enabled :)

dprokop:
@kaydelaney please don't forget about docs update before we get this merged :)

packages/scenes/src/components/layout/Splitter.tsx
```diff
@@ -0,0 +1,494 @@
...
+import { useUniqueId } from './grid/LazyLoader';
+import { SceneFlexItemLike, SceneFlexItemPlacement } from './SceneFlexLayout';
+
+interface SplitLayoutState extends SceneObjectState, SceneFlexItemPlacement {
+  primary: SceneFlexItemLike;
+  secondary: SceneFlexItemLike;
+  direction: 'row' | 'column';
+}
+
+export class SplitLayout extends SceneObjectBase<SplitLayoutState> {
```
torkelo, line 18: think it could sense to break out the this scene object into it's own file (named SplitLayout) (this file is getting a bit long)

torkelo (approved):
think this looks good! 

Some pretty complex untested code in the Splitter callbacks (for good reasons), I think this is fine, writing tests for this would be too messy.

PR #238, opened 2023-06-26
SceneQueryRunner: Cancel previous request when starting new one

### LINKED ISSUE #233: SceneQueryRunner: Previous request not cancelled if when new one is started
`SceneQueryRunner` does not cancel previous request when starting a new one. In some cases previous one might still be running, leading to race conditions or excess work

kaydelaney:
Not 100% sure this is the right approach, but it seems to work, I think?

Closes #233


dprokop (approved):
This should be sufficient!

PR #242, opened 2023-07-04
POC: Time comparison

dprokop:
POC for handling time-over-time comparison.

Introduces `SceneTimeRangeWithComparison` that behaves like a regular `SceneTimeRange` object but also implements `MultiTimeRangeProvider` interface for providing multiple time ranges to consumers. In this case the consumer is `SceneQueryRunner` which will detect whether or not its time range is multi time range provider, and run the queries accordingly. In case of SQR being in scope of `MultiTimeRangeProvider` it will execute two query requests with the corresponding time ranges. Later, when the data is received it will re-adjust compare series timestamps based on the comparison diff and also modify returned fields' field config in order to indicate the comparison series. 

Not sure this is the best approach, but seems to be working just fine :)

### DIFF
--- a/packages/scenes/src/components/SceneTimeRangeWithComparison.tsx
+++ b/packages/scenes/src/components/SceneTimeRangeWithComparison.tsx
@@ -1,0 +1,137 @@
+import { dateTime, getTimeZone, rangeUtil, TimeRange } from '@grafana/data';
+import { sceneGraph } from '../core/sceneGraph';
+import { SceneObjectBase } from '../core/SceneObjectBase';
+import { evaluateTimeRange } from '../core/SceneTimeRange';
+import { SceneObjectUrlValues, SceneTimeRangeLike, SceneTimeRangeState } from '../core/types';
+import { SceneObjectUrlSyncConfig } from '../services/SceneObjectUrlSyncConfig';
+
+interface SceneTimeRangeWithComparisonState extends SceneTimeRangeState {
+  compareTo?: string;
+}
+
+interface MultiTimeRangeProvider extends SceneTimeRangeLike {
+  getTimeRanges(): [TimeRange, TimeRange | undefined];
+}
+
+export class SceneTimeRangeWithComparison
+  extends SceneObjectBase<SceneTimeRangeWithComparisonState>
+  implements MultiTimeRangeProvider
+{
+  protected _urlSync = new SceneObjectUrlSyncConfig(this, { keys: ['compareTo'] });
+
+  public constructor(state: Partial<SceneTimeRangeWithComparisonState>) {
+    super({
+      ...state,
+      // Fake time range, it's actually provided via closest time range object on activation
+      from: 'now-6h',
+      to: 'now',
+      value: evaluateTimeRange('now-6h', 'now', state.timeZone || getTimeZone()),
+    });
+
+    this.addActivationHandler(this._activationHandler);
+  }
+
+  public getUrlState() {
+    return { from: this.state.from, to: this.state.to };
+  }
+
+  public updateFromUrl(values: SceneObjectUrlValues) {
+    // TODO
+  }
+
+  private _activationHandler = () => {
+    const timeRangeObject = this.getTimeRangeObject();
+    const { from, to } = timeRangeObject.state;
+    this.setState({
+      from,
+      to,
+      value: evaluateTimeRange(from, to, timeRangeObject.getTimeZone()),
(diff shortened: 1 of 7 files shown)


packages/scenes/src/components/SceneTimeRangeWithComparison.tsx
```diff
@@ -0,0 +1,137 @@
...
+      from: 'now-6h',
+      to: 'now',
+      value: evaluateTimeRange('now-6h', 'now', state.timeZone || getTimeZone()),
+    });
+
+    this.addActivationHandler(this._activationHandler);
+  }
+
+  public getUrlState() {
+    return { from: this.state.from, to: this.state.to };
```
dprokop, line 35: TODO: fix

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -252,7 +261,56 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
...
+            ...request,
+            range: primaryTimeRange,
+          };
+        }
+      }
+
+      if (secondaryRequest) {
+        const diff = request.range.from.diff(secondaryRequest.range.from);
+
+        this._querySub = forkJoin([runRequest(ds, request), runRequest(ds, secondaryRequest)]).subscribe(([p, s]) => {
```
bfmatei, line 285: Is it possible that `forkJoin` could cause issues for the data streaming case here? In my PoC I relied on `combineLatestWith` operator as whenever one of the observables emits data, we would also receive the last emitted data from the other observable.

bfmatei (commented):
Awesome one @dprokop! This is actually pretty similar to what I implemented in a PoC as well. Just done right 😄.

torkelo (commented):
This looks promising! 

Just some minor issues that I think can be fixed with some refactorings. 

* Quite a bit of duplication between normal SceneTimeRange and SceneTimeRangeWithComparison 
* Breakout the double request & timeshift transform to separate files/functions (I know this is POC so not an issue with the approach just a todo for turning it into mergable PR) 

The biggest question I think we need to decide is how / where to represent the comparison time range window state. 

* Is it a special version of SceneTimeRangeLike (implements MultiTimeRangeProvider) like in this PR. Then how to fix duplication? 
* Could it be just a normal option/state for the default SceneTimeRange?  (would need to extend the SceneTimeRangeLike interface). 
* Can it be a separate scene object/state? (One that SceneQueryRunner looks specifically for) 

But generally approach looks really promising, and when the time series panel supports timeshifts natively (ie we do not need to shift time points) we can update this to set the corresponding meta property.

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -252,7 +261,56 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
...
+            series.fields.forEach((field) => {
+              // Aligne compare series time stamps with reference series
+              if (field.type === FieldType.time) {
+                field.values = field.values.map((v) => {
+                  return v + diff;
+                });
+              }
+              return (field.config = {
+                ...field.config,
+                displayName: `${field.name} (compare)`,
```
torkelo, line 297: this is a bit tricky, we should really use getFieldDisplayName here but the problem is that the DataFrame has not yet been passed through applyFieldOverrides so any display name set via field config is not yet set. Hm..

packages/scenes/src/querying/SceneQueryRunner.ts
torkelo: But I think we want to wait for both to emit (to avoid double rendering/flickering)

packages/scenes/src/querying/SceneQueryRunner.ts
dprokop: Yes, very good point!

packages/scenes/src/querying/SceneQueryRunner.ts
bfmatei: Good point!

packages/scenes/src/querying/SceneQueryRunner.ts
torkelo: wonder if we can set some kind of meta property on the field, that can be used in in getFieldDisplayName to add a `(compare)` suffix

dprokop:
closing in favor of #244

PR #243, opened 2023-07-04
SceneGridLayout: Fixes issues with unmount on every re-render

torkelo:
When testing core dashboards loaded in scenes runtime I noticed that every panel grid resize change caused big flickering due to all panels unmounting and re-mounting. 

* Root cause was a new component instance LazyLoader was created on every render (when isLazy was false)
* But to fix I ended up refactoring the LazyLoader so to avoid the unnecessary wrapping div (it just needed forwardRef)
* Also fixed the lazy demo which had two issues:
** Lazy was not enabled!! :) 
** Was just not assigning any query on every panel so was just using same data  (this likely introduced in recent refactor to use PanelBuilders)

### DIFF
--- a/packages/scenes/src/components/layout/grid/SceneGridLayoutRenderer.tsx
+++ b/packages/scenes/src/components/layout/grid/SceneGridLayoutRenderer.tsx
@@ -3,13 +3,12 @@ import ReactGridLayout from 'react-grid-layout';
 import AutoSizer from 'react-virtualized-auto-sizer';
 import { SceneComponentProps } from '../../../core/types';
 import { GRID_CELL_VMARGIN, GRID_COLUMN_COUNT, GRID_CELL_HEIGHT } from './constants';
-import { LazyLoader, Props } from './LazyLoader';
+import { LazyLoader } from './LazyLoader';
 import { SceneGridLayout } from './SceneGridLayout';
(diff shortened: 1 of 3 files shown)


packages/scenes/src/components/layout/grid/SceneGridLayoutRenderer.tsx
```diff
@@ -3,13 +3,12 @@ import ReactGridLayout from 'react-grid-layout';
...
 import { SceneComponentProps } from '../../../core/types';
 import { GRID_CELL_VMARGIN, GRID_COLUMN_COUNT, GRID_CELL_HEIGHT } from './constants';
-import { LazyLoader, Props } from './LazyLoader';
+import { LazyLoader } from './LazyLoader';
 import { SceneGridLayout } from './SceneGridLayout';
 import { SceneGridItemLike } from './types';
 
 export function SceneGridLayoutRenderer({ model }: SceneComponentProps<SceneGridLayout>) {
   const { children, isLazy } = model.useState();
-  const LazyWrapper = isLazy ? LazyLoader : ({ style, children }: Props) => <div style={style}>{children}</div>;
```
torkelo, line 12: here was the root cause of the unmounting bug, as when isLazy is false we create a new functional component here on every render

dprokop (approved):

PR #244, opened 2023-07-05
Scene: Allow time range comparison 

dprokop:
Alternative to #242 

Actually a simpler approach to the other PR. This relies on a single, new scene object `SceneTimeRangeCompare` that serves only as a compare with state provider. It also implements `MultiTimeRangeProvider` interface that allows an object to return, well, multiple (2) time ranges. 

This introduces a helper function, `getTimeCompare` that will find the closest comparer object. It's not part of the `sceneGraph` utils because it's very specific to `SceneQueryRunner` that needs multi time range resolving.

See the demos to observe different scenarios of applying the `SceneTimeRangeCompare` object.

TODO:
- [x] Support for compare options based on currently selected time range
- [x] Support for relative and absolute compare periods
- [x] Figure out a way in the field config or DataFrame to indicate series that are the comparison ones.
- [x] Tests
- [x] UI
- [x] Docs
- [x] Demo

## Release notes

You can now automatically perform queries against a secondary time range to visualize time-over-time comparisons. Use `SceneTimeRangeCompare` as in the example below:

```ts
const queryRunner = new SceneQueryRunner({
  datasource: {
    type: 'prometheus',
    uid: 'gdev-prometheus',
  },
  queries: [
    {
      refId: 'A',
      expr: 'rate(prometheus_http_requests_total{handler=~"/metrics"}[5m])',
    },
  ],
});

const scene = new EmbeddedScene({
  $data: queryRunner,
  $timeRange: new SceneTimeRange({ from: 'now-5m', to: 'now' }),
  controls: [
    new SceneTimePicker({}), 
    new SceneTimeRangeCompare({}) // Use this object to enable time frame comparison UI
  ],
  body: new SceneFlexLayout({
    direction: 'row',
    children: [
      new SceneFlexItem({
        width: '100%',
        height: '100%',
        body: PanelBuilders.timeseries().setTitle('Panel using global time range').build(),
      }),
    ],
  }),
 });
```

<details>
  <summary>📦 Published PR as canary version: <code>0.27.0--canary.244.6011359925.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@0.27.0--canary.244.6011359925.0
  # or 
  yarn add @grafana/scenes@0.27.0--canary.244.6011359925.0
  ```
</details>


packages/scenes/src/components/SceneTimeRangeCompare.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  public onCompareWithChanged = (compareWith: string) => {
+    this.setState({ compareWith });
+  };
+
+  public onClearCompare = () => {
+    this.setState({ compareWith: undefined });
+  };
+
+  public getTimeRanges(): [TimeRange, TimeRange | undefined] {
```
torkelo, line 50: I think a signature that looks like this makes more sense to me:

```
public getComparisonTimeRange(sourceTimeRange: TimeRange): TimeRange 
```

The time range to compare with source be local to the source (caller / SceneQueryRunner) not local to the SceneTimeRangeCompare object. 

A function name like that also makes more sense for the isMultiTimeRangeProvider function, if we rename the interface to TimeRangeComparer and the check function to isTimeRangeComparer (or something along those lines)

dprokop: Agree, pushed an update

dprokop:
@kaydelaney thanks for the updates!

Few cents from my side.

1. When comparing **small time ranges 0-24h** I suggest providing the following options:
  - Previous period (time range preceding selected range)
  - Day before 
  - Week before
  - Month before
  - 3 months before
  - 6 months before
  - 1 year before
2. When comparing **24h-7d** time ranges
  - Previous period (time range preceding selected range)
  - A week before
  - A month before
  - 3 months before
  - 6 months before
  - 1 year before
3. When selecting 7d+ ranges
  - Previous period (time range preceding selected range)
  - Selected options applicable based on the selected range (month before, 3 months before, 6 months before, 1 year before)

@kaydelaney since I'm not a native English speaker, I would use some of your help regarding _ago_ / _before_ wording. Feels like _ago_ applies more to relative time ranges, while _before_ sounds more suitable for absolute ranges.

Any thoughts on the above approach @bfmatei and @torkelo?

torkelo:
👍

bfmatei:
Sounds good from my PoV as well!

dprokop:
Talked with @kaydelaney yesterday, and we also gonna add possibility to customize available options. 

Talking with @Rootzreggae about app o11y requirements for this it turns out that only a limited amount of options will be necessary for their needs (1d ago, 1w ago, 1M ago).

dprokop:
Pushed some changes:
- updates according to https://github.com/grafana/grafana/pull/72715
- demos showcasing different scenarios and panels support

Please test this against Grafana's PR:  https://github.com/grafana/grafana/pull/72715

dprokop:
@torkelo, @bfmatei  - I've pushed some updates. Feels like UI and docs are the missing pieces now. I'm working on it together with @Rootzreggae

packages/scenes-app/src/demos/timeRangeComparison.tsx
```diff
@@ -0,0 +1,99 @@
...
+} from '@grafana/scenes';
+import { getQueryRunnerWithRandomWalkQuery } from './utils';
+
+export function getTimeRangeComparisonTest(defaults: SceneAppPageState) {
+  const getTimeseriesPanelWithComparison = (refId: string) =>
+    PanelBuilders.timeseries().setOverrides((b) =>
+      b
+        .matchComparisonQuery(refId)
+        .overrideCustomFieldConfig('lineStyle', {
+          fill: 'dash',
```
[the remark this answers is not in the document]
dprokop: Agree, but consider this just an example. I'd like to have this somehow predefined, but feels wrong to configure this in the SceneQueryRunner via field's `fieldConfig` given this can be overridden down the road (similarly to display name).

packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
```diff
@@ -72,8 +72,18 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
...
-    if (isSceneObject(headerActions)) {
-      actionsElement = <headerActions.Component model={headerActions} />;
+    if (Array.isArray(headerActions)) {
+      actionsElement = (
+        <>
+          {headerActions.map((action, index) => {
+            return (
+              <React.Fragment key={`headerAction-${model.state.key}-${index}`}>
+                <action.Component model={action} />
+              </React.Fragment>
```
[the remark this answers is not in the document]
dprokop: Unnecessary - old code to be removed

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -47,7 +47,7 @@ export interface VizPanelState<TOptions = {}, TFieldConfig = {}> extends SceneOb
   menu?: VizPanelMenu;
   isDraggable?: boolean;
   isResizable?: boolean;
-  headerActions?: React.ReactNode | SceneObject;
+  headerActions?: React.ReactNode | SceneObject[];
```
bfmatei, line 50: This can be a breaking change for people who already defined a SceneObject as a `headerAction`. Given that at the time being we're using minor versions, this can cause failures in build processes.

I don't know what is the right thing to do here: either clearly state it in changelog, implement this change in a separate PR for visibility purposes or allow `SceneObject[]` in addition to `SceneObject`.

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -207,7 +227,65 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
...
+
+      if (secondaryRequest) {
+        const diff = secondaryRequest.range.from.diff(request.range.from);
+
+        this._querySub = forkJoin([runRequest(ds, request), runRequest(ds, secondaryRequest)]).subscribe(([p, s]) => {
+          s.series.forEach((series) => {
+            series.refId = getCompareSeriesRefId(series.refId || '');
+            series.meta = {
+              ...series.meta,
+              // @ts-ignore Remove when https://github.com/grafana/grafana/pull/71129 is released
```
bfmatei, line 246: The PR was merged. Should this also wait for an official release that includes the change?

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -207,7 +227,65 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
...
+          this.onDataReceived({
+            ...p,
+            series: [...p.series, ...s.series],
+          });
+        });
+      } else {
+        this._querySub = runRequest(ds, request).subscribe(this.onDataReceived);
+      }
+    } catch (err) {
+      console.error('PanelQueryRunner Error', err);
```
bfmatei, line 272: Should this use the scenes logger rather than plain `console` logging?

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -226,41 +304,46 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
...
-      request.scopedVars = Object.assign({}, request.scopedVars, {
-        __interval: { text: norm.interval, value: norm.interval },
-        __interval_ms: { text: norm.intervalMs.toString(), value: norm.intervalMs },
-      });
-
-      request.interval = norm.interval;
-      request.intervalMs = norm.intervalMs;
+    request.targets = request.targets.map((query) => {
+      if (!query.datasource) {
+        query.datasource = ds.getRef();
```
bfmatei, line 309: This also mutates the original object and it might have side effects. Should it be something like the following?

```
return {
  ...query,
  datasource: query.datasource ?? ds.getRef(),
};
```

bfmatei (commented):
A couple of mentions here and there, but other than that, it looks great!

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -207,7 +227,65 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
...
+            series.fields.forEach((field) => {
+              // Align compare series time stamps with reference series
+              if (field.type === FieldType.time) {
+                field.values = field.values.map((v) => {
+                  return diff < 0 ? v - diff : v + diff;
+                });
+              }
+              return field;
+            });
+          });
```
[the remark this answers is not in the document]
dprokop: ✔️

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -283,3 +366,22 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
...
+      comparer = child;
+    }
+  });
+  if (comparer) {
+    return comparer;
+  }
+  if (sceneObject.parent) {
+    return getTimeCompare(sceneObject.parent);
+  }
+  return null;
```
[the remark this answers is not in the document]
dprokop: Not necessarily - mind that this fn looks across object's children in order to find a comparer that's supposed to be used as a control or header action.

packages/scenes/src/querying/SceneQueryRunner.ts
dprokop: It doesn't have to wait technically. As this change is backward compatible. The core change is all about providing the means for time comparison series detection later on, as well as display name calculation. 

This scenes PR works without it, it's just a display name that's not calculated. Hence if you have series A that has the comparison series, the latter will not have the `(comparison)` suffix added until Grafana 10.2.

packages/scenes/src/querying/SceneQueryRunner.ts
dprokop: Scenes logger is just a local debug util that needs to be explicitly enabled via `window.grafanaSceneLogging`.

packages/scenes/src/querying/SceneQueryRunner.ts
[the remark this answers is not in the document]
torkelo: @dprokop then I think you can use sceneGraph.findObject, it will search down and up.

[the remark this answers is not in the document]
torkelo: and I think you could start with the parent (as the time range comparer will never be directly on the query runner)

packages/scenes/src/querying/SceneQueryRunner.ts
dprokop: Nah, it's only executed when `datasource` is not set (L308).

packages/scenes/src/components/VizPanel/VizPanel.tsx
torkelo: I think can make this `React.ReactNode | SceneObject | SceneObject[]` and support both the array and non array case

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -207,7 +229,77 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
...
+      } else {
+        this._querySub = runRequest(ds, request).subscribe(this.onDataReceived);
+      }
+    } catch (err) {
+      console.error('PanelQueryRunner Error', err);
+    }
+  }
+
+  private timeShiftQueryResponse =
+    (diff: number): OperatorFunction<[PanelData, PanelData], PanelData> =>
```
torkelo, line 256: why arrow function? looks like it is called directly (and not passed as a callback)

packages/scenes/src/querying/SceneQueryRunner.ts
torkelo: @bfmatei no request.targets is a deep clone of the queries state model, so this does not mutate any state

packages/scenes/src/querying/SceneQueryRunner.ts
dprokop: To provide the time range diff via closure. But we can figure this out based on PanelData in operator, updated

packages/scenes/src/querying/SceneQueryRunner.ts
bfmatei: Got it! Thanks!

packages/scenes/src/querying/SceneQueryRunner.ts
bfmatei: Got it! Thanks!

packages/scenes/src/components/VizPanel/VizPanel.tsx
dprokop: Good catch @bfmatei  - updated in [c73b632](https://github.com/grafana/scenes/pull/244/commits/c73b632415e7ef809a28b74c9bd53a81d17ae060)

packages/scenes-app/src/demos/timeRangeComparison.tsx
```diff
@@ -0,0 +1,99 @@
...
+        })
+        .overrideColor({
+          mode: 'fixed',
+          fixedColor: 'gray',
+        })
```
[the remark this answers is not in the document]
dprokop: Approached it in https://github.com/grafana/scenes/pull/244/commits/07b0a29dd0cf6a7e4c754f020c7afe78c25b47eb


This is how comparison series look in light/dark theme:

Screenshot 2023-08-28 at 17 39 17


Screenshot 2023-08-28 at 17 38 59

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -207,7 +230,82 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
...
+          });
+        });
+
+        return {
+          ...p,
+          series: [...p.series, ...s.series],
+        };
+      })
+    );
+  }
```
torkelo, line 294: nitpick: I think this could live in a separate file/function. Needs no access to SQR or anything else in this file, it's a very simple pure function.

torkelo (commented):
Looks awesome! :rocket:

torkelo (approved):

packages/scenes/src/querying/timeShiftQueryResponseOperator.ts
```diff
@@ -0,0 +1,46 @@
...
+    map(([p, s]) => {
+      const diff = s.timeRange.from.diff(p.timeRange.from);
+      s.series.forEach((series) => {
+        series.refId = getCompareSeriesRefId(series.refId || '');
+        series.meta = {
+          ...series.meta,
+          // @ts-ignore Remove when https://github.com/grafana/grafana/pull/71129 is released
+          timeCompare: {
+            diffMs: diff,
+            isTimeShiftQuery: true,
```
leeoniya, line 17: what's the purpose of this boolean? is it not enough to see that diffMs > 0 or simply that timeCompare object exists?

PR #248, opened 2023-07-06
Tabs: Add icon and suffix

pbaumard:
Adding icon and suffix available in [PageTabs](https://github.com/grafana/grafana/blob/main/public/app/core/components/Page/PageTabs.tsx).

### DIFF
--- a/packages/scenes/src/components/SceneApp/types.ts
+++ b/packages/scenes/src/components/SceneApp/types.ts
@@ -1,3 +1,4 @@
+import { ComponentType } from 'react';
 import { SceneObject, SceneObjectState } from '../../core/types';
 import { EmbeddedScene } from '../EmbeddedScene';
 import { IconName } from '@grafana/data';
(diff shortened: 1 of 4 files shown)


packages/scenes/src/components/SceneApp/types.ts
```diff
@@ -32,8 +33,10 @@ export interface SceneAppPageState extends SceneObjectState {
   renderTitle?: (title: string) => React.ReactNode;
   /** For an image before title */
   titleImg?: string;
-  /** For an icon before title */
+  /** For an icon before title or tab label */
   titleIcon?: IconName;
+  /** For a tab label suffix */
+  tabSuffix?: ComponentType<{ className?: string; }>,
```
torkelo, line 39: Let's add tabCounter while we are at it

pbaumard: Tab counter is not in [PageTabs](https://github.com/grafana/grafana/blob/main/public/app/core/components/Page/PageTabs.tsx), but only in [Tab](https://github.com/grafana/grafana/blob/main/packages/grafana-ui/src/components/Tabs/Tab.tsx).

In addition adding a counter is as easy as:

```js
tabSuffix: () => <Counter value={1} />,
```

So I believe adding tab counter property is outside this PR scope.

torkelo: ah i see.

torkelo (approved):

PR #254, opened 2023-07-11
Behaviors: Enabled type stateless behavior params

torkelo:
I can't figure out any other way than add this any here to
make it possible to have behavior functions that take in a typed SceneObject

### DIFF
--- a/packages/scenes/src/core/types.ts
+++ b/packages/scenes/src/core/types.ts
@@ -24,7 +24,7 @@ export interface SceneObjectState {
    * Can be used to add extra behaviors to a scene object.
    * These are activated when the their parent scene object is activated.
    */
-  $behaviors?: Array<SceneObject | SceneStatelessBehavior>;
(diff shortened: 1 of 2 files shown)


packages/scenes/src/core/types.ts
```diff
@@ -24,7 +24,7 @@ export interface SceneObjectState {
    * Can be used to add extra behaviors to a scene object.
    * These are activated when the their parent scene object is activated.
    */
-  $behaviors?: Array<SceneObject | SceneStatelessBehavior>;
+  $behaviors?: Array<SceneObject | SceneStatelessBehavior<any>>;
```
dprokop, line 27: `unknown` maybe?

dprokop (approved):

packages/scenes/src/core/types.ts
torkelo: @dprokop  sadly not :(
Screenshot 2023-07-11 at 14 04 51

PR #259, opened 2023-07-12
Behaviors: Provide behavior for visualization cursor sync

### LINKED ISSUE #256: Panels: Add support for shared tooltip and crosshair
We'd like to add shared crosshairs/tooltips to the panels in the app plugin we're creating with scenes. In Grafana, this is set with the `graphTooltip: number` field. 

Screenshot 2023-07-11 at 8 25 50 AM

dprokop:
Enables configuring cursor sync in Scenes. Closes #256 

Allows configuring multiple sync context within a scene, just attach a `behaviors.CursorSync` instance to $behaviors array in your scene, and all visualizations under that scene object will become a subject of cursor sync. To configure global cursor sync, attach behavior to the root of a scene.

Requires https://github.com/grafana/grafana/pull/71561 for scoped syncs to work. This is backward compatible, but will sync globally until https://github.com/grafana/grafana/pull/71561 is released.

https://github.com/grafana/scenes/assets/2376619/cb96a72c-f661-46d9-b8c4-8a459bb998db

<details>
  <summary>📦 Published PR as canary version: <code>0.23.0--canary.259.5599018850.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@0.23.0--canary.259.5599018850.0
  # or 
  yarn add @grafana/scenes@0.23.0--canary.259.5599018850.0
  ```
</details>


packages/scenes/src/behaviors/EnableCursorSync.ts
```diff
@@ -0,0 +1,43 @@
+import { DashboardCursorSync } from '@grafana/schema';
+import { SceneObjectBase } from '../core/SceneObjectBase';
+import { SceneObject, SceneObjectState } from '../core/types';
+
+interface EnableCursorSyncState extends SceneObjectState {
+  // The name of the variable to subscribe to changes to.
```
dprokop, line 6:
```suggestion
```

packages/scenes-app/src/demos/cursorSync.tsx
```diff
@@ -0,0 +1,98 @@
...
+    getScene: () => {
+      return new EmbeddedScene({
+        ...getEmbeddedSceneDefaults(),
+        key: 'Flex layout embedded scene',
+        body: new SceneFlexLayout({
+          direction: 'column',
+          children: [
+            new SceneFlexLayout({
+              direction: 'row',
+              $behaviors: [new behaviors.EnableCursorSync({ key: 'sync1', sync: DashboardCursorSync.Tooltip })],
```
dprokop, line 27: Cursor sync behavior creates a scope for cursor sync. So as you can see in this demo, there are cursor syncs defined for different fragments of a scene. Think that's neat!

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -284,4 +275,29 @@ export class VizPanel<TOptions = {}, TFieldConfig = {}> extends SceneObjectBase<
       legend: { ...legendOptions, sortBy, sortDesc },
     } as TOptions);
   };
+
+  private buildPanelContext() {
+    const sync = getCursorSyncScope(this);
+
+    this._panelContext = {
+      eventBus: sync ? sync.getEventsBus() : getAppEvents(),
```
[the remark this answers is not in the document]
dprokop: @torkelo  This is taking advantage of this. Mind that sync.getEventBus is using scoped bus from the sync object. There is no custom event bus created - this is basically the same implementation we have in core. 

I'm using the scoped bus just because I need an addressable bus in order to configure uPlot correctly, see https://github.com/grafana/grafana/pull/71434/files/e5c8aff36eae5dfd87af4729a619e23c67d9c034#diff-57e07daf7609a601c0e85dc9c2122c20dc0f81ce10240849066aecec310a4f0eR604

packages/scenes/src/behaviors/EnableCursorSync.ts
```diff
@@ -0,0 +1,42 @@
...
+
+interface EnableCursorSyncState extends SceneObjectState {
+  sync: DashboardCursorSync;
+}
+
+/**
+ * This behavior will provide a cursor sync context within a scene.
+ */
+
+export class EnableCursorSync extends SceneObjectBase<EnableCursorSyncState> {
```
dprokop, line 13: Not happy with naming here. Should be just `behaviors.CursorSync` maybe?

packages/scenes/src/behaviors/EnableCursorSync.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+  });
+
+  if (sync) {
+    return sync;
+  }
+  if (sceneObject.parent) {
+    return getCursorSyncScope(sceneObject.parent);
+  }
+  return null;
```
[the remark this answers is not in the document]
dprokop: Yeah, damn always forgetting about this util 😬

packages/scenes/src/behaviors/EnableCursorSync.ts
torkelo: @dprokop yea, was thinking the same thing, the word Enable does not add much here

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -284,4 +275,29 @@ export class VizPanel<TOptions = {}, TFieldConfig = {}> extends SceneObjectBase<
...
+      app: CoreApp.Unknown, // TODO,
+      sync: () => {
+        if (sync) {
+          return sync.state.sync;
+        }
+        return DashboardCursorSync.Off;
```
kaydelaney, line 289:
```suggestion
        return sync?.state.sync ?? DashboardCursorSync.Off
```

kaydelaney (approved):
Looks good and works well! Left one small nit suggestion, but that's it I think

packages/scenes/src/behaviors/CursorSync.ts
```diff
@@ -0,0 +1,77 @@
...
+    });
+  }
+
+  public getEventsBus = (panel: SceneObject) => {
+    if (!this.parent) {
+      throw new Error('EnableCursorSync cannot be used as a standalone scene object');
+    }
+    // Since EnableCursorSync is a behavior, it is not a parent to any object in the scene graph.
+    // We need to get it's parent in order to provide correct EventBus context to the children.
+    return new PanelContextEventBus(this.parent, panel);
```
dprokop, line 30: @torkelo - so I did not find a way to avoid creating this proxy per panel. Feels immposible to allow a single bus to publish events with different origins without doing changes to core as well.

But - I think with the current implementation that under the hood using the events bus from the parent, the scoping is guaranteed and the filtering done self-origin filtering done in PanelContextEventBus is working as expected, making https://github.com/grafana/grafana/blob/panel-context-events-scope/packages/grafana-ui/src/components/GraphNG/GraphNG.tsx#L181 irrelevant in Scenes.

packages/scenes/src/behaviors/CursorSync.ts
```diff
@@ -0,0 +1,77 @@
...
+  public getStream<T extends BusEvent>(eventType: BusEventType<T>): Observable<T> {
+    return new Observable<T>((observer) => {
+      const handler = (event: T) => {
+        observer.next(event);
+      };
+
+      const sub = this._source.subscribeToEvent(eventType, handler);
+
+      return () => sub.unsubscribe();
+    }).pipe(filter((event) => (event as any).origin !== this._eventsOrigin));
```
torkelo, line 61:
```suggestion
    return new Observable<T>((observer) => {
      const handler = (event: T) => {
         if (event.origin !== this) {
           observer.next(event);
         } 
      };

      const sub = this._source.subscribeToEvent(eventType, handler);

      return () => sub.unsubscribe();
    });
```

less rxjs the better :)

packages/scenes/src/behaviors/CursorSync.ts
```diff
@@ -0,0 +1,77 @@
...
+    return this.parent.state.key;
+  };
+}
+
+// This serves as a shared EventsBus that is shared by children or CursorSync behavior.
+class PanelContextEventBus implements EventBus {
+  public constructor(private _source: SceneObject, private _eventsOrigin: SceneObject) {}
+
+  public publish<T extends BusEvent>(event: T): void {
+    (event as any).origin = this._eventsOrigin;
```
torkelo, line 48:
```suggestion
    (event as any).origin = this;
```

I think origin needs to be the bus, as that is what we GraphNG is comparing against, but since that filtering is done internally here I guess we can decide on making origin something else

packages/scenes/src/behaviors/CursorSync.ts
dprokop: Yes, but this assumes the origin is events bus - no filtering it here means that the comparison in GraphNG https://github.com/grafana/grafana/blob/panel-context-events-scope/packages/grafana-ui/src/components/GraphNG/GraphNG.tsx#L181 will fail.

packages/scenes/src/behaviors/CursorSync.ts
dprokop: Yeah, with this change the filtering in GraphNG is never called. I'll open a separate work that will address that in the core Grafana.

PR #261, opened 2023-07-13
Allow template variables to be cancelled

kaydelaney:
Allows the loading indicator to be clicked in order to cancel a loading template variable.

Follow up to https://github.com/grafana/scenes/pull/220

### DIFF
--- a/packages/scenes/src/variables/variants/query/QueryVariable.tsx
+++ b/packages/scenes/src/variables/variants/query/QueryVariable.tsx
@@ -25,6 +25,8 @@ import { createQueryVariableRunner } from './createQueryVariableRunner';
 import { metricNamesToVariableValues } from './utils';
 import { toMetricFindValues } from './toMetricFindValues';
 import { getDataSource } from '../../../utils/getDataSource';
+import { SceneVariableSet } from '../../sets/SceneVariableSet';
+import { getClosest } from '../../../core/sceneGraph/utils';
 
 export interface QueryVariableState extends MultiValueVariableState {
   type: 'query';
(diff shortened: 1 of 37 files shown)


packages/scenes/src/variables/variants/query/QueryVariable.tsx
```diff
@@ -100,6 +102,12 @@ export class QueryVariable extends MultiValueVariable<QueryVariableState> {
     );
   }
 
+  public cancel() {
+    this.setState({ loading: false });
+    const sceneVarSet = getClosest(this, (s) => s instanceof SceneVariableSet ? s : undefined);
+    sceneVarSet?.cancel(this);
+  }
```
torkelo, line 109: think you can move this cancel function implementation to the abstract class

packages/scenes/src/variables/variants/query/QueryVariable.tsx
```diff
@@ -100,6 +102,12 @@ export class QueryVariable extends MultiValueVariable<QueryVariableState> {
     );
   }
 
+  public cancel() {
+    this.setState({ loading: false });
+    const sceneVarSet = getClosest(this, (s) => s instanceof SceneVariableSet ? s : undefined);
```
torkelo, line 107:
```suggestion
    const sceneVarSet = this.parent as SceneVariableSet;
```

We assume variables are inside variable sets (at least for now) :)

torkelo (commented):
Looks good, tricky typing issue with the onCancel method. 

Only other option I can see is wrapping the Select in a React context that contains the variable, and accessing the context from the LoadingIndicator.

packages/scenes/src/variables/variants/query/QueryVariable.tsx
kaydelaney: Hmm, I get some type errors related to
```typescript
this.setState({ loading: false });
```
if I move it to the abstract class:
"Argument of type '{ loading: false; }' is not assignable to parameter of type 'Partial\<TState\>'. ts(2345)"

torkelo: @kaydelaney  try using setStateHelper (function on MultiValueVariable that works around a typescript issue)

dprokop (approved):
Looks and works well. One thing missing is some tests of the cancellation flow, think it would be worth adding those to make sure that cancellation is propagated correctly.

PR #262, opened 2023-07-13
SceneDebugger: Scene graph explore & state viewer

torkelo:
The idea is to be able to inspect the scene graph and inspect state (see video below). 

When the query parameter "scene-debugger" is detected the debug button is added to SceneAppPage controls (top right). When clicked it opens a drawer  

Debugger can 
* Show full scene graph 
* Select scene object and show scene object details 

Todo
* [ ] Show nested object states (inspect data etc, queries, etc) 

Screenshot 2023-07-13 at 17 23 38

[Grafana-Monitoring---Scenes---Dashboards---Grafana.webm](https://github.com/grafana/scenes/assets/10999/c39a2559-cabe-4183-b113-b59ad4249131)


bfmatei:
I was talking to @dprokop pretty recently that something like this would be extremely useful. I was planning to build a Chrome extension for it but having it "inline" is way better.

This would help with understanding the structure of a Scenes project as well as determining if there's a hierarchical issue caused by custom objects.

packages/scenes/src/index.ts
```diff
@@ -45,6 +45,7 @@ export { SceneFlexLayout, SceneFlexItem, type SceneFlexItemState } from './compo
 export { SceneGridLayout, SceneGridItem } from './components/layout/grid/SceneGridLayout';
 export { SceneGridRow } from './components/layout/grid/SceneGridRow';
 export { SplitLayout } from './components/layout/split/SplitLayout';
+export { SceneDebugger } from './components/SceneDebugger/SceneDebugger';
```
dprokop, line 48: Maybe let's wait with making this a public API until it's a bit more usable?

dprokop (approved):
Nice, this is a good start!

packages/scenes/src/index.ts
torkelo: @dprokop good catch! not sure why I exported this in the first place :)

PR #276, opened 2023-08-11
Adding developer portal dev workflow and config

tolzhabayev:
This PR:

- Adds a github action that deploys dataplane docusaurus content into dev portal dev bucket, triggered manually
- Adds a docusaurus config necessary to remove the landing page so it can work from a subdirectory

We will simplify the docusaurus configs later once production deployment is in place


tolzhabayev:
merging, it only has affect on the dev portal deployment.

PR #284, opened 2023-08-16
SceneObject: Warn if parent is already set to another SceneObject

torkelo:
I think something like this is good to have as I found some bugs in scene apps related to reusing scene instances in multiple scenes.

### DIFF
--- a/packages/scenes/src/core/SceneObjectBase.tsx
+++ b/packages/scenes/src/core/SceneObjectBase.tsx
@@ -81,6 +81,14 @@ export abstract class SceneObjectBase<TState extends SceneObjectState = SceneObj
 
   private _setParent() {
     this.forEachChild((child) => {
+      // If we already have a parent and it's not this, then we likely have a bug
+      if (child._parent && child._parent !== this) {
+        console.warn(
+          'SceneObject already has a parent that is different from new parent. You cannot share the same instance in multiple scenes or in multiple different places in the same scene. Use SceneObject.clone() to duplicate a SceneObject or store a key reference to the scene object.',
+          child,
+          this
+        );
+      }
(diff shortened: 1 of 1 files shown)


packages/scenes/src/core/SceneObjectBase.tsx
```diff
@@ -81,6 +81,14 @@ export abstract class SceneObjectBase<TState extends SceneObjectState = SceneObj
 
   private _setParent() {
     this.forEachChild((child) => {
+      // If we already have a parent and it's not this, then we likely have a bug
+      if (child._parent && child._parent !== this) {
+        console.warn(
+          'SceneObject already has a parent that is different from new parent. You cannot share the same instance in multiple scenes or in multiple different places in the same scene. Use SceneObject.clone() to duplicate a SceneObject or store a key reference to the scene object.',
```
dprokop, line 87: > or store a key reference to the scene object

@torkelo  - I'm not sure what you mean by that.

dprokop (approved):

packages/scenes/src/core/SceneObjectBase.tsx
torkelo: I mean store the key of a scene object and use sceneGraph.findObject to get it

torkelo: @dprokop updated the text to:

> SceneObject already has a parent that is different from the new parent. You cannot share the same SceneObject instance in multiple scenes or in multiple different places of the same scene graph. Use SceneObject.clone() to duplicate a SceneObject or store a state key reference and use sceneGraph.findObject to locate it.

kaydelaney (approved):

PR #287, opened 2023-08-17
VizPanel: Handle plugin not found scenario correctly

dprokop:
I've noticed a runtime error appearing when there is plugin not found error comming from panel plugin import. That's because [an error thrown from `importPanelPlugin` is not wrapped in a promise](https://github.com/grafana/grafana/blob/dash-loaders-lib-panel/public/app/features/plugins/importPanelPlugin.ts#L20), hence we cannot rely on `Promise.catch` in VizPanel.

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanel.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.tsx
@@ -13,6 +13,7 @@ import {
   compareArrayValues,
   compareDataFrameStructures,
   applyFieldOverrides,
+  PluginType,
 } from '@grafana/data';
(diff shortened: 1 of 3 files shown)


packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -303,3 +307,31 @@ export class VizPanel<TOptions = {}, TFieldConfig = {}> extends SceneObjectBase<
     };
   }
 }
+
+function getPanelPluginNotFound(id: string, silent?: boolean): PanelPlugin {
```
dprokop, line 311: This is pretty much copied from the core, but it does not render any component, as error status message is handled by PanelChrome which seems sufficient.

torkelo (approved):

PR #288, opened 2023-08-17
SceneQueryRunner: Fixes issues when being cloned 

torkelo:
* I noticed a bug when you drag a panel from one row to another the panel is blank because SceneQueryRunner is cloned in this scenario and we lose the instance prop state of hasFetchedData, so I moved this to scene object state 

I am trying a new pattern to mark internal scene object state props with an underscore. 

<details>
  <summary>📦 Published PR as canary version: <code>0.25.0--canary.288.5940611667.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@0.25.0--canary.288.5940611667.0
  # or 
  yarn add @grafana/scenes@0.25.0--canary.288.5940611667.0
  ```
</details>


packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
```diff
@@ -349,10 +340,10 @@ function sortGridLayout(layout: ReactGridLayout.Layout[]) {
...
-
-function isSceneGridLayout(child: SceneObject): child is SceneGridLayout {
-  return child instanceof SceneGridLayout;
+function forceRenderChildren(child: SceneGridItemLike) {
+  if (child instanceof SceneGridItem) {
+    child.state.body?.forceRender();
+  } else if (child instanceof SceneGridRow) {
+    child.forEachChild(forceRenderChildren);
+  }
 }
```
torkelo, line 349: this won't work if we have object that intern wrap VizPanels so I think we have to make it so that VizPanel subscribes to the layout state, so we have to change the SceneLayout and remove the isDraggable function, and move it to SceneLayoutState 

We could also revert to the old SceneComponentProps interface where isEditing is always passed to all scene objects (this made sure going into edit mode always triggered a full tree re-render)

torkelo, line 349: this won't work if we have object that intern wrap VizPanels so I think we have to make it so that VizPanel subscribes to the layout state, so we have to change the SceneLayout and remove the isDraggable function, and move it to SceneLayoutState 

We could also revert to the old SceneComponentProps interface where isEditing is always passed to all scene objects (this made sure going into edit mode always triggered a full tree re-render)

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -29,9 +29,10 @@ export interface QueryRunnerState extends SceneObjectState {
   minInterval?: string;
   maxDataPoints?: number;
   liveStreaming?: boolean;
-  // Non persisted state
   maxDataPointsFromWidth?: boolean;
-  isWaitingForVariables?: boolean;
+  // Private runtime state
+  _isWaitingForVariables?: boolean;
+  _hasFetchedData?: boolean;
```
torkelo, line 35: trying a new pattern of marking private internal state with a undescore prefix

packages/scenes/src/components/layout/grid/SceneGridItem.tsx
```diff
@@ -0,0 +1,36 @@
...
+  return <body.Component model={body} />;
+}
+
+export function isSceneGridRow(child: SceneObject): child is SceneGridRow {
+  return child instanceof SceneGridRow;
+}
+
+function isSceneGridLayout(child: SceneObject): child is SceneGridLayout {
+  return child instanceof SceneGridLayout;
+}
```
torkelo, line 36: no changes here, just moved it to it's own file

packages/scenes/src/components/layout/grid/SceneGridLayoutRenderer.tsx
```diff
@@ -50,15 +50,17 @@ export function SceneGridLayoutRenderer({ model }: SceneComponentProps<SceneGrid
               onLayoutChange={model.onLayoutChange}
               isBounded={false}
             >
-              {layout.map((gridItem) => {
+              {layout.map((gridItem, index) => {
                 const sceneChild = model.getSceneLayoutChild(gridItem.i)!;
+                // This it have panels higher up the page have higher z-index
+                const style = { zIndex: layout.length - index };
```
torkelo, line 56: implements: https://github.com/grafana/grafana/pull/73401

dprokop:
Implementation wise I think this looks good.

My concern is mostly about where this utility (`forceRenderChilden`) should live, given it solves a problem that is and should not be very common. Given that, I would expect `forceRenderChilden` utility to live close to Dashboard Edit in core, and be graduated to Scenes if we see a broader need for that. 

This approach also may have performance implications - we don't know the structure of the scene, or how deep in the structure the object that needs to react to change is. I wonder if we are not opening a path for overuse of this pattern and expensive re-renders of deep sub-trees of a Scene. 

With the [TExtraProps](https://github.com/grafana/scenes/pull/294) we would follow a familiar React pattern of just passing props where necessary, simple (from API user perspective) and declarative (aligned with general Scenes pattern).

torkelo:
@dprokop updated the PR (removed the sceneGraph.forceRenderChildren and moved it to the main repo PR)

PR #289, opened 2023-08-17
SceneGridLayout: Alternative isDraggable propagation 

torkelo:
Alt to https://github.com/grafana/scenes/pull/288 

* This changes so that we propagate isDraggable (and dragClass and dragClassCancel via) via props instead. We can move back dragClass and dragClassCancel to functions if we want they don't really change but figured it be good to move all the layout specific things VizPanel need to props
* Removes isDraggable from VizPanel (it already exists on SceneGridItem I think this is a leftover) 

There is a third option and that is to introduce a new type of edit manager like we had before that components can subscribe to and instead of having isDraggable / isResizable state on the SceneGridLayout we have a the edit manager have the isEditing state, but we still migtht need isDraggable and isResizable on per SceneGridItem level and then  isDraggable needs to be propagated somehow to VizPanel


packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -44,8 +44,6 @@ export interface VizPanelState<TOptions = {}, TFieldConfig = {}> extends SceneOb
   displayMode?: 'default' | 'transparent';
   hoverHeader?: boolean;
   menu?: VizPanelMenu;
-  isDraggable?: boolean;
-  isResizable?: boolean;
```
torkelo, line 48: I think these are leftovers

PR #294, opened 2023-08-22
SceneObject: Typed extra props to SceneComponents

torkelo:
In some rare scenarios we have situations where a state change in a parent needs to trigger re-renders of children (that also care about this state change). 

Example here is isDraggable on SceneGridLayout, this property is something VizPanel cares about, and it can also be overriden by a state property on SceneGridItem level. So we need some system to cascade this state change down to VizPanel. 

Options
* We can manually force render layout children (Explored in https://github.com/grafana/scenes/pull/288) 
* We can pass untyped extra props to children via util function renderSceneComponentWithExtraProps, explored in https://github.com/grafana/scenes/pull/289 

This PR explores a third-typed approach to extra props. Sadly this complicated SceneObject, SceneObjectBase, SceneComponent, and SceneComponentProps types by requiring a second generic param. I don't like this extra type complexity for what should be a uncommon scenario, so really torn on the best approach here.

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
@@ -9,8 +9,9 @@ import { sceneGraph } from '../../core/sceneGraph';
 import { isSceneObject, SceneComponentProps } from '../../core/types';
 
 import { VizPanel } from './VizPanel';
+import { SceneGridItemExtraProps } from '../layout/grid/types';
 
-export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
+export function VizPanelRenderer({ model, isDraggable }: SceneComponentProps<VizPanel, SceneGridItemExtraProps>) {
   const {
     title,
     description,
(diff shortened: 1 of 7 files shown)


dprokop:
@torkelo - should we close this for now and possibly reopen it in the future when we see more need for this?

bfmatei:
Hey yo @torkelo @dprokop !

I finally found an edge case for this: I need to render a Scene component in one place or another due to how the layout is built. Unfortunately, force-rendering the parent scene will cause a nasty re-render on the screen as well, hence my workaround for this would be to do something like this:

```jsx
function parentSceneRender({ model }: SceneComponentProps<ParentScene>) {
  const { myScene } = model.useState();

  return (
    <div>
      {/* This is where is should be rendered when myScene is expanded */}
      <myScene.Component model={myScene} isExpanded={true} />

      <div>
        {/* This is where is should be rendered when myScene is NOT expanded */}  
        <myScene.Component model={myScene} isExpanded={false} />
      </div>
    </div>
  );
}
```

Then in the renderer of `MyScene`:
```jsx
function mySceneRender({ model, renderExpanded }: SceneComponentProps<MyScene> & { renderExpanded: boolean }) {
  const { isExpanded } = model.useState();

  // Prevent rendering when the expected type of renderer is not the same as the internal flag
  if (isExpanded !== renderExpanded) {
    return null;
  }

  return (
    <>
      {/* ... */}
    </>
  );
}
```

Do you have an alternative for this?

torkelo:
Not sure i understand, do you want to render the same scene object in two locations at the same time but  be rendered differently?

PR #295, opened 2023-08-23
DO NOT MERGE: Use latest schema API

dprokop:
This is just prep for an eventual bump of peer dependencies when they are released.


leventebalogh:
I have just cut a [pull request](https://github.com/grafana/scenes/pull/532) which updates the peer deps to `10.2.3`, however it doesn't have the `defaultFieldConfig` import change - would that be necessary, too?

dprokop:
@leventebalogh - no, this change won't be necessary. I'll close this PR in favor of yours.

PR #297, opened 2023-08-28
sceneUtils: cloneSceneObjectState

torkelo:
Have a scenario where I want to only clone the state of a scene object, and not it's containing scene object instance.

This is for the discard mechanic found in https://github.com/grafana/grafana/pull/73873, If I do a full clone and then try to reuse that state with another
instance I hit my setParent warning of changing scene parent.

DashboardScene want's to copy it's current state so it can revert back to it, so it wants to clone it's state (but not itself)

### DIFF
--- a/packages/scenes/src/core/sceneGraph/utils.ts
+++ b/packages/scenes/src/core/sceneGraph/utils.ts
@@ -9,7 +9,15 @@ export function cloneSceneObject<T extends SceneObjectBase<TState>, TState exten
   sceneObject: SceneObjectBase<TState>,
   withState?: Partial<TState>
 ): T {
-  const clonedState = { ...sceneObject.state };
+  const clonedState = cloneSceneObjectState(sceneObject.state, withState);
+  return new (sceneObject.constructor as any)(clonedState);
+}
+
+export function cloneSceneObjectState<TState extends SceneObjectState>(
+  sceneState: TState,
+  withState?: Partial<TState>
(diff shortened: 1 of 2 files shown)


packages/scenes/src/core/sceneGraph/utils.ts
```diff
@@ -9,7 +9,15 @@ export function cloneSceneObject<T extends SceneObjectBase<TState>, TState exten
...
   withState?: Partial<TState>
 ): T {
-  const clonedState = { ...sceneObject.state };
+  const clonedState = cloneSceneObjectState(sceneObject.state, withState);
+  return new (sceneObject.constructor as any)(clonedState);
+}
+
+export function cloneSceneObjectState<TState extends SceneObjectState>(
+  sceneState: TState,
+  withState?: Partial<TState>
```
dprokop, line 18: This arg seems not used neither in scenes nor in core. Do we need it? This can be handled (if needed) outside of the library.

torkelo: @dprokop 
we use here: https://github.com/grafana/scenes/blob/main/packages/scenes/src/components/layout/grid/SceneGridLayout.tsx#L218

dprokop: ah yeah, indirecty :) Missed that

dprokop (approved):

PR #303, opened 2023-08-30
SceneQueryRunner: Allow external DataQueryRequest context configuration

dprokop:
<details>
  <summary>📦 Published PR as canary version: <code>0.27.1--canary.303.6024412821.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@0.27.1--canary.303.6024412821.0
  # or 
  yarn add @grafana/scenes@0.27.1--canary.303.6024412821.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.test.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.test.ts
@@ -138,6 +138,93 @@ describe('SceneQueryRunner', () => {
         }
       `);
     });
+
+    describe('DataQueryRequest decoration', () => {
+      it('should allow DataQueryRequest object decoration', async () => {
+        Date.now = jest.fn(() => 1689063488000);
+        const queryRunner = new SceneQueryRunner({
+          queries: [{ refId: 'A' }],
+          $timeRange: new SceneTimeRange(),
+          getDataQueryRequestContext: () => {
+            return {
+              app: 'my-app',
+              panelId: 100,
+            };
+          },
+        });
+
+        queryRunner.activate();
+
+        await new Promise((r) => setTimeout(r, 1));
+
+        expect(sentRequest).toBeDefined();
+        const { scopedVars, ...request } = sentRequest!;
+
+        expect(request.app).toBe('my-app');
+        expect(request.panelId).toBe(100);
+        expect(request).toMatchInlineSnapshot(`
+          {
+            "app": "my-app",
+            "interval": "30s",
+            "intervalMs": 30000,
+            "liveStreaming": undefined,
+            "maxDataPoints": 500,
+            "panelId": 100,
+            "range": {
+              "from": "2023-07-11T02:18:08.000Z",
+              "raw": {
+                "from": "now-6h",
+                "to": "now",
+              },
+              "to": "2023-07-11T08:18:08.000Z",
+            },
+            "rangeRaw": {
+              "from": "now-6h",
+              "to": "now",
+            },
+            "requestId": "SQR101",
+            "startTime": 1689063488000,
+            "targets": [
+              {
+                "datasource": {
+                  "uid": "test",
+                },
+                "refId": "A",
+              },
+            ],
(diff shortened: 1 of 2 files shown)


packages/scenes/src/querying/SceneQueryRunner.test.ts
```diff
@@ -138,6 +138,61 @@ describe('SceneQueryRunner', () => {
         }
       `);
     });
+
+    it('should allow DataQueryRequest object decoration', async () => {
+      Date.now = jest.fn(() => 1689063488000);
+      const queryRunner = new SceneQueryRunner({
+        queries: [{ refId: 'A' }],
+        $timeRange: new SceneTimeRange(),
+        getDataQueryRequestContext: () => ({
```
torkelo, line 147: This function could not be cloned as it will always need to rely on function scope captured variables.

torkelo (commented):
hm.. wonder if this really should be on SceneQueryRunner itself, or if it should be something that is defined once higher up. 

Thinking of the app plugin use case and having to remember to set this on every SceneQueryRunner etc. 

One other option would be for SQR to look up it's parent stack for any scene object that has a DataQueryContextProvider. hm.. but it would probably mean adding another state prop to SceneObjectState (something like $queryContext: DataQueryContextProvider). 

Another option is to have scene objects that have a context provider via the SceneObject interface (similar to urlSync and variableDependency). This would work great for DashboardScene as it could then simply set & implement that. But for scene app plugins it's not so straight forward as they usually don't implement a custom scene object root (but use SceneAppPage or EmbeddedScene). 

But wait we could maybe add this feature to SceneApp (that implements this provider). 

If you think this is worth testing I can throw together a poc to show what I mean

torkelo (commented):
hm.. wonder if this really should not be SceneQueryRunner itself, or if it should be something that is defined once higher up. 

Thinking of the app plugin use case and having to remember to set this on every SceneQueryRunner etc. 

One other option would be for SQR to look up it's parent stack for any scene object that has a DataQueryContextProvider. hm.. but it would probably mean adding another state prop to SceneObjectState (something like $queryContext: DataQueryContextProvider). 

Another option is to have scene objects that have a context provider via the SceneObject interface (similar to urlSync and variableDependency). This would work great for DashboardScene as it could then simply set & implement that. But for scene app plugins it's not so straight forward as they usually don't implement a custom scene object root (but use SceneAppPage or EmbeddedScene). 

But wait we could maybe add this feature to SceneApp (that implements this provider). 

If you think this is worth testing I can throw together a poc to show what I mean

torkelo:
Something like https://github.com/grafana/scenes/compare/encricher-test?expand=1 
we could provide maybe some good default implementations where scene app plugins can specify the app name, and we have some default heuristic of looking for the closest VizPanel and use the title (after we add something other than panelId for scene apps to use)

dprokop:
> or are we over-enginerring this? could just add app & panelId to SQR state and leave it up to user code to set and sync. For core dashboards, panelId should not really change during the lifetime of a dashboard scene.

I don't think so. For the app, I think it would be worth to have some context provider so that SQR could read that property from it. Otherwise one need to remember to provide correct context to each scene query runner.

The only thing i don't know is how the enricher results should be collected. I.e. :
- VizPanel can be considered an enricher that provides `panelId`
- DashboardPage provides `dashboardId`
- SceneApp provides `app `

So IMO we need to collect the enriched data query request all way up to the root of the scene in order to get the full object, rather than collecting from the closest parent that implements an enricher interface.

But i'm not sure we need https://github.com/grafana/scenes/compare/encricher-test?expand=1#diff-8e640d1b532fe751df12b895885faed4a0f9e33d4feee2e0353413e296e6f7cbR14 function, just reading from the object state should be sufficient I think.

dprokop:
@torkelo something like this built on top of your enricher: https://github.com/grafana/scenes/commit/ed064da83c57c5ae35f441e4943c02f5eaebf782

PR #304, opened 2023-08-30
SceneObjectRef: Provide a way to have references to other scene objects without impacting parent 

torkelo:
Can't figure out how to make this be clonable (without making clone a lot more complicated), and even then there is no guarantee it will work as the reference could be to a scene object in a different scene graph tree.  

But not sure if this SceneObjectRef is any better than simple having a function in state that returns the instance alla
```ts
    getDashboard: () => dashboard,
```

<details>
  <summary>📦 Published PR as canary version: <code>0.27.1--canary.304.6034526594.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@0.27.1--canary.304.6034526594.0
  # or 
  yarn add @grafana/scenes@0.27.1--canary.304.6034526594.0
  ```
</details>

### DIFF
--- a/docusaurus/docs/core-concepts.md
+++ b/docusaurus/docs/core-concepts.md
@@ -254,7 +254,7 @@ const scene = new EmbeddedScene({
 It's very important that you do not reuse the same instance to a scene object in multiple different scenes or locations within the same scene. A scene object's parent is automatically set when it's part of the state of another scene object. So if you want to use the same scene object instance as the part of the state of more than one scene object you have two options.
 
 * Clone the source scene object. This will create a separate instance with no connection to the source object.
-* Use SceneObjectRef to wrap the instance. This makes sure the object's original parent is not changed while allowing you to store a reference to the instance in the state of another scene object.
+* Use `SceneObjectRef` to wrap the instance. This makes sure the object's original parent is not changed while allowing you to store a reference to the instance in the state of another scene object.
 
 
 ## Source code
--- a/docusaurus/docs/core-concepts.md
+++ b/docusaurus/docs/core-concepts.md
@@ -251,8 +251,7 @@ const scene = new EmbeddedScene({
 
 ## References to SceneObject's and the parent reference
 
-It's very important that you do not reuse the same instance to a scene object in multiple different scenes or locations within the same scene. A SceneObject's parent is automatically set when it's part of the state of another scene object. So
-if you want to use the same scene object instance as the part of the state of more than one scene object you have two options.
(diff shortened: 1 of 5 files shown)


torkelo:
Testing this here: https://github.com/grafana/grafana/pull/74126/files

torkelo:
after testing this I think it makes some sense

dprokop (approved):
I agree it makes sense. Also easier to reason about from i.e. InspectDrawer code where panel is explicitly typed as ref, rather than function 👍

docusaurus/docs/core-concepts.md
```diff
@@ -249,6 +249,15 @@ const scene = new EmbeddedScene({
...
 
+## References to SceneObject's and the parent reference
+
+It's very important that you do not reuse the same instance to a scene object in multiple different scenes or locations within the same scene. A SceneObject's parent is automatically set when it's part of the state of another scene object. So
+if you want to use the same scene object instance as the part of the state of more than one scene object you have two options.
+
+* Clone the source scene object. This will create a separate instance with no connection to the source object.
+* Use SceneObjectRef to wrap the instance. This makes sure the object's original parent is not changed while allowing you store a reference to the instance in the state of another scene object.
+
+
```
torkelo, line 260: @dprokop added this to core concepts doc as I think this fact that you cannot reuse the same object instance is very core

docusaurus/docs/core-concepts.md
```diff
@@ -249,6 +249,14 @@ const scene = new EmbeddedScene({
 });
 ```
 
+## References to SceneObject's and the parent reference
+
+It's very important that you do not reuse the same instance to a scene object in multiple different scenes or locations within the same scene. A SceneObject's parent is automatically set when it's part of the state of another scene object. So if you want to use the same scene object instance as the part of the state of more than one scene object you have two options.
```
dprokop, line 254:
```suggestion
It's very important that you do not reuse the same instance to a scene object in multiple different scenes or locations within the same scene. A scene object's parent is automatically set when it's part of the state of another scene object. So if you want to use the same scene object instance as the part of the state of more than one scene object you have two options.
```

docusaurus/docs/core-concepts.md
```diff
@@ -249,6 +249,14 @@ const scene = new EmbeddedScene({
...
+
+* Clone the source scene object. This will create a separate instance with no connection to the source object.
+* Use SceneObjectRef to wrap the instance. This makes sure the object's original parent is not changed while allowing you to store a reference to the instance in the state of another scene object.
```
dprokop, line 257:
```suggestion
* Use `SceneObjectRef` to wrap the instance. This makes sure the object's original parent is not changed while allowing you to store a reference to the instance in the state of another scene object.
```

docusaurus/docs/core-concepts.md
dprokop: yup, looks good

Issue #298, opened 2023-08-29
SceneQueryRunner: Provide a way to pass dashboardUID and panelId 


torkelo:
For core dashboards tracking / request headers to keep working we need to preserve these, currently SceneQueryRunner has no way to provide this

PR #309, opened 2023-08-31
DataQueryRequest enricher

dprokop:
Something like this as an alt to https://github.com/grafana/scenes/pull/303

Closes https://github.com/grafana/scenes/issues/298

PR #311, opened 2023-08-31
DataQueryRequest enricher

torkelo:
Trying to simplify (and optimize) https://github.com/grafana/scenes/pull/309 a bit by 

* Skipping having to walk the full scene graph and progressively setting & resetting props (to make deeper props override) 
* Skip providing the enricher via SceneApp state (I don't think this is needed right now). For scene app plugin just setting the app name should be enough. We do not have any other prop they can set. And when we do I think we should have some built-in mechanism to set panel or query tracking name directly on SQR or VizPanel. 

To skip the direct instanceof SceneAppPage and check getParentPage I do that instead in SceneAppPage by implementing the enrichDataRequest method. 

Closes https://github.com/grafana/scenes/issues/298

### DIFF
--- a/packages/scenes-app/src/pages/DemoListPage.tsx
+++ b/packages/scenes-app/src/pages/DemoListPage.tsx
@@ -19,6 +19,7 @@ const getScene = () => {
   const demos = getDemos();
 
   return new SceneApp({
+    name: 'Demos',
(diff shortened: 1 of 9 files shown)


dprokop (approved):
🥳

PR #312, opened 2023-08-31
SceneTimeRangeTransformerBase 

torkelo:
I tried creating a base class that could be used for SceneTimeZoneOverride and for PanelTime overrides.

Not 100% sure we need it, it's pretty simple code and having 1-3 duplicates of is ok.

Having base clases like this is a bit tricky with setState and requiring typescript ignore :)

<details>
  <summary>📦 Published PR as canary version: <code>0.29.0--canary.312.6063013465.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@0.29.0--canary.312.6063013465.0
  # or 
  yarn add @grafana/scenes@0.29.0--canary.312.6063013465.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/core/SceneTimeZoneOverride.tsx
+++ b/packages/scenes/src/core/SceneTimeZoneOverride.tsx
@@ -34,4 +34,8 @@ export class SceneTimeZoneOverride
   public getTimeZone(): TimeZone {
     return this.state.timeZone;
   }
+
+  public onTimeZoneChange(timeZone: string): void {
+    this.setState({ timeZone, value: evaluateTimeRange(this.state.from, this.state.to, this.state.timeZone) });
+  }
 }
--- a/packages/scenes/src/core/SceneTimeZoneOverride.tsx
+++ b/packages/scenes/src/core/SceneTimeZoneOverride.tsx
@@ -1,7 +1,8 @@
 import { TimeZone } from '@grafana/schema';
 import { evaluateTimeRange } from './SceneTimeRange';
 import { SceneTimeRangeLike, SceneTimeRangeState } from './types';
-import { SceneTimeRangeModifierBase } from './SceneTimeTransformer';
+import { SceneTimeRangeModifierBase } from './SceneTimeRangeModifierBase';
+import { getDefaultTimeRange } from '@grafana/data';
 
 interface SceneTimeZoneOverrideState extends SceneTimeRangeState {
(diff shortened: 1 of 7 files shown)


dprokop:
> Not 100% sure we need it, it's pretty simple code and having 1-3 duplicates of is ok.

I agree, but it gets complicated when the duplication is on two separate repos. I don't see a huge risk with this modifier, and if we can make reusable things I think we should do it.

torkelo:
@dprokop pushed some updates to get rid of the typescript ignore (Added a util function setBaseClassState)

packages/scenes/src/core/SceneTimeZoneOverride.tsx
```diff
@@ -1,70 +1,41 @@
...
+export class SceneTimeZoneOverride
+  extends SceneTimeRangeTransformerBase<SceneTimeZoneOverrideState>
+  implements SceneTimeRangeLike
+{
   public constructor(state: Omit<SceneTimeZoneOverrideState, 'from' | 'to' | 'value'>) {
     super({
       ...state,
       timeZone: state.timeZone,
-      // Fake time range, it's actually provided via closest time range object on activation
+      // Real time range is deifined
```
dprokop, line 19: // This time range is not valid until activation

packages/scenes/src/utils/utils.ts
```diff
@@ -0,0 +1,9 @@
+import { SceneObject, SceneObjectState } from '../core/types';
+
+/**
+ *  This function works around the problem of Contravariance of the SceneObject.setState function
+ *  Contravariance is not enforce by interfaces and here we use the SceneObject interface to access the setState function
+ */
+export function setBaseClassState<T extends SceneObjectState>(sceneObject: SceneObject<T>, newState: Partial<T>) {
```
dprokop, line 7: Don't see how this change is related to the scope of this PR?

dprokop (approved):

packages/scenes/src/utils/utils.ts
dprokop: I see it used only in multi value variable, what am I missing?

packages/scenes/src/core/SceneTimeZoneOverride.tsx
dprokop, line 19:
```suggestion
  // This time range is not valid until activation
```

PR #317, opened 2023-09-04
Variables: New LocalValueVariable to better support repeating panels 

torkelo:
This is to block query execution for repeating panels where the variable is refreshed on time range change. 

- LocalValueVariable: Special variable value to better support repeating panels
- Update comment

### DIFF
--- a/packages/scenes/src/variables/sets/SceneVariableSet.test.tsx
+++ b/packages/scenes/src/variables/sets/SceneVariableSet.test.tsx
@@ -524,6 +524,10 @@ describe('SceneVariableList', () => {
       scene.state.nested?.activate();
 
       expect(innerSet.isVariableLoadingOrWaitingToUpdate(scopedA)).toBe(true);
+
+      A.signalUpdateCompleted();
+
+      expect(innerSet.isVariableLoadingOrWaitingToUpdate(scopedA)).toBe(false);
     });
   });
 });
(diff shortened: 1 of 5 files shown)


packages/scenes/src/variables/sets/SceneVariableSet.test.tsx
```diff
@@ -483,6 +485,51 @@ describe('SceneVariableList', () => {
...
+
+      // Should start variables with no dependencies
+      expect(A.state.loading).toBe(true);
+      expect(B.state.loading).toBe(undefined);
+
+      expect(set.isVariableLoadingOrWaitingToUpdate(A)).toBe(true);
+      expect(set.isVariableLoadingOrWaitingToUpdate(B)).toBe(true);
+
+      A.signalUpdateCompleted();
+      expect(set.isVariableLoadingOrWaitingToUpdate(A)).toBe(false);
```
dprokop, line 505: Just for clarity maybe:
```suggestion
      expect(set.isVariableLoadingOrWaitingToUpdate(A)).toBe(false);
      expect(set.isVariableLoadingOrWaitingToUpdate(B)).toBe(true);
```

dprokop (commented):
Interesting approach, do you have an example how this will be used in a DashboardScene?

dprokop (approved):

PR #320, opened 2023-09-04
Annotations alt: composite runner

dprokop:


packages/scenes/src/querying/CompositeQueryRunner.ts
```diff
@@ -0,0 +1,125 @@
...
+      acc.series = (acc.series || []).concat(value.series || []);
+
+      // We would need to sensibly recude states here
+      acc.state = value.state;
+      acc.alertState = value.alertState;
+
+      acc.timeRange = value.timeRange;
+      acc.annotations = (acc.annotations || []).concat(value.annotations || []);
+      // THIS IS GOING TO BREAK A LOADS OF THINGS
+      acc.structureRev = value.structureRev;
```
torkelo, line 118: we don't have structureRev here as far as I know (I don't think backends / datasources return this). 

We overwrite it here and just increment it every time we see a new frame, hm.. that logic looks a bit strange 
https://github.com/grafana/scenes/blob/main/packages/scenes/src/components/VizPanel/VizPanel.tsx#L214

torkelo (commented):
The concept of a composite runner I don't think is strictly needed for annotations, The thing we need is the concept of searching up scene graph looking for "DataLayer" $data provider and subscribing to it. 

Sort of a mix of this your other PR? 

A  CompositeDataLayer I think would be useful if we need multiple query runners on the same level (For example a DashboardPanelAnnotationLayer, a PanelAlertStateLayer etc).

packages/scenes/src/querying/CompositeQueryRunner.ts
```diff
@@ -0,0 +1,125 @@
+import { AnnotationEvent, arrayToDataFrame, DataFrame, DataTopic, PanelData } from '@grafana/data';
+import { map, merge, mergeAll } from 'rxjs';
+import { SceneObjectBase } from '../core/SceneObjectBase';
+import { CancelActivationHandler, SceneDataProvider, SceneObjectState } from '../core/types';
+
+interface CompositeQueryRunnerState extends SceneObjectState {
+  runners: SceneDataProvider[];
```
torkelo, line 7: This would reset parent, so would have to be SceneObjectRef

packages/scenes-app/src/demos/dataLayers.tsx
```diff
@@ -78,26 +79,30 @@ export function getDataLayersTestTest(defaults: SceneAppPageState) {
...
-                  //   scenarioId: 'slow_query',
-                  //   stringInput: '5s',
-                }),
+              // $dataLayers: [globalAnnotations],
+              $data: new CompositeQueryRunner({
+                runners: [
+                  getQueryRunnerWithRandomWalkQuery({}),
+                  getQueryRunnerWithRandomWalkQuery({ scenarioId: 'slow_query', stringInput: '5s' }),
+                  globalAnnotations as any,
+                  nestedAnnotationsLayer,
```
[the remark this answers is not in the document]
dprokop: this is just an example one can mix multple runners

packages/scenes/src/querying/CompositeQueryRunner.ts
dprokop: That's true, but this quick POC does not accommodate for sharing on different levels. It just assumes that on whatever layer you want SQR + Annotations, one would use CompositeQueryRunner. But this has limitations that you cannot have global annotations & local SQR.

dprokop:
Closing in favor of https://github.com/grafana/scenes/pull/325 https://github.com/grafana/scenes/pull/328

PR #321, opened 2023-09-05
SceneGridRow: Small design change and fixes, add actions support 

torkelo:
* Interpolate title 
* Fix alignment of drag handle (removes SceneDragHandle, only used from SceneGridRow) 
* Simplify markup a bit for SceneGridRow (as unnecessarily nested probably from old nested grid solution where there was need for a row header) 
* Remove background of row when collapsed, instead add border 
* Add actions SceneObject extension prop that core dashboards can use for row settings & remove action 

[row_design.webm](https://github.com/grafana/scenes/assets/10999/c98812f3-01dc-47e9-8dfb-8375cd8a01b2)


dprokop (approved):
I like the style a lot!

PR #325, opened 2023-09-06
Data layers:  Isolated change

dprokop:
Data layer is a collection of `SceneDataProvider` attached to an arbitrary scene object using `$data` property. Thata layer results will be combined into `SceneQueryRunner` results under a specified `DataTopic`.

- [x] Add handling for transformation layer that as of now can retrieve SceneDataLayers as data provider
- [ ] [Update documentation](https://github.com/grafana/scenes/issues/331)
- [x] Get rid of getDataTopic from `SceneDataLayerProvider` interface
- [x] Allow cancellation per panel that will not cause cancellation for other panels using the same layer


packages/scenes/src/querying/SceneDataLayers.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    const resultsMap: Record<DataTopic, Map<string, DataFrame[]>> = {
+      annotations: new Map(),
+    };
+
+    if (layers.length > 0) {
+      layers.forEach((layer) => {
+        observables.push(layer.getResultsStream());
+      });
+
```
torkelo, line 65: Does this need to happen here? since SQR subscribes to all the layers, this won't really be used?

packages/scenes/src/querying/SceneDataLayers.ts
```diff
@@ -0,0 +1,110 @@
...
+      // possibly we want to combine the results from all layers and only then update, but this is tricky ;(
+      this._subs.add(
+        merge(observables)
+          .pipe(
+            mergeAll(),
+            map((v) => {
+              if (v.origin.getDataTopic() === DataTopic.Annotations) {
+                // Is there a better, rxjs only way to combine multiple same-data-topic observables?
+                // Indexing by origin state key is to make sure we do not duplicate/overwrite data from the different origins
+                resultsMap[v.origin.getDataTopic()].set(v.origin.state.key!, v.data);
```
torkelo, line 75: would not be a bit cleaner / simpler if dataTopic was part of the result object? (instead of a function on origin). that way a provider could technically return data for any topic.

torkelo (commented):
This is looking pretty good!

packages/scenes/src/querying/SceneDataLayers.ts
dprokop: Yes, I agree! I think there best way is to make sure the data layer returns DataFrame with the topic set, then we won't really need the topic on the layer itself I think.

packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -58,6 +59,9 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
 
   private getSourceData(): SceneDataProvider {
     if (this.state.$data) {
+      if (this.state.$data instanceof SceneDataLayers) {
+        throw new Error('SceneDataLayers can not be used as data provider for SceneDataTransformer.');
+      }
```
torkelo, line 64: Maybe we can add this check to SceneDataLayers activation instead? (it can check it's parent). Not sure why it's better but feels like SceneDataLayers is the "special data provider" so it feels a bit better for it to handle the edge cases where it could be misused

dprokop: Agree

packages/scenes/src/querying/SceneDataLayers.ts
dprokop: SQR subscribes to all `SceneDataLayers`, not individual layers provided to `SceneDataLayers`. I think we could subscribe to all collected SceneDataLayers.state.layers. But I like this PR proposed separation where:
1. A layer provides DataFrame[],
2. that's combined into PanelData (in SceneDataLayers) 
3. and composed in SceneQueryRunner into results passed to a consumer

I think thanks to 2), with multiple layers scenario, we will be able to provide individual loading states per layers group, and reduce those in SQR to provide SQR's loading state. Also, if we do all the shebang in SQR, what's the point of SceneDataLayers, as it will be literally empty implementation - just responsible for individual layers activation. The SceneDataProvider interface is also going to be a fake one, as there will be no results. 

So I think this needs to happen here. This makes sure on a given level all layers are combined, while SQR makes sure results from multiple levels of layers are collected.

packages/scenes/src/querying/SceneDataTransformer.ts
dprokop: Had to revert it as it created a cricular dependency that surfaced in tests. Not sure yet how to solve it

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -85,6 +117,55 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
...
+    if (dataLayers.length > 0) {
+      dataLayers.forEach((layer) => {
+        observables.push(layer.getResultsStream());
+      });
+
+      // possibly we want to combine the results from all layers and only then update, but this is tricky ;(
+      this._dataLayersSub = merge(observables)
+        .pipe(
+          mergeAll(),
+          takeUntil(this._cancelationStream.asObservable()),
```
torkelo, line 138: what is the point of the cancellationStream? 

I can see cancelQuery calls ,  this._cancelationStream.next 

but that will complete the this subscription, so we wont' see any annotations after the cancellation (if I know my rxjs)

dprokop: There is a re-subscription happening, see this commit: https://github.com/grafana/scenes/pull/325/commits/8978e858563cd60546a1cf9dc06b374942b115bf

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -188,12 +274,21 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
...
       data: { ...this.state.data!, state: LoadingState.Done },
     });
   }
 
   private async runWithTimeRange(timeRange: SceneTimeRangeLike) {
+    // If data layers subscription doesn't exist, create one
+    //
+    if (!this._dataLayersSub) {
+      this._handleDataLayers();
+    }
```
torkelo, line 290: this this is called here, we should justt be able to call `!this._dataLayersSub.unsubscribe()` from cancelQuery instead of using the cancelationStream?

dprokop: yeah, think we can skip the canncellation stream

packages/scenes/src/querying/SceneQueryRunner.ts
torkelo: I see now that a new sub will be created,  if (!this._dataLayersSub) {  (but that should not be true if the cancellation stream is triggered, that will just complete the observable)

packages/scenes/src/querying/SceneQueryRunner.ts
dprokop: removed the cancellation stream

torkelo (approved):
:tada:

PR #326, opened 2023-09-06
SceneGridLayout: Fix toggle row issue

torkelo:
Noticed panels appearing below other rows when expanding a collapsed row.

Happens if you collapse a row. and then collapse another row above it.

Then try to expand the first row. The children it contains have to high y position
and needs to be corrected. We had a condition that only corrected the row child position
if it needed to be made higher not smaller.

I checked the current arch toggleRow logic and it did not have this condition that only correct y pos when greater 

<details>
  <summary>📦 Published PR as canary version: <code>1.2.0--canary.326.6111681445.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@1.2.0--canary.326.6111681445.0
  # or 
  yarn add @grafana/scenes@1.2.0--canary.326.6111681445.0
  ```
</details>


dprokop (approved):

dprokop:
Change `minor` to `patch` release.

PR #327, opened 2023-09-06
License: Switch to Apache 2.0

torkelo:
Decided that the AGPL license is going to limit adoption too much, as it will basically
force any plugin that uses this library to also be OSS.

So we decided to switch to Apache 2.0

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,661 +1,202 @@
-                    GNU AFFERO GENERAL PUBLIC LICENSE
-                       Version 3, 19 November 2007
-
- Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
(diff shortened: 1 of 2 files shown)


dprokop (approved):
🥳

dprokop:
FYI added minor release label.

cbos:
@torkelo @dprokop 

I think the switch of the license is only partly done. 
The package.json the sub packages are still "AGPL-3.0-only"
https://github.com/grafana/scenes/blob/main/packages/scenes/package.json
Same holds for the scene-app package.

In the NPM store the package is also listed as "AGPL-3.0-only"
https://www.npmjs.com/package/@grafana/scenes

Issue #313, opened 2023-09-01
Provide a mechanism for performing annotation queries in a scene


dprokop:
```[tasklist]
### Tasks
- [x] [Provide infrastructure for executing annotation queries and merging them to panel data](https://github.com/grafana/scenes/pull/325)
- [x] [Implement `AnnotationsDataLayer` that is responsible for querying and processing annotations](https://github.com/grafana/scenes/pull/328)
- [x] [Provide API to filter annotations by panelId](https://github.com/grafana/scenes/pull/329)
- [x] [Allow toggling individual layer on/off](https://github.com/grafana/scenes/pull/333)
- [x] [Support individual layer cancellation](https://github.com/grafana/scenes/pull/337)
- [x] Handle annotation queries errors
- [x] [Add support for template variables](https://github.com/grafana/scenes/pull/358)
- [x] [Deduplicate annotations](https://github.com/grafana/scenes/pull/351)
- [x] Cancel annotation queries on time range change
- [x] Support legacy annotations runner
- [ ] ~Handle annotation config and mappings without the transient AnnotationEvents model~
```

PR #328, opened 2023-09-08
Data layers:  Annotations

dprokop:
Adds `AnnotationsDataLayer` to enable annotation queries in scenes.


packages/scenes/src/querying/layers/annotations/AnnotationsDataLayer.ts
```diff
@@ -0,0 +1,142 @@
...
+
+interface AnnotationsDataLayerState extends SceneObjectState {
+  data?: PanelData;
+  queries: AnnotationQuery[];
+}
+
+export class AnnotationsDataLayer extends SceneObjectBase<AnnotationsDataLayerState> implements SceneDataLayerProvider {
+  private _querySub?: Unsubscribable;
+  private _results = new ReplaySubject<SceneDataLayerProviderResult>();
+
```
torkelo, line 24: need to add variables dependency here so you can have annotation queries that use variables, but we can add that as a follow-up PR but good to add to a todo list :)

packages/scenes/src/querying/layers/annotations/standardAnnotationsSupport.ts
```diff
@@ -0,0 +1,267 @@
...
+  { key: 'prevState' },
+  { key: 'newState' },
+  { key: 'data' as any },
+  { key: 'panelId' },
+  { key: 'alertId' },
+  { key: 'dashboardId' },
+  { key: 'dashboardUID' },
+];
+
+export function getAnnotationsFromData(
```
torkelo, line 150: do we have any tests for this and other functions here we can copy from core?

dprokop: not, still on the todos :)

packages/scenes/src/querying/layers/annotations/AnnotationsDataLayer.ts
dprokop: yup, added https://github.com/grafana/scenes/issues/313

packages/scenes/src/querying/layers/annotations/standardAnnotationsSupport.ts
dprokop: done now, moved what I could :)

PR #330, opened 2023-09-12
SceneObject: Add getRef for easier SceneObjectRef usage

torkelo:
Just an idea that could make SceneObjectRef a bit easier to use and discover


dprokop (approved):
Nice!

PR #333, opened 2023-09-13
DataLayers: Allow toggling individual layers on/off

dprokop:
Introduces:
- `SceneDataLayerBase` class that handles individual layer toggling on/off and common logic for sharing results.
- `SceneDataLayerControls` object that renders a layer(s) name with a Switch toggle to allow toggling layers on/off.
- Extends `SceneDataLayerProviderState` interface with `name` and `isEnabled`.
- Improves `sceneGraph.getDataLayers` to allow resolving the closest layers only (needed for nested controls) and adds tests.
- Introduces internal `ControlsLabel` component shared for controls like data layers toggles or variables.
- Updates demo

https://github.com/grafana/scenes/assets/2376619/907624fd-c461-4e9b-ae28-59f2247a0fdb


packages/scenes/src/querying/layers/SceneDataLayerBase.ts
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  public cancelQuery() {
+    if (this.querySub) {
+      this.querySub.unsubscribe();
+      this.querySub = undefined;
+    }
+  }
+
+  public publishResults(data: PanelData, topic: DataTopic) {
```
torkelo, line 104: should this be public or just protected?

dprokop: protected it should be. No need for external publishing.

packages/scenes/src/variables/components/VariableValueSelectors.tsx
```diff
@@ -1,106 +1,55 @@
...
     return null;
   }
 
   return (
-    <div className={styles.container}>
-      <VariableLabel state={state} styles={styles} />
+    <div
+      style={{
+        display: 'flex',
+      }}
```
torkelo, line 38: a static variable defined (outside the function / file global scope)

const flexContainer = css({display: 'flex'}); 

is nice simple way to avoid creating new css objects inline

packages/scenes/src/querying/layers/SceneDataLayerControls.tsx
```diff
@@ -0,0 +1,60 @@
...
+  public constructor() {
+    // Holds the state of the layers, to avoid force re-rendering
+    super({ layersMap: {} });
+
+    this.addActivationHandler(() => this._onActivate());
+  }
+
+  private _onActivate() {
+    const layers = sceneGraph.getDataLayers(this);
+    this.setState({ layersMap: layers.reduce((acc, l) => ({ ...acc, [l.state.key!]: l.state.isEnabled }), {}) });
```
torkelo, line 24: should this not also get the layers closest to the controls?

torkelo (approved):
This looks great, just a few nits and a question on SceneDataLayersControls localOnly param

PR #335, opened 2023-09-14
NestedScene: Update design to match grid row, add controls property and update demo scene to include variables 

torkelo:
* Updates design to match SceneGridRow (sharing some styles)
* Rename actions => controls to match EmbeddedScene naming
* Updated demo scene to include variables on the NestedScene (showing that they work, but also includes an example that does not work: ie a variable on the nested row  that depends on a parent level variable, this should be relatively easy to add support for but leaving that for a future PR) 

This is needed by https://github.com/grafana/grafana/pull/74783


dprokop (approved):
LGTM

PR #337, opened 2023-09-14
DataLayers: Allow cancelling layers from layer control

dprokop:
Introduces the possibility of canceling data layer queries using a button (spinner) in the data layer control.

- `SceneDataLayerBase` now has `data: PanelData` state so that layer doesn't have to implement it.
- `SceneDataLayerBase` now has `runLayer` abstract method that should implement query execution logic and subscription setup. `runLayer` is now executed also in the base class, so that layers don't have to implement `shouldRunQueriesOnActivate`. This is now handled in the base class, so activation handlers should not be needed in the layers, unless some peculiar logic is required.
- `SceneDataLayerControls` now renders a loading spinner for layers that take >500ms to resolve results. Clicking on the icon will cancel layer query.

https://github.com/grafana/scenes/assets/2376619/e03a071f-9d60-47b4-b0be-08bb8cf4b9c6

### DIFF
--- a/packages/scenes/src/querying/layers/SceneDataLayerControls.tsx
+++ b/packages/scenes/src/querying/layers/SceneDataLayerControls.tsx
@@ -1,13 +1,12 @@
 import { css } from '@emotion/css';
 import { LoadingState } from '@grafana/schema';
-import { InlineSwitch, useTheme2 } from '@grafana/ui';
+import { InlineSwitch } from '@grafana/ui';
 import React, { useEffect } from 'react';
 import { map, of, switchMap, timer } from 'rxjs';
 import { sceneGraph } from '../../core/sceneGraph';
 import { SceneObjectBase } from '../../core/SceneObjectBase';
 import { SceneComponentProps, SceneDataLayerProvider, SceneObjectState } from '../../core/types';
 import { ControlsLabel } from '../../utils/ControlsLabel';
-import { LoadingIndicator } from '../../utils/LoadingIndicator';
 
 interface SceneDataLayerControlsState extends SceneObjectState {
   layersMap: Record<string, boolean>;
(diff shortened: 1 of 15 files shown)


packages/scenes/src/querying/layers/SceneDataLayerControls.tsx
```diff
@@ -44,17 +48,74 @@ function SceneDataLayerControlsRenderer({ model }: SceneComponentProps<SceneData
...
+        })
+      )
+      .subscribe((v) => {
+        setShowLoading(v);
+      });
+
+    return () => {
+      loadingIndicatorSub.unsubscribe();
+    };
+  }, [layer]);
```
torkelo, line 93: this to me hints at a potential problem, why is this not just
```
const { isLoading } = layer.useState();
```

Maybe we need to make layers single query for that?

dprokop: @torkelo - yeah, I think you're correct. I mean - for the controls to make sense one would have to model layers as independent anyways. Think we can limit the queries to 1, and figure out multi-query layers when there's a a need for that.

packages/scenes/src/querying/layers/annotations/standardAnnotationQuery.ts
```diff
@@ -31,7 +31,7 @@ export function executeAnnotationQuery(
   datasource: DataSourceApi,
   timeRange: SceneTimeRangeLike,
   query: AnnotationQuery
-): Observable<AnnotationEvent[]> {
+): Observable<{ state: LoadingState; events: AnnotationEvent[] }> {
```
dprokop, line 34: @torkelo  - I'm changing the emitted type, as this is proxying to `runRequest`, which already handles the loading state and timeout for setting emitting empty results with LoadingState correctly set.

packages/scenes/src/querying/layers/SceneDataLayerControls.tsx
dprokop: simplified that bit

torkelo (approved):
Looks good

PR #338, opened 2023-09-14
Unify layers and variables loading/cancellation UI

dprokop:
Based on #337 

This unifies the loading indicator (that serves as a cancellation trigger) between data layer controls and variable controls.

Before:
Screenshot 2023-09-14 at 22 40 23

After:
Screenshot 2023-09-14 at 22 39 57

### DIFF
--- a/packages/scenes/src/variables/components/VariableValueSelect.tsx
+++ b/packages/scenes/src/variables/components/VariableValueSelect.tsx
@@ -6,53 +6,42 @@ import { MultiSelect, Select } from '@grafana/ui';
 import { SceneComponentProps } from '../../core/types';
 import { MultiValueVariable } from '../variants/MultiValueVariable';
 import { VariableValue, VariableValueSingle } from '../types';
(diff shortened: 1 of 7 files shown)


packages/scenes/src/variables/components/VariableValueSelect.tsx
```diff
@@ -6,53 +6,42 @@ import { MultiSelect, Select } from '@grafana/ui';
...
   return (
-    <MultiSelect<VariableValueSingle, { onCancel: () => void }>
+    <MultiSelect<VariableValueSingle>
       id={key}
       placeholder="Select value"
       width="auto"
       value={arrayValue}
       tabSelectsValue={false}
       allowCustomValue
-      isLoading={loading}
```
torkelo, line 47: There are variable selectors with hidden label, so in that scenario you loose loading state, but think that is ok.

torkelo (approved):
Looks good, I kind of want to try this spining icon for cancel query in the PanelChrome as well as it's static cancel query icon feels worse

PR #339, opened 2023-09-14
Update link to plugins docs in scene-app.md

josmperez:


dprokop (commented):
I don't think the proposed link is valid in this context. It drops a person right into the plugin creation guide, without explaining what the Grafana plugin even is.

dprokop:
> but today I just want to get the address correct

But this address ain't correct. I mean - it's fine, but it's throwing people into deep water, straight into plugin creation, without letting them know what plugins are. While the old one is basically Grafana plugin developer’s guide, which is an intro to what plugins are and how they can help.

dprokop:
> I'm heads down on the plugin site, but if it would be helpful I can take a look at this doc next week and offer suggestions.

Yes, definitely, please!

PR #340, opened 2023-09-15
QueryVariable: Fixes queries with older model

### LINKED ISSUE #69: Variables: Legacy queries don't work (in some cases)
I tested a devenv dashboard with variables 
http://localhost:3000/grafana/d/O6GmNPvWk/templating-nested-variables-drilldown?orgId=1 

It has some testdata variables, that don't work in scenes. 

I tracked it down to being caused by scenes variables not having something like the function upgradeLegacyQueries which updates the query model before variableSupport.toDataQuery is called. 

https://github.com/grafana/grafana/blob/main/public/app/features/variables/state/actions.ts#L929 

In core (old dashboards) when toDataQuery is called the query is an object with query prop, but with scenes query prop is undefined

torkelo:
Fixes #69

Noticed the variables in this dashboard did not work:
http://localhost:3000/grafana/d/-Y-tnEDWk/templating-nested-template-variables?orgId=1

Tracked it down some a migration that happens here: 
https://github.com/grafana/grafana/blob/main/public/app/features/variables/state/actions.ts#L1064

Not sure where we should place this migration, here, when we load the dashboard in core?

### DIFF
--- a/packages/scenes/src/variables/variants/query/createQueryVariableRunner.ts
+++ b/packages/scenes/src/variables/variants/query/createQueryVariableRunner.ts
@@ -29,6 +29,11 @@ class StandardQueryRunner implements QueryRunner {
 
   public getTarget(variable: QueryVariable) {
     if (hasStandardVariableSupport(this.datasource)) {
+      // Migrate legacy queries
+      if (!isDataQueryType(variable.state.query)) {
+        variable.setState({ query: { query: variable.state.query, refId: `variable-${variable.state.name}` } });
+      }
+
       return this.datasource.variables.toDataQuery(variable.state.query);
     }
 
@@ -138,4 +143,12 @@ export let createQueryVariableRunner = createQueryVariableRunnerFactory;
  */
 export function setCreateQueryVariableRunnerFactory(fn: (datasource: DataSourceApi) => QueryRunner) {
   createQueryVariableRunner = fn;
+}
+
+function isDataQueryType(query: any): query is DataQuery {
+  if (!query) {
+    return false;
+  }
+
+  return query.hasOwnProperty('refId') && typeof query.refId === 'string';
 }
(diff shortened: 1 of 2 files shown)


packages/scenes/src/variables/variants/query/createQueryVariableRunner.ts
```diff
@@ -29,6 +29,11 @@ class StandardQueryRunner implements QueryRunner {
 
   public getTarget(variable: QueryVariable) {
     if (hasStandardVariableSupport(this.datasource)) {
+      // Migrate legacy queries
+      if (!isDataQueryType(variable.state.query)) {
```
dprokop, line 33: Think I would put it in the core before the query even hits the scene. I mean - I don't think we should anyhow incentivize legacy queries in scenes apps. Maybe here we should just have a warning that the query used is a legacy one, rather than handling the migration.

torkelo: @dprokop where? A schema migration? it would need to load the datasource and check what kind of variable support it has. 

I can try, I think we have other schema migrations that do need to load the datasouce plugin so maybe not so complicated

dprokop: I was thinking a naive path - making sure Scenes variables are created with a proper model

torkelo: @dprokop but the migration should only happen for data sources that are using the new "standard" variable interface

torkelo: @dprokop it's a really strange schema migration, that happens on every getOptions (get variable values) call, and on every call to variableUpdated (redux action) 
https://github.com/grafana/grafana/blob/main/public/app/features/variables/state/upgradeLegacyQueries.test.ts 

I think a schema migration should be a better fit, but looking at DashboardMigrator we have 0 async dashboard migrations and this would have to be async (wait for plugin to load), hm.. dam

dprokop: yhym, let's keep it as is here for now?

dprokop (approved):

PR #346, opened 2023-09-19
AdHocFiltersSet and AdhocFiltersVariable with manual and automatic modes 

torkelo:
*  [x] Adds a new AdHocFiltersSet type
*  [x] Adds new UI components that handles the filter UI and rewritten state logic moved to scene object (greatly reduced the code & complexity) 
*  [x] Works without any changes to core or data sources via patched `templateSrv.getAdhocFilters` (This is just for backwards compatability). Will also do a new/future proof solution where we pass adhocFilters to datasources via DataQueryRequest. But that requires change in core and in data sources (to use). So need this base solution first. 
* [x] Unit tests for UI logic 
* [x] Url sync
* [x] A way to provide custom getTagKey and getTagValues logic to handle the custom scene app scenarios (listed below) 
* [x] Can be used as a variable and then be used in queries to have more control over which queries.

Here o11y frontend app is doing some custom stuff:
https://github.com/grafana/app-o11y-kwl/blob/main/plugin/src/scenes/components/apps/filters/AdHocFilterKey.tsx#L56 
https://github.com/grafana/app-o11y-kwl/blob/main/plugin/src/scenes/components/apps/filters/AdHocFilterValue.tsx#L63

o11y services filter 
https://github.com/grafana/app-observability-plugin/blob/87334b1577564dcb92a760462b35bdd7e995e9d7/src/modules/services/components/ServicesFilter.tsx#L18

both those scene app plugins render the filters to a variable that is then used in the query, I cannot adopt this approach by default to stay backward compatible with existing dashboards. Currently, data sources get current adhoc filters via calling `templateSrv.getAdhocFilters` and applying the filters to all queries. not sure yet how to opt out of that and have a more manual approach where specific queries are only affected (yet). 

Redid the UX using Select components, using same design as label filters in the prometheus / loki query builder 
Screenshot from 2023-09-25 09-35-39


packages/scenes/src/variables/adhoc/patchGetAdhocFilters.ts
```diff
@@ -0,0 +1,41 @@
...
+  const templateSrv: any = getTemplateSrv();
+  if (!templateSrv.getAdhocFilters) {
+    console.log('Failed to patch getAdhocFilters');
+  }
+
+  if (!originalGetAdhocFilters) {
+    originalGetAdhocFilters = templateSrv.getAdhocFilters;
+  }
+
+  templateSrv.getAdhocFilters = function (dsName: string): AdHocVariableFilter[] {
```
dprokop, line 19: Think this is not gonna work if one uses multiple AdHocFIltersVariables. `originalGetAdhocFilters` may be another patched `getAdhocFilters` already.

torkelo: @dprokop you are right, this is just for the legacy / backward compatability. I will also add a way to pass adhocFilters to DataSourceAPI.query via DataQueryRequest, that will be the future proof method that will support multiple adhoc filters on different levels 

So this patch here is to support core dashboards and existing / old data sources using `  templateSrv.getAdhocFilters `

packages/scenes-app/src/demos/adhocFiltersDemo.tsx
```diff
@@ -0,0 +1,114 @@
...
+      }),
+      new SceneAppPage({
+        title: 'As variable',
+        url: `${defaults.url}/manual`,
+        getScene: () => {
+          return new EmbeddedScene({
+            ...getEmbeddedSceneDefaults(),
+            $variables: new SceneVariableSet({
+              variables: [
+                AdHocFiltersVariable.create({
```
dprokop, line 72: Not sure how `create` is helping us compared to regular constructor:

```
private constructor(state: AdHocFiltersVariableCreateHelperArgs) {
    super({
      type: 'adhoc',
      hide: VariableHide.hideLabel,
      name: state.name ?? 'Filters',
      set: new AdHocFilterSet({
        ...state,
        // Main reason for this helper factory functyion
        applyMode: 'manual',
      }),
    });

```

Since we expect the `applyMode` to be set to manual anyways, I think we don't need to allow providing AdHocFilterSet so the constructor anyways. 

Overall - I like the idea of having the factory function, but not for this variable exclusively. It creates inconsistency in the API that I think we should avoid until we decide on a pattern for all variables. WDYT?

packages/scenes/src/core/PanelBuilders/index.ts
```diff
@@ -79,7 +79,7 @@ import {
 import {
   Options as XYChartOptions,
   defaultOptions as defaultXYChartOptions,
-  defaultScatterFieldConfig as defaultXYChartFieldConfig,
+  defaultFieldConfig as defaultXYChartFieldConfig,
```
dprokop, line 82: This looks weird - I think the rename is not in version < 10.1.x (ref https://github.com/grafana/grafana/pull/67878), but I cannot see package versions updates.

packages/scenes/src/querying/layers/annotations/standardAnnotationsSupport.ts
```diff
@@ -133,7 +133,7 @@ export const publicDashboardEventNames: AnnotationFieldInfo[] = [
 // Given legacy infrastructure, alert events are passed though the same annotation
 // pipeline, but include fields that should not be exposed generally
 const alertEventAndAnnotationFields: AnnotationFieldInfo[] = [
-  ...(config.isPublicDashboardView ? publicDashboardEventNames : []),
+  ...(config.publicDashboardAccessToken ? publicDashboardEventNames : []),
```
dprokop, line 136: good catch

packages/scenes/src/variables/adhoc/AdHocFilterSeparator.tsx
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+export function AdHocFilterSeparator() {
+  return <div>{/** AND */}</div>;
```
dprokop, line 4: This empty div creates some uneven spacing between filters:

image

Since the separator is not used at all now, maybe just remove it?

packages/scenes/src/variables/adhoc/AdHocFiltersSet.tsx
```diff
@@ -0,0 +1,281 @@
...
+  filters: AdHocVariableFilter[];
+  /** Base filters to always apply when looking up keys*/
+  baseFilters?: AdHocVariableFilter[];
+  /** Datasource to use for getTagKeys and getTagValues and also controls which scene queries the filters should apply to */
+  datasource: DataSourceRef | null;
+  /** Controls if the filters can be changed */
+  readOnly?: boolean;
+  /**
+   * Defaults to same-datasource which means filters will automatically be applied to all queries with the same data source as this AdHocFilterSet.
+   * In manual mode no queries are rerun on changes and you have to manual apply the filter to what ever queries you want.
```
dprokop, line 36: nit 
```suggestion
   * In manual mode no queries are re-run on changes, and you have to manually apply the filter to whatever queries you want.
```

packages/scenes/src/variables/adhoc/AdHocFiltersSet.tsx
```diff
@@ -0,0 +1,281 @@
...
+   */
+  applyMode?: 'same-datasource' | 'manual';
+  /**
+   * Extension hook for customizing the key lookup.
+   * Return replace: true if you want to override the default lookup and
+   * Return replace: false will just combine the results.
```
dprokop, line 42:
```suggestion
   * Return replace: true if you want to override the default lookup
   * Return replace: false if you want to combine the results with the default lookup
```

packages/scenes/src/variables/adhoc/AdHocFiltersSet.tsx
```diff
@@ -0,0 +1,281 @@
...
+   */
+  getTagKeysProvider?: getTagKeysProvider;
+  /**
+   * Extension hook for customizing the value lookup.
+   * Return replace: true if you want to override the default lookup.
+   * Return replace: false to combine the results.
```
dprokop, line 48:
```suggestion
   * Return replace: true if you want to override the default lookup.
   * Return replace: false if you want to combine the results with the default lookup
```

packages/scenes/src/variables/adhoc/AdHocFiltersSet.tsx
```diff
@@ -0,0 +1,281 @@
...
+      applyMode: 'same-datasource',
+      ...initialState,
+    });
+
+    if (this.state.applyMode === 'same-datasource') {
+      patchGetAdhocFilters(this);
+    }
+  }
+
+  public _updateFilter(filter: AdHocVariableFilter, prop: keyof AdHocVariableFilter, value: string | undefined | null) {
```
dprokop, line 91: What's your thinking for having the _ for a public method?

packages/scenes/src/variables/adhoc/AdHocFiltersSet.tsx
```diff
@@ -0,0 +1,281 @@
...
+      label: value,
+      value,
+    }));
+  }
+
+  private _runSceneQueries() {
+    // In manual mode we do not trigger any queries
+    if (this.state.applyMode === 'manual') {
+      return;
+    }
```
dprokop, line 200: Just to double-check check iIunderstand correctly - this basically delegates queries execution to `SceneVariablesSet`?

packages/scenes/src/variables/adhoc/AdHocFiltersVariableUrlSyncHandler.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    if (filters.length === 0) {
+      return {};
+    }
+
+    const value = filters.map((filter) => toArray(filter).map(escapeDelimiter).join('|'));
+    return { [this.getKey()]: value };
+  }
+
+  public updateFromUrl(values: SceneObjectUrlValues): void {
```
dprokop, line 27: I've noticed that with multiple filters applied, only the first one is applied on page refresh.

dprokop (approved):
Massive work, think it looks good overall, left couple of comments.

packages/scenes-app/src/demos/adhocFiltersDemo.tsx
torkelo: >I think we don't need to allow providing AdHocFilterSet so the constructor anyways.

@dprokop not sure what you mean? that would break the SceneObjectBase interface, and cloning 

The cloning full trees forces us to be consistent with constructors only taking in state (and full state)

torkelo: > Overall - I like the idea of having the factory function, but not for this variable exclusively. It creates inconsistency in the API that I think we should avoid until we decide on a pattern for all variables. WDYT?

I can remove it and add docs to set applyMode to manual for the inner set. So would have to look like this 

```ts 
new new AdHocFiltersVariable({
  name: 'Filters',   
  set: new AdHocFilterSet({ 
     name: 'Filters',
     mode: 'manual', 
      baseFilters: [{ key: '__name__', operator: '=', value: 'ALERTS' }],
      datasource: { uid: 'gdev-prometheus' },
     ....
 })
})   
```

dprokop: I've meant that `AdHocFiltersVariableCreateHelperArgs` guarantees one will not provide the `AdHocFilterSet` to the constructor. 

> The cloning full trees forces us to be consistent with constructors only taking in state (and full state)

How come? We have several objects that take `Partial<State>` as constructor arg.

packages/scenes/src/core/PanelBuilders/index.ts
torkelo: @dprokop why? everything seems to work. they renamed this export and so just fixed the import

dprokop: Seems like we have an invalid lock. In scenes-app we are using `canary` while in the scenes repo we are still on `10.0.3`. In the lock there `10.2.0-pre....`.

So this will surface in app plugins that use `/schema` < 10.1.0 which is sort of enforced by scenes peer dependency on 10.0.3

packages/scenes/src/variables/adhoc/AdHocFiltersSet.tsx
torkelo: @dprokop well yes and no. It only does that when it's used inside a AdHocFiltersVariable

packages/scenes/src/variables/adhoc/AdHocFiltersSet.tsx
torkelo: @dprokop we have problem I feel with the need to make a lot of "internal" functions public because the UI component needs to call them. So was thinking using underscore as a way to signal to plugin developers "please do not call this"

packages/scenes/src/variables/adhoc/AdHocFiltersSet.tsx
dprokop: yeah, was thinking primarily of the variable case, thanks.

packages/scenes/src/variables/adhoc/AdHocFiltersVariableUrlSyncHandler.ts
torkelo: @dprokop good catch! fixed that

packages/scenes-app/src/demos/adhocFiltersDemo.tsx
dprokop: We figured out the above comments async, I ignored the fact that `AdHocFiltersVariable` constructor cannot use a mixture of `AdHocFiltersVariableState` and `AdHocFilterSetState`. Because of how SceneObjectBase is typed.

PR #356, opened 2023-09-21
POC: Breakdown demo 

torkelo:
A very quick POC exploring a breakdown workflow. 

A few cool interesting take aways 

* A tab layout that is basically a variable value selector 
* I really wish variable queries/values could include a count so the variable query could be powered by an instant query, and then provide a counter for each label for how many series each label has. 
* The split layout is also interesting and could be potentially extracted to be reusable, a bit complicated as the UI for it is actually part of the tabbed layout.


packages/scenes-app/src/demos/breakdownDemo.tsx
```diff
@@ -0,0 +1,380 @@
...
+      <div className={styles.container}>
+        {loading && <div>Loading...</div>}
+        <TabsBar>
+          <div className={styles.tabHeading}>Breakdown by label</div>
+          {options.map((option, index) => (
+            <Tab
+              key={index}
+              label={option.label}
+              active={option.value === variable.state.value}
+              onChangeTab={() => variable.changeValueTo(option.value, option.label)}
```
torkelo, line 202: this is kind of cool, the simplicity of this line and what it triggers (Value changes, query runner will re-run query, and then panel will re-render)

PR #357, opened 2023-09-21
SceneApp: Introduce a useSceneApp hook that should replace useMemo as method of caching SceneApp instance 

torkelo:
The problem with our current useMemo recommended approach is that the whole scene app and all the data/scene caches are lost when you leave the app (say click on a link to explore, and then go back)

So this adds a new useSceneApp that adds a global cache that stays put even when you leave the app

### DIFF
--- a/docusaurus/docs/scene-app.md
+++ b/docusaurus/docs/scene-app.md
@@ -21,10 +21,11 @@ Scenes come with the following objects that make it easy to build highly interac
 Define a new Scenes app using the `SceneApp` object :
 
 ```tsx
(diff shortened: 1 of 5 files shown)


dprokop (commented):
Let's update the docs: https://github.com/grafana/scenes/blob/main/docusaurus/docs/scene-app.md#L43

torkelo:
@dprokop ah, good catch, I though I had updated all the docs mentioning this but missed that line

domasx2:
Nice!

PR #358, opened 2023-09-21
AnnotationsDataLayer: Add variables support

dprokop:
This mostly does the same thing as `SceneQueryRunner` to support variables in annotation queries. Tests are literally the same.

`SceneDataLayerBase` handles basic behaviors (like skipping execution on activation when the variable is not loaded), while AnnotationsDataLayer implements the variables to support itself. `SceneDataLayerBase` also handles variable values recording, hence the subtle change in `VariableValueRecorder`.

<details>
  <summary>📦 Published PR as canary version: <code>1.6.0--canary.358.6274180040.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@1.6.0--canary.358.6274180040.0
  # or 
  yarn add @grafana/scenes@1.6.0--canary.358.6274180040.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/querying/layers/SceneDataLayerBase.ts
+++ b/packages/scenes/src/querying/layers/SceneDataLayerBase.ts
@@ -11,6 +11,7 @@ import {
 import { setBaseClassState } from '../../utils/utils';
 import { writeSceneLog } from '../../utils/writeSceneLog';
 import { SceneVariable } from '../../variables/types';
+import { VariableDependencyConfig } from '../../variables/VariableDependencyConfig';
 import { VariableValueRecorder } from '../../variables/VariableValueRecorder';
 
 /**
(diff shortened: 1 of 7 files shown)


packages/scenes/src/querying/layers/annotations/AnnotationsDataLayer.ts
```diff
@@ -21,6 +23,15 @@ export class AnnotationsDataLayer
...
   public topic = DataTopic.Annotations;
 
+  protected _variableDependency: VariableDependencyConfig<AnnotationsDataLayerState> = new VariableDependencyConfig(
+    this,
+    {
+      statePaths: ['query'],
+      onVariableUpdatesCompleted: (variables, dependencyChanged) =>
+        this.onVariableUpdatesCompleted(variables, dependencyChanged),
+    }
+  );
```
dprokop, line 33: @torkelo - I am thinking about moving this config to the base class and allowing the `statePaths` configuration via `super.constructor` call... What do you think? This would require us to extend `VariableDependencyConfig` with a setter for `_statePaths`.

packages/scenes/src/querying/layers/SceneDataLayerBase.ts
```diff
@@ -105,6 +110,20 @@ export abstract class SceneDataLayerBase<T extends SceneDataLayerProviderState =
...
+  protected onVariableUpdatesCompleted(variables: Set<SceneVariable>, dependencyChanged: boolean): void {
+    writeSceneLog('SceneDataLayerBase', 'onVariableUpdatesCompleted');
+    if (this.state._isWaitingForVariables && this.shouldRunLayerOnActivate()) {
+      this.runLayer();
+      return;
+    }
+
+    if (dependencyChanged) {
+      this.runLayer();
+    }
```
torkelo, line 126: wonder if there is a way to build in the variableValueRecorder functionality into VariableDependencyConfig

packages/scenes/src/querying/layers/annotations/AnnotationsDataLayer.ts
torkelo: yea, I think that could work.

torkelo (approved):

PR #360, opened 2023-09-22
SceneTimeRangeCompare: Enable URL sync

dprokop:
<details>
  <summary>📦 Published PR as canary version: <code>1.6.0--canary.360.6273406733.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@1.6.0--canary.360.6273406733.0
  # or 
  yarn add @grafana/scenes@1.6.0--canary.360.6273406733.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/SceneTimeRangeCompare.tsx
+++ b/packages/scenes/src/components/SceneTimeRangeCompare.tsx
@@ -2,8 +2,10 @@ import React from 'react';
 import { DateTime, dateTime, rangeUtil, TimeRange } from '@grafana/data';
 import { ButtonGroup, ButtonSelect, Checkbox, Icon, ToolbarButton } from '@grafana/ui';
 import { SceneObjectBase } from '../core/SceneObjectBase';
-import { SceneComponentProps, SceneObjectState } from '../core/types';
+import { SceneComponentProps, SceneObjectState, SceneObjectUrlValues } from '../core/types';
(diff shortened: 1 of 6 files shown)


packages/scenes/src/components/SceneTimeRangeCompare.tsx
```diff
@@ -109,6 +112,27 @@ export class SceneTimeRangeCompare
...
+    }
+
+    const update: Partial<SceneTimeRangeCompareState> = {};
+    const compareWith = parseUrlParam(values.compareWith);
+
+    if (compareWith) {
+      update.compareWith = compareWith;
+    }
+
+    this.setState(update);
```
bfmatei, line 134: Should we check the value against the available options? I'm thinking about an edge case where a developer removes one of the options for various reasons and some users still try to access bookmarked/published links.

It may not be the case, it's just an afterthought.

bfmatei (approved):
LGTM with only one comment regarding an edge case.

packages/scenes/src/components/SceneTimeRangeCompare.tsx
dprokop: @bfmatei great catch. Updated. So if the compare URL value is not valid, it will fallback to previous period comparison. Sounds good?

bfmatei: Perfect!

PR #363, opened 2023-09-25
VizPanel: Allow options and field config updates

dprokop:
Closes https://github.com/grafana/scenes/pull/266

This PR implements logic for dynamically updating panel options and field config. 

It also refactors `VizPanelBuilder` a bit, so that field config and options are built using a respective, composed builder. I think this makes sense as for programmatic options/field config updates it's good to have typed builders for `Options` and `FieldConfig` types - at least for the built-in Grafana visualizations. So the user can use exactly the same API as they would use for building panel using `PanelBuilders`.

<details>
  <summary>📦 Published PR as canary version: <code>1.8.0--canary.363.6298066504.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@1.8.0--canary.363.6298066504.0
  # or 
  yarn add @grafana/scenes@1.8.0--canary.363.6298066504.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanel.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.tsx
@@ -30,15 +30,16 @@ import { emptyPanelData } from '../../core/SceneDataNode';
 import { changeSeriesColorConfigFactory } from './colorSeriesConfigFactory';
 import { loadPanelPluginSync } from './registerRuntimePanelPlugin';
 import { getCursorSyncScope } from '../../behaviors/CursorSync';
+import { cloneDeep, merge } from 'lodash';
 
 export interface VizPanelState<TOptions = {}, TFieldConfig = {}> extends SceneObjectState {
-  title: string;
-  description?: string;
   /**
    * This is usually a plugin id that references a core plugin or an external plugin. But this can also reference a
    * runtime registered PanelPlugin registered via function registerScenePanelPlugin.
    */
   pluginId: string;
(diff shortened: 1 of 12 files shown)


packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -171,12 +174,52 @@ export class VizPanel<TOptions = {}, TFieldConfig = {}> extends SceneObjectBase<
...
+
+    // When replace is true, we want to replace the entire field config. Default will be applied.
+    const nextFieldConfig = replace ? fieldConfigUpdate : merge(cloneDeep(fieldConfig), fieldConfigUpdate);
+
+    const withDefaults = getPanelOptionsWithDefaults({
+      plugin: this._plugin!,
+      currentOptions: options,
+      currentFieldConfig: nextFieldConfig,
+      isAfterPluginChange: false,
+    });
```
torkelo, line 218: technically we only need this if replace: true, but might as well always do it

torkelo (approved):
Looks good!

packages/scenes/src/components/VizPanel/VizPanel.tsx
dprokop: Not really. `getPanelOptionsWithDefaults` applies plugin defaults so I think it's quite necessary to always call it. If one wanted to completely replace the config, they would have to provide a complete config to `onFieldConfigChange`.

Issue #211, opened 2023-05-25
Variables: Add support for interval variable type


dprokop:
Reference code in core Grafana: https://github.com/grafana/grafana/blob/main/public/app/features/variables/interval/adapter.ts#L14

PR #365, opened 2023-09-25
Variables: Implement Interval Variable

axelavargas:
Implement Interval Variable based on Core Grafana

Reference code: https://github.com/grafana/grafana/blob/main/public/app/features/variables/interval/adapter.ts#L14

Fixes: #211 

Task 
- [x] Create base interval variable implementation
- [x] Figure out auto interval option logic
- [x] implement URL synchronization 
- [x] Add unit test
- [x] Update scenes-app using interval variable
- [x] Create a better PR Description

Notes:

In the variables demo, I noticed a delay in the interpolation of the interval variable when it is configured to auto. This seems to be related to the current architecture for processing variable dependencies

https://github.com/grafana/scenes/assets/239999/d66a2d4b-42a3-4dad-866f-ddbd01057548


packages/scenes/src/variables/variants/IntervalVariable.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { config, Observable, of } from 'rxjs';
+import { sceneGraph } from '../../core/sceneGraph';
+import { SceneComponentProps } from '../../core/types';
+import { renderSelectForVariable } from '../components/VariableValueSelect';
+import { VariableValueOption } from '../types';
+import { VariableDependencyConfig } from '../VariableDependencyConfig';
+import { MultiValueVariable, MultiValueVariableState, VariableGetOptionsArgs } from './MultiValueVariable';
+
+export interface IntervalVariableState extends MultiValueVariableState {
+  query: string;
```
dprokop, line 12: Why query for interval variable option?

packages/scenes/src/variables/variants/IntervalVariable.tsx
```diff
@@ -0,0 +1,81 @@
...
+  auto_min: string;
+  auto_count: number;
+  refresh: VariableRefresh;
+}
+
+export class IntervalVariable extends MultiValueVariable<IntervalVariableState> {
+  // does it have dependencies?
+  protected _variableDependency = new VariableDependencyConfig(this, {
+    statePaths: ['query'],
+  });
```
dprokop, line 23: I don't think it does.

packages/scenes/src/variables/variants/IntervalVariable.tsx
dprokop: Ah yeah, I see we use it for the actual options in the core. NVM comment above. We do the same thing in the custom variable.

Think it would be maybe wise to have the interfaces here a little bit more verbose, rather than using `query` which in case of variables like interval or custom is meaningless IMO.

packages/scenes/src/variables/variants/IntervalVariable.tsx
```diff
@@ -0,0 +1,81 @@
...
+      options.push({ label: 'No interval found', text: '', value: '', selected: false });
+    }
+
+    return of(options);
+  }
+
+  private getAutoRefreshInteval(minRefreshInterval) {
+    const timeRange = sceneGraph.getTimeRange(this).state.value || getDefaultTimeRange();
+    // where is this window coming from?
+    const resolution = window?.innerWidth ?? 2000;
```
dprokop, line 70: `window` is a global object.

Why's this function in the Interval variable implementation?

axelavargas: It was my failed attempt to understand how the auto-configuration was working, I saw a similar function being called from core Grafana for the interval variable :)

packages/scenes/src/variables/variants/IntervalVariable.tsx
axelavargas: thank you ❤️

packages/scenes/src/variables/variants/IntervalVariable.tsx
dprokop: I think it will still make sense when the `getValue` is implemented, so maybe keep it here for now until you start working on resolving the auto value.

packages/scenes/src/variables/variants/IntervalVariable.tsx
```diff
@@ -1,11 +1,10 @@
...
+import { Observable, of } from 'rxjs';
 import { sceneGraph } from '../../core/sceneGraph';
 import { SceneComponentProps } from '../../core/types';
 import { renderSelectForVariable } from '../components/VariableValueSelect';
-import { VariableValueOption } from '../types';
-import { VariableDependencyConfig } from '../VariableDependencyConfig';
+import { SceneVariableValueChangedEvent, VariableValue, VariableValueOption } from '../types';
 import { MultiValueVariable, MultiValueVariableState, VariableGetOptionsArgs } from './MultiValueVariable';
 
 export interface IntervalVariableState extends MultiValueVariableState {
```
torkelo, line 10: I think it's better not to extend MultiValueVariableState and implement the SceneVariable Interface directly, see TextBoxVariable for example. 

You can then implement validateAndUpdate, instead of subscribing to the time range in the activation handler (you can remove this bit). validateAndUpdate will be called when time range changes by SceneVariableSet if you set refresh: VariableRefresh.onTimeRangeChanged (which you already do). 

In validateAndUpdate you can do this

 ```ts
  public validateAndUpdate(): Observable<ValidateAndUpdateResult> {
    const { value } = this.state;

    // If auto the value can change (can optimize this bit later and only publish this when the calculated value actually changed) 
    if (value === '$auto') {
      this.publishEvent(new SceneVariableValueChangedEvent(this), true);
    } else if (!this.state.value) {
      // Todo set to first option in this.state.intervals
      this.setState({ value: '1m' });
      this.publishEvent(new SceneVariableValueChangedEvent(this), true);
    }

    return of({});
  }
```

You also need to implement UI component, something like this 


 ```tsx

public _onChange = (value: SelectableValue<string>) => {
    this.setState({ value: value.value! });
    this.publishEvent(new SceneVariableValueChangedEvent(this), true);
  };

 public static Component = ({ model }: SceneComponentProps<IntervalVariable>) => {
    const { key, value } = model.useState();

    return (
      <Select
        id={key}
        placeholder="Select value"
        width="auto"
        value={value}
        tabSelectsValue={false}
        options={model.getOptionsForSelect()}
        onChange={model._onChange}
      />
    );
  };
```

packages/scenes/src/variables/variants/IntervalVariable.tsx
```diff
@@ -0,0 +1,103 @@
...
+      query: '1m,10m,30m,1h,6h,12h,1d,7d,14d,30d',
+      value: '',
+      text: '',
+      name: '',
+      auto_count: 30,
+      auto_min: '10s',
+      options: [],
+      auto: false,
+      refresh: VariableRefresh.onTimeRangeChanged,
+      ...initialState,
```
torkelo, line 31: instead of query, put the intervals in an intervals array (string[]),  and remove options (this is not needed here since you should extend MultiValueVariable). 

auto_count, auto_min, & auto => can go to a 

```
autoEnabled?: boolean;
autoStepCount: number;  
autoMinInterval: string
```

We don't need to use the old property names associated with the legacy variable system (we can map to those when we translate from and to save model in main repo)

torkelo:
@axelavargas looks close just some tests missing

torkelo (approved):
This looks great! 

The tests are nice! Esp compared to the old messy redux tests that are so hard to read: 
https://github.com/grafana/grafana/blob/main/public/app/features/variables/interval/actions.test.ts

PR #370, opened 2023-09-26
Demos: Client side data filtering

dprokop:
Adds a simple demo showcasing how to implement basic regex filtering of the data from `SceneQueryRunner`.


torkelo (approved):
Started thinking about how we can package this as a reusable component but it's pretty tricky (unless the SearchBox component takes in say ref to data provider, and modifies the parent to wrap in a SceneDataTransformer (if not already wrapped). etc. becomes a bit messy but would be nice to find away.

PR #371, opened 2023-09-27
SceneVariableSet: Show and log errors

### LINKED ISSUE #349: SceneVariableSet: Improve error handling & logging
Currently when an error happens in the variable getValueOptions process (rxjs) it ends up in handleVariableError in SceneVariableSet that just logs it using scene info logging (which is by default disabled), so it becomes silent in the console log. 

We should do a console.error here and show an error icon+tooltip in the variable label UI 

https://github.com/grafana/scenes/blob/main/packages/scenes/src/variables/sets/SceneVariableSet.ts#L237

torkelo:
Fixes #349

Shows and logs variable errors happening in the validateAndUpdate observable / rxjs process 

image


torkelo:
added a test

dprokop (approved):

Issue #299, opened 2023-08-29
SceneTimeRange: Implement onZoom behavior


dprokop:
Zoom-out button in `SceneTimePicker` has no effect on the time range. 

Corresponding implementation in core Grafana that needs to be replicated in Scenes: https://github.com/grafana/grafana/blob/main/public/app/features/dashboard/services/TimeSrv.ts#L347

Corresponding code that's misssing in the library: https://github.com/grafana/scenes/blob/main/packages/scenes/src/components/SceneTimePicker.tsx#L37

PR #374, opened 2023-09-28
SceneTimeRange: Implement onZoom behavior

polibb:
Zoom-out button in `SceneTimePicker` has an effect on the time range.

To do:
- [x] tests

This implementation only handles zooming out,
I'm not sure what to do with the refresh and setting of the URL that happen on zoom-out in core: https://github.com/grafana/grafana/blob/main/public/app/features/dashboard/services/TimeSrv.ts#L290. 

Fixes https://github.com/grafana/scenes/issues/299

<details>
  <summary>📦 Published PR as canary version: <code>1.11.2--canary.374.6394808989.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@1.11.2--canary.374.6394808989.0
  # or 
  yarn add @grafana/scenes@1.11.2--canary.374.6394808989.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/SceneTimePicker.tsx
+++ b/packages/scenes/src/components/SceneTimePicker.tsx
@@ -25,6 +25,10 @@ function SceneTimePickerRenderer({ model }: SceneComponentProps<SceneTimePicker>
     return null;
   }
 
+  function onZoom() {
+    timeRange.onZoom(2);
+  }
+
   return (
     <TimeRangePicker
       isOnCanvas={isOnCanvas ?? true}
@@ -34,7 +38,7 @@ function SceneTimePickerRenderer({ model }: SceneComponentProps<SceneTimePicker>
       fiscalYearStartMonth={0}
       onMoveBackward={() => {}}
       onMoveForward={() => {}}
-      onZoom={() => {}}
+      onZoom={onZoom}
       onChangeTimeZone={timeRange.onTimeZoneChange}
       onChangeFiscalYearStartMonth={() => {}}
     />
--- a/packages/scenes/src/components/SceneTimePicker.tsx
+++ b/packages/scenes/src/components/SceneTimePicker.tsx
@@ -20,6 +20,11 @@ export class SceneTimePicker extends SceneObjectBase<SceneTimePickerState> {
     const zoomedTimeRange = getZoomedTimeRange(timeRange.state.value, 2);
     timeRange.onTimeRangeChange(zoomedTimeRange);
   };
+
+  public onChangeFiscalYearStartMonth = (month: number) => {
+    const timeRange = sceneGraph.getTimeRange(this);
+    timeRange.setState({ fiscalYearStartMonth: month });
+  };
 }
 
 function SceneTimePickerRenderer({ model }: SceneComponentProps<SceneTimePicker>) {
@@ -38,12 +43,12 @@ function SceneTimePickerRenderer({ model }: SceneComponentProps<SceneTimePicker>
       value={timeRangeState.value}
       onChange={timeRange.onTimeRangeChange}
       timeZone={timeZone}
-      fiscalYearStartMonth={0}
+      fiscalYearStartMonth={timeRangeState.fiscalYearStartMonth}
       onMoveBackward={() => {}}
       onMoveForward={() => {}}
       onZoom={model.onZoom}
       onChangeTimeZone={timeRange.onTimeZoneChange}
-      onChangeFiscalYearStartMonth={() => {}}
+      onChangeFiscalYearStartMonth={model.onChangeFiscalYearStartMonth}
     />
   );
 }
--- a/packages/scenes/src/components/SceneTimePicker.tsx
+++ b/packages/scenes/src/components/SceneTimePicker.tsx
@@ -8,7 +8,6 @@ import { SceneComponentProps, SceneObjectState } from '../core/types';
 import { TimeRange, toUtc } from '@grafana/data';
 
 export interface SceneTimePickerState extends SceneObjectState {
-  onZoom?: (timeRange: TimeRange, callback: (timeRange: TimeRange) => void) => void;
   hidePicker?: boolean;
   isOnCanvas?: boolean;
 }
@@ -16,14 +15,14 @@ export interface SceneTimePickerState extends SceneObjectState {
 export class SceneTimePicker extends SceneObjectBase<SceneTimePickerState> {
   public static Component = SceneTimePickerRenderer;
 
-  public onZoom = (timeRange: TimeRange, callback: (timeRange: TimeRange) => void) => {
+  public onZoom(timeRange: TimeRange, callback: (timeRange: TimeRange) => void): void {
     const zoomedTimeRange = getZoomedTimeRange(timeRange, 2);
     callback(zoomedTimeRange);
-  };
+  }
 }
 
 function SceneTimePickerRenderer({ model }: SceneComponentProps<SceneTimePicker>) {
-  const { hidePicker, isOnCanvas, onZoom } = model.useState();
+  const { hidePicker, isOnCanvas } = model.useState();
   const timeRange = sceneGraph.getTimeRange(model);
   const timeZone = timeRange.getTimeZone();
   const timeRangeState = timeRange.useState();
@@ -41,7 +40,7 @@ function SceneTimePickerRenderer({ model }: SceneComponentProps<SceneTimePicker>
       fiscalYearStartMonth={0}
       onMoveBackward={() => {}}
       onMoveForward={() => {}}
-      onZoom={() => onZoom?.(timeRangeState.value, timeRange.onTimeRangeChange)}
+      onZoom={() => model.onZoom(timeRangeState.value, timeRange.onTimeRangeChange)}
       onChangeTimeZone={timeRange.onTimeZoneChange}
       onChangeFiscalYearStartMonth={() => {}}
     />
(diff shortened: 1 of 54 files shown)


packages/scenes/src/core/SceneTimeRange.tsx
```diff
@@ -109,6 +109,34 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
...
+    const zoomedTimeRange = this.getZoomedTimeRange(scale);
+    const rawZoomedTimeRange = {
+      from: toUtc(zoomedTimeRange.from),
+      to: toUtc(zoomedTimeRange.to),
+    };
+
+    update.value = evaluateTimeRange(rawZoomedTimeRange.from, rawZoomedTimeRange.to, this.getTimeZone());
+
+    this.setState(update);
+  };
```
torkelo, line 138: not sure we need this on the SceneTimeRange , I think it can live on the SceneTimePicker , and then it can just call  onTimeRangeChange  (and we can let SceneTimeRangeLike remain slim)

torkelo: but it's close, I don't think zoom out will be a popular use case do call programmatically  / from other places (other than the keyboard shortcut, but the picker can registera and handle that)

torkelo: so basically move this to function on SceneTimeRangePicker and have it call onTimeRangeChange on the timeRange object

dprokop: I agree, think it's time picker specific action, rather than SceneTimeRangeLike interface. 

Nice first stab @polibb BTW!

packages/scenes/src/components/SceneTimePicker.tsx
```diff
@@ -25,6 +26,12 @@ function SceneTimePickerRenderer({ model }: SceneComponentProps<SceneTimePicker>
     return null;
   }
 
+  function onZoom() {
```
torkelo, line 29: think this needs to be an arrow function :)

torkelo: ah it's inside the render / react function.

torkelo (commented):
Looks super good and close to mergable :tada:  just needs 1-2 tests for the zoom out logic (which you can probably copy from core

packages/scenes/src/components/SceneTimePicker.tsx
```diff
@@ -5,18 +5,25 @@ import { TimeRangePicker } from '@grafana/ui';
 import { SceneObjectBase } from '../core/SceneObjectBase';
 import { sceneGraph } from '../core/sceneGraph';
 import { SceneComponentProps, SceneObjectState } from '../core/types';
+import { TimeRange, toUtc } from '@grafana/data';
 
 export interface SceneTimePickerState extends SceneObjectState {
+  onZoom?: (timeRange: TimeRange, callback: (timeRange: TimeRange) => void) => void;
```
dprokop, line 11: I think this is not necessary - one will not configure the time picker with a callback to zoom event.

packages/scenes/src/components/SceneTimePicker.tsx
```diff
@@ -5,18 +5,25 @@ import { TimeRangePicker } from '@grafana/ui';
...
+
+  public onZoom = (timeRange: TimeRange, callback: (timeRange: TimeRange) => void) => {
+    const zoomedTimeRange = getZoomedTimeRange(timeRange, 2);
+    callback(zoomedTimeRange);
+  };
 }
 
 function SceneTimePickerRenderer({ model }: SceneComponentProps<SceneTimePicker>) {
-  const { hidePicker, isOnCanvas } = model.useState();
+  const { hidePicker, isOnCanvas, onZoom } = model.useState();
```
dprokop, line 26: you don't neet to have onZoom in state. It can be just model method, accessed down ther as `model.onZoom`

packages/scenes/src/components/SceneTimePicker.tsx
```diff
@@ -34,9 +41,21 @@ function SceneTimePickerRenderer({ model }: SceneComponentProps<SceneTimePicker>
       fiscalYearStartMonth={0}
       onMoveBackward={() => {}}
       onMoveForward={() => {}}
-      onZoom={() => {}}
+      onZoom={() => onZoom?.(timeRangeState.value, timeRange.onTimeRangeChange)}
```
dprokop, line 44: I think this can be as simple as `model.onZoom` :)

packages/scenes/src/components/SceneTimePicker.tsx
```diff
@@ -5,18 +5,25 @@ import { TimeRangePicker } from '@grafana/ui';
...
+  onZoom?: (timeRange: TimeRange, callback: (timeRange: TimeRange) => void) => void;
   hidePicker?: boolean;
   isOnCanvas?: boolean;
 }
 
 export class SceneTimePicker extends SceneObjectBase<SceneTimePickerState> {
   public static Component = SceneTimePickerRenderer;
+
+  public onZoom = (timeRange: TimeRange, callback: (timeRange: TimeRange) => void) => {
+    const zoomedTimeRange = getZoomedTimeRange(timeRange, 2);
```
dprokop, line 20: The arguments for this function may be unnecessary. Instead, you can use `sceneGraph.getTimeRange` directly in this method, and call `timeRange.onTimeRangeChange` directly here

polibb: sure, but I have to setup sceneGraph both here and in the Renderer. Isn't that too much - twice for one thing?

polibb: Besides setting it up twice, today it's sceneGraph's `timeRange.onTimeRangeChange` that happens after the zooming is calculated, but tomorrow something else might be called after the zoom is calculated. So I suppose a callback is not that bad.

packages/scenes/src/components/SceneTimePicker.tsx
```diff
@@ -34,9 +40,21 @@ function SceneTimePickerRenderer({ model }: SceneComponentProps<SceneTimePicker>
       fiscalYearStartMonth={0}
       onMoveBackward={() => {}}
       onMoveForward={() => {}}
-      onZoom={() => {}}
+      onZoom={() => model.onZoom(timeRangeState.value, timeRange.onTimeRangeChange)}
```
torkelo, line 43:
```suggestion
      onZoom={model.onZoom}
```

No no need to create a new function here on every render :)

packages/scenes/src/utils/evaluateTimeRange.ts
```diff
@@ -1,15 +1,21 @@
...
+  const raw = {
+    from: isDateTime(from) ? dateTime(from) : from,
+    to: isDateTime(to) ? dateTime(to) : to,
+  };
+
   return {
-    from: dateMath.parse(from, false, timeZone, fiscalYearStartMonth)!,
-    to: dateMath.parse(to, true, timeZone, fiscalYearStartMonth)!,
+    from: dateMath.parse(raw.from, false, timeZone, fiscalYearStartMonth)!,
+    to: dateMath.parse(raw.to, true, timeZone, fiscalYearStartMonth)!,
```
torkelo, line 18: Don't think we need these changes? from and to should always be strings the way this function is designed (SceneTimeRangeState from & to are always strings)

packages/scenes/src/components/SceneTimePicker.tsx
polibb: okay, nevermind 😊 I got a response on that

packages/scenes/src/utils/evaluateTimeRange.ts
polibb: yeah, no need to do that here, it's leftover from other changes

packages/scenes/src/components/SceneTimePicker.tsx
```diff
@@ -13,6 +14,14 @@ export interface SceneTimePickerState extends SceneObjectState {
 
 export class SceneTimePicker extends SceneObjectBase<SceneTimePickerState> {
   public static Component = SceneTimePickerRenderer;
+
+  public onZoom = () => {
+    const timeRange = sceneGraph.getTimeRange(this);
+    const timeRangeState = timeRange.useState();
```
torkelo, line 20: useState should only be called from the react component (it's a hook)

packages/scenes/src/components/SceneTimePicker.tsx
```diff
@@ -13,6 +14,14 @@ export interface SceneTimePickerState extends SceneObjectState {
...
+
+    const zoomedTimeRange = getZoomedTimeRange(timeRangeState.value, 2);
```
torkelo, line 22:
```suggestion
    const zoomedTimeRange = getZoomedTimeRange(timeRange.state.value, 2);
```

torkelo (approved):
Very close! just need to remove the useState (line 20)

packages/scenes/src/components/SceneTimePicker.tsx
```diff
@@ -34,9 +43,21 @@ function SceneTimePickerRenderer({ model }: SceneComponentProps<SceneTimePicker>
...
       onMoveForward={() => {}}
-      onZoom={() => {}}
+      onZoom={model.onZoom}
       onChangeTimeZone={timeRange.onTimeZoneChange}
       onChangeFiscalYearStartMonth={() => {}}
     />
   );
 }
+
+function getZoomedTimeRange(timeRange: TimeRange, factor: number): TimeRange {
```
dprokop, line 53: A test for this fn would be good.

dprokop (approved):
This is looking good @polibb! Think one last comment from Torkel is there to be addressed and we should be good to merge :)

PR #380, opened 2023-10-02
SceneAppPage: Custom fallback page

domasx2:
* Adds ability to specify custom fallback page on `SceneAppPage`
* Update `dynamicPage` demo to use this feature to show loading message while tab is loading

Context:
App o11y service overview needs to load some metadata from metrics data source before it can figure out what tabs to show for a particular service.  For this purpose it uses a custom "loader" scene that initializes required variables and a query runner to load the metadata. But if user cold opens URL for a tab, tabs are not initialized yet. Currently we rely on a routing "hack" to show the loader scene instead of "page not  found" fallback scene. This can be nicely solved by allowing `SceneAppPage` to provide a custom fallback page.

Related discussion: https://raintank-corp.slack.com/archives/C04FRQ2S0AV/p1695390081582359

I pondered if a high level component could be made for this case, but not sure how to implement it to solve multiple possible use cases and actually add value. With custom fallback page the solution is already pretty terse and easy. May combe back to it later

### DIFF
--- a/packages/scenes-app/src/demos/dynamicPage.tsx
+++ b/packages/scenes-app/src/demos/dynamicPage.tsx
@@ -8,6 +8,7 @@ import {
   SceneRefreshPicker,
   SceneAppPageState,
   PanelBuilders,
+  SceneReactObject,
 } from '@grafana/scenes';
 import React from 'react';
 import { demoUrl } from '../utils/utils.routing';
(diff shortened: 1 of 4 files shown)


torkelo (approved):
This looks great! 

Maybe it would be good with a default "Loading" page in the scene lib at some point (we can wait and see what UI page we end up with here and how reusable it is). 

So you could define
`getFallbackPage: fallbackPages.LoadingPage`  (or something like that)

PR #381, opened 2023-10-02
SceneAppPageView: Fixes react and scene state missmatch

torkelo:
If the SceneApp is created with useMemo which is invalidated shortly after initial loading (causing unmount & remount of the embedded scene). This caused a state mismatch in SceneAppPageView where the local react state initialized did not match the scene state of the page containerState.initializedScene 

https://github.com/grafana/grafana-k8s-plugin/pull/819/files#diff-3bfbfb0a9f4447beaacffbc8799fc57325273a246e52244ba42dcb8d420966fdR43

Fix

* Reverted back to using only page state to know if it it initialized. And to solve the issue of re-initializing it (needed for Url sync) when coming back I clear initializedState when SceneAppPageView is unmounted. 

Not sure how to write a unit test for this, was only able to replicate it with this change to how we create the demo app

```ts

export const DemoListPage = () => {
  const [isLoading, setIsLoading] = useState(false);
  const scene = useMemo(() => getDemoSceneApp(isLoading), [isLoading]);

  useEffect(() => {
    if (!isLoading) {
      setTimeout(() => {
        setIsLoading(true);
      }, 1);
    }
  }, [isLoading]);

  return <scene.Component model={scene} />;
};
```

<details>
  <summary>📦 Published PR as canary version: <code>1.9.1--canary.381.6378419140.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@1.9.1--canary.381.6378419140.0
  # or 
  yarn add @grafana/scenes@1.9.1--canary.381.6378419140.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/SceneApp/SceneAppPageView.tsx
+++ b/packages/scenes/src/components/SceneApp/SceneAppPageView.tsx
@@ -1,6 +1,6 @@
 import { NavModelItem, UrlQueryMap } from '@grafana/data';
 import { PluginPage } from '@grafana/runtime';
-import React, { useLayoutEffect, useState } from 'react';
+import React, { useEffect, useLayoutEffect } from 'react';
 
 import { RouteComponentProps } from 'react-router-dom';
 import { SceneObject } from '../../core/types';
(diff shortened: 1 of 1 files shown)


packages/scenes/src/components/SceneApp/SceneAppPageView.tsx
```diff
@@ -20,17 +20,21 @@ export function SceneAppPageView({ page, routeProps }: Props) {
...
     }
-  }, [initialized, scene, containerPage]);
+  }, [scene, containerPage, isInitialized]);
 
-  if (!initialized) {
+  // Clear initializedScene when unmounting
+  useEffect(() => {
+    return () => containerPage.setState({ initializedScene: undefined });
+  }, [containerPage]);
+
```
torkelo, line 36: So I originally used only the `containerState.initializedScene === scene;` to know if the scene was initialized but I added the React.useState instead to make sure that `containerPage.initializeScene(scene);` is called when coming back from a drilldown. As calling containerPage.initializeScene is needed to initialize urlsync.

dprokop (approved):

PR #382, opened 2023-10-02
AdhocFilters: Pass filters via request object

torkelo:
With the change in core to pass filters via request object, and to start using it in Prometheus made it so that the new adhoc filters feature was not working with scenes anymore (for prometheus). 

So I have to make this change now (before updating latest packages, hence the typescript ignore).

related to 
https://github.com/grafana/grafana/pull/75552

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.ts
@@ -72,8 +72,9 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
   private _results = new ReplaySubject<SceneDataProviderResult>();
   private _scopedVars = { __sceneObject: { value: this, text: '__sceneObject' } };
 
-  // Closest filter set (if any)
-  private _filterSet?: AdHocFilterSet;
+  // Closest filter set if found)
+  private _adhocFilterSet?: AdHocFilterSet;
+  private _adhocFilterSub?: Unsubscribable;
 
   public getResultsStream() {
     return this._results;
@@ -242,6 +243,10 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
       this._dataLayersSub = undefined;
     }
 
+    if (this._adhocFilterSub) {
+      this._adhocFilterSub.unsubscribe();
+    }
+
     this._variableValueRecorder.recordCurrentDependencyValuesForSceneObject(this);
   }
 
@@ -332,6 +337,10 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
       const datasource = this.state.datasource ?? findFirstDatasource(queries);
       const ds = await getDataSource(datasource, this._scopedVars);
 
+      if (!this._adhocFilterSet) {
+        this.findAndSubscribeToAdhocFilters(ds.uid);
+      }
+
       const [request, secondaryRequest] = this.prepareRequests(timeRange, ds);
 
       writeSceneLog('SceneQueryRunner', 'Starting runRequest', this.state.key);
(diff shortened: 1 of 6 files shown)


packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -486,6 +491,45 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
...
+
+    const setDataSource = getDataSourceSrv().getInstanceSettings(set?.state.datasource, this._scopedVars);
+
+    if (setDataSource?.uid !== ourDataSourceUid) {
+      return undefined;
+    }
+
+    // Subscribe to filter set state changes so that queries are re-issued when it changes
+    this._filterSet = set;
+    this._subs.add(this._filterSet?.subscribeToState(() => this.runQueries()));
```
dprokop, line 529: Are you sure we should subscribe here? Basically, we are creating a new subscription every time a request is issued. Think it should be set in the activation handler.

dprokop (approved):
Looks good, but there's one thing we should change I think.

packages/scenes/src/querying/SceneQueryRunner.ts
torkelo: @dprokop no, that would be bad.  check the first few lines of the function, we return the filterSet if we wound it already (and subscribed to it). So the subscription here should only happen once during the lifetime of this query runner 

What is needed here is some support for changing filterSets (but think we can wait until that need arise)

dprokop: > @dprokop no, that would be bad. check the first few lines of the function, we return the filterSet if we wound it already (and subscribed to it). So the subscription here should only happen once during the lifetime of this query runner

Yeah, I see the early return. But still, this feels wrong to have a side effect in a getter. Sort of against what we say about activation handlers. 
The `getAdhocFilters` is similar to `getTimeCompare`, IMO we should use the same way for solving the same problem, at least within a single class.

torkelo: @dprokop yea I agree. not sure why I did it this way, I had it in the activation handler first. think it was trying to support changing filter sets (Which I later choose not to handle anyway)

torkelo: will try to refactor this a bit

torkelo: @dprokop now I remember why, it's because I need the scene query runner's resolved data source and that code is in runWithTimeRange

torkelo: @dprokop there is a similarity here to how collect and subscribe to data layers which also happens in runWithTimeRange

dprokop: > now I remember why, it's because I need the scene query runner's resolved data source and that code is in runWithTimeRange

aha, I see. And true, this is similar to data layer handling. Still feels wrong we now have actually 3 _slightly_ different implementations of such a subscription. 

Since this is resembling data layers, maybe an implementation similar to those would make sense? Basically, outside check (in runWithTimeRange) whether or not a subscription to filters exist, and then `handleFilters....` implementation. My only point here actually is just the side effect in the getter, the rest is fine.

torkelo: @dprokop refactored this to be a more clear and more similar to how the data layers sub is handled

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -525,10 +528,8 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
     }
 
     // Subscribe to filter set state changes so that queries are re-issued when it changes
-    this._filterSet = set;
-    this._subs.add(this._filterSet?.subscribeToState(() => this.runQueries()));
-
-    return this._filterSet.state.filters;
+    this._adhocFilterSet = set;
```
dprokop, line 531: weird `set` ain't a reserved word

dprokop (approved):
Looks good now with that small refactor

PR #389, opened 2023-10-03
SceneTimeRange: onTimeZoneChange and onTimeRangeChange to set

tskarhed:
This step will help a little bit, but I think I want to make another PR to change these to `setTimeZone` and `setTimeRange`.

Here is my suggestion for doing so, please give me your thoughts:

1. Mark these as deprecated
2. Add `setTimeZone` and `setTimeRange`.
3. Call the new methods from the deprecated methods
4. Update unit test with new methods

Then it is possible to do a gradual change. Thoughts?

### DIFF
--- a/docusaurus/docs/advanced-data.md
+++ b/docusaurus/docs/advanced-data.md
@@ -65,6 +65,30 @@ The subscription returned from `sourceData.subscribeToState` is added to `this._
 
 Similarly to data, you can use the closest time range in a custom scene object using `sceneGraph.getTimeRange(model)`. This method can be used both in the custom object class and the renderer, as described previously in the [Use data](#use-data) section.
 
+Sometimes you may need to update a time range dynamically. You can do this by using the `setTimeRange` method.
+
+```tsx
+import { SceneTimeRange } from '@grafana/scenes';
+import { toUtc } from '@grafana/data';
+...
+
+const localTimeRange = new SceneTimeRange(); // Timerange defaults to the last six hours
+
+...
+
+const from = toUtc(1696405657);
+const to = toUtc(1696405687);
+
+localTimeRange.setTimeRange({
+  raw: {
+    from,
+    to
+  }
+  from,
+  to
+});
+```
+
 ## Source code
 
 [View the example source code](https://github.com/grafana/scenes/tree/main/docusaurus/docs/advanced-data.tsx)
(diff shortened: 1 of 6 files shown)


dprokop:
Thanks @tskarhed . This sounds like a good plan! Mind adding docs as well for what you've just learned about setting time range externally? Think adding it to https://grafana.com/developers/scenes/advanced-data#use-time-range would make sense.

torkelo (commented):
would this not break the guideline we have for react callback handlers to begin with onXX ? to match when we hook them up like this. 

image

I do agree that this naming does make it feel like something you should maybe not call manually (outside a react callback handler). 

Do we have any other onXX interface/ scene object methods in scenes (want to make sure we do not add some inconsistency here). I think MultiValueVariable has changeValueTo (so not using onXX). 

Not seeing and change to stop using the deprecated versions here

docusaurus/docs/advanced-data.md
```diff
@@ -65,6 +65,30 @@ The subscription returned from `sourceData.subscribeToState` is added to `this._
...
+import { SceneTimeRange } from '@grafana/scenes';
+import { toUtc } from '@grafana/data';
+...
+
+const localTimeRange = new SceneTimeRange(); // Timerange defaults to the last six hours
+
+...
+
+const from = toUtc(1696405657);
+const to = toUtc(1696405687);
```
tskarhed, line 80: Use `dateTime` when it is an epoch

tskarhed:
> would this not break the guideline we have for react callback handlers to begin with onXX ? to match when we hook them up like this.
> 
> image
> 
> I do agree that this naming does make it feel like something you should maybe not call manually (outside a react callback handler).
> 
> Do we have any other onXX interface/ scene object methods in scenes (want to make sure we do not add some inconsistency here). I think MultiValueVariable has changeValueTo (so not using onXX).
> 
> Not seeing and change to stop using the deprecated versions here

Hmm, I see your point. Would it make sense for these two to just exist in parallel, just for the sake of good naming? You would use `setTimeRange` whenever you need to do it manually, and `onTimeRangeChange` for callback handling.

dprokop:
>Hmm, I see your point. Would it make sense for these two to just exist in parallel, just for the sake of good naming? You would use setTimeRange whenever you need to do it manually, and onTimeRangeChange for callback handling.

@torkelo  @tskarhed I don't think having two makes sense. API is one thing, the callback is another. For the `SceneTimePicker` and any other usages that are callbacks the model can wrap the API in a higher-level function that will follow the `on...` convention. But since we are talking about API that can be used in multiple contexts, I think we need to name it in a way that is understandable and context agnostic. With that in mind I would suggest callin it `updateTimeRange` or `setTimeRange` and follow this for all the public model methods that can be used in an ambiguous way.

torkelo:
@dprokop yea, agree.

PR #390, opened 2023-10-04
Compatability: Add global window object that points to the current active EmbeddedScene 

torkelo:
* This refactors the timeSrv "patching" (not really patching) so that it instead of setting a time range object as on the window it just sets the active EmbeddedScene. This way we can re-use the same object to access the current scene time range and variables (From TimeSrv and TemplateSrv). 

I tried to update the tests but found they where really testing unrelated things. It was testing if SceneAppPage and the EmbeddedScene was activated and connected to it's parent. So added some tests in that vein to SceneApp (where we test SceneAppPage). 

Will create a corresponding PR in main to leverage this change . 

<details>
  <summary>📦 Published PR as canary version: <code>1.13.0--canary.390.6403804001.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@1.13.0--canary.390.6403804001.0
  # or 
  yarn add @grafana/scenes@1.13.0--canary.390.6403804001.0
  ```
</details>


packages/scenes/src/utils/compatibility/setWindowGrafanaScene.ts
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Adds the scene object to the global state window state so that templateSrv in core can interpolate strings using the scene interpolation engine with the correct SceneObject scope.
+ * This is needed for old data sources that call templateSrv.replace without passing scopedVars. For example in DataSourceAPI.metricFindQuery.
+ */
+export function setWindowGrafanaScene(activeScene: SceneObject) {
+  const prevScene = (window as any).__grafanaScene;
+
+  writeSceneLog('setWindowGrafanaScene', 'set window.__grafanaScene', activeScene);
+  (window as any).__grafanaScene = activeScene;
```
torkelo, line 12: any better ideas for the naming of this variable?

dprokop: `__appSceneRoot`, `__sceneRoot` ? I don't think this name really matters given the verbose description above.

dprokop (approved):

PR #392, opened 2023-10-04
[cr] creates SceneCSSGridLayout to use CSS Grid with SceneFlexItems

jewbetcha:
This custom scene object will let you build a CSS Grid experience while using SceneFlexItem inside. Setting as a draft right now since I'm sure there will be some issues to iron out.

### DIFF
--- a/packages/scenes/src/components/layout/CSSGrid/SceneCSSGridLayout.tsx
+++ b/packages/scenes/src/components/layout/CSSGrid/SceneCSSGridLayout.tsx
@@ -1,0 +1,119 @@
+import { css, CSSObject } from '@emotion/css';
+import { config } from '@grafana/runtime';
+import React, { ComponentType, CSSProperties, useMemo } from 'react';
+
+import { SceneObjectBase } from '../../../core/SceneObjectBase';
+import { SceneComponentProps, SceneLayout, SceneObjectState, SceneObject } from '../../../core/types';
+
+export interface SceneFlexItemStateLike extends SceneObjectState {}
+
+export interface SceneFlexItemLike extends SceneObject<SceneFlexItemStateLike> {}
+
+interface SceneCSSGridLayoutState extends SceneObjectState, SceneCSSGridItemPlacement {
+  children: SceneFlexItemLike[];
+}
+
+export class SceneCSSGridLayout extends SceneObjectBase<SceneCSSGridLayoutState> implements SceneLayout {
+  public static Component = SceneCSSGridLayoutRenderer;
+
+  public isDraggable(): boolean {
+    return false;
+  }
+}
+
+function SceneCSSGridLayoutRenderer({ model, parentState }: SceneCSSGridItemRenderProps<SceneCSSGridLayout>) {
+  const { children, isHidden } = model.useState();
+  const style = useLayoutStyle(model.state);
+
+  if (isHidden) {
+    return null;
+  }
+
+  return (
+    <div className={style}>
+      {children.map((item) => {
+        const Component = item.Component as ComponentType<SceneCSSGridItemRenderProps<SceneObject>>;
+        return <Component key={item.state.key} model={item} parentState={model.state} />;
+      })}
+    </div>
+  );
+}
+
+export interface SceneCSSGridItemPlacement {
+  rows: CSSProperties['gridTemplateRows'];
+  columns: CSSProperties['gridTemplateColumns'];
+  rowGap?: CSSProperties['rowGap'];
+  columnGap?: CSSProperties['columnGap'];
+  justifyItems?: CSSProperties['justifyItems'];
+  alignItems?: CSSProperties['alignItems'];
+  justifyContent?: CSSProperties['justifyContent'];
+
+  /**
+   * True when the item should rendered but not visible.
+   * Useful for conditional display of layout items
+   */
+  isHidden?: boolean;
+
+  md?: SceneCSSGridItemPlacement;
+}
+
+export interface SceneCSSGridItemState extends SceneCSSGridItemPlacement, SceneObjectState {
+  body: SceneObject | undefined;
+}
+
+interface SceneCSSGridItemRenderProps<T> extends SceneComponentProps<T> {
+  parentState?: SceneCSSGridItemPlacement;
+}
+
+export class SceneCSSGridItem extends SceneObjectBase<SceneCSSGridItemState> {
+  public static Component = SceneCSSGridItemRenderer;
+}
+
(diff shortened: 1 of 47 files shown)


packages/scenes/src/components/layout/CSSGrid/SceneCSSGridLayout.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+import { SceneObjectBase } from '../../../core/SceneObjectBase';
+import { SceneComponentProps, SceneLayout, SceneObjectState, SceneObject } from '../../../core/types';
+
+export interface SceneFlexItemStateLike extends SceneObjectState {}
+
+export interface SceneFlexItemLike extends SceneObject<SceneFlexItemStateLike> {}
+
+interface SceneCSSGridLayoutState extends SceneObjectState, SceneCSSGridItemPlacement {
+  children: SceneFlexItemLike[];
```
torkelo, line 13: I think we need other item types here as the CSS grid items have very different props right? Maybe we can be more forgiving here and accept:
children: Array<SceneCSSGridItem | SceneObject>;  (because for many cases we might even need a item I think as item size constraints can be specific to such extend on the layout)

torkelo:
Nice! I think this could be really useful, does require a bit more work probably from the developer to add responsive layouts. 

Added a demo page (could need more work  to add options and show case the strength of this layout) 

[CSS-Grid-Layout---Demos---Scenes-Test-App---Apps---Grafana.webm](https://github.com/grafana/scenes/assets/10999/f1e6c604-9af0-4101-96e0-93e78b405d91)

packages/scenes/src/components/layout/CSSGrid/SceneCSSGridLayout.tsx
```diff
@@ -0,0 +1,120 @@
...
+}
+
+export interface SceneCSSGridItemPlacement {
+  rows: CSSProperties['gridTemplateRows'];
+  columns: CSSProperties['gridTemplateColumns'];
+  rowGap?: CSSProperties['rowGap'];
+  columnGap?: CSSProperties['columnGap'];
+  justifyItems?: CSSProperties['justifyItems'];
+  alignItems?: CSSProperties['alignItems'];
+  justifyContent?: CSSProperties['justifyContent'];
```
torkelo, line 49:
```suggestion
export interface SceneCSSGridItemPlacement {
  templateRows: CSSProperties['gridTemplateRows'];
  templateColumns: CSSProperties['gridTemplateColumns'];
  rowGap?: number; // This should probably be in Grafana theme grid units 
  columnGap?: number; This should probably be in Grafana theme grid units 
  justifyItems?: CSSProperties['justifyItems'];
  alignItems?: CSSProperties['alignItems'];
  justifyContent?: CSSProperties['justifyContent'];
```

I would move these state props to the layout object itself, as they only make sense on the layout object (not the items)

torkelo: and the reason for adding templateRows, templateColumn naming , was just to make it a bit clear the connection to the css prop names

jewbetcha:
Updated with the optional flex property on `EmbeddedScene` and moved types around.

> // This should probably be in Grafana theme grid units 

I wasn't able to find hard units for this... the type `ThemeSpacing` in `GrafanaTheme2` leads me to `string | number`. Is there somewhere else I should look?

torkelo:
@jewbetcha can you maybe add a scenario to the demo where you had the issue with flex on the parent element? I could not replicate that, could not make rows 100% make any sense no matter the parent style (as it made the first row 100% high )

jewbetcha:
> @jewbetcha can you maybe add a scenario to the demo where you had the issue with flex on the parent element? I could not replicate that, When I tried a template rows value of 100% I could not make the layout work no matter what the parent style was (as it made the first row 100% high )

Sure, you're thinking a `<Select>` similar to the column choices?

jewbetcha:
Added row selector to demo

jewbetcha:
@torkelo just checking that this can be good to go

torkelo:
@jewbetcha looks close, but want to check that flex issue. But been traveling for SLT offsite today.


Want to try it and understand the flex issue before merge , will try to tomorrow

jewbetcha:
Sounds good, thank you!

torkelo:
@jewbetcha I gave it a try and removed the flex option on EmbeddedScene (as without flex can't get the css grid row to work at all). having the container be flex is super important for the grid to take up available space.  so can't think of a single scenario where the body layout (flex or grid) should not take up all available space by default. Then it's just a matter of configuring the SceneFlexLayout or SceneCSSGridLayout to use the space or not (or wrap it in a flex layout and flex item with size constraints you need). 

I added flexGrow: 1 to SceneCSSGridLayout to grid styles to make get any width. 

I really think the SceneCSSGridLayout needs a prop for grid-auto-rows as the grid-template-rows option is so limited when you don't know how many rows you need.

jewbetcha:
> I really think the SceneCSSGridLayout needs a prop for grid-auto-rows as the grid-template-rows option is so limited when you don't know how many rows you need.

You're thinking `grid-auto-rows` rather than `grid-template-rows`? we could default to `grid-template-rows: auto`

torkelo:
@jewbetcha yea, I was thinking of having both since after experimenting with grid-template-rows it seems really limited to only situations where you know exactly how many rows there are. Maybe you can build the template-rows value based on how many child items you have. 

grid-template-rows seems really useful  in that you can then define constraints and the number of rows will be dynamic

jewbetcha:
So with `flex-grow` back in there, my grid is back to being broken. This is a case where we don't want the grid to take up all the available space. See screenshots with `flex-grow` enabled/disabled.

Screenshot 2023-10-18 at 7 51 16 AM
Screenshot 2023-10-18 at 7 51 30 AM

jewbetcha:
I also don't seem to be able to get `gridAutoRows` as a CSS property to be reflected in the demo app...

torkelo:
@jewbetcha yes, but that seems so ease to fix with settings on the grid (either an option for flexGrow, or auto rows), seems so much better to control this on the grid level instead of options on the parent, and if you do not want that just wrap the grid in a SceneFlexLayout and SceneFlexItem with ySizing: 'content'

torkelo:
@kaydelaney @dprokop have any last minute feedback on this? I think this is ready, only missing docs but maybe we can add that later (it's just a wrapper around css grid)

dprokop (approved):
I think this is good to merge, worth adding some docs, created an issue for it https://github.com/grafana/scenes/issues/421

PR #393, opened 2023-10-05
ActWhenVariableChanged: Add behavior to onChange callback

torkelo:
This helps making it possible to use the behavior without declaring a variable of the behavior and the behavior parent and capturing their instances in a lambda.

Having the behavior be part of the callback makes it possible to write the onChange callback as a pure function.


dprokop (approved):
I think this makes sense, but how/where did you use it?

PR #394, opened 2023-10-05
TextBoxVariable: Fixes and make it auto size

torkelo:
* [x] Fixes bug in TextBoxVariable where it always caused variable change event a few milliseconds after mount (debounce fired on mount for some reason)
* [x] Remove debounce and make it change variable onBlur and onKeyDown (enter key) (like the current main repo implementation). I think the short 250ms debounce is to short for most slow query scenarios. I think we can add a "debounce" option later to opt into this behavior for scenarios where the textbox is used for very fast queries (or in memory filtering). 
* [x] Make the input auto size
* [x] Add test 
* [x] Move some test classes to be shared


dprokop (approved):
This makes sense

PR #395, opened 2023-10-05
QueryVariable: Support for queries that contain "$__searchFilter" 

### LINKED ISSUE #208: QueryVariable: Add support for $__searchFilter
The `QueryVariable` currently does not support `$__searchFilter` (https://github.com/grafana/scenes/blob/main/packages/scenes/src/variables/variants/query/QueryVariable.tsx#L106)

torkelo:
Closes #208

Tested with Graphite (and TestData data source). 

This is such a messy feature that barely works (in the old template variable system implementation) for some legacy data sources. 

Discovered some strange things.
On the old code we only check the query if it's a string or if it has a "target" property (only graphite query editor has this)
https://github.com/grafana/grafana/blob/main/public/app/features/variables/pickers/OptionsPicker/actions.ts#L81 

I solved this in this implementation by stringifying the whole query model then looking for "__searchFilter" .

Second strange / buggy thing in the old implementation is: For Custom and Standard variable support we pass it as scopedVars but not with leading underscores (so it wont work as way to interpolate the query).
https://github.com/grafana/grafana/blob/main/public/app/features/variables/query/VariableQueryRunner.ts#L175 

I fixed that in this implementation by passing __searchFilter as scoped var. 

* [x] __searchFilter support for legacy runner
* [x] __searchFilter support for custom & standard runners
* [x] unit tests
* [x] trying to rid of the query: any type (would love to make it only be DataQueryExtended but so hard to know if we have changed string => object and need to change back when calling metricFindQuery).

### DIFF
--- a/packages/scenes/src/variables/variants/query/createQueryVariableRunner.ts
+++ b/packages/scenes/src/variables/variants/query/createQueryVariableRunner.ts
@@ -1,6 +1,13 @@
 import { from, mergeMap, Observable, of } from 'rxjs';
 
-import { DataQueryRequest, DataSourceApi, getDefaultTimeRange, LoadingState, PanelData } from '@grafana/data';
+import {
+  DataQueryRequest,
+  DataSourceApi,
+  getDefaultTimeRange,
+  LoadingState,
+  PanelData,
+  StandardVariableQuery,
+} from '@grafana/data';
 import { getRunRequest } from '@grafana/runtime';
 
 import { hasCustomVariableSupport, hasLegacyVariableSupport, hasStandardVariableSupport } from './guards';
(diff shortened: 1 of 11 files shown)


packages/scenes/src/variables/variants/query/createQueryVariableRunner.ts
```diff
@@ -14,7 +21,7 @@ export interface RunnerArgs {
 }
 
 export interface QueryRunner {
-  getTarget: (variable: QueryVariable) => DataQuery;
+  getTarget: (variable: QueryVariable) => DataQuery | string;
```
dprokop, line 24: Oh no :( Legacy data sources return strings for variable queries?

dprokop (approved):
Turned out quite simple implementation-wise, awesome work!

packages/scenes/src/variables/variants/query/createQueryVariableRunner.ts
torkelo: @dprokop yea, so hard to get rid this. I think there is a way, if we remembered that it was a string when we build the QueryVariable we can turn it back into a string when calling metricFindQuery

PR #396, opened 2023-10-06
VizPanel: Add support for a custom migration handler

torkelo:
This is one required to handle the auto migration of old angular panels. 

Because the panel type change migration handlers already work by mutating a "PanelModel" like object I am continuing to use that model here as it makes the logic very simple, we just build this mutatable panel model-like object once and pass it through the custom handler, and then through the normal migration handler. 

used in https://github.com/grafana/grafana/pull/76100

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanel.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.tsx
@@ -50,8 +50,11 @@ export interface VizPanelState<TOptions = {}, TFieldConfig = {}> extends SceneOb
   // internal state
   pluginLoadError?: string;
   pluginInstanceState?: any;
-  /** Only for use from core to handle migration from old angular panels */
-  customMigrationHandler?: (panel: PanelModel, plugin: PanelPlugin) => void;
+  /**
+   * @internal
+   * Only for use from core to handle migration from old angular panels
+   **/
+  UNSAFE_customMigrationHandler?: (panel: PanelModel, plugin: PanelPlugin) => void;
 }
 
 export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends SceneObjectBase<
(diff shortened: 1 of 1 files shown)


packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -50,6 +50,8 @@ export interface VizPanelState<TOptions = {}, TFieldConfig = {}> extends SceneOb
   // internal state
   pluginLoadError?: string;
   pluginInstanceState?: any;
+  /** Only for use from core to handle migration from old angular panels */
+  customMigrationHandler?: (panel: PanelModel, plugin: PanelPlugin) => void;
```
dprokop, line 54: An idea - what if we had some convention on what's the public API vs what's core specific one - rather than just a comment? 


Maybe sth like double underscore `__customMigrationHandler` or React-like `unsafe` prefix: `UNSAFE_customMigrationHandler`?

dprokop (approved):
LGTM - But let's discuss the naming :)

packages/scenes/src/components/VizPanel/VizPanel.tsx
dprokop: I mean - we could have a clear statement in the documentation that such methods/properties are not intended to use and will cause breaking changes if used in apps.

torkelo: @dprokop what do you think of using underscore and "@internal" property? We do this in a few places already 

```ts
 /**
   * @internal state of the new filter being added
   */
  _wip?: AdHocVariableFilter;
```ts

dprokop: It's OK, but hard to document. I would be as verbose as possible in the naming convention - the ie. `UNSAFE` prefix just craves not to use it, while `_` can be ambiguous and requires doc comments. 

With underscore properties we could have an eslint rule maybe that would produce errors if those props are used in the app context.  But this is more work I think than just naming methods/properties what they are :)

torkelo: @dprokop yea, agree. renamed it to `UNSAFE_customMigrationHandler`

torkelo: added a underscore, so it follows the same pattern at least as the "private" state `_UNSAFE_customMigrationHandler`

dprokop: great!

Issue #383, opened 2023-10-02
DataSourceVariable: Improve default option 


torkelo:
The default option is a bit old and the way it is returned now without showing the name of the data source is pretty bad. 

We should change this so that it returns the name and real uid instead:
https://github.com/grafana/scenes/blob/main/packages/scenes/src/variables/variants/DataSourceVariable.tsx#L70

PR #400, opened 2023-10-09
DataSourceVariable: Value should be uid, and other fixes

torkelo:
old (current) template variable system DataSourceVariable was changed in May in the main repo to put uid as the value (https://github.com/grafana/grafana/pull/69259) 

* [x] Fixes value issue (so it's set to uid and not name). Old implementation in core sets value to the uid so the url state also has the uid. So when opening the dashboard in scenes the uid was shown in the variable dropdown. 
* [x] Fixes issue with validateAndUpdate when the value matches a option but the current text value is not correct. This can happen after initial url sync sets both text & value (due to no options as url sync happens before validateAndUpdate). Added corrective logic to fix text state. 
* [x] Move pluginId fitlering to DataSourceSrv.getList
* [x] Adds option to only add default option when defaultOptionEnabled (Not 100% sure we needed, definitely not for scene apps where this option is just causing problems). It is causing problems in core dashbords as well. If I enable Multi value and All value and repeat a row I get a duplicate row for "default".  I think we can remove this option and keep url compatability with old dashboards if we make sure the default data source is first in the options array? 

Fixes https://github.com/grafana/scenes/issues/383

### DIFF
--- a/packages/scenes/src/variables/variants/DataSourceVariable.test.ts
+++ b/packages/scenes/src/variables/variants/DataSourceVariable.test.ts
@@ -6,16 +6,17 @@ import { SceneObject } from '../../core/types';
 
 import { DataSourceVariable } from './DataSourceVariable';
 import { VariableCustomFormatterFn } from '../types';
+import { DataSourceSrv, GetDataSourceListFilters, setDataSourceSrv } from '@grafana/runtime';
 
-function getDataSource(name: string, type: string, isDefault = false): DataSourceInstanceSettings {
+function getDataSource(overrides: Partial<DataSourceInstanceSettings>): DataSourceInstanceSettings {
   return {
     id: 1,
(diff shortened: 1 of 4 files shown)


dprokop:
> I think we can remove this option and keep URL compatability with old dashboards if we make sure the default data source is first in the options array?

So in this case the URL compatibility is about resolving the value that's `default`? I would love that option to be removed from the list, and the variable just to pick the correct data source from the list instead - it can even have some UI indicator in the list that it's the default one. All those special values in the variables (i.e. the `$__auto` that @axelavargas recently worked on) are PITA.

torkelo:
> I think we can remove this option and keep URL compatability with old dashboards if we make sure the default data source is first in the options array

> So in this case the URL compatibility is about resolving the value that's default

Yea, we can just have some backward compatibility with urls then I think we can remove it

Issue #213, opened 2023-05-25
Variables: Individual selections in multi value variable are causing cascading updates


dprokop:
When a variable is specified as `isMulti` each individual selection triggers a variables update chain which is too eager. 

We need to figure out a way of applying all selections together( i.e. in the `onBlur`handler of variable value selector).

PR #410, opened 2023-10-13
Variables: Multi select batch update

torkelo:
Fixes https://github.com/grafana/scenes/issues/213 

A possible way to fix selecting or removing values without triggering cascading updates or dashboard data queries for every change


dprokop (approved):
Looks good 👍 And works as advertised :D

PR #415, opened 2023-10-17
Remove checkbox from time window comparison

javiruiz01:
Fixes [#447](https://github.com/grafana/app-observability-plugin/issues/447).

Update Time Range comparison component:
- Removed checkbox
- Add new `No comparison` option
- Display chosen comparison option on select

### DIFF
--- a/packages/scenes/src/components/SceneTimeRangeCompare.tsx
+++ b/packages/scenes/src/components/SceneTimeRangeCompare.tsx
@@ -1,11 +1,12 @@
 import React from 'react';
-import { DateTime, dateTime, rangeUtil, TimeRange } from '@grafana/data';
-import { ButtonGroup, ButtonSelect, Checkbox, Icon, ToolbarButton } from '@grafana/ui';
+import { DateTime, dateTime, GrafanaTheme2, rangeUtil, TimeRange } from '@grafana/data';
+import { ButtonGroup, Checkbox, Dropdown, Menu, ToolbarButton, useStyles2 } from '@grafana/ui';
 import { SceneObjectBase } from '../core/SceneObjectBase';
 import { SceneComponentProps, SceneObjectState, SceneObjectUrlValues } from '../core/types';
 import { sceneGraph } from '../core/sceneGraph';
 import { SceneObjectUrlSyncConfig } from '../services/SceneObjectUrlSyncConfig';
 import { parseUrlParam } from '../utils/parseUrlParam';
+import { css } from '@emotion/css';
 
 export interface TimeRangeCompareProvider {
   getCompareTimeRange(timeRange: TimeRange): TimeRange | undefined;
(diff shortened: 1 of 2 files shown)


dprokop:
Implementation wise I think this looks good. 

I personally miss the checkbox a bit to have a clear indication the comparison is enabled. So how about leaving the checkbox as it is, but do not require checking it in order to enable the dropdown? Something like this: https://github.com/grafana/scenes/commit/bd22a9b4b5f926c8db1808e233fba0c65f8c05eb


https://github.com/grafana/scenes/assets/2376619/f3760661-b95d-4760-992d-b00913a51bb7

torkelo:
@dprokop it looks like the checkbox does not do anything? it looks like opens the dropdown which looks very strange. 

Can we change the wording to just be "Comparison", "Time frame comparison" is way to long :) 

I kind of like how elastic does it, they have a checkbox, when clicked it will enable the comparison and use last window you used (or the default). So if's only two clicks if you want to change the window. 
image

dprokop:
> @dprokop it looks like the checkbox does not do anything? it looks like opens the dropdown which looks very strange.

I think it does quite a lot - indicates that the comparison is on. However, functionally I think we can still improve it's behavior so that when the checkbox is active and clicked it would disable the comparison rather than open the popover to choose the No comparison option.

javiruiz01:
Yes! That was my first approach also, let me make the change and we'll see how it looks 👌

torkelo:
I updated the list of comaprison windows to just be "Day before" "Week before" ,"Month before", reads so much better. Not sure how common "3 days before" is going to be. If we need more maybe we add a "custom" option that changes the dropdown to an input. 
Screenshot from 2023-10-26 09-07-54

One last change that I think would make this better is to just have the checkbox disable the period select (and the time range comparison). So that when you click the checkbox to enable the comparison you see what period will be enabled. It also helps explain the checkbox a bit better with the window value being seen (but disabled).

Screenshot from 2023-10-26 09-05-44

javiruiz01:
I like it! It's a lot more clear now, and you know what to expect when you turn it on

packages/scenes/src/components/SceneTimeRangeCompare.tsx
```diff
@@ -166,25 +180,38 @@ function SceneTimeRangeCompareRenderer({ model }: SceneComponentProps<SceneTimeR
...
+        </ToolbarButton>
       )}
     </ButtonGroup>
   );
 }
+
+function getStyles(theme: GrafanaTheme2) {
+  return {
+    previewButton: css({
+      cursor: 'not-allowed',
```
torkelo, line 208: is this not set for a disabled button?

packages/scenes/src/components/SceneTimeRangeCompare.tsx
```diff
@@ -166,25 +180,38 @@ function SceneTimeRangeCompareRenderer({ model }: SceneComponentProps<SceneTimeR
...
         />
       ) : (
-        <ToolbarButton
-          icon={<Icon name="angle-down" size="md" />}
-          style={{ cursor: !enabled ? 'not-allowed' : 'pointer' }}
-          variant="canvas"
-        />
+        <ToolbarButton className={styles.previewButton} disabled variant="canvas">
+          {previousValue.label}
+          <Icon name="angle-down" size="md" />
```
torkelo, line 198: can't you say isOpen={false} ? instead

torkelo:
Pushed an update that fixed the styling, a disabled ToolbarButton did not look right in a ButtonGroup

torkelo (approved):

javiruiz01:
Great, thanks for the fix! Just to confirm, we want the default value to be the previous value right?

PR #419, opened 2023-10-18
UrlSync: Fixes and one update

torkelo:
Noticed some issues with UrlSync clearing state when moving to another route (before scene was unmounted) as
location update fires before the current scene has been unmounted / and new route mounted,

### DIFF
--- a/packages/scenes/src/services/UrlSyncManager.ts
+++ b/packages/scenes/src/services/UrlSyncManager.ts
@@ -11,6 +11,7 @@ import { Unsubscribable } from 'rxjs';
 export interface UrlSyncManagerLike {
   initSync(root: SceneObject): void;
   cleanUp(root: SceneObject): void;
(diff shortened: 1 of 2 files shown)


packages/scenes/src/services/UrlSyncManager.ts
```diff
@@ -139,6 +147,31 @@ export class UrlSyncManager implements UrlSyncManagerLike {
 
     sceneObject.forEachChild((child) => this._syncSceneStateFromUrl(child, urlParams));
   }
+
+  public getUrlState(root: SceneObject): SceneObjectUrlValues {
+    const urlKeyMapper = new UniqueUrlKeyMapper();
+    urlKeyMapper.rebuildIndex(root);
+
```
dprokop, line 154: Why not sing the instance's url mapper?

dprokop: ah, makes sense!

PR #420, opened 2023-10-19
SceneTimeRange: Respect time zone when updating time range

dprokop:
Alt to https://github.com/grafana/scenes/pull/413

In SceneTimeRange we are storing the range in two places `state.from` / `state.to` and `state.value` which holds the actual evaluated time range. The former holds state that's used in the url or comes from configuration, while the `state.value` holds the parsed time range, used downstream in quiery runners, pickers and more.. When updating the abs time range from the UI, the value was calculated against the UTC value of the selected range. 

This PR makes sure the evaluation happens against the value that the user actually selected in the UI, not the UTC-converted one.

<details>
  <summary>📦 Published PR as canary version: <code>1.19.1--canary.420.6573494215.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@1.19.1--canary.420.6573494215.0
  # or 
  yarn add @grafana/scenes@1.19.1--canary.420.6573494215.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/core/SceneTimeRange.tsx
+++ b/packages/scenes/src/core/SceneTimeRange.tsx
@@ -97,20 +97,32 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
 
   public onTimeRangeChange = (timeRange: TimeRange) => {
     const update: Partial<SceneTimeRangeState> = {};
+    const updateToEval: Partial<SceneTimeRangeState> = {};
 
     if (typeof timeRange.raw.from === 'string') {
       update.from = timeRange.raw.from;
+      updateToEval.from = timeRange.raw.from;
     } else {
       update.from = timeRange.raw.from.toISOString();
+      // @ts-expect-error Need an update in core API
+      updateToEval.from = timeRange.raw.from.toISOString(true);
     }
 
     if (typeof timeRange.raw.to === 'string') {
       update.to = timeRange.raw.to;
+      updateToEval.to = timeRange.raw.to;
     } else {
       update.to = timeRange.raw.to.toISOString();
+      // @ts-expect-error Need an update in core API
+      updateToEval.to = timeRange.raw.to.toISOString(true);
     }
 
-    update.value = evaluateTimeRange(update.from, update.to, this.getTimeZone(), this.state.fiscalYearStartMonth);
+    update.value = evaluateTimeRange(
+      updateToEval.from,
+      updateToEval.to,
+      this.getTimeZone(),
+      this.state.fiscalYearStartMonth
+    );
 
     // Only update if time range actually changed
     if (update.from !== this.state.from || update.to !== this.state.to) {


packages/scenes/src/core/SceneTimeRange.tsx
```diff
@@ -97,20 +97,32 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
...
     const update: Partial<SceneTimeRangeState> = {};
+    const updateToEval: Partial<SceneTimeRangeState> = {};
 
     if (typeof timeRange.raw.from === 'string') {
       update.from = timeRange.raw.from;
+      updateToEval.from = timeRange.raw.from;
     } else {
       update.from = timeRange.raw.from.toISOString();
+      // @ts-expect-error Need an update in core API
+      updateToEval.from = timeRange.raw.from.toISOString(true);
```
torkelo, line 108: what is this true option here? and what core API change is referenced?

dprokop: The `true` indicates that the time zone offset is preserved in the generated string. It's [moment's `keepOffset` API](https://www.geeksforgeeks.org/moment-js-moment-toisostring-method/) that we [do not expose in core `toISOString` signature](https://github.com/grafana/grafana/blob/scenes-for-viewers/packages/grafana-data/src/datetime/moment_wrapper.ts#L70)

torkelo: @dprokop is there a main repo change that adds this param?

torkelo (approved):

packages/scenes/src/core/SceneTimeRange.tsx
dprokop: It's not necessary as it works like this in core already. We do not store UTC values in the time srv / dashboard state.

dprokop: ah wait, you're asking if the API change is there? Not yet, will be in a a sec

dprokop: https://github.com/grafana/grafana/compare/to-iso-api?expand=1

torkelo: @dprokop not sure I understand this change, it changes the format of from/to when onTimeRangeChange is called vs when we initialize SceneTimeRange (in the constructor we also call evaluate) it update it from URL. 

It's only here we call toISOString(true), which kind of break the assumption that from/to should always be UTC iso strings. 

if I revert this timezone is still respected when I change timerange via graphs, do you remember what the scenario was?

cbarreholm: I suspect that this was raised as a result of something I reported (since I follow this ticket). Hopefully the following could shed some light on it.


I wrote in our internal company chat in the context of Frontend observability: 
> I have noticed some odd behaviour in the date picker as well, where I write e.g. 11 am and it becomes 9 am. 
I also wrote:
> A reflection on the time picker. I don't know if it is by design. But it differs from how it works in the on-prem Grafana 9.x.
> 1. I pick a date. The range is then displayed as 2023-10-25 00:00:00 - 2023-10-25 23:59:59.
> 2. Click Apply Time Range
> 3. Open the time picker. The range is expressed in UTC.
> I didn't see the trailing Z to begin with, so I got really confused when pasting  2023-10-25 09:00:00 and applying and  it becomes 2023-10-25T07:00:00.000Z.

PR #422, opened 2023-10-19
SceneQueryRunner: Fixes adhoc filters when using a variable data source

torkelo:
I am removing this condition as I don't think it's needed, at least not if you use the data source variable consistently (in both the set and queries)

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.ts
@@ -13,7 +13,7 @@ import {
   rangeUtil,
   transformDataFrame,
 } from '@grafana/data';
-import { getDataSourceSrv, getRunRequest, toDataQueryError } from '@grafana/runtime';
+import { getRunRequest, toDataQueryError } from '@grafana/runtime';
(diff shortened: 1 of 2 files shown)


packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -520,12 +520,6 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
       return;
     }
 
-    const setDataSource = getDataSourceSrv().getInstanceSettings(set?.state.datasource, this._scopedVars);
```
torkelo, line 523: getInstanceSettings returned the variable expr, for example `$ds` , thought it was always the resolved real uid

eskirk (approved):
if I understand correctly - we no longer pass along a DS UID rather we just find the closest DS in the scene graph? sounds good to me - thanks for fixing

torkelo:
@eskirk not really, we still check uid of adhoc filter set match query ds uid. If one use a variable ds both need to in order to match

PR #427, opened 2023-10-23
Vertical layout for variable selectors and a simple mode for adhoc filters 

torkelo:
* The horizontal / vertical mode for variable selector is pretty self evident 
  * One benefit of the vertical mode is that the variable loading spinner now does not make the select move/shift 
* Adhoc filters need more options to have a simple mode without an operator selector, maybe individual filters can have an option to specify that they need the operator selector or it's a mode on the whole set. 

image
image


packages/scenes/src/utils/ControlsLabel.tsx
```diff
@@ -40,34 +43,48 @@ export function ControlsLabel(props: ControlsLabelProps) {
...
+        {props.icon && <Icon name={props.icon} className={styles.normalIcon} />}
+        {props.label}
+        {loadingIndicator}
+      </label>
+    );
+  }
 
   if (props.description) {
     return (
       <Tooltip content={props.description} placement={'bottom'}>
```
dprokop, line 77: Think we need to change the alignment of the tooltip for vertical layout, otherwise the descriotion renders like this:

image

dprokop (approved):
Looks good, but I see the controls are not vertically aligned, think we should align all of them to the bottom of the controls container:

Screenshot 2023-11-06 at 12 11 32

packages/scenes/src/utils/ControlsLabel.tsx
torkelo: fixed

PR #428, opened 2023-10-24
SceneObjectBase: Add utility function getAncestor

torkelo:
After developing the data trails https://github.com/grafana/grafana/pull/77019https://github.com/grafana/grafana/pull/77019
I found myself needing to get a parent object of a specific type a lot.

I think a utility function like this could be useful.

### DIFF
--- a/packages/scenes/src/core/SceneObjectBase.tsx
+++ b/packages/scenes/src/core/SceneObjectBase.tsx
@@ -325,6 +325,23 @@ export abstract class SceneObjectBase<TState extends SceneObjectState = SceneObj
     }
   }
 
+  /**
+   * A utility function to find the closest ancestor of a given type. This function expects
+   * to find it and will throw an error if it does noit.
+   */
+  public getAncestor<ParentType>(type: { new (...args: never[]): ParentType }): ParentType {
+    let parent: SceneObject | undefined = this;
+
+    while (parent) {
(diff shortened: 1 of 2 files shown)


packages/scenes/src/core/SceneObjectBase.tsx
```diff
@@ -325,6 +325,23 @@ export abstract class SceneObjectBase<TState extends SceneObjectState = SceneObj
     }
   }
 
+  /**
+   * A utility function to find the closest ancestor of a given type. This function expects
+   * to find it and will throw an error if it does noit.
+   */
+  public getAncestor<ParentType>(type: { new (...args: never[]): ParentType }): ParentType {
```
dprokop, line 332: I think this should be part of the `sceneGraph`, not the base class. Feels weird that some traversing is available here, and some in other place.

torkelo: @dprokop yea your right, was just nice to skip one argument for this to make it a bit clearer. `sceneGraph.getAncestor(model, DashboardScene)`  did not feel as clear to me somehow. But your right it's a bit inconsistent to have some graph lookups in sceneGraph and some in base class

jewbetcha:
Agree with @dprokop but this will be really useful!

dprokop (approved):

PR #430, opened 2023-10-25
VizPanel: Make VizPanel usable without relative position parent 

torkelo:
Currently VizPanel requires a relative positioned parent (something SceneFlexItem and SceneCSSGridItem both add). 

SceneFlexItem is really needed anyway for flex layouts as it has size constraints. But for CSS Grid layout the size constraints are on the grid itself and you rarely need the item level props (grid-column-start, etc). 

So I think it could make sense to add this relative wrapper to VizPanel so that i can be used directly as a child anywhere (but esp in css grid layouts)

### DIFF
--- a/packages/scenes-app/src/demos/cssGridLayoutDemo.tsx
+++ b/packages/scenes-app/src/demos/cssGridLayoutDemo.tsx
@@ -72,15 +72,11 @@ export function getCssGridLayoutDemo(defaults: SceneAppPageState) {
 }
 
 function getLayoutChildren(count: number) {
-  return Array.from(
-    Array(count),
-    (v, index) =>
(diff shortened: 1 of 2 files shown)


torkelo:
oh for those that wonder why the absolute wrapper is needed around PanelChrome, I wish it was not needed. but without it, the panel will resize to be smaller when the window or other panels make the space available for it change. 

The use of useMeasure to get the width and height then sets these dimensions PanelChrome / visualization, making the panel really take up all that space, so when the space available changes it does not make useMeasure change as PanelChrome is then making use of the space previously assigned, locking it in.

dprokop (approved):

jewbetcha (approved):

PR #433, opened 2023-10-25
Variables: Fixes SQL formatting and escaping double quotes 

piggito:
Sqlstring format should escape double quotes since the formatted value is nested in a JSON object sent to query API, see https://github.com/grafana/grafana/blob/main/packages/grafana-runtime/src/utils/DataSourceWithBackend.ts#L193

Without escaping double quotes the JSON gets malformed and we get errors like the following:

Screenshot from 2023-10-25 11-39-24


torkelo (approved):
Thanks for contributing! 

pushed a minor refactor and fixed the single value case

PR #434, opened 2023-10-26
Urls: Make sure urls include sub path

torkelo:
I really hate the complexity of the sub path logic where everywhere we have to either add the sub path or remove it :cry: 

Noticed tabs and breadcrumbs built using the scenes app never include the instance sub path. As well as urls coming from `${__url}` 

Wish there was a more centralized way to handle this. The irony is that the sub path is removed in TextLink/Link component (because react-router does not want it), or if we need to use locationServce.push it also needs to be removed


axelavargas (approved):
Looks good to me :)

PR #437, opened 2023-10-27
VizPanel: Allow title items configuration

dprokop:
Adds API to VizPanel to allow panel links rendering. There is no component for links rendering in the library, as the dashboard's link renderer heavily depends on core-only code.  https://github.com/grafana/grafana/pull/77295

We may consider adding some generic link renderer, but I would wait with that until we get request for panel links support in the lib.

<details>
  <summary>📦 Published PR as canary version: <code>1.20.1--canary.437.6693096080.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@1.20.1--canary.437.6693096080.0
  # or 
  yarn add @grafana/scenes@1.20.1--canary.437.6693096080.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanel.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.tsx
@@ -14,6 +14,7 @@ import {
   compareDataFrameStructures,
   applyFieldOverrides,
   PluginType,
+  LinkModel,
 } from '@grafana/data';
 import { PanelContext, SeriesVisibilityChangeMode, VizLegendOptions } from '@grafana/ui';
 import { config, getAppEvents, getPluginImportUtils } from '@grafana/runtime';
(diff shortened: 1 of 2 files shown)


packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -53,6 +59,10 @@ export interface VizPanelState<TOptions = {}, TFieldConfig = {}> extends SceneOb
    * So the best way to add dynamic menu actions and links is by adding them in a behavior attached to the menu.
    */
   menu?: VizPanelMenu;
+  /**
+   * Defines a menu that renders panel link.
+   **/
+  panelLinks?: SceneObject<VizPanelLinksState>;
```
torkelo, line 65: since it's up to the consumer of how to render these links, it feels more like this should be "titleItems" and have no enforced state associated with the scene object.

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -40,6 +45,7 @@ export interface VizPanelState<TOptions = {}, TFieldConfig = {}> extends SceneOb
   pluginId: string;
   title: string;
   description?: string;
+  links?: LinkModel[];
```
torkelo, line 48: this seems to be unused?

packages/scenes/src/components/VizPanel/VizPanel.tsx
dprokop: You think we should expose title items? I was considering this but wasn't convinced. But maybe that's a good idea to allow customization of this section basically.

torkelo (approved):

PR #438, opened 2023-10-30
SceneTimeRangeCompare: UI improvements proposal

dprokop:
I think https://github.com/grafana/scenes/pull/415 was prematurely merged, resulting in the scene time range compare picker looking broken (to me maybe) due to the disabled styles usage on the dropdown button: 

image

I'm getting back to my proposal of not showing the selected comparison range when the comparison is disabled. I think it makes more sense than the partially disabled button which is a pattern that I can't recall we use anywhere in the product (cc @grafana/design-system .

https://github.com/grafana/scenes/assets/2376619/e2df9149-32f3-40b9-b649-41306f48e682

<details>
  <summary>📦 Published PR as canary version: <code>1.20.2--canary.438.6696445069.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@1.20.2--canary.438.6696445069.0
  # or 
  yarn add @grafana/scenes@1.20.2--canary.438.6696445069.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/SceneTimeRangeCompare.tsx
+++ b/packages/scenes/src/components/SceneTimeRangeCompare.tsx
@@ -155,7 +155,6 @@ function SceneTimeRangeCompareRenderer({ model }: SceneComponentProps<SceneTimeR
 
   const value = compareOptions.find(({ value }) => value === compareWith);
   const onClick = () => {
-    console.log('onClick', previousCompareWith, compareWith);
     if (enabled) {
       model.onClearCompare();
     } else if (!enabled) {
--- a/packages/scenes/src/components/SceneTimeRangeCompare.tsx
+++ b/packages/scenes/src/components/SceneTimeRangeCompare.tsx
@@ -1,12 +1,11 @@
-import { DateTime, dateTime, GrafanaTheme2, rangeUtil, TimeRange } from '@grafana/data';
-import { ButtonGroup, ButtonSelect, Checkbox, ToolbarButton, useStyles2 } from '@grafana/ui';
+import { DateTime, dateTime, rangeUtil, TimeRange } from '@grafana/data';
+import { ButtonGroup, ButtonSelect, Checkbox, ToolbarButton } from '@grafana/ui';
 import React from 'react';
 import { sceneGraph } from '../core/sceneGraph';
 import { SceneObjectBase } from '../core/SceneObjectBase';
 import { SceneComponentProps, SceneObjectState, SceneObjectUrlValues } from '../core/types';
 import { SceneObjectUrlSyncConfig } from '../services/SceneObjectUrlSyncConfig';
 import { parseUrlParam } from '../utils/parseUrlParam';
-import { css } from '@emotion/css';
 
 export interface TimeRangeCompareProvider {
   getCompareTimeRange(timeRange: TimeRange): TimeRange | undefined;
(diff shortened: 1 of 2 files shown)


torkelo:
what is wrong with the disabled button? I think it's nice to see what will happen when you enable it, my preference was that it would default to "Day before", so it's much more clear what is disabled and what will happen when you check the checkbox. 

this was the same ux the comparison checkbox & select had in kibana.

torkelo:
But I am fine with a the selected value being hidden as well, it saves some space

torkelo:
One idea I had is that this component should not be using ToolbarButton but should be a ControlsLabel + Select, and look like our variable selects (but with checkbox inside label), thoughts? 

But that is not super nice as it will make it look a bit different to TimeRangePicker and RefreshPicker, so it's probably best to maintain the ToolbarButton look

torkelo (approved):

packages/scenes/src/components/SceneTimeRangeCompare.tsx
```diff
@@ -1,19 +1,19 @@
...
 import { parseUrlParam } from '../utils/parseUrlParam';
-import { css } from '@emotion/css';
 
 export interface TimeRangeCompareProvider {
   getCompareTimeRange(timeRange: TimeRange): TimeRange | undefined;
 }
 
 interface SceneTimeRangeCompareState extends SceneObjectState {
   compareWith?: string;
+  previousCompareWith?: string;
```
torkelo, line 16: this feels a bit awkward, I think storing isEnabled in state would be sooo much cleaner

dprokop:
> what is wrong with the disabled button?

I >think< it looks like it's a UI bug. Think if the disabled part was a disabled select, rather than button, then it would have look better, somehow like in Kibana. But again, this is pure aestethincs thing, I don't have to be right here, this PR is just a proposal of an improvement.



> Oh no, I'm sorry about that! I'm fine with the change too, as long as clicking on the checkbox activates it I like it 👍

@javiruiz01 don't be sorry, I think the behavior that's merged is fine, this is just trying to tweak it a bit.

packages/scenes/src/components/SceneTimeRangeCompare.tsx
dprokop: If we want to restore the previously selected one  then we need it. Or, we can decide to elways select "Previous perid" when enabled - which is fine too and will simplify this a bit.

dprokop:
>One idea I had is that this component should not be using ToolbarButton but should be a ControlsLabel + Select, and look like our variable selects (but with checkbox inside label), thoughts?

@torkelo I think ToolbatButton + Select would work best, but it's gonna be problematic too, as the disabled elements in grafana/ui have borders that are more subtle than in enabled state. This creates this weird impression IMO that the control is built with two, not matching UI elements. That's why I'm trying to push to have the button always enabled not to create this weird look.

torkelo:
@dprokop yea, that was why I added the custom disabled button styling to have the border match the toolbar button. 

but we can hide the select, it saves space.

packages/scenes/src/components/SceneTimeRangeCompare.tsx
torkelo: @dprokop if the value does not change when you disable it (only the isEnabled state would change), then I do not see why we need to store the previous value. 

I mean the checkbox would only enable/disable it, not change the value. It's a bit awkward that the checkbox state is in the react renderer and not in the component state. So when disabling it we are forced to clear the value, since it currently controls the checkbox state as well

domasx2 (approved):

dprokop:
Clising this for now, will not have time to focus on this in the upcoming weeks.

PR #439, opened 2023-10-30
feat: add store field to SceneAppState interface

erj826:
Adds a simple key value store to the SceneAppState that allows any nested SceneObject in the graph to access the store data. Useful for adding global data to the scene graph.


erj826:
Closing as we don't want to store global state on the SceneApp object

PR #443, opened 2023-11-04
Variables: Support for variables on lower levels to depend on variables on higher levels 

### LINKED ISSUE #301: Dropdown in SceneAppPage()
Hi! 

Similar to https://github.com/grafana/scenes/issues/292 I have some problems when I use a dropdown in a SceneAppPage.

```typescript
new SceneApp({
        pages: [
            new SceneAppPage({
                title: "Host",
                // Important: Mind the page route is ambiguous for the tabs to work properly
                url: prefixRoute(`${ROUTES.Overview}`),
                hideFromBreadcrumbs: true,
                getScene: getOverviewScene,
                tabs: [
                    new SceneAppPage({
                        title: "Overview",
                        url: prefixRoute(`${ROUTES.Overview}`),
                        getScene: getOverviewScene
                    }),
                    new SceneAppPage({
                        title: "Network",
                        url: prefixRoute(`${ROUTES.Overview}/network`),
                        getScene: getNetworkScene
                    }),
                ],
                $variables: new SceneVariableSet({
                    variables: [
                        host_name,
                    ]
                }),
                controls: [
                    new VariableValueSelectors({}),
                ],
            }),
        ],
    });
``` 

With this code the variable `host_name` is visible on the right hand side of the page. If i add this variable only to the `SceneAppPage()` the dropdown works and returns values **but does not update the other contents (table, textbox, ...) on the page**. The update does only happen when I change tabs or reload the page)
If I add this variable to the `SceneAppPage()` and all Tabs on this page the dropdown does not work at all (the query is been cancelled in the network tab of the browser console). 
I create a new variable using this function 

```typescript
function getHostNameVariable(hide = 0) {
    return new QueryVariable({
        name: 'host_name',
        label: 'Host',
        datasource: DATASOURCE,
        query: "SELECT name from host ",
        sort: 1,
        isMulti: false,
        includeAll: false,
        placeholder: "No Host selected",
        hide: hide
    });
}
```

as already suggestedby @torkelo  (and working for https://github.com/grafana/scenes/issues/292). 
No related warnings in the console. 

Any suggests on this issue? 

Thx!

### LINKED ISSUE #416: QueryRunner executes query multiple times on initial load of a scene
Version: 1.18.0

When navigating to a SceneAppPage for the first time when the scene variables are not already stored in the url we see the query runner make 2 identical requests rapidly causing the first request to be canceled when the new variable is appended to the URL.

Bug was noticed when using a [TextBoxVariable and a QueryVariable](https://github.com/grafana/security-vulnerability-obs/blob/main/plugin/src/scenes/pages/issues/IssuesPage.tsx#L30-L68) that are both being used in the same [QueryRunner](https://github.com/grafana/security-vulnerability-obs/blob/main/plugin/src/scenes/pages/issues/queryRunners/getIssuesQueryRunner.tsx#L91-L94).

Screenshot 2023-10-13 at 9 08 25 AM

@dprokop to provide repro repo

torkelo:
Closes #301 
Fixes #416

* [x] Discovered a bug sceneGraph.hasVariableDependencyInLoadingState (it returned after the first variable). Fixes #416
* [x] Make sure variables wait for dependencies on higher levels and local level
* [x] SceneVariableSet implements the `variableDependency` "mixin"  (property) to get notified when a variable batch update completes on a higher level. It can then use this notification to kick off an update batch if a local variable was waiting for a variable on a higher level. 
* [x] The same notification mechanism is used to check if the changed variables include a variable that a local variable depends on. The local variable is then added to the _variablesToUpdate. 

[Nested-scenes-and-variables---Demos---Scenes-Test-App---Apps---Grafana.webm](https://github.com/grafana/scenes/assets/10999/0ad4ab77-9198-41d5-92b4-5fa245af5cbe)


packages/scenes/src/core/sceneGraph/sceneGraph.ts
```diff
@@ -70,7 +70,9 @@ export function hasVariableDependencyInLoadingState(sceneObject: SceneObject) {
     }
 
     const set = variable.parent as SceneVariables;
-    return set.isVariableLoadingOrWaitingToUpdate(variable);
+    if (set.isVariableLoadingOrWaitingToUpdate(variable)) {
+      return true;
+    }
   }
```
torkelo, line 76: an important bug fix, fixes https://github.com/grafana/scenes/issues/416

packages/scenes/src/variables/sets/SceneVariableSet.ts
```diff
@@ -183,7 +196,7 @@ export class SceneVariableSet extends SceneObjectBase<SceneVariableSetState> imp
       }
 
       // Wait for variables that has dependencies that also needs updates
-      if (this._hasDependendencyInUpdateQueue(variable)) {
+      if (sceneGraph.hasVariableDependencyInLoadingState(variable)) {
```
torkelo, line 199: I was very pleased that I could simply reusing `sceneGraph.hasVariableDependencyInLoadingState` here, an not only remove _hasDependendencyInUpdateQueue but also get the benefit of the support of checking variables loading state on higher levels that sceneGraph.hasVariableDependencyInLoadingState has

packages/scenes/src/variables/sets/SceneVariableSet.ts
```diff
@@ -331,3 +343,21 @@ export interface VariableUpdateInProgress {
...
+
+  public getNames(): Set<string> {
+    return this._emptySet;
+  }
+
+  public hasDependencyOn(name: string): boolean {
+    return false;
+  }
+
+  public variableUpdatesCompleted(changedVariables: Set<SceneVariable>): void {
```
dprokop, line 360: So - the "outer" set, when processed all of his variables, will call this, and trigger "inner" set updates. 

But - does this mean, that we need to wait for all "outer" scope variables to complete before updating the "inner" scope?

dprokop (commented):
It works as described - i've just left one question about a thing that I'm not certain about.

dprokop (approved):

packages/scenes/src/variables/sets/SceneVariableSet.ts
torkelo: @dprokop we do this with line change on 199, 

```ts
if (sceneGraph.hasVariableDependencyInLoadingState(variable)) {
```

this will make variables on the local level wait for dependant on variables on a higher level that are loading or waiting to load. 

But this waiting will only happen for variables that do have this dependency.

PR #444, opened 2023-11-06
VizPanel: Fixes series visibility toggling

### LINKED ISSUE #424: VizPanel: TimeSeries legend series toggling does not work
I can click a series item to focus (only show) that series but when I click the same legend item again the other series are not shown again (it does not reset).

Works in old dashboard but not in scenes so assume it's something with the toggle legend series in VizPanel

torkelo:
Fixes #424


axelavargas (approved):
Works as expected 🙌🏾

PR #445, opened 2023-11-07
PanelContext: Eventbus should not filter out local events

torkelo:
The pie chart legend hover events are not working. They highlight slices on the same panel and using the standard DataHoverEvent. The panel relies on being able to subscribe to local events. But the scenes PanelContextEventBus filters out any local events. 

Removing this filtering, panels filter out local events they don't care about using:

```ts
 if (eventBus === evt.origin) {
   return;
  }
```

Trying to make the needed changes to get that to work. But when testing the cursor sync demo I am unable to get it to work (with or without these changes). All cursors show on all panels. Think uPlot cursor sync scope key has broken. 

We could also try to revive https://github.com/grafana/grafana/pull/71866

### DIFF
--- a/packages/scenes-app/src/demos/cursorSync.tsx
+++ b/packages/scenes-app/src/demos/cursorSync.tsx
@@ -14,7 +14,7 @@ import { getEmbeddedSceneDefaults, getQueryRunnerWithRandomWalkQuery } from './u
 
 export function global() {
   return new SceneAppPage({
-    title: 'Cursor sync test',
+    title: 'Global sync scope',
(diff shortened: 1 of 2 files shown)


torkelo:
ah, the first demo scene has global scope, got confused when panel titles said `Panel 1 - syncs tooltip with Panel 2'`

dprokop (approved):
Works good against core dashboards as well. Think we can land this, and work on https://github.com/grafana/grafana/pull/71866 some time in the next few week.s

PR #446, opened 2023-11-07
VizPanel: Support passing legacyPanelId to PanelProps

### LINKED ISSUE #404: VizPanel: Some way to pass correct id to react panels
Currently PanelProps has an id property (which we set to panel id for dashboard). 

Not sure what the solution is here, for repeated panels we don't really have a numeric id in scene-based dashboards. 

We could add a new property to PanelProps (like `key`) or add it to PanelContext 

Location:
https://github.com/grafana/scenes/blob/main/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx#L117

torkelo:
Closes #404

Provides a way to pass panel id to PanelProps via PanelContext instanceState


dprokop (approved):
Think a few people realize there's `instanceState` on PanelContext ;)


But - this is looking good!

torkelo:
The general idea is we deprecate the id property on PanelProps, not sure what to replace with. Maybe a string key.

Anyway thinking is this is a temporary thing to reduce breaking  behavior.

PR #449, opened 2023-11-08
Variables: Support __org and __user variable macros

### LINKED ISSUE #448: Variables: Support for system variables (url macros) like __user and __org and __dashboard
Currently, we are missing url macros for:

* __user
* __org
* __dashboard

torkelo:
Closes #448 

* [x] Adds a new __org macro
* [x] Adds a new __user macro 

Question how to implement the `${__dashboard.name}` and `${__dashboard.uid}` macros

* Implement here in scenes lib by going to scene root and look for state properties names `uid` and `name`
* Or provide a way for consumers of the lib to register custom macros (could be very powerful)

### DIFF
--- a/packages/scenes/src/variables/macros/contextMacros.test.ts
+++ b/packages/scenes/src/variables/macros/contextMacros.test.ts
@@ -1,0 +1,41 @@
+import { config } from '@grafana/runtime';
+import { TestScene } from '../TestScene';
+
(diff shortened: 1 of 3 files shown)


dprokop (approved):
Looks good!

> Or provide a way for consumers of the lib to register custom macros (could be very powerful)

Think this is the way to go, we should decouple the lib from the dashboard concept as much as possible. And agree this may become a very powerful addition to the lib.

PR #452, opened 2023-11-08
Variables: Register variable macro

torkelo:
Exports a `sceneUtils.registerVariableMacro` that makes it possible for scene app plugins to register and unregister custom macros


dprokop (approved):
Nice! Probably deserves a note in advanced variable usage docs :)

axelavargas (approved):

PR #454, opened 2023-11-09
SceneQueryRunner: Handle alert states data layer

dprokop:
Required work for https://github.com/grafana/grafana/pull/77945

Adds handling for `alertState` property of the panel data that is going to be provided by appropriate data layer.

<details>
  <summary>📦 Published PR as canary version: <code>1.22.0--canary.454.6852070351.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@1.22.0--canary.454.6852070351.0
  # or 
  yarn add @grafana/scenes@1.22.0--canary.454.6852070351.0
  ```
</details>


packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -150,24 +153,64 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
...
+                  ...this.state.data!,
+                  alertState,
+                },
+              });
+            }
+          }
+        );
+      }
     } else {
       this.setState({
```
torkelo, line 215: this bit I did not understand really (at least not the use of async data transformations), it results in two state updates (so two re-renders of VizPanel) but everything here is sync and pretty simple filtering. The AlertStateInfo extraction with reduce also feels very hard to understand. 

I tried to make sense it, and ended up with this refactor: https://github.com/grafana/scenes/compare/data-layers-alert-states...data-layers-alert-states-refactor?expand=1

torkelo: In my refactor I used `DataFameView<AlertStateInfo>`  simplify the extraction of filtering of the alertState

dprokop: But no matter if it's async/sync it will still result in two state updates, as data layers emit values independently.

Ref your refactor proposal- I think it's OK as it does exactly the same thing, but in a sync way :) I personally found it appealing to handle two cases using the same, let's say standard method, which I consider to be transformations. 

Do you mind merging your branch into this PR @torkelo?

torkelo: @dprokop merged. 

> personally found it appealing to handle two cases using the same, let's say standard method, which I consider to be transformations.

Not sure they are ideal like this, as they are mainly designed for declarative and persisted transformations, so they become very verbose (and type unsafe), for very simple transformations/filterings I think they are not worth it. Here that filter values by panel id was ~10-15 lines of very hard to understand declarative model, but in the code verision it's a simple if statement on & 3 lines of code.

dprokop: yup yup, agree!

torkelo (approved):

PR #455, opened 2023-11-13
Variables: Fixes url sync issue for key/value multi value variables

torkelo:
Noticed that dashboards with variables that have different value / text representation, after a full page reload the panel repeats only showed the "id" (value) representation and was not showing the correct text representation. 

The issue as that we did not compare the text values when deciding to update or not


axelavargas (approved):
Looks good to me :)

kaydelaney (approved):

PR #459, opened 2023-11-14
Variables: Add natural sort from core grafana to query variables

axelavargas:
In core grafana we received a community contribution to implement natural sort. https://github.com/grafana/grafana/pull/78024 

This PR just adds their changes and a few unit tests.

### DIFF
--- a/packages/scenes/src/variables/variants/query/utils.ts
+++ b/packages/scenes/src/variables/variants/query/utils.ts
@@ -79,31 +79,23 @@ export const sortVariableValues = (options: any[], sortOrder: VariableSort) => {
     return options;
   }
 
-  const sortType = Math.ceil(sortOrder / 2);
-  const reverseSort = sortOrder % 2 === 0;
+  // @ts-ignore
+  const sortByNumeric = (opt) => {
+    if (!opt.text) {
+      return -1;
+    }
+    const matches = opt.text.match(/.*?(\d+).*/);
+    if (!matches || matches.length < 2) {
+      return -1;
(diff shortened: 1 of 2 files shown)


packages/scenes/src/variables/variants/query/utils.ts
```diff
@@ -101,6 +101,19 @@ export const sortVariableValues = (options: any[], sortOrder: VariableSort) => {
...
+      if (!a.text) {
+        return -1;
+      }
+
+      if (!b.text) {
+        return 1;
+      }
+
+      return a.text.localeCompare(b.text, undefined, { numeric: true });
+    });
```
torkelo, line 116: const sortType = Math.ceil(sortOrder / 2); 

and these if statements are pure cryptic gold, can we switch to using the enum VariableSort (and switch/case statement)? (you will have to add a //@ts-expect-error for the newly added enum that, add a TODO for that so we can find it and remove the ts-expect-error when we update the schema package in the future

axelavargas: I agree with that, it's confusing , I will do the changes 😄

axelavargas: @torkelo I did the refactor, but I couldn't find a way to avoid hardcoding the value of the new options. Also, there is some `//ts-ignore` around because options is `any[]`

torkelo (approved):
Nice! Do much easier to read and understand now!!

dprokop (approved):

PR #463, opened 2023-11-15
VizPanel: Add possibility to dynamically change panel type

kaydelaney:
Just some initial code to address the first task in https://github.com/grafana/grafana/issues/78152
Will have a follow-up PR in the main repo

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanel.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.tsx
@@ -91,7 +91,10 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
   private _prevData?: PanelData;
   private _dataWithFieldConfig?: PanelData;
   private _structureRev = 0;
-  private _cachedPluginOptions: Record<string, { options: DeepPartial<TOptions>, fieldConfig: FieldConfigSource<DeepPartial<TFieldConfig>> }> = {};
+  private _cachedPluginOptions: Record<
+    string,
+    { options: DeepPartial<TOptions>; fieldConfig: FieldConfigSource<DeepPartial<TFieldConfig>> } | undefined
+  > = {};
 
   public constructor(state: Partial<VizPanelState<TOptions, TFieldConfig>>) {
     super({
@@ -168,6 +171,7 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
       options: withDefaults.options,
       fieldConfig: withDefaults.fieldConfig,
       pluginVersion: currentVersion,
+      pluginId: plugin.meta.id,
     });
   }
 
@@ -176,27 +180,26 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
   }
 
   public changePlugin(newPluginId: string) {
+    const { title, options, fieldConfig, pluginId } = this.state;
     this._cachedPluginOptions[this.state.pluginId] = {
-      options: this.state.options,
-      fieldConfig: this.state.fieldConfig
+      options,
+      fieldConfig,
     };
 
+    const panel: PanelModel = {
+      title,
+      options,
+      fieldConfig,
+      id: 1,
+      type: newPluginId,
+    };
+    const newOptions = this._plugin?.onPanelTypeChanged?.(panel, pluginId, options, fieldConfig);
     const cachedPluginOptions = this._cachedPluginOptions[newPluginId];
-    if (cachedPluginOptions) {
-      this.setState({
-        pluginId: newPluginId,
-        pluginVersion: undefined,
-        options: cachedPluginOptions.options,
-        fieldConfig: cachedPluginOptions.fieldConfig,
-      })
-    } else {
-      this.setState({
-        pluginId: newPluginId,
-        pluginVersion: undefined,
-        options: {},
-        fieldConfig: { defaults: {}, overrides: [] },
-      });
-    }
+    const newState = {
+      options: { ...cachedPluginOptions?.options, ...newOptions },
+      fieldConfig: cachedPluginOptions?.fieldConfig ?? { defaults: {}, overrides: [] },
+    };
+    this.setState(newState);
 
     this._loadPlugin(newPluginId);
     this._structureRev++;
--- a/packages/scenes/src/components/VizPanel/VizPanel.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.tsx
@@ -90,7 +90,8 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
   private _plugin?: PanelPlugin;
   private _prevData?: PanelData;
   private _dataWithFieldConfig?: PanelData;
-  private _structureRev: number = 0;
+  private _structureRev = 0;
+  private _cachedPluginOptions: Record<string, { options: DeepPartial<TOptions>, fieldConfig: FieldConfigSource<DeepPartial<TFieldConfig>> }> = {};
 
   public constructor(state: Partial<VizPanelState<TOptions, TFieldConfig>>) {
     super({
@@ -101,9 +102,7 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
       ...state,
     });
 
-    this.addActivationHandler(() => {
-      this._onActivate();
-    });
+    this.addActivationHandler(() => this._onActivate());
   }
 
   private _onActivate() {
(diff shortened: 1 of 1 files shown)


packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -176,6 +175,33 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
     return plugin && plugin.meta.info.version ? plugin.meta.info.version : config.buildInfo.version;
   }
 
+  public changePlugin(newPluginId: string) {
+    this._cachedPluginOptions[this.state.pluginId] = {
```
dprokop, line 179: Why caching by plugin id?

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -176,6 +175,33 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
...
+        fieldConfig: cachedPluginOptions.fieldConfig,
+      })
+    } else {
+      this.setState({
+        pluginId: newPluginId,
+        pluginVersion: undefined,
+        options: {},
+        fieldConfig: { defaults: {}, overrides: [] },
+      });
+    }
```
dprokop, line 199: This will require panel type change migration handler - some panels have those handlers implemented: https://github.com/grafana/grafana/blob/main/public/app/features/dashboard/state/PanelModel.ts#L493

dprokop (commented):
Would be nice to have a demo in the demos app of changing panel type :)

packages/scenes/src/components/VizPanel/VizPanel.tsx
kaydelaney: Just copying what's already done

https://github.com/grafana/grafana/blob/384db8e0cab3356090ce6f7cba15663a7490a945/public/app/features/dashboard/state/PanelModel.ts#L505-L514

I guess it's for if you switch between panel types so your options don't get trashed

packages/scenes/src/components/VizPanel/VizPanel.tsx
kaydelaney: Ah, right, forgot this

torkelo: @kaydelaney @dprokop 

I am torn if this should be here or not, when we change plugin we could just change VizPanel instance, and this code to remember old state (in case user switches back) could be handled in panel edit (core). 

The only benefit of making it a feature of VizPanel is if it's something other plugins might want to use?

dprokop: @torkelo  - I think I'd rather have that implemented in the lib so that one can easily implement for instance timeseries / table view without having to implement this themselves. It's a nice addition that I think plugins can benefit from.

torkelo: @dprokop time series table view?

dprokop: sth like "View underlying data"  type of visualization.

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -176,6 +175,33 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
...
+        options: cachedPluginOptions.options,
+        fieldConfig: cachedPluginOptions.fieldConfig,
+      })
+    } else {
+      this.setState({
+        pluginId: newPluginId,
+        pluginVersion: undefined,
+        options: {},
+        fieldConfig: { defaults: {}, overrides: [] },
+      });
```
torkelo, line 198: think  you need to clear / unset `this._plugin` before doing this state change,

packages/scenes/src/components/VizPanel/VizPanel.tsx
torkelo: what would be the use case for doing this in plugins? I guess for building panel edit like UIs? 

If a scene app plugin wanted to change the visualization they would likely not want this function that only allow changing pluginId and not option, and if they wanted to change pluginId + options + fieldConfig, why not just replace the VizPanel with another instance built using PanelBuilders?

packages/scenes/src/components/VizPanel/VizPanel.tsx
torkelo: I think we can cache this in panel edit.

packages/scenes/src/components/VizPanel/VizPanel.tsx
dprokop: For the described case (type + options + field config) yes, instance change makes sense of course. 

But considering how would that change look in the core - can we do a quick POC of that and see how much complexity is going to add there? The more I think about it the more torn I am honestly.

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -169,13 +171,40 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
...
+    };
+
+    const panel: PanelModel = {
+      title,
+      options,
+      fieldConfig,
+      id: 1,
+      type: newPluginId,
+    };
+    const newOptions = this._plugin?.onPanelTypeChanged?.(panel, pluginId, options, fieldConfig);
```
torkelo, line 196: this is the wrong way around, the onPanelTypeChanged should be called on the new plugin, it can then check the pluginId of the old plugin and re-use some of the options from the old plugin. Useful for example for stat, bar gauge and gauge panels, they share a lot of options. 

I think we can wait putting this in VizPanel, when we get to this in panel edit we can review what the best option is. I am not sure the way switch visualization in panel edit is going to be that common in scene apps.

packages/scenes/src/components/VizPanel/VizPanel.tsx
torkelo: I am torn also, but I would start with trying to do this outside VizPanel, and when we change plugin type in panel edit we do this by replacing the VizPanel instance, (And remembering / caching the old instance).

kaydelaney: Currently the way the panel edit scene is built is like this: 
```typescript

export function buildPanelEditScene(dashboard: DashboardScene, panel: VizPanel): PanelEditor {
  const panelClone = panel.clone();
  const dashboardStateCloned = sceneUtils.cloneSceneObjectState(dashboard.state);

  return new PanelEditor({
    dashboardRef: dashboard.getRef(),
    sourcePanelRef: panel.getRef(),
    panelRef: panelClone.getRef(),
    controls: dashboardStateCloned.controls,
    $variables: dashboardStateCloned.$variables,
    $timeRange: dashboardStateCloned.$timeRange,
    body: new SplitLayout({
      direction: 'row',
      primary: new SceneFlexLayout({
        direction: 'column',
        children: [panelClone],
      }),
      secondary: new SceneFlexLayout({
        direction: 'column',
        children: [new PanelOptionsPane(panelClone), new PanelVizTypePicker(panelClone)],
      }),
    }),
  });
}
```

Even if the panel editor's vizpanel is changed, i.e
```typescript
panelEditor.setState({ panelRef: myNewPanel.getRef() });
```

it seems like the body will still use the original panel clone:
``` typescript
{
  // ...
      primary: new SceneFlexLayout({
        direction: 'column',
        children: [panelClone], // <- here
      }),
  // ...
}
```

I'm probably missing something, but would I have to write code to manually update the children so they now refer to the new panel?

torkelo: @kaydelaney hm.. interesting. We could solve it by having some kind of wrapper SceneObject around the VizPanel that can manage the plugin changes. 

But if you think it makes sense to have it here I we could try a new take on this changePlugin function that takes in 
more than just pluginId. 

```ts
interface ChangePluginArgs<TOptions, TFieldConfig> {
   pluginId: string;
   options?: TOptions;
   fieldConfig?: TFieldConfig;   
}
``

but the thing that feels akward is that VizPanel is generic with TOptions / TFieldConfig and this function would basically change the typing of the VizPanel. I know we don't really use the typed version in core dashboards. 

Hm.. torn. But at least the caching of old options I think belongs outside VizPanel.

kaydelaney: Draft PR based on the wrapper suggestion: https://github.com/grafana/grafana/pull/78477

kaydelaney:
Dropping in favor of https://github.com/grafana/grafana/pull/78477

PR #464, opened 2023-11-15
VIzPanel: Support markdown in panel description

dprokop:


packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
```diff
@@ -134,7 +134,7 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
         {width > 0 && height > 0 && (
           <PanelChrome
             title={titleInterpolated}
-            description={description ? () => model.interpolate(description) : ''}
+            description={descriptionInterpolated}
```
torkelo, line 137:
```suggestion
            description={model.getDescription}
```

or some version of that, so we do not have to parse the markdown and interpolate the description string on every render, only when you hover over the info icon

torkelo: PanelChrome description prop does take a function

torkelo (approved):

packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
dprokop: Indeed, glad that you brought this becuase ..... I think we need to allow function for `description` for a viz panel as well. Overall, some core implementations report interactions, via function props (i.e. decription).

PR #466, opened 2023-11-16
DynamicApp: Demo of a SceneApp with loading phase, a first setup phase and settings page that rebuilds pages

torkelo:
I have noticed that it's tricky to use SceneApp and SceneAppPages when you need a loading and setup phase where you either need to load settings and/or have the user make some initial configuration before all the pages can be built.  

This is an exploration in how we can solve this. 

Instead of extending SceneApp I am using a behavior attached to the SceneApp. This is because extending SceneObjects is problematic due to Typescript and generics co-variance and problematic from a backward compatible way (makes it hard to make changes to the library). 

* [x] SceneApp that initially starts with no pages
* [x] When activated it will add a loading page that has a fallback (so all urls will show loading page) 
* [x] Once settings are loaded we check if the the app has been configured, if not we show the initial setup page 
* [x] Once the app has been configured we add all the app pages (with app settings being passed around) 
* [x] If settings change we rebuild the app pages (not ideal, as we loose the scene cache, but workable for this scenario I think).

https://github.com/grafana/scenes/assets/10999/c3eda879-1dc7-4f73-8424-72c623205906

## Futher SceneApp exploration 

I think SceneApp could be improved. Maybe support a cleaner way to extend than a behavior like in this PR and a easier way to access it (without having to call getParent on SceneAppPage)

*  Support redirects
*  Allow for more plain pages (that do not implement the full SceneAppPageLike interface) 
*  Support variables or state that can be easily accessed and subscribed to by any part of the app

### DIFF
--- a/packages/scenes-app/src/pages/DynamicApp.tsx
+++ b/packages/scenes-app/src/pages/DynamicApp.tsx
@@ -1,0 +1,262 @@
+import {
+  EmbeddedScene,
+  SceneApp,
+  SceneAppPage,
+  SceneFlexLayout,
+  SceneFlexItem,
+  SceneReactObject,
+  useSceneApp,
+  SceneObjectState,
+  SceneObjectBase,
+  sceneGraph,
+  PanelBuilders,
+  SceneVariableSet,
+  DataSourceVariable,
+  SceneComponentProps,
+  SceneObject,
+} from '@grafana/scenes';
+import React from 'react';
+import { prefixRoute } from '../utils/utils.routing';
+import { DataSourceInstanceSettings, GrafanaTheme2 } from '@grafana/data';
+import { css } from '@emotion/css';
+import { Button, Field, Spinner, Switch, VerticalGroup } from '@grafana/ui';
+import { getEmbeddedSceneDefaults, getQueryRunnerWithRandomWalkQuery } from '../demos/utils';
+import { SceneAppState } from '@grafana/scenes/src/components/SceneApp/types';
+import { DataSourcePicker } from '@grafana/runtime';
+
+interface DynamicAppState extends SceneObjectState {
+  initialDataSource?: string;
+  showPanelDescriptions?: boolean;
+  isInitialized?: boolean;
+  isConfigured?: boolean;
+  settings: DynamicAppSettings;
+}
+
+class DynamicApp extends SceneObjectBase<DynamicAppState> {
+  constructor(state: DynamicAppState) {
+    super(state);
+    this.addActivationHandler(this._activationHandler.bind(this));
+  }
+
+  private _activationHandler() {
+    if (this.state.isInitialized) {
+      return;
+    }
+
+    // Set initializing (loading) page
+    this._updateState({
+      pages: [getInitializingPage()],
+    });
+
+    setTimeout(() => this._initializeCompleted(), 2000);
+  }
+
+  private _updateState(appState: Partial<SceneAppState>, customState?: Partial<DynamicAppState>) {
+    const app = sceneGraph.getAncestor(this, SceneApp);
+    app.setState(appState);
+
+    if (customState) {
+      this.setState(customState);
+    }
+  }
+
+  private _initializeCompleted() {
+    if (!this.state.isConfigured) {
+      this._updateState(
+        {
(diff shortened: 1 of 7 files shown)


packages/scenes-app/src/pages/shared.ts
```diff
@@ -0,0 +1,28 @@
+import { SceneAppPage, SceneObject } from '@grafana/scenes';
+import { DynamicApp } from './DynamicApp';
+
+export interface AppSettings {
+  initialDataSource?: string;
+  showPanelDescriptions?: boolean;
+  isConfigured?: boolean;
+}
+
+export function getDynamicApp(model: SceneObject): DynamicApp {
```
domasx2, line 10: Interesting trick!
 I'm facing the same issue in a refactoring I'm working on, but solved it by dispatching an event to bubble up from the "landing page" scene and listening to it in the `SceneApp`.

It feels though like we're at a point where we'd benefit from a state management solution that's loosely coupled with the scene object hierarchy, to avoid a flimsy tangle of interconnections. "context for scenes"  or "redux for scenes" or some such. Already seen cases of variables being (mis)used for this

jewbetcha: Not only some state management, but some ability to store plugin settings :D

packages/scenes-app/src/pages/loadSettings.ts
```diff
@@ -0,0 +1,24 @@
+import { AppSettings } from './shared';
+
+export function loadSettings(): Promise<AppSettings> {
+  let settings: AppSettings = {};
+
+  const settingsJson = localStorage.getItem('dynamic_app');
```
jewbetcha, line 6: My first question was where the settings would be stored. This could be a good time to light a fire under some of the efforts to provide plugins with some persistent storage, then this would be _extra_ helpful by providing an API to read/write.

torkelo: @jewbetcha this use of localStorage was just a "simulated" real store (using an http API to fetch / save settings). 

Still waiting on app platform work to support new storage & apis for app plugins

jewbetcha: Nice, yeah I'm all for this!

packages/scenes-app/src/pages/DynamicApp.tsx
```diff
@@ -0,0 +1,201 @@
...
+    if (this.state.settings?.isConfigured) {
+      this._updatePages(buildAppPages(this.state.settings!));
+    }
+  };
+
+  public onCompleteSetup = () => {
+    this.onChangeSettings({ isConfigured: true });
+    locationService.push(prefixRoute('dynamic-app'));
+  };
+}
```
dprokop, line 71: This is very interesting. Wonder how much we should consider this an API vs a pattern? I mean - this implementation is quite loosely coupled with functions (i.e. `getLoadingPage` or `getFirstSetupPage`).

Thinking of a different scenario, where the first setup is not necessary, as the configuration is provided via default settings. I assume then it's the API that would be just return `settings.isConfigured = true`, which defines an API of the persistent app settings to some extend.

My point is to learn how we want to solve this problem. 

@domasx2 / @jewbetcha - would a DynamicApp API be something that you would fancy in your apps, or rather a pattern/guideline for solving a certain problem?

domasx2: I thought about it, but couldn't figure out an abstraction that felt useful

torkelo: @domasx2 @dprokop I had the same idea, that maybe this "loading/user setup/running" phase could be built into SceneApp some way. I think that could be interesting if we can find the abstraction / base class or what ever to make that possible. 

Still struggling finding the right future for SceneApp, discussed some ideas here https://github.com/grafana/scenes/issues/470

jewbetcha: I think DynamicApp API sounds really nice, especially since I know a lot of plugin teams (including us) have to do things like make a query for some metrics to see if setup has been complete. 

Something like this could get rid of those types of workarounds 🎉

PR #468, opened 2023-11-16
fix: compare time ranges using actual unix time vs objects

darrenjaneczek:
We want to avoid false detection of stale time lines and compare using the underlying `moment.unix()` time stamp for the `from`s and the `to`s, so that we don't accidentally compare object instances.

See:
- https://github.com/grafana/grafana/issues/78185

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.test.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.test.ts
@@ -539,7 +539,9 @@ describe('SceneQueryRunner', () => {
 
   describe('when time range changed while in-active', () => {
     it('It should re-issue new query', async () => {
-      const timeRange = new SceneTimeRange();
+      const from = '2000-01-01';
+      const to = '2000-01-02';
+      const timeRange = new SceneTimeRange({from, to});
       const queryRunner = new SceneQueryRunner({
         queries: [{ refId: 'A' }],
         $timeRange: timeRange,
@@ -556,6 +558,10 @@ describe('SceneQueryRunner', () => {
 
       deactivateQueryRunner();
 
+      await new Promise((r) => setTimeout(r, 1));
+
+      const differentTo = '2000-01-03'
+      timeRange.setState({from, to: differentTo});
       timeRange.onRefresh();
 
       queryRunner.activate();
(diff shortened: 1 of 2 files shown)


packages/scenes/src/querying/SceneQueryRunner.test.ts
```diff
@@ -556,6 +558,10 @@ describe('SceneQueryRunner', () => {
 
       deactivateQueryRunner();
 
+      await new Promise((r) => setTimeout(r, 1));
+
+      const differentTo = '2000-01-03'
+      timeRange.setState({from, to: differentTo});
       timeRange.onRefresh();
```
darrenjaneczek, line 565: The previous `onRefresh` was using relative times, but for some reason, the `onRefresh` wasn't actually changing to a different time, so I use explicit dates to fix this test.

torkelo: @darrenjaneczek when I run this test in main the time range changes, due to the timeout we wait for 1 millesecond and then refresh, the refresh will re-evaluate the relative time range and update timeRange.state.value. So when I check  `timeRange.state.value.to.valueOf()`  it has changed after that refresh.

torkelo: The reason your where not seeing a change was that your calling `.unix()` which has not millisecond precision

packages/scenes/src/querying/SceneQueryRunner.test.ts
```diff
@@ -556,6 +558,10 @@ describe('SceneQueryRunner', () => {
 
       deactivateQueryRunner();
 
+      await new Promise((r) => setTimeout(r, 1));
```
torkelo, line 561: since your changing the time range you do not need this (this was only there to wait 1 millisecond and then refresh time range)

packages/scenes/src/querying/SceneQueryRunner.test.ts
torkelo: but I think using unix is fine here we don't need millisecond precision on this time range comparison

torkelo (approved):

torkelo:
@darrenjaneczek added a release label, so there should be a new release soon of "@grafana/scenes"

Issue #462, opened 2023-11-15
SceneQueryRunner: Sets invalid PanelData state with undefined series prop


torkelo:
I am seeing Panel errors flicker (shown just 1 frame) for panels with transformations. 

The error is in SceneDataTransformer where it passes undefined to transformDataFrame. data.series is undefined (which it should not be). 

The cause is data layers coming in before the first query response, and updating PanelData but only setting annotations array, leaving series prop undefined. 

https://github.com/grafana/scenes/blob/main/packages/scenes/src/querying/SceneQueryRunner.ts#L196

PR #474, opened 2023-11-21
Fix non-null assertion in SceneQueryRunner

dprokop:
Fixes https://github.com/grafana/scenes/issues/462

Collecting data layers assumed queries have been resolved already. This is not true, layers can definitely finish sooner than queries, there is no guarantee. And using non-null assertions is bad anyway ;)

Fixes:
-  transformation errors we all saw during bug  bash today 
- [data.timeRange](https://github.com/grafana/scenes/blob/main/packages/scenes/src/querying/SceneQueryRunner.ts#L251) being undefined if layers finish sooner than queries 

<details>
  <summary>📦 Published PR as canary version: <code>1.24.1--canary.474.6946948769.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@1.24.1--canary.474.6946948769.0
  # or 
  yarn add @grafana/scenes@1.24.1--canary.474.6946948769.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.ts
@@ -152,6 +152,7 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
   }
 
   private _onLayersReceived(results: Map<string, PanelData>) {
+    const timeRange = sceneGraph.getTimeRange(this);
     const dataLayers = sceneGraph.getDataLayers(this);
     const { dataLayerFilter } = this.state;
     let annotations: DataFrame[] = [];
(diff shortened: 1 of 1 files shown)


kaydelaney (approved):

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -193,9 +193,11 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
       }
     }
 
+    const baseStateUpdate = this.state.data ? this.state.data : emptyPanelData;
+
     this.setState({
       data: {
-        ...this.state.data!,
+        ...baseStateUpdate,
```
torkelo, line 200: hm.., I wonder if it's not better to somehow defer this data update and add the annotations and alert state update in onDataReceived. 

thinking this will cause "no data" flickering but actually it won't, this is fine. isDataReadyToDisplay checks _hasFetchedData which is set in onDataReceived, so this solution of setting empty panel data when layers finish before query should be fine

torkelo (approved):

packages/scenes/src/querying/SceneQueryRunner.ts
dprokop: Agree, this should be fine as you described.

BTW - this PR is also fixing an issue with `data.timeRange` being undefined in https://github.com/grafana/scenes/blob/main/packages/scenes/src/querying/SceneQueryRunner.ts#L251

PR #477, opened 2023-11-23
VizPanel: Only pass data layers when plugin supports them

dprokop:
Fixes an issue where alert states and annotations would be passed to panels that does not support them. This resulted in i.e. alert state icon being shown in Stat panels.

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanel.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.tsx
@@ -15,6 +15,7 @@ import {
   applyFieldOverrides,
   PluginType,
   renderMarkdown,
+  PanelPluginDataSupport,
 } from '@grafana/data';
 import { PanelContext, SeriesVisibilityChangeMode, VizLegendOptions } from '@grafana/ui';
 import { config, getAppEvents, getPluginImportUtils } from '@grafana/runtime';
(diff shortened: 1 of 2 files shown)


torkelo (approved):
How come the stat panel had an alertstate associated with it? Did it used to be a graph with an alert rule on that dashboard with the same panel id? 

This looks ok, but feels like it should be on the SceneQueryRunner somehow, but a bit more complex to add the logic there.

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -306,6 +309,11 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
       }),
     };
 
+    if (!(pluginDataSupport.alertStates || pluginDataSupport.annotations)) {
+      this._dataWithFieldConfig.alertState = undefined;
+      this._dataWithFieldConfig.annotations = undefined;
+    }
```
ivanortegaalba, line 315: Why do we mark as undefined both of them as `undefined`?

If there is pluginDataSupport.alertStates, but not pluginDataSupport.annotations. Should we mark `pluginDataSupport.alertStates` as `undefined`?

ivanortegaalba (changes requested):
Small comment on the disabling the annotations and alerts. 

I think we should not assume if there is no annotations, there is not alert state since the viz panel can be extended by a plugin that can be anything. WDYT?

dprokop:
>How come the stat panel had an alertstate associated with it? Did it used to be a graph with an alert rule on that dashboard with the same panel id?

@torkelo  - most probably yes.

> This looks ok, but feels like it should be on the SceneQueryRunner somehow, but a bit more complex to add the logic there.

Adding this to SQR would mean it needs to now what is the data support from the plugin. So we could [not rely directly on state as we do now](https://github.com/grafana/scenes/blob/fix-layers-passed-regardless-data-support/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx#L33), but we would need an additional function that would clean up unnecessary layers. I think this behavior is peculiar to VizPanel, and it's OK to have it implemented there. 

The concern I have is whether or not it should belong to `applyFieldConfig` function.

packages/scenes/src/components/VizPanel/VizPanel.tsx
dprokop: This is how it currently works https://github.com/grafana/grafana/blob/main/public/app/features/query/state/PanelQueryRunner.ts#L332.

But actually, we can improve this and handle those layers independently.

ivanortegaalba (approved):
Looks good to me! 💯

PR #482, opened 2023-11-24
SceneQueryRunner: Fixes issue with cloned scene query runner would issue new query

torkelo:
Working on a new way to handle the "View panel" feature in dashboards (where I need to clone the panel). And it's issuing a new query when it gets the containerWidth.


dprokop (approved):
This looks good!

PR #483, opened 2023-11-27
VizPanelMenu: Fix auto focus / keyboard navigation issue

torkelo:
Alt fix to https://github.com/grafana/grafana/pull/78683

PR #484, opened 2023-11-28
SceneObject: Allow component rendering to override the "lazy" rendering and render content before activation 

torkelo:
Alt alt fix :) 
https://github.com/grafana/scenes/pull/483 

The logic that staggers rendering to after activation is useful for higher-level scene objects to make sure that parents activate before children. But for leaf nodes it does not really matter, and so far we have two instances where this has caused issues. 

SceneControlsSpacer and VizPanelMenu. This adds a new property `UNSAFE_renderBeforeActive` to override the condition that renders null on first mount so that the scene object component is rendered before activation.


torkelo:
alt name could be instead of `UNSAFE_renderBeforeActive` UNSAFE_renderOnMount` ?

dprokop:
See my comment https://github.com/grafana/scenes/pull/483#issuecomment-1829669102

Issue #486, opened 2023-11-28
VizPanel/SceneQueryRunner: $__interval should work in title and displayName


torkelo:
* ` $__interval ` currently does not work in title 
* ` $__interval ` does not work in displayName in fieldConfig defaults

PR #487, opened 2023-12-01
Macros: Support $_interval[_ms] variable

dprokop:
Closes https://github.com/grafana/scenes/issues/486

An approach to support `__interval[_ms]` macros. 

I've also tried a different approach which would resolve the inteval on the SceneQueryRunner without the need for storing `_lastRequest`, but it has become super tricky as this function would have to by async (in oroder to get interval from the data source) and this would require us to also have async `FormatVariable` :(

But with this `SceneDataProvide` API addition, we can easily implement interval retrieval for SharedDataQueryProvider - see PR in core: https://github.com/grafana/grafana/pull/78952

@torkelo  any thoughts on the default 1s interval returned?

<details>
  <summary>📦 Published PR as canary version: <code>1.25.0--canary.487.7060253941.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@1.25.0--canary.487.7060253941.0
  # or 
  yarn add @grafana/scenes@1.25.0--canary.487.7060253941.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.ts
@@ -59,7 +59,6 @@ export interface QueryRunnerState extends SceneObjectState {
   // Private runtime state
   _isWaitingForVariables?: boolean;
   _hasFetchedData?: boolean;
-  _lastRequest?: [DataQueryRequest, DataQueryRequest | undefined];
 }
 
 export interface DataQueryExtended extends DataQuery {
@@ -381,13 +380,9 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
         // change subscribe callback below to pipe operator
         this._querySub = forkJoin([runRequest(ds, request), runRequest(ds, secondaryRequest)])
           .pipe(timeShiftQueryResponseOperator)
-          .subscribe((data) => {
-            this.onDataReceived(data, [request, secondaryRequest]);
-          });
+          .subscribe(this.onDataReceived);
       } else {
-        this._querySub = runRequest(ds, request).subscribe((data) => {
-          this.onDataReceived(data, [request, undefined]);
-        });
+        this._querySub = runRequest(ds, request).subscribe(this.onDataReceived);
       }
     } catch (err) {
       console.error('PanelQueryRunner Error', err);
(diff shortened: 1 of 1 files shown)


packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -476,7 +481,7 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
     return [request, secondaryRequest];
   };
 
-  private onDataReceived = (data: PanelData) => {
+  private onDataReceived = (data: PanelData, request: [DataQueryRequest, DataQueryRequest | undefined]) => {
```
torkelo, line 484: PanelData already has the request I think

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -560,6 +565,13 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
     this._adhocFilterSet = set;
     this._adhocFilterSub = this._adhocFilterSet?.subscribeToState(() => this.runQueries());
   }
+
+  public getInterval(): IntervalValues {
+    return {
+      interval: this.state._lastRequest?.[0].interval ?? '1s',
```
torkelo, line 571: I think we can just do this.state.data.request (but on mobile so not sure)

torkelo (approved):
Nice idea with using a macro! Looks great.

But not sure we need the changes to scene query runner, PanelData should have the request and the interval, but not 100% sure.

packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -100,6 +100,10 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
     return this._results;
   }
 
+  public getInterval() {
+    return this.getSourceData().getInterval?.() || { interval: '1s', intervalMs: 1000 };
```
torkelo, line 104: Ah, do we preserve PanelData.request after transform, I think that should be preserved

dprokop: yup, it's preserved

packages/scenes/src/core/types.ts
```diff
@@ -202,6 +203,7 @@ export interface SceneDataProvider extends SceneObject<SceneDataState> {
   isDataReadyToDisplay?: () => boolean;
   cancelQuery?: () => void;
   getResultsStream?(): Observable<SceneDataProviderResult>;
+  getInterval?(): IntervalValues;
```
torkelo, line 206: Do we really need this? Can't we just read it from data.request?

PR #491, opened 2023-12-05
VizPanel: Expose events for interactions

dprokop:
This PR introduces some events specific to `VizPanel` that we can use in core (and apps) for tracking certain interactions that users can take with the panels.

Spent some time thinking about an alternative approach, but couldn't figure out anything that would have a smaller impact. 

My point here is to have **a common way for handling interactions** that may be triggered from a VizPanel (i.e. description interaction is handled indirectly by panel chrome, panel menu interaction is handled by the menu rendering etc...) It creates a mess in core to report those interactions in different ways.

Corresponding PR in core: https://github.com/grafana/grafana/pull/79078

<details>
  <summary>📦 Published PR as canary version: <code>1.27.0--canary.491.7102227881.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@1.27.0--canary.491.7102227881.0
  # or 
  yarn add @grafana/scenes@1.27.0--canary.491.7102227881.0
  ```
</details>


torkelo (approved):
first when looking that the menu shown event I was thinking, wait this can easily be done from core. But then looking at the other events, not so easy. And it's nice to have these events together. 

I guess there is one slight alternative to this approach and that would be for all these events to use the same event type "UserActionEvent" or something. To more easily subscribe to all in one go. And then have some kind of action type property. Not sure if it's better :)

dprokop:
> first when looking that the menu shown event I was thinking, wait this can easily be done from core

yeah, it's trivial for menu (via behavior), but not for other events :( 

> these events to use the same event type "UserActionEvent" or something.

Yes, this is nice alternative, would expose event types only, to do the correct reporting in core, while only crating a single subscription!

dprokop:
@torkelo pushed a minor refactor to a single event, looks way better.

PR #492, opened 2023-12-06
VizPanel Re-render skipDataQuery panels when time range change

torkelo:
Fixes issue with dash list panel, text panel (in case it uses variables that include time range) 

Because you can't do conditional subscriptions (effects) in render I could not find any other way than do this subscription from the model. We could always subscribe to time range in VizPanelRenderer (via useState), but it would cause unnecessary re-renders for data panels


packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -179,6 +179,12 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
       fieldConfig: withDefaults.fieldConfig,
       pluginVersion: currentVersion,
     });
+
+    // Non data panels needs to be re-rendered when time range change
+    if (plugin.meta.skipDataQuery) {
```
ivanortegaalba, line 184: When @axelavargas was investigating about using this parameter to split between widgets and vizs, we realized there are many custom plugins that don't use `skipDataQuery: true` even if they don't relay on querying data, and data modifiers. 

I don't remember specific plugins, but maybe @axelavargas does.

What would happen if a plugin creator do not set `skipDataQuery: true` but they don't use query runners? They won't get the plugin re-rendered in the dashboard, right? 

Anyways I don't see any better way to solve this, without triggering an extra re-render for every plugin (I wouldn't go in this direction for missing a configuration).

Do you see it as a valid concern?

packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
```diff
@@ -260,3 +260,15 @@ const getAlertStateStyles = (theme: GrafanaTheme2) => {
     }),
   };
 };
+
+function getPanelTimeRange(sceneTimeRange: SceneTimeRangeLike, plugin: PanelPlugin | undefined, data?: PanelData) {
```
ivanortegaalba, line 264: Would you move this to the `VizPanel` scene object? It would help to test this logic and simplify the `Component` render function.

ivanortegaalba (approved):
Works as expected! I left a couple of comments

packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
torkelo: @ivanortegaalba  @dprokop I tried moving it but feels awkward as VizPanel model function as it only really needs the plugin (from the model/this), the other two params I want to pass in, I have to make the function public and exposed to plugins. 

I think I prefer it as simple render function util that controls what time range to pass to PanelProps. 

I added 2 new render (react testing) tests to VizPanel, we had no tests for VizPanelRenderer. They become simple and pretty nice!

ivanortegaalba: Neat!

packages/scenes/src/components/VizPanel/VizPanel.test.tsx
```diff
@@ -454,30 +469,81 @@ describe('VizPanel', () => {
     });
   });
 
-  describe('Non data plugin', () => {
+  describe('VizPanel panel rendering ', () => {
+    beforeEach(() => {
+      panelRenderCount = 0;
+      panelProps = undefined;
```
ivanortegaalba, line 475: Clever way to count renders! I used to render them as text and query through selectors when it is a react component, but I like this one more

ivanortegaalba (approved):
Ready to go @torkelo! I like how simple it is!

PR #498, opened 2023-12-07
Splitter: Fixes small issue with keyboard control

kaydelaney:
Fixes a small issue where if the panel was being altered with arrows keys and focus was lost (e.g. by clicking somewhere), panel size would continue to be modified when arrow keys were lifted.


axelavargas (approved):
Code looks good to me :)

PR #499, opened 2023-12-11
Make updateValueGivenNewOptions protected

bfmatei:
One edge case that we encountered during development is that we need to extend `QueryVariable` and override `updateValueGivenNewOptions` in order to auto-select a value based on some custom criteria. The problem is that the named function is private in `MultiValueVariable`, causing TS to complain about incorrectly extending another class.

We suppressed the TS for the time being but it would be nice to avoid having `ts-expect-error` directive.

### DIFF
--- a/packages/scenes/src/variables/variants/MultiValueVariable.ts
+++ b/packages/scenes/src/variables/variants/MultiValueVariable.ts
@@ -73,7 +73,7 @@ export abstract class MultiValueVariable<TState extends MultiValueVariableState 
   /**
    * Check if current value is valid given new options. If not update the value.
    */
-  private updateValueGivenNewOptions(options: VariableValueOption[]) {
+  protected updateValueGivenNewOptions(options: VariableValueOption[]) {
     // Remember current value and text
     const { value: currentValue, text: currentText } = this.state;
 


torkelo (commented):
It's a bit scary to allow extending this, such critical complex logic in this function. 

And we are still missing a critical behavior here (To override/set custom value via URL) 

@bfmatei how are you extending it? are you calling it from the overriden function?

bfmatei:
Hey @torkelo !

So, our use case is the following:
- we have a query variable that has a custom behaviour for pre-selecting the default option like so:
   - it is either selectable by the user in a configuration page
   - it can also be "auto-detectable", as in the system tries to guess what the correct value is
   - it falls back to "All" option if neither of the above is applicable
- there is also a singleton service that is used in the workflow described above. It is used to inform the user that the value they selected is auto-selected or it is forced via config/URL
- there is also a custom behaviour of the `All` option in terms of how we interpolate it in queries (we are using custom transformers for this approach)
- there is also a hardcoded value that is neither `All` nor provided as a result of the query. This value is called `None` (having the value `$__none`, similar to `$__all`) and this one is also interpolated using custom transformers

What we do can be found [here](https://github.com/grafana/app-observability-plugin/blob/main/src/components/EnvironmentValueVariable.ts#L87) but as a short answer: we have a class that extends QueryVariable in order to add the custom logic.

The "demo" part of this is the environment selector in the app observability plugin.

torkelo:
@bfmatei thanks, maybe some of this can be extracted and incorporated into the core QueryVariable? not sure what makes sense though. handling a "none" / no value state is a big missing feature. 

> and this one is also interpolated using custom transformers

what does this mean? 

[Looking at your overriden implementatio](https://github.com/grafana/app-observability-plugin/blob/main/src/components/EnvironmentValueVariable.ts#L87 )n, it look like it's missing a lot of the default implementation behaviors (as well as the crucial SceneVariableValueChangedEvent which without it will be pretty broken).

I don't think making updateValueGivenNewOptions protected is enough for what your after here, you can't call it and then do something else that changes value, would cause double value changes, which should be fine but feels hacky. 

Maybe a new optional function "validateValueState" that is empty does nothing that you can override and interest the special logic at the end of updateValueGivenNewOptions  (ie the default updateValueGivenNewOptions would call validateValueState before doing the state change)

bfmatei:
Closing this in favor of a more approachable solution.

PR #504, opened 2023-12-18
SceneInterpolator: Ignore interval interpolation for prometheus queries

dprokop:
Due to the lack of a better idea at the moment, I'm proposing this hack for Prometheus queries (ref core code: https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/prometheus/datasource.ts#L888)

### DIFF
--- a/packages/scenes/src/variables/interpolation/sceneInterpolator.ts
+++ b/packages/scenes/src/variables/interpolation/sceneInterpolator.ts
@@ -9,6 +9,8 @@ import { formatRegistry, FormatVariable } from './formatRegistry';
 import { VARIABLE_REGEX } from '../constants';
 import { lookupVariable } from '../lookupVariable';
 import { macrosIndex } from '../macros';
+import { IntervalMacro } from '../macros/timeMacros';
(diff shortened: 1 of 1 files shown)


packages/scenes/src/variables/interpolation/sceneInterpolator.ts
```diff
@@ -34,6 +36,15 @@ export function sceneInterpolator(
     const fmt = fmt2 || fmt3 || format;
     const variable = lookupFormatVariable(variableName, match, scopedVars, sceneObject);
 
+    // This condition is handling a special case for Prometheus data source that relies on interval variables interpolation in the backend./
+    if (
+      variable instanceof IntervalMacro &&
+      sceneObject instanceof SceneQueryRunner &&
+      sceneObject.state.datasource?.type === 'prometheus'
```
torkelo, line 43: We can't rely on this type property, very rarely set

dprokop:
Closing in favor of https://github.com/grafana/grafana/pull/79645

PR #505, opened 2023-12-19
Demos: Adds an example with InteractiveTable with exandable rows that shows sub scenes 

torkelo:
Adds a demo using InteractiveTable that renders sub scenes in the expandable rows. 

This I think highlights one scenario where scenes lib is making something a bit harder than it should be. 

To render the expanded row scene we need to first attach it it to the scene. It's not too complicated but still feels like there is something here we could do to make this scenario easier. 

image

### DIFF
--- a/packages/scenes-app/src/demos/interactiveTableDemo.tsx
+++ b/packages/scenes-app/src/demos/interactiveTableDemo.tsx
@@ -49,10 +49,6 @@ interface TableVizState extends SceneObjectState {
 }
 
 class TableViz extends SceneObjectBase<TableVizState> {
-  public addExpandedRow(row: SceneObject) {
-    this.setState({ expandedRows: [...(this.state.expandedRows ?? []), row] });
-  }
-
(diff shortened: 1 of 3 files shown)


packages/scenes-app/src/demos/interactiveTableDemo.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const { expandedRows } = tableViz.useState();
+
+  const rowScene = expandedRows?.find((scene) => scene.state.key === row.handler);
+
+  useEffect(() => {
+    if (!rowScene) {
+      const newRowScene = buildExpandedRowScene(row.handler);
+      tableViz.setState({ expandedRows: [...(tableViz.state.expandedRows ?? []), newRowScene] });
+    }
+  }, [row, tableViz, rowScene]);
```
dprokop, line 109: So this is for caching purposes mainly? I men - it's quite tricky :)

dprokop (approved):
Nice demo!

PR #509, opened 2023-12-19
SceneTimeRange: Support delay now to avoid data drops in charts

ivanortegaalba:
Dashboards PR: https://github.com/grafana/grafana/pull/79703

This enables support for `UNSAFE_nowDelay` in Dashboards. It is prefixed as `UNSAFE` because we don't know how used and useful is this.

This allows us to override the now time by entering a time delay. This option accommodates known delays in data aggregation to avoid null values.

b2aa17d125dfc0aa8caee873760a8f44ebf2dcab

### DIFF
--- a/packages/scenes/src/core/SceneTimeZoneOverride.tsx
+++ b/packages/scenes/src/core/SceneTimeZoneOverride.tsx
@@ -32,7 +32,7 @@ export class SceneTimeZoneOverride
         timeRange.to,
         this.state.timeZone,
         timeRange.fiscalYearStartMonth,
-        timeRange.delayNow
+        timeRange.nowDelay
       ),
     });
   }
@@ -49,7 +49,7 @@ export class SceneTimeZoneOverride
         this.state.to,
         this.state.timeZone,
         this.getAncestorTimeRange().state.fiscalYearStartMonth,
-        this.state.delayNow
+        this.state.nowDelay
       ),
     });
   }
--- a/packages/scenes/src/core/SceneTimeZoneOverride.tsx
+++ b/packages/scenes/src/core/SceneTimeZoneOverride.tsx
@@ -27,7 +27,13 @@ export class SceneTimeZoneOverride
     this.setState({
       ...timeRange,
       timeZone: this.state.timeZone,
-      value: evaluateTimeRange(timeRange.from, timeRange.to, this.state.timeZone, timeRange.fiscalYearStartMonth),
+      value: evaluateTimeRange(
+        timeRange.from,
+        timeRange.to,
+        this.state.timeZone,
+        timeRange.fiscalYearStartMonth,
+        timeRange.delayNow
+      ),
     });
   }
 
@@ -42,7 +48,8 @@ export class SceneTimeZoneOverride
         this.state.from,
         this.state.to,
         this.state.timeZone,
-        this.getAncestorTimeRange().state.fiscalYearStartMonth
+        this.getAncestorTimeRange().state.fiscalYearStartMonth,
+        this.state.delayNow
       ),
     });
   }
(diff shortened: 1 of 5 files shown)


torkelo:
I wonder how much this option is actually for used. It has worked a bit badly for years , wonder if we should consider depreciating it.

packages/scenes/src/core/SceneTimeZoneOverride.tsx
```diff
@@ -42,7 +48,8 @@ export class SceneTimeZoneOverride
         this.state.from,
         this.state.to,
         this.state.timeZone,
-        this.getAncestorTimeRange().state.fiscalYearStartMonth
+        this.getAncestorTimeRange().state.fiscalYearStartMonth,
+        this.state.nowDelay
```
torkelo, line 52:
```suggestion
         this.getAncestorTimeRange().state.nowDelay,
```

torkelo (approved):
I think this looks good (with one minor issue)

it is a feature I am interested in how many actually use, and if we can deprecate it maybe, or find a simpler implementation for. Wonder if we can mark it as internal here in scenes lib?  

The problem with how it's imlemented now (and here) s that it causes simple time ranges like `Last 15 minutes` to read like  `now-30m to now-1m` . I think this now delay should not change how intervals are displayed, just how "now" is evaluated. 

An alternative change would be to introduce a global "nowDelay" state that we check in dateMath.parse , so it only changes how we interprete "now", the time range picker and the display of the selected time range does not need to change. 

thoughts @dprokop @thanos-karachalios ?

ivanortegaalba:
> I wonder how much this option is actually for used. It has worked a bit badly for years, wonder if we should consider depreciating it.

I opened this PR to track the usage of this setting, so we can track how many dashboards that are actually used contains those settings https://github.com/grafana/grafana/pull/79744

dprokop:
> Wonder if we can mark it as internal here in scenes lib?

I think this is a good idea. Let's gather some insights on the usage (thanks @ivanortegaalba for opening the PR!) and decide based on that if deprecation is a way to go.

@torkelo  @ivanortegaalba  do we agree on continuing with the `_UNSAFE_internalPropertyName` naming convention for such things? Ref https://github.com/grafana/scenes/pull/396

ivanortegaalba:
> > Wonder if we can mark it as internal here in scenes lib?
> 
> I think this is a good idea. Let's gather some insights on the usage (thanks @ivanortegaalba for opening the PR!) and decide based on that if deprecation is a way to go.
> 
> @torkelo @ivanortegaalba do we agree on continuing with the `_UNSAFE_internalPropertyName` naming convention for such things? Ref #396

I'm ok with that. I like the way to mark it as "do not use it". I'll apply the changes

packages/scenes/src/core/SceneTimeZoneOverride.tsx
ivanortegaalba: Why should we use the `nowDelay` from the parent only if the timezone is gotten from it?

If we do that, the `nowDelay` will be only inherited from the ancestor if the timezone is not defined and if the ancestor time range changes.

The delay should be always the defined one, although we want to inherit any delay from parents, which implies many more changes, since we need to always get the delay from the ancestor as we do for the timezone, not only in the evaluation of the value when the timezone changes.

Am I missing something @torkelo? 🤔

ivanortegaalba:
@dprokop I renamed the state property from `delayNow` to `UNSAFE_delayNow`. Once you give me the ✅, I'll merge

ivanortegaalba:
> The problem with how it's imlemented now (and here) s that it causes simple time ranges like Last 15 minutes to read like now-30m to now-1m . I think this now delay should not change how intervals are displayed, just how "now" is evaluated.

@torkelo I also fixed the display issue with the time range

torkelo:
@ivanortegaalba great, I think this is ready for merge then!

PR #512, opened 2023-12-20
Moving the grafana monitoring demo app to scenes-app

torkelo:
Thinking of cleaning up (removing) the /scenes folder from the main repo. 

Don't want to delete this demo app as it's kind of useful.


dprokop (approved):
Good idea!

PR #513, opened 2023-12-21
Scene query state & cancel all queries 

torkelo:
So this is just not a new feature but something we need for solo panel & reporting (to know when all queries are completed). 

Was not super trivial, tried a few different ideas. 

* [x] A new query controller concept that everything that issues queries can find and register queries with 
* [x] Refresh picker finds closest query controller and updates it's loading & text state
* [x] Adds a new withText option for refresh picker . Shows "Refresh" when not running or "Cancel" when running. Think this is better than a pure text option as we want this text to be the same in all places? We still have a problem in that internationalization functions are not available to scenes/plugins. 
* [x] Adds primary option to Refresh picker to make it primary (blue) 
* [x] Use queryVaraible 
* [x] Use in annotation query layer

https://github.com/grafana/scenes/assets/10999/adc0a999-ce49-4f68-a873-1f3ebb44862a


ivanortegaalba (commented):
✅ I like the approach, it is simple and keeps complexity away from `SceneQueryRunner`. This is the way to go IMO.

Something that came to my mind is the future usage of query manager as a component overlooking all the queries moving around. For example, it could be used to do performance measurements or error reporting. How could we extend the current API to support those use cases?

ivanortegaalba:
My two cents on the open questions:

> Default withText to true?
I'd say no. It is the real state we can use for anything else. The refresh button is clear enough. However, the cancel is good to have it as text but not required.

>Default primary to true?
No. It looks like an add button because of the position, which adds too much importance to a refresh button as the main action of a dashboard.

> Would be nice to update the Refresh picker tooltip with a number of queries running and what type they are, but the refresh picker does not support making this tooltip dynamic so we would have to re-render RefreshPicker every time a query started/completed. Not the whole world. But the biggest problem is the lack of internationalization here as the default refresh picker tooltips all have translations.

About internationalization, this is something we will have for any text we want to introduce in a scene component. We can't live only with icons. Probably we need to find a way to provide a translation within scenes. Some ideas are on top of my mind. We could use the behavior to get something like `const t = nodeGraph.getTranslator()` to get a scene object in charge of providing a CB to be used by any scene object to process a string. We could pass a callback that receives a string and returns the translated one.

packages/scenes/src/querying/SceneQueryController.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export interface SceneQueryControllerEntry {
+  request?: DataQueryRequest;
+  type: SceneQueryType;
+  origin: SceneObject;
+  cancel?: () => void;
+}
+
+export class SceneQueryController
```
dprokop, line 35: This is just a nit, but can we move this to `behaviors` directoy and expose it under `behaviors` namespace, similar to how we do with other behaviors?

packages/scenes/src/querying/SceneQueryController.ts
```diff
@@ -0,0 +1,68 @@
...
+  extends SceneObjectBase<SceneQueryStateControllerState>
+  implements SceneQueryControllerLike
+{
+  public isQueryController: true = true;
+
+  #running = new Set<SceneQueryControllerEntry>();
+
+  public constructor(state: Partial<SceneQueryStateControllerState>) {
+    super({ isRunning: state.isRunning ?? false });
+  }
```
dprokop, line 45: How about no argument in the constructor? And defaulting to `isRunning: false`? Can't think of how we would use the initial `isRunning: true` state at the moment. Maybe in the future when capturing half-baked state.

packages/scenes/src/querying/SceneQueryController.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  queryStarted(entry: SceneQueryControllerEntry): void;
+  queryCompleted(entry: SceneQueryControllerEntry): void;
+}
+
+export function isQueryController(s: SceneObject | SceneStatelessBehavior): s is SceneQueryControllerLike {
+  return 'isQueryController' in s;
+}
+
+export type SceneQueryType = 'data' | 'annotations' | 'panel' | 'variable' | 'alerts';
```
dprokop, line 22: Isn't `panel` the same as `data`?  Would keep `data` simply

dprokop (approved):
Looks good, just some nits to consider.

packages/scenes/src/querying/SceneQueryController.ts
torkelo: @dprokop yea, was thinking of marking query runners attached to VizPanels specifically as "panel" queries vs say a query on a future repeater object. but lets keep it simple for now and just use "data"

packages/scenes/src/querying/SceneQueryController.ts
torkelo: @dprokop good idea, done!

packages/scenes/src/querying/SceneQueryController.ts
torkelo: yea, good idea.

PR #515, opened 2023-12-21
Fix issue with duplicate annotations

domasx2:
In case annotations come with result of query instead of from layers, new annotations would be appended to list of previous annotations instead of replacing them. Eg, when cahnging time range, annotations from previous time range remain visible. If time ranges overlap, annotations get duplicated, triplicated etc. 

This PR refactors query runner to merge annotations from layers and query result without keeping annotations from previous runs.

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.ts
@@ -59,6 +59,8 @@ export interface QueryRunnerState extends SceneObjectState {
   // Private runtime state
   _isWaitingForVariables?: boolean;
   _hasFetchedData?: boolean;
+  _layerAnnotations?: DataFrame[];
+  _resultAnnotations?: DataFrame[];
 }
 
 export interface DataQueryExtended extends DataQuery {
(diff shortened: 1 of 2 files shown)


torkelo (commented):
Feels like there is a refactoring that can simplify this  (and reduce state updates)

In case the layers complete before the main query we could perhaps postpone the state update, and do it in later in onDataReceived.

torkelo:
I guess that does not work really, the data layer query and local query runner query are issued in sync like that (they could depend on different triggers)

torkelo (approved):
thinking more I think your right in that we need both   _layerAnnotations and _resultAnnotations. 

But not sure we need them in state, could just be private instance props on SceneQueryRunner, but does not really matter in this case I think (could be either). 

would like to wait for @dprokop review as well as he knows the data layer / annotations system best.

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -59,6 +59,8 @@ export interface QueryRunnerState extends SceneObjectState {
   // Private runtime state
   _isWaitingForVariables?: boolean;
   _hasFetchedData?: boolean;
+  _layerAnnotations?: DataFrame[];
+  _resultAnnotations?: DataFrame[];
```
dprokop, line 63: Don't think we need this in the state. The relevant and used downstream state update (`data`) happens as a result of layers/SQR data retrieval, so we can just keep this as a private instance property as @torkelo suggested.

dprokop (approved):
@domasx2 - think this looks good, would only think of removing the layers/results annotations from the state and keeping them as a private instance prorperty.

packages/scenes/src/querying/SceneQueryRunner.ts
domasx2: done!

PR #518, opened 2023-12-28
Adhoc filters variable improvements

javiruiz01:
We're using `AdHocFiltersVariable` and found a couple of places where we can improve the component/variable:
- Add explicit return type to the `getValue` function, right now it's being narrowed to `string | undefined`
- Expose `renderFilters` class method, it's useful for cases where we want to make some transformation before getting the value of the filters
- When selecting a key from the dropdown, what gets shown as selected is the value instead of the label as one would expect

Related to [this PR](https://github.com/grafana/app-observability-plugin/pull/705)

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 
 import { AdHocFilterSet } from './AdHocFiltersSet';
 import { AdHocVariableFilter, GrafanaTheme2, SelectableValue, toOption } from '@grafana/data';
@@ -27,8 +27,21 @@ export function AdHocFilterRenderer({ filter, model }: Props) {
     isValuesOpen?: boolean;
   }>({});
 
-  const keyValue = filter.key !== '' ? toOption(filter.key) : null;
+  const keyValue =
+    filter.key !== '' ? state?.keys?.find((key) => key.value === filter.key) ?? toOption(filter.key) : null;
   const valueValue = filter.value !== '' ? toOption(filter.value) : null;
+
+  useEffect(() => {
+    async function getKeys() {
+      setState({ ...state, isKeysLoading: true });
+      const keys = await model._getKeys(filter.key);
+      setState({ ...state, isKeysLoading: false, keys });
+    }
+
+    if (state.keys === undefined) {
+      getKeys();
+    }
+  }, [filter.key, model, state]);
 
   const valueSelect = (
     <Select
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -61,7 +61,7 @@ export class AdHocFiltersVariable
     });
   }
 
-  public getValue() {
+  public getValue(): VariableValue | undefined {
     return this.state.filterExpression;
   }
 
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { AdHocVariableFilter } from '@grafana/data';
 import { SceneObjectBase } from '../../core/SceneObjectBase';
-import { SceneVariable, SceneVariableState, SceneVariableValueChangedEvent } from '../types';
+import { SceneVariable, SceneVariableState, SceneVariableValueChangedEvent, VariableValue } from '../types';
 import { AdHocFilterSet, AdHocFilterSetState } from './AdHocFiltersSet';
 import { SceneComponentProps } from '../../core/types';
 import { VariableHide } from '@grafana/schema';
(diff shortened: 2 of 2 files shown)


packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
```diff
@@ -27,9 +27,22 @@ export function AdHocFilterRenderer({ filter, model }: Props) {
...
+    async function getKeys() {
+      setState({ ...state, isKeysLoading: true });
+      const keys = await model._getKeys(filter.key);
+      setState({ ...state, isKeysLoading: false, keys });
+    }
+
+    if (state.keys === undefined) {
+      getKeys();
+    }
+  }, [filter.key, model, state]);
```
torkelo, line 44: not sure I understand this change, looks to be dupicating the logic inside onOpenMenu={async () => { on the key select. 

Also this forces the UI to call getKeys to just show the already added filters

torkelo: So when loading a view from url with 4 already defined filters there should be no need to call getKeys

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -42,13 +42,12 @@ export class AdHocFiltersVariable
...
   public constructor(state: AdHocFiltersVariableState) {
-    super({
-      ...state,
-      filterExpression: state.filterExpression ?? renderFilters(state.set.state.filters),
-    });
+    super({ ...state });
 
     // Subscribe to filter changes and up the variable value (filterExpression)
     this.addActivationHandler(() => {
+      this.setState({ filterExpression: state.filterExpression ?? this.renderFilters(state.set.state.filters) });
```
torkelo, line 49: this looks to duplicate logic found a few lines below (inside the activation handler), there is a call to _updateFilterExpression

torkelo:
> Expose renderFilters class method, it's useful for cases where we want to make some transformation before getting the value of the filters

Can you describe what you mean here? are you extending AdHocFiltersVariable ?

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -61,12 +60,25 @@ export class AdHocFiltersVariable
     });
   }
 
-  public getValue() {
+  public getValue(): VariableValue | undefined {
     return this.state.filterExpression;
   }
 
+  public renderFilters(filters: AdHocVariableFilter[]) {
```
torkelo, line 67: why this move? 

If a consumer want's to customize this, might be better to have some way to pass in a render function?

packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
javiruiz01: The issue I found is that, when a filter is selected, we're showing the `value` and not the `label`, so the dropdown and the selected text is actually different and might lead to confusion.

In your example, loading a view from a url with 4 filters _should_ call `getKeys` since we need to show the label, and not the value IMO

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
javiruiz01: Humm so in our case, it was not so much about customising it, but rather re-using it. 
We're extending AdHocFiltersVariable to customise the formatting, but still want to use the by default rendering logic (with all the proper escaping)

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
javiruiz01: 100% right, I removed that line 👌

packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
torkelo: @javiruiz01 not sure it's worth it, for adhoc filters we assume the label/value are the same

torkelo: just don't want a bunch of network calls when loading a page with many adhoc filters defined via URL, definitely not have to call the getKeys to load ALL keys for each filter

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
torkelo: @javiruiz01 how are you customizing the formatting and still re-using it?

torkelo: I don't think this function belong as part of AdHocFiltersVariable, maybe we can export it as a separate util function named "renderPrometheusLabelFilters" (under sceneUtils)

packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
javiruiz01: Humm yeah I understand what you mean, I'll remove the change

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
javiruiz01: That's actually a great idea! I'll move it there 👌

torkelo (approved):
Looks good

PR #519, opened 2024-01-02
SceneQueryRunner: Fix redundant execution on variable change

domasx2:
Noticed some intermittent unexepcted cancelled queries in network logs.
Found that the problem is that if `SceneQueryRunner` has `maxDataPointsFromWidth=true` and variables change/complete before width is received, it will run queries without waiting for width. Then  width is received immediately after and new execution is triggered.

This PR prevents variable update callback from executing queries if container width has not been received yet.


torkelo (approved):
Looks good

PR #524, opened 2024-01-05
SceneQueryRunner: Fix for variable & setContainerWidth timing issue 

torkelo:
So this is the quick fix for the scenario of 

* Variables (A, B), Query (depends on A). 
* A completes right away, but SceneVariableSet is still waiting for B to complete to notify scene
* VizPanel get's the container width and calls setContainerWidth, which starts a new query as the scene query runner no longer depends on on any loading variables.
* B completes and SceneVariableSet completes and notifies scene, SceneQueryRunner issues a new query as SceneVariableSet says variable A has changed. 

Resulting in double query on the initial load. 

Fixing this required some minor changes. But ideally, we should figure out a way to avoid this by having SceneVariableSet notify scene as soon as the variable update is complete. Will give this a try, will require some changes I think to variableDependency interface

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.test.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.test.ts
@@ -648,7 +648,7 @@ describe('SceneQueryRunner', () => {
 
     it('Should not issue query when setContainerWidth is called and we are waiting for variables', async () => {
       const varA = new TestVariable({ name: 'A', value: 'AA', query: 'A.*' });
-      const varB = new TestVariable({ name: 'A', value: 'AA', query: 'A.$A.*' });
+      const varB = new TestVariable({ name: 'B', value: 'AA', query: 'A.$A.*' });
 
       // Query only depends on A
       const queryRunner = new SceneQueryRunner({
--- a/packages/scenes/src/querying/SceneQueryRunner.test.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.test.ts
@@ -8,6 +8,7 @@ import {
   LoadingState,
   PanelData,
   toDataFrame,
+  VariableRefresh,
 } from '@grafana/data';
 
 import { SceneTimeRange } from '../core/SceneTimeRange';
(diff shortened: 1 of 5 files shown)


packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -213,10 +213,7 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
    * the query execution on activate was stopped due to VariableSet still not having processed all variables.
    */
   private onVariableUpdatesCompleted(_variablesThatHaveChanged: Set<SceneVariable>, dependencyChanged: boolean) {
-    // If no maxDataPoints specified we might need to wait for container width to be set from the outside
-    if (!this.state.maxDataPoints && this.state.maxDataPointsFromWidth && !this._containerWidth) {
-      return;
-    }
+    console.log('completed', _variablesThatHaveChanged);
```
domasx2, line 216: `console.log` left

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -298,7 +290,7 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
       if (this.state.maxDataPointsFromWidth && !this.state.maxDataPoints) {
         // As this is called from render path we need to wait for next tick before running queries
         setTimeout(() => {
-          if (this.isActive && !this.state._hasFetchedData) {
+          if (this.isActive && !this.state._hasFetchedData && !this.state._isWaitingForVariables) {
```
domasx2, line 293: I don't get why this works :thinking: isn't `_isWaitingForVariables` only dependent on A, due to `sceneGraph.hasVariableDependencyInLoadingState(this)` only checking individual variables that query depends on? So shouldn't prevent it from running even if B is not resolved

packages/scenes/src/querying/SceneQueryRunner.test.ts
```diff
@@ -645,6 +646,44 @@ describe('SceneQueryRunner', () => {
       expect(runRequestMock.mock.calls.length).toBe(2);
     });
 
+    it('Should not issue query when setContainerWidth is called and we are waiting for variables', async () => {
+      const varA = new TestVariable({ name: 'A', value: 'AA', query: 'A.*' });
+      const varB = new TestVariable({ name: 'A', value: 'AA', query: 'A.$A.*' });
```
domasx2, line 651:
```suggestion
      const varB = new TestVariable({ name: 'B', value: 'AA', query: 'A.$A.*' });
```

packages/scenes/src/querying/SceneQueryRunner.ts
domasx2: maybe it only works because both variables in test were named `A`

torkelo: @domasx2 this PR depends on https://github.com/grafana/scenes/pull/523

domasx2: ahh, didn't notice :+1:

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -229,11 +226,6 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
   }
 
   private shouldRunQueriesOnActivate() {
-    // If no maxDataPoints specified we might need to wait for container width to be set from the outside
-    if (!this.state.maxDataPoints && this.state.maxDataPointsFromWidth && !this._containerWidth) {
-      return false;
-    }
```
dprokop, line 235: Why's this check now moved to the `runWithTimeRange`?

packages/scenes/src/querying/SceneQueryRunner.test.ts
```diff
@@ -645,6 +646,44 @@ describe('SceneQueryRunner', () => {
...
+
+      const scene = new TestScene({
+        $variables: new SceneVariableSet({ variables: [varA, varB] }),
+        $timeRange: new SceneTimeRange(),
+        $data: queryRunner,
+      });
+
+      scene.activate();
+
+      // should execute query when variable completes update
```
dprokop, line 667:
```suggestion
      // should not execute query when variable completes update
```

dprokop (approved):
Looks good, but lest one question about a change.

packages/scenes/src/querying/SceneQueryRunner.ts
torkelo: @dprokop because it was duplicated here and in  onVariableUpdatesCompleted, so figured it's better to move it to runWithTimeRange and we can have this condition only once

dprokop: ah, makes sense, didn't notice the duplication

dprokop:
@torkelo - should we close this in favor of https://github.com/grafana/scenes/pull/525?

torkelo:
@dprokop yes !

PR #526, opened 2024-01-05
Variables: Move handling of "waitingForVariables" to VariableDependencyConfig 

torkelo:
a continuation of https://github.com/grafana/scenes/pull/525  (base) 

Exploration unifying / moving handling of "waitingForVariable" state to VariableDependencyConfig. 

This flag / state is just to know that we wanted to do something but could not, for example: 

* Issue query on activation 
* Issue query on time range change 

But we can't do that as one of the variable dependencies (or a dependency of a dependency) where in a loading state. But this means that we now want to know when any variable update is completed not just a direct dependency as just because a dependency of a dependency where in loading state does not mean any dependency changed value (just that they could have). 

TODO
* Add some more tests to VariableDependencyConfig

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.ts
@@ -351,11 +351,6 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
       return;
     }
 
-    // If we were waiting for variables, clear that flag
-    // if (this.state._isWaitingForVariables) {
-    //   this.setState({ _isWaitingForVariables: false });
-    // }
-
     const { queries } = this.state;
 
     // Simple path when no queries exist
--- a/packages/scenes/src/querying/SceneQueryRunner.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.ts
@@ -220,10 +220,8 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
    * only run if all variables are in a non loading state so in other scenarios where a query depends on many variables this will
    * be called many times until all dependencies are in a non loading state.   *
    */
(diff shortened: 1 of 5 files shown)


packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -349,17 +346,11 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
     this._querySub?.unsubscribe();
 
     // Skip executing queries if variable dependency is in loading state
-    if (sceneGraph.hasVariableDependencyInLoadingState(this)) {
+    if (this._variableDependency.hasDependencyInLoadingState()) {
       writeSceneLog('SceneQueryRunner', 'Variable dependency is in loading state, skipping query execution');
-      this.setState({ _isWaitingForVariables: true });
       return;
     }
```
torkelo, line 352: I really do not like that this call to `this._variableDependency.hasDependencyInLoadingState` is the thing that causes the 
VariableDependencyConfig to set it's internal _isWaitingForVariables state, which changes when it calls onVariableUpdateCompleted. 

But the alternative is requires so much more duplicated logic.

```ts
 if (this._variableDependency.hasDependencyInLoadingState()) {
      this._variableDependency.setIsWaitingForVariables(true);
      writeSceneLog('SceneQueryRunner', 'Variable dependency is in loading state, skipping query execution');
      return;
    } else {
      this._variableDependency.setIsWaitingForVariables(false);
    }
```

So this logic needs to be completed pretty much verbatim everywhere this scenario comes up. This is why I am going with hiding this logic inside hasDependencyInLoadingState and VariableDependencyConfig. But it feels wrong to have a function named hasDependencyInLoadingState to have a side effect

torkelo:
@dprokop merged this into https://github.com/grafana/scenes/pull/525 and added some docs

PR #529, opened 2024-01-15
UrlSync: Export new util functions 

torkelo:
for the EmbeddedDashboard use case, I need some new lower-level url utils, enables:

* Sync scene state from search params (that could come from a string that is not the actual current url)
* Get url state (already existed but now exported as a separate util, as this should not really be done via UrlSyncManager as that is a singleton and only deals with one scene at a time).


packages/scenes/src/services/UrlSyncManager.test.ts
```diff
@@ -103,6 +117,42 @@ describe('UrlSyncManager', () => {
...
+
+    //   urlManager = new UrlSyncManager();
+    //   urlManager.initSync(scene);
+
+    //   deactivate = scene.activate();
+
+    //   scene.setState({ children: [new SceneFlexItem({ body: obj })] });
+
+    //   expect(obj.state.name).toEqual('name-from-url');
+    // });
```
torkelo, line 153: this is a scenario that we do not support currently but I want to support, it's a limitation right now, but needs a bit refactor

ivanortegaalba (approved):
Very useful, I like the API and naming. Very intuitive! Do we usually document these utils in Scenes docs?

dprokop (approved):

PR #546, opened 2024-01-25
Make secondary part optional in Splitter

cbos:
With this PR the secondary part is optional.
If it is put as `undefined`, it will be removed. The splitter / handle will not be visible anymore.

I adjusted the example.

Here is a short demo on how it works with both situations
https://github.com/grafana/scenes/assets/255404/a39f59d7-a342-4e08-93cd-c3f1f1865041

### DIFF
--- a/packages/scenes-app/src/demos/split.tsx
+++ b/packages/scenes-app/src/demos/split.tsx
@@ -147,11 +147,20 @@ const getDynamicSplitScene = () => {
                       })
                     )
                     .setHeaderActions(
-                      <IconButton
-                        name="x"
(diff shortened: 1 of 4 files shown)


cbos:
@kaydelaney 
If any changes to required on the modifications done, please let me know, then I can change the code in the PR.

dprokop:
@cbos  - sorry for a long wait. I've pushed a cosmetic change so that the resize handle and secondary panes are not rendered when there is no secondary pane at all.

dprokop (approved):

cbos:
@dprokop 
Thanks for finishing the pull request!

PR #548, opened 2024-01-25
Introduce GroupBy variable to allow passing group by dimensions to data sources

ashharrison90:
TODO:
- [x] Provide selected group by dimensions to DataQueryRequest in SceneQueryRunner (similar to how filters are provided via subscription). This requires updating the type in core Grafana to accept groupBy dimensions on the `DataQueryRequest`.
- [x] Get AggregationsSetUrlSync in place
- [x] Implement tests
- [x] Pass scene queries to `getTagKeys` calls (see https://github.com/grafana/scenes/pull/544)
- [x] Implement AdHocAggredationVariable


packages/scenes/src/variables/groupby/AggregationsSet.tsx
```diff
@@ -0,0 +1,182 @@
...
+        tabSelectsValue={false}
+        loadOptions={model._getKeys}
+        closeMenuOnSelect={false}
+        isClearable
+        // onInputChange={onInputChange}
+        onBlur={() => {
+          model._update(uncommittedValue);
+        }}
+        // TODO refactor to remove defaultOptions
+        defaultOptions
```
dprokop, line 158: What's this about?

ashharrison90: @dprokop setting this retrieves the options on mount instead of when opening the menu.

we don't want that, but i haven't had time yet to dig into why it doesn't work without this, and i figured having options showing is more important for now 😅

i'll add an extra comment to explain 👍

dprokop: OK, makes sense

dprokop:
@ashharrison90  - I've added some todos to make this mergable.

ashharrison90:
@dprokop did you forget to apply the TODOs? i don't see them 😅

packages/scenes/src/variables/groupby/AggregationsSet.tsx
```diff
@@ -0,0 +1,185 @@
...
+import React, { useEffect, useState } from 'react';
+import { ControlsLabel } from '../../utils/ControlsLabel';
+import { css } from '@emotion/css';
+
+export interface AggregationsSetState extends SceneObjectState {
+  /** Defaults to "Group" */
+  name?: string;
+  /** The visible keys to group on */
+  // TODO review this type
+  aggregations: string[];
```
dprokop, line 17: Think we can rename this to `dimension` and use `MetricFindValue` here.

packages/scenes-app/src/demos/groupByDemo.tsx
```diff
@@ -0,0 +1,86 @@
...
+                  text: 'baz',
+                  value: 'baz',
+                },
+              ],
+            }),
+            new GroupByVariable({
+              name: 'groupBy',
+              label: 'Group By (from data source)',
+              datasource: { uid: 'gdev-prometheus' },
+            }),
```
dprokop, line 48: We need to separate to two demos, as otherwise SceneQueryRunner will always take the first variable as the source of dimensions

dprokop (approved):

torkelo (approved):
Looks good

PR #549, opened 2024-01-26
LocalValueVariable: Provide backwards compatibility for SQL-ish data source

dprokop:
Provides a fix for SQL data source queries used in a repeated row scenario (discovered when dogfooding dashboard scene for viewers). The root cause problem lies in the `SqlDatasource` implementation which [relies on variable.multi and vairbale.includeAll ](https://github.com/grafana/grafana/blob/main/packages/grafana-sql/src/datasource/SqlDatasource.ts#L80) to format the string correctly. 

This PR brings two  properties to the `LocalValueVariable`. Marking them as ommited in LocalValueVariabel for a reason being LocalValueVariable is used ONLY in the repeats scenatio which assume a variable being a multi one.

@zoltanbedi  I would love to fix this in `SqlDatasource` as the dependency there on the multi/include all sounds not solid - think about a single value variable that has include all option enabled, but not selected. How does formatted the value differs in such a scenario compared to selecting the same option selected from a variable without include all enabled? In the firs scenario the value will be wrapped in `''` while in the other it won't.

### DIFF
--- a/packages/scenes/src/variables/interpolation/sceneInterpolator.ts
+++ b/packages/scenes/src/variables/interpolation/sceneInterpolator.ts
@@ -9,7 +9,6 @@ import { formatRegistry, FormatVariable } from './formatRegistry';
 import { VARIABLE_REGEX } from '../constants';
 import { lookupVariable } from '../lookupVariable';
 import { macrosIndex } from '../macros';
-import { LocalValueVariable } from '../variants/LocalValueVariable';
 
 /**
  * This function will try to parse and replace any variable expression found in the target string. The sceneObject will be used as the source of variables. It will
@@ -91,8 +90,8 @@ function formatValue(
     return formatNameOrFn(value, {
       name: variable.state.name,
       type: variable.state.type as VariableType,
-      multi: variable instanceof LocalValueVariable ? variable.state.UNSAFE_isMulti : variable.state.isMulti,
-      includeAll: variable instanceof LocalValueVariable ? variable.state.UNSAFE_includeAll : variable.state.includeAll,
+      multi: variable.state.isMulti,
+      includeAll: variable.state.includeAll,
     });
   }
 
--- a/packages/scenes/src/variables/interpolation/sceneInterpolator.ts
+++ b/packages/scenes/src/variables/interpolation/sceneInterpolator.ts
@@ -9,6 +9,7 @@ import { formatRegistry, FormatVariable } from './formatRegistry';
 import { VARIABLE_REGEX } from '../constants';
 import { lookupVariable } from '../lookupVariable';
 import { macrosIndex } from '../macros';
+import { LocalValueVariable } from '../variants/LocalValueVariable';
 
 /**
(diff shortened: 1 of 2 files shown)


packages/scenes/src/variables/interpolation/sceneInterpolator.ts
```diff
@@ -90,8 +91,8 @@ function formatValue(
     return formatNameOrFn(value, {
       name: variable.state.name,
       type: variable.state.type as VariableType,
-      multi: variable.state.isMulti,
-      includeAll: variable.state.includeAll,
+      multi: variable instanceof LocalValueVariable ? variable.state.UNSAFE_isMulti : variable.state.isMulti,
+      includeAll: variable instanceof LocalValueVariable ? variable.state.UNSAFE_includeAll : variable.state.includeAll,
```
torkelo, line 95: not sure about this, it kind of breaks the logic of variables with multi should always have array values. 

Can't we fix this inside the SQL data source instead?

torkelo: hm.. tricky 

from SqlDataSource

```ts
 interpolateVariable = (value: string | string[] | number, variable: VariableWithMultiSupport) => {
    if (typeof value === 'string') {
      if (variable.multi || variable.includeAll) {
        return this.getQueryModel().quoteLiteral(value);
      } else {
        return String(value).replace(/'/g, "''");
      }
    }

    if (typeof value === 'number') {
      return value;
    }

    if (Array.isArray(value)) {
      const quotedValues = value.map((v) => this.getQueryModel().quoteLiteral(v));
      return quotedValues.join(',');
    }

    return value;
  };
```

so values are only surrounded in quotes when the value is an array, or when a string when multi || includeAll. 

and when not multi/includeAll it assumes the variable is inside a single quoted string so escapes single quotes. 

Can't think of a way to fix this, I kind of understand why it is like this. If was differently you would have to change your queries when going from single panel with multi value variable to a repeated panel by said variable (which turn array values into single values)

dprokop: I'd love to fix it in SQL data source, but the abstraction is used by several data sources and I don't want to touch that untested and IMO sort of fragile code. I would rather merge this PR to stop the bleeding and continue with the feature toggle, and ask team reposible for SQL ds to try to address the problem.

dprokop: > Can't think of a way to fix this, I kind of understand why it is like this. If was differently you would have to change your queries when going from single panel with multi value variable to a repeated panel by said variable (which turn array values into single values)

Exactly. I think this code can be imporved somehow maybe, but until it's changed we need a backwards compatible solution. Plus - I'm not sure if there isn't data sources that rely on variable multi/includeAll properties anyways.

torkelo (approved):

zoltanbedi:
@dprokop are you going to fix this or should we?

PR #550, opened 2024-01-26
AdHocFiltersVariable: AdHocFilterSet is not activated when the parent variable is mounted

ivanortegaalba:
**Problem**
When the `AdHocFilterSet` is activated, this registers the filters into a global object to be findable by the SceneQueryRunner.

When rendering the components for the `AdHocFiltersVariable` because we were not using the component from the model, the `AdHocFilterSet` scene object was not activated, so they were not findable by the SceneQueryRunner.

**Solution**
Use the component from the scene object instead of importing and using the static class property.

_This was found while implementing: https://github.com/grafana/grafana/pull/81318_ 

<details>
  <summary>📦 Published PR as canary version: <code>2.2.1--canary.550.7712757823.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@2.2.1--canary.550.7712757823.0
  # or 
  yarn add @grafana/scenes@2.2.1--canary.550.7712757823.0
  ```
</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
   "devDependencies": {
     "@auto-it/omit-commits": "^10.43.0",
     "@auto-it/released": "^10.43.0",
+    "@testing-library/react": "^14.1.2",
     "auto": "^10.43.0",
     "lerna": "^6.5.1",
     "lint-staged": "^13.2.0"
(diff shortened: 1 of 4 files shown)


dprokop (approved):

dprokop (commented):
@ivanortegaalba - adding a test to make sure the internal set is activated would be a good idea here I think if anyone ever thinks about reverting this change ;)

ivanortegaalba:
@dprokop I added RLT and the test

dprokop (approved):

PR #553, opened 2024-01-30
AdHocFiltersVariable: The `create()` factory supports `applyMode` as parameter

ivanortegaalba:
**Problem**
When creating a variable using `AdHocFiltersVariable.create()` the `applyMode` state prop was not configurable. This was forcing to use AdHocFiltersVariable in `manual` only. But in the Dashboard use case reflected here https://github.com/grafana/grafana/pull/81318 where we use AdHocFiltersVariable to manage filters for dashboard panels, we need `applyMode: same-dashboard`

**Solution**
Accept `applyMode` as the initial state but default to `manual` if it is not defined. This will avoid breaking changes.

<details>
  <summary>📦 Published PR as canary version: <code>2.2.1--canary.553.7712958297.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@2.2.1--canary.553.7712958297.0
  # or 
  yarn add @grafana/scenes@2.2.1--canary.553.7712958297.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -30,6 +30,7 @@ export type AdHocFiltersVariableCreateHelperArgs = Pick<
   | 'getTagValuesProvider'
   | 'name'
   | 'layout'
+  | 'applyMode'
 >;
 
 export class AdHocFiltersVariable
@@ -43,9 +44,9 @@ export class AdHocFiltersVariable
       hide: VariableHide.hideLabel,
       name: state.name ?? 'Filters',
       set: new AdHocFilterSet({
-        ...state,
-        // Main reason for this helper factory functyion
+        // The applyMode defaults to 'manual' when used through the variable as it is the most frecuent use case
         applyMode: 'manual',
+        ...state,
       }),
     });
   }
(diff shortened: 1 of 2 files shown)


packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -43,9 +44,9 @@ export class AdHocFiltersVariable
       hide: VariableHide.hideLabel,
       name: state.name ?? 'Filters',
       set: new AdHocFilterSet({
-        ...state,
-        // Main reason for this helper factory functyion
+        // The applyMode defaults to 'manual' when used through the variable as it is the most frecuent use case
         applyMode: 'manual',
```
torkelo, line 48: What is the default applyMode ? Won't this change the default ?

ivanortegaalba: If I'm not wrong, the `create()` was intended to default to `manual`, but when creating AdHocFilterSet through the constructor, the default is `same-datasource`. 

I didn't want to provoke a breaking change, since `create` by default was creating the variable object with manual by default, my change is only about consider `applyMode` as part of the accepted state.

Do we want to keep `same-datasource` as default, no matter if the AdHocFiltersVariable is created through the factory or through the constructor?

dprokop: @ivanortegaalba  - i think you are right, let's keep `manual` as default and allow overriding as you did in this PR.

dprokop (approved):

PR #554, opened 2024-01-30
Fix dataLayer subscriptions refresh on reactivation

mdvictor:
Fixes issues with reactivating annotation dataLayers, where the `SceneQueryRunner` subs would get stale and needed to be refreshed.

`SceneDataLayerBase - querySub` also needs to be reinitialised on dataLayer reactivation.

<details>
  <summary>📦 Published PR as canary version: <code>2.2.2--canary.554.7713830844.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@2.2.2--canary.554.7713830844.0
  # or 
  yarn add @grafana/scenes@2.2.2--canary.554.7713830844.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.ts
@@ -119,6 +119,10 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
       this.runQueries();
     }
 
+    if (!this._dataLayersSub) {
+      this._handleDataLayers();
+    }
+
     return () => this._onDeactivate();
   }
 
--- a/packages/scenes/src/querying/layers/SceneDataLayerBase.ts
+++ b/packages/scenes/src/querying/layers/SceneDataLayerBase.ts
@@ -172,6 +172,10 @@ export abstract class SceneDataLayerBase<T extends SceneDataLayerProviderState>
       return true;
     }
 
+    if (!this.querySub) {
+      return true;
+    }
+
     if (this.state.data) {
       return false;
     }


packages/scenes/src/querying/layers/SceneDataLayerBase.ts
```diff
@@ -172,6 +172,10 @@ export abstract class SceneDataLayerBase<T extends SceneDataLayerProviderState>
       return true;
     }
 
+    if (!this.querySub) {
```
mdvictor, line 175: @dprokop I know we talked about modifying `subscribeToState` to fix the anno toggle not working after discarding modifications, but that modification resulted in all annotations dissapearing from all panels.

I am not sure why but I reckon after discarding and reactivating all annotation layers with no `querySub` defined, when the dashboard repaints there are no queries to pull data from to show annotations, maybe?

This check in `shouldRunLayerOnActivate` makes sure that, on activation, a query will be set. But then I wonder about the next statement checking if `state.data` is set. I think that would be populated by using that query, or with empty data... 

Looking for more feedback on this as I am not entirely sure.

dprokop: >@dprokop I know we talked about modifying subscribeToState to fix the anno toggle 

Thinkning about it and I don't think that would work actually, as the subscription needs to be first set up via the activation handler. I think what you are doing here makes sense.

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -119,6 +119,10 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
       this.runQueries();
     }
 
+    if (!this._dataLayersSub) {
+      this._handleDataLayers();
+    }
```
dprokop, line 124: Ok, so this creates data layer subscription independently from the regular queries execution flow. Looks good to me  too.

dprokop (approved):

PR #555, opened 2024-01-31
UrlSync: Fixes overwrite issue where later state change overwrites earlier change

torkelo:
Finally after being unable to replicate this url overwrite scenario in a unit test I managed to figure out how to replicate it. 

It's when you subscribe to a scene object's state (a scene object that has url sync), and in that subscription you change another object's state (which also has url sync). 

So what was happening was this 

Two scene objects: A (name=A)  B (other=B).   both with url sync for properties name & other respectively 

* Change state of Object A (name=update) -> 
* Triggers state change handler =>
* State change handler changes state of Object B (other=muu)  =>
* UrlSyncManager updates state url with (other=muu) => 
* Url location change triggers sync from URL for Object A as the url as state name=A and state is `name=updated`, so the old value is used to overwrite the new state update 

The fix was to never let a location change trigger a URL => state sync. So that UrlSyncManager will ignore the location update it triggers itself.

Fixes a bug in data trails. 

<details>
  <summary>📦 Published PR as canary version: <code>2.2.3--canary.555.7728272785.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@2.2.3--canary.555.7728272785.0
  # or 
  yarn add @grafana/scenes@2.2.3--canary.555.7728272785.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/services/UrlSyncManager.test.ts
+++ b/packages/scenes/src/services/UrlSyncManager.test.ts
@@ -9,6 +9,7 @@ import { SceneObjectState, SceneObjectUrlValues } from '../core/types';
 
 import { SceneObjectUrlSyncConfig } from './SceneObjectUrlSyncConfig';
 import { UrlSyncManager } from './UrlSyncManager';
+import { activateFullSceneTree } from '../utils/test/activateFullSceneTree';
 
 interface TestObjectState extends SceneObjectState {
   name: string;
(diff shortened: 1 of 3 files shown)


torkelo:
test this with data trails and it fixes the breakdown add filters action (it sets the adhoc filter state and changes the group by variable, and that overwrote the adhoc filter change)

darrenjaneczek (approved):
Interesting URL bug.

Verified that this fixes the following:
> test this with data trails and it fixes the breakdown add filters action

PR #557, opened 2024-02-01
SceneVariableSet: Cancel query when dependency changes

### LINKED ISSUE #556: SceneVariableSet: Automatically cancel variable queries that running when a dependency changes value
Say you have variable A, and B (B depends on A). If B is very slow and loading and you change A it should cancel B and add it to update queue. 

I think what is happening now is that since B is already running we wont start a new "update" as it will be removed from update queue when B completes.

torkelo:
Fixes #556 

Say you have variable A, and B (B depends on A). If B is very slow and loading and you change A it should cancel B and add it to update queue.


axelavargas (approved):
The code looks good to me 🙌🏾

mdvictor (approved):
LGTM

PR #558, opened 2024-02-01
POC: Group by variable

dprokop:
Exploration of an alternative approach to https://github.com/grafana/scenes/pull/548. 

This brings aggreagations as a Scene Variable. Seems to be working fine, just had to do minor changes to the multi value variable interface in order allow clearing the value.

Todo:
- [ ] Demos update to show both static and dynamic version
- [x] Add variable tests

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.ts
@@ -38,8 +38,8 @@ import { filterAnnotations } from './layers/annotations/filterAnnotations';
 import { getEnrichedDataRequest } from './getEnrichedDataRequest';
 import { AdHocFilterSet } from '../variables/adhoc/AdHocFiltersSet';
 import { findActiveAdHocFilterSetByUid } from '../variables/adhoc/patchGetAdhocFilters';
-import { findActiveAggregationsSetVariablesByUid } from '../variables/groupby/findActiveAggregationsSetByUid';
-import { AggregationsSetVariable } from '../variables/groupby/AggregationsSetVariable';
+import { findActiveGroupByVariablesByUid } from '../variables/groupby/findActiveGroupByVariablesByUid';
+import { GroupByVariable } from '../variables/groupby/GroupByVariable';
 
 let counter = 100;
 
@@ -79,9 +79,9 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
   private _adhocFilterSet?: AdHocFilterSet;
   private _adhocFilterSub?: Unsubscribable;
 
-  // Closest aggregations set if found
-  private _adhocAggregationsSet?: AggregationsSetVariable;
-  private _adhocAggregationsSub?: Unsubscribable;
+  // Closest group by variable if found
+  private _groupBySource?: GroupByVariable;
+  private _groupBySourceSub?: Unsubscribable;
 
   public getResultsStream() {
     return this._results;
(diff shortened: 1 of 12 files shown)


packages/scenes/src/variables/components/VariableValueSelect.tsx
```diff
@@ -74,7 +74,10 @@ export function VariableValueSelectMulti({ model }: SceneComponentProps<MultiVal
       onBlur={() => {
         model.changeValueTo(uncommittedValue);
       }}
-      onChange={(newValue) => {
+      onChange={(newValue, action) => {
+        if (action.action === 'clear' && noValueOnClear) {
+          model.changeValueTo([]);
+        }
```
dprokop, line 80: @torkelo  @ash this piece of code enables clearing the value of a variable. But, to preserve the current behavior of other multi value variables, it requries an explicit config of the variable `noValueOnClear`

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -442,7 +442,7 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
 
     if (this._adhocAggregationsSet) {
       // @ts-ignore (Temporary ignore until we update @grafana/data)
-      request.aggregations = this._adhocAggregationsSet.state.dimensions;
+      request.aggregations = this._adhocAggregationsSet.state.value;
```
torkelo, line 445: this is a a confusing name to me, I thought it was request.groupBy or request.groupByDimensions ?  

aggregations to me imply aggregation functions, not dimensions. And if it did imply dimensions it is not clear if the dimensions should be excluded from aggregation (by) or included (without).

torkelo (commented):
Looks good. Think skipping with the variable / set separation is the right thing, want to try to refactor and do the same to AdHocFilterSet. 

the only thing that does not feel 100% is the naming of the variable as AggregationsSetVariable. 

AggregationsSet feels like it would a set of aggregation functions (min, max, avg). 

Would not GroupByVariable make more sense? The request property we send to the data source was named groupBy in the design doc i I recall. I think naming this aggregations is very confusing.

packages/scenes/src/variables/components/VariableValueSelect.tsx
torkelo: looks good

ashharrison90 (approved):
think this is good, we can iterate 🤓

packages/scenes/src/querying/SceneQueryRunner.ts
dprokop: @torkelo  - just POC, we have not landed the required change in the core repo yet. I agree with you, `groupByDimensions` is probably the closes tho what it should be. I'll update as soon as we land this in core.

dprokop:
> do we need to preserve the old behaviour, or could we just have this every time? 🤔

For now let's preserve, mainly fo backwards compatibility needs. We need to figure out empty state for variables as an independent thing.

packages/scenes/src/querying/SceneQueryRunner.ts
torkelo: @dprokop maybe we use groupByKeys instead to use the same word as in getTagKeys

torkelo (approved):

packages/scenes/src/variables/groupby/AggregationsSet.tsx
```diff
@@ -1,6 +1,6 @@
 import { SceneObjectBase } from '../../core/SceneObjectBase';
 import { AdHocVariableFilter, GrafanaTheme2, MetricFindValue, SelectableValue } from '@grafana/data';
-import { allActiveAggregationsSets } from './findActiveAggregationsSetByUid';
+import { allActiveAggregationsSets } from './findActiveGroupByVariablesByUid';
```
ashharrison90, line 3: tiny bit of renaming still to do here, but we can do that in the other branch once this is merged in 👍

ashharrison90 (approved):
lgtm, let's get this merged into the other branch so we can tidy everything up 😄

PR #560, opened 2024-02-01
Fix empty state adhoc filters

javiruiz01:
Problem:
- When having `AdHocFilterSet` in two different views, they sync properly except when removing all filters in one of the views.

Fix: 
- Set filters to empty array when no url value present
- Update filter expression should publish event since filters might've changed from url

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -67,7 +67,7 @@ export class AdHocFiltersVariable
         })
       );
 
-      this._updateFilterExpression(this.state.set.state.filters, false);
+      this._updateFilterExpression(this.state.set.state.filters, true);
     });
   }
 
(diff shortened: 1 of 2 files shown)


packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -67,7 +67,7 @@ export class AdHocFiltersVariable
         })
       );
 
-      this._updateFilterExpression(this.state.set.state.filters, false);
+      this._updateFilterExpression(this.state.set.state.filters, true);
```
torkelo, line 70: this this change needed?

javiruiz01: Turns out it isn't, removed it 👌

packages/scenes/src/variables/adhoc/AdHocFiltersVariableUrlSyncHandler.ts
```diff
@@ -14,7 +14,12 @@ export class AdHocFiltersVariableUrlSyncHandler implements SceneObjectUrlSyncHan
   }
 
   public getUrlState(): SceneObjectUrlValues {
-    let filters = this._variable.state.filters;
+    const filters = this._variable.state.filters;
+
+    if (filters.length === 0) {
+      return { [this.getKey()]: [''] };
+    }
```
torkelo, line 21: no updated or new tests needed?

torkelo (approved):
only one comment about tests, otherwise this should be ok

packages/scenes/src/variables/adhoc/AdHocFiltersVariableUrlSyncHandler.ts
javiruiz01: Working on it 👌

Issue #559, opened 2024-02-01
Preserving selected tab after clicking off the page


xforman2:
I have a an App with more pages in the left navigation menu. After clicking off the page and returning back, the first tab is always selected.
[Screencast from 2024-02-01 12-25-14.webm](https://github.com/grafana/scenes/assets/129223585/d7c3c368-6ec1-4ef4-99e4-6c8f0ab45b09)

Is there a way to preserve the last selected tab ?

PR #567, opened 2024-02-05
implemented preserving tabs logic

xforman2:
So after discovering #559, I thought it would be nice to have a option to preserve the tabs. 
I came up with an idea.

Now the SceneAppPageRenderer always renders the first tab with two urls: the parent page url and the actual tab url.
But if we can remember the last selected tab, we can match the parent page url to any tab. This is what `_lastTabIndex` is for.
And you also have an option to choose not preserving tabs with `preserveTab` property.

Let me know what you think.


torkelo (commented):
Not sure this makes sense, tabs have urls. so if you want to change the default route/tab for page you will need to do that through some redirect.

torkelo:
that seems a bit wrong that different pages are rendered for the same url

xforman2:
I agree that it is not good to have different pages render with the same url, so I tried what you suggested in the first answer. What do you think?

xforman2:
@torkelo I am not sure if you looked at my other implementation but it would be really helpful if this could get resolved

torkelo:
I have, the solution looks a bit complex / messy. Should be a simpler way to do this

PR #572, opened 2024-02-06
SceneTimeRange: Support timezone url sync

torkelo:
Embedded (Solo) page and report page needs to support url timezone sync so instead of doing that via some custom init code I think adding it to SceneTimeRange could make sense. 

It does mean that when users in dashboards (or scene apps) change to a custom / non system/user default timezone that it will be added to URL and included in shared urls. Not sure if this is good or bad. Alt implementation is to only support a one way sync (only check it in updateFromUrl but never return it in getUrlState).


dprokop (approved):
I think this makes sense. I can see no harm of including the tz in the URL

ivanortegaalba:
There is an issue about it in Dashboards repo: https://github.com/grafana/grafana/issues/81538 

They won't see this until we enable the Scenes toggle, but mentioning it here to track it easily.

PR #575, opened 2024-02-07
query controller wrapps observable

torkelo:
Had a timing issue when I wrote a test for scene query runner and it's registration with the query controller. 

SceneQueryRunner called queryStarted after the subscription was created, and if the query completed right away that happens before queryStarted. 

So wanted to try to make QueryController wrap the request observable and that way it can cancel and know state of query automatically, but not so sure this is a good idea, makes QueryController a bit more complex and requires that all queries we register with it are observables that it can wrap

### DIFF
--- a/packages/scenes/src/querying/SceneQueryController.ts
+++ b/packages/scenes/src/querying/SceneQueryController.ts
@@ -1,6 +1,8 @@
 import { SceneObjectBase } from '../core/SceneObjectBase';
 import { SceneObject, SceneObjectState, SceneStatelessBehavior } from '../core/types';
 import { DataQueryRequest } from '@grafana/data';
+import { Observable } from 'rxjs';
+import { LoadingState } from '@grafana/schema';
 
 export interface SceneQueryStateControllerState extends SceneObjectState {
   isRunning: boolean;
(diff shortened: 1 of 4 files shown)


packages/scenes/src/querying/SceneQueryController.ts
```diff
@@ -19,11 +25,16 @@ export function isQueryController(s: SceneObject | SceneStatelessBehavior): s is
...
+  state: LoadingState;
+}
+
+export interface SceneQueryControllerEntry<T extends QueryResultWithState = QueryResultWithState> {
   request?: DataQueryRequest;
   type: SceneQueryType;
-  source?: SceneObject;
-  cancel: () => void;
+  sceneObject?: SceneObject;
+  runStream: Observable<T>;
```
dprokop, line 36: nit: `sourceStream` or `cancellableStream` maybe? `runStream` sound like a method.

dprokop (approved):

packages/scenes/src/querying/SceneQueryController.ts
torkelo: yea, good point. I struggled naming this. Maybe just

* queryStream
* queryObservable
* stream

dprokop: `stream` would be just fine i think

dprokop (approved):

PR #580, opened 2024-02-08
AdHoc filter: Don't render label twice when used as a variable

dprokop:
A hacky way to make sure filter's control label ain't rendered twice (see https://github.com/grafana/grafana/pull/81318#issuecomment-1932233163). 

This is until we get rid of `AdhocFilterSet` and replace it with only a variable as we did with [group by variable](https://github.com/grafana/scenes/pull/548).

<details>
  <summary>📦 Published PR as canary version: <code>2.6.5--canary.580.7828999283.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@2.6.5--canary.580.7828999283.0
  # or 
  yarn add @grafana/scenes@2.6.5--canary.580.7828999283.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersSet.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersSet.tsx
@@ -13,6 +13,7 @@ import { AdHocFilterBuilder } from './AdHocFilterBuilder';
 import { css } from '@emotion/css';
 import { sceneGraph } from '../../core/sceneGraph';
 import { DataQueryExtended, SceneQueryRunner } from '../../querying/SceneQueryRunner';
+import { AdHocFiltersVariable } from './AdHocFiltersVariable';
 
 export interface AdHocFilterSetState extends SceneObjectState {
   /** Defaults to Filters */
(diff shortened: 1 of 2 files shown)


packages/scenes/src/variables/adhoc/AdHocFiltersSet.tsx
```diff
@@ -227,10 +228,11 @@ export class AdHocFilterSet extends SceneObjectBase<AdHocFilterSetState> {
 export function AdHocFiltersSetRenderer({ model }: SceneComponentProps<AdHocFilterSet>) {
   const { filters, readOnly, layout, name } = model.useState();
   const styles = useStyles2(getStyles);
+  let skipLabel = model.parent instanceof AdHocFiltersVariable;
 
   return (
     <div className={styles.wrapper}>
-      {layout !== 'vertical' && <ControlsLabel label={name ?? 'Filters'} icon="filter" />}
+      {layout !== 'vertical' && !skipLabel && <ControlsLabel label={name ?? 'Filters'} icon="filter" />}
```
torkelo, line 235: @dprokop With this change there is no label. 

The AdHocFiltersVariable.create always sets the variable label to hidden, so now both are hidden. 

Hm... 

Screenshot 2024-02-08 at 12 40 44

dprokop: yes.. But that's how it always been :)  We need to add that to the factory I think in order to allow the initial state to be changed. Otherwise one can only change it programatically (which happens when the var is edited).

dprokop:
@torkelo  - actually I'm closing this one, talked with @ivanortegaalba and I think we can just disable the label config in core for the adhoc variable and rely on the adhoc filter set built-in label. Maybe not the best, but at least we can avoid this hacking.

PR #585, opened 2024-02-09
AdHocFiltersSet: Add `hide` property to control controls visibility

ivanortegaalba:
**Problem**
When the AdHocFiltersSet is used as part of AdHocFiltersVariable, the label is rendered twice.

VariableValueSelectors is in charge of rendering the controls, including label and value control. But AdHocFiltersVariable is different since the AdHocFiltersSet can be rendered out of the variable, and it contains its own label and control rendering.

**Solution**
In the same way, every variables has a `hide` property to control the visibility, the AdHocFiltersSet should have its own, so we can decide what part of the control we want to render. This allows when using a AdHocFiltersSet as apart of AdHocFiltersVariable state, hiding the controls to not render them twice.

**Notes**
* I also updated the AdHocFiltersVariable.create to create the filter set as hidden by default but get the hide value from the initial state.

**TODO:**
- [ ] Add tests

Related to #580
Part of https://github.com/grafana/grafana/pull/81318


torkelo:
I think we rather bite the bullet now and unify AdHocFilterSet as a variable, https://github.com/grafana/scenes/pull/586

ivanortegaalba:
Close in favor of #586

PR #586, opened 2024-02-09
AdHocFiltersSet/Variable: Unify both objects as a scene variable (breaking change)

torkelo:
The split implementation of AdHocFilterSet and AdHocFilterVariable just adds complexity, unifying them simplifíes things a lot and avoids confusion. It also opens the door to subscribe to changes to these using variable system even in auto mode (think I will do that in a separate PR). 

And this change makes options like https://github.com/grafana/scenes/pull/585 not needed 

* Removes AdHocFilterSet, replaced by AdHocFiltersVariable (that works the same way) 
* Rename applyMode = "same-datasource" to "auto" as it was a bit confusing name
* Remove the AdHocFilterVariable.create factory function (mostly because it's not needed anymore and we are breaking some defaults so this makes it  more explicit when you upgrade to set the right options)
* Update docs
* Add a new VariableValueControl that allows you to render only a specific variable control (no matter the variable hide option). Gives scene apps more control if they want to render the adhoc filters separatly from the main variables. 

Open question 
* What is the correct default for applyMode? Before the AdHocFilterSet defaulted to "auto" (or same-datasource as it was named), and AdHocFiltersVariable defaulted to "manual" so no matter what we default to it's a breaking change. 

# Release notes

AdHocFilterSet is now removed from the library. AdHocFiltersVariable can now be used in both modes (auto and manual).  

To migrate replace AdHocFilterSet with AdHocFiltersVariable , the `applyMode` defaults to `auto` which is the new renamed value that was previously `same-datasource`.  Instead of adding this directly to a controls array add it to the variables array of a SceneVariableSet. It will then be rendered along with other variables via the VariableValueSelectors controls component. If you want to render ad hoc filters separately you can set `hide: VariableHide.hideVariable` so that the filters are not rendered by VariableValueSelectors and use the new component VariableValueControl that can render a specific variable.  

`AdHocFiltersVariable.create` is also removed as this separate factory function is no longer needed. If you where using `AdHocFiltersVariable.create` then switch to the normal constructor but be sure to pass in `applyMode: 'manual'` when you create it to preserve the same behavior as before. 

<details>
  <summary>📦 Published PR as canary version: <code>3.0.0--canary.586.7870305924.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@3.0.0--canary.586.7870305924.0
  # or 
  yarn add @grafana/scenes@3.0.0--canary.586.7870305924.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -185,7 +185,7 @@ export class AdHocFiltersVariable
       return [];
     }
 
-    const otherFilters = this.state.filters.filter((f) => f.key !== currentKey).concat(this.state.baseFilters!);
+    const otherFilters = this.state.filters.filter((f) => f.key !== currentKey).concat(this.state.baseFilters ?? []);
     const queries = this._getSceneQueries();
     let keys = await ds.getTagKeys({ filters: otherFilters, queries });
 


ivanortegaalba:
> What is the correct default for applyMode? Before the AdHocFilterSet defaulted to "automatic" (or same-datasource as it was named), and AdHocFiltersVariable defaulted to "manual" so no matter what we default to it's a breaking change.

As a library consumer, I'd expect for them to work out of the box. If we use them as variable, and you're familiar with what they do, I'd expect for them to affect automatically, in the same way other elements from controls does, like time range picker or refresh.

Probably we need to add breaking changes notes here with some mini migration guide to do the replacement

packages/scenes/src/variables/groupby/GroupByVariable.tsx
```diff
@@ -119,7 +119,7 @@ export class GroupByVariable extends MultiValueVariable<GroupByVariableState> {
       options: [],
       datasource: null,
       baseFilters: [],
-      applyMode: 'same-datasource',
+      applyMode: 'automatic',
```
ivanortegaalba, line 122: Maybe so used to CSS; but I think `auto` sounds better. WDYT?

torkelo: @ivanortegaalba yea, agree! good idea

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.test.tsx
```diff
@@ -107,7 +299,101 @@ describe('AdHocFiltersVariable', () => {
...
+    ],
+    ...overrides,
+  });
+
+  const scene = new EmbeddedScene({
+    $timeRange: new SceneTimeRange(),
+    $variables: new SceneVariableSet({
+      variables: [],
+    }),
+    controls: [filtersVar],
```
ivanortegaalba, line 373: Should this go into `$variables`?
```suggestion
    $variables: new SceneVariableSet({
      variables: [filtersVar],
    }),
    controls: [],
```

torkelo: well it should be a VariableValueSelectors component

packages/scenes-app/src/demos/adhocFiltersDemo.tsx
```diff
@@ -27,15 +26,16 @@ export function getAdhocFiltersDemo(defaults: SceneAppPageState) {
...
-                datasource: { uid: 'gdev-prometheus' },
-              }),
-              ...getEmbeddedSceneDefaults().controls,
-            ],
+            $variables: new SceneVariableSet({
+              variables: [
+                new AdHocFiltersVariable({
+                  name: 'Filters',
+                  // Only want keys for this series
+                  baseFilters: [{ key: '__name__', operator: '=', value: 'ALERTS', condition: '' }],
```
ivanortegaalba, line 34:
```suggestion
                  baseFilters: [{ key: '__name__', operator: '=', value: 'ALERTS' }],
```
`condition` is optional and it [is deprecated](https://github.com/grafana/grafana/commit/6a37a56d68f4c47dd6880c457c31eb07c9a96b04).

We need to update the version of `@grafana/data` to the latest.

ivanortegaalba (approved):
I tested with [the dashboard PR](https://github.com/grafana/grafana/pull/81318), and it works like a charm. Very easy to migrate. 💯

packages/scenes/src/index.ts
```diff
@@ -44,6 +44,7 @@ export * from './variables/types';
 export { VariableDependencyConfig } from './variables/VariableDependencyConfig';
 export { formatRegistry, type FormatVariable } from './variables/interpolation/formatRegistry';
 export { VariableValueSelectors } from './variables/components/VariableValueSelectors';
+export { VariableValueControl } from './variables/components/VariableValueControl';
```
dprokop, line 47: Do we need this this be public API? Can't see it used in the demos app.

dprokop (commented):
Code wise i think this looks good, this is a great unification.

However I found one bug I think. In the manual mode demo (as variable -  which btw we may want to rename). When you remove the filters after visiting that demo, and try adding a new one, ther runtime error is thrown:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'key')
    at datasource.ts:678:16
    at Array.map (<anonymous>)
    at PrometheusDatasource.getTagKeys (datasource.ts:677:69)
    at eval (AdHocFiltersVariable.js:114:25)
    at Generator.next (<anonymous>)
    at asyncGeneratorStep (AdHocFiltersVariable.js:25:28)
    at _next (AdHocFiltersVariable.js:43:17)
```

Investigating this it seems like the `filters` passed to `getTagKeys` is `filters:[undefined]`. My suspicion is this is related to the `var-Filters=` in URL. IMO we may want to allow clearing the value from the URL completely as we do with the GroupBy variable.

packages/scenes/src/index.ts
torkelo: @dprokop This allow scene apps to show just the adhoc filters separately from the variables (which some scene apps do via AdHocFilterSet),

torkelo:
> Investigating this it seems like the filters passed to getTagKeys is filters:[undefined]. My suspicion is this is related to the var-Filters= in URL. IMO we may want to allow clearing the value from the URL completely as we do with the  GroupBy variable.

Was a bug in getTagKeys, assumed baseFilters was alway set with a this.state.baseFilter! in the concat expression, (in getTagValues we correctly did this.state.baseFilter ?? []). 

Also fixed the var-Filters= blank url state on activation, added back so that filterExpression is set in the ctor so that in sync when activated (so no state change on activation anymore) 

@dprokop

packages/scenes/src/index.ts
dprokop: k, that's fine then

dprokop:
> Was a bug in getTagKeys, assumed baseFilters was alway set with a this.state.baseFilter! in the concat expression, (in getTagValues we correctly did this.state.baseFilter ?? []).

@torkelo i think you did push the getValuesFor fox, but not keys 🙈 .  I've pushed the fix  just now

dprokop (approved):

torkelo:
@dprokop yea, thanks I just noticed, sure I changed getTagKeys and got it working, super strange :)

PR #587, opened 2024-02-09
SceneQueryRunner: decouple time range comparisons

sd2k:
Prior to this commit, `SceneQueryRunner` had special handling for
`SceneTimeRangeCompare` objects, explicitly searching for them in the
scene graph, adding additional queries, and transforming resulting
queries. This made it difficult to re-use the general behaviour of
'running additional queries' in other objects.

This commit introduces a new interface, `SceneRequestAdder`, which
can be implemented to inform the query runner that it should run
additional requests (returned by `getExtraRequests`) and transform the
results in some fashion.

Instead of searching the graph for `SceneTimeRangeCompare` objects,
the query runner searches for implementors of `SceneRequestAdder`
and uses those instead. The specifics of how it searches for these
is a little bit fuzzy and should probably be improved: it walks up the
graph until it finds at least one adder at the current level or in
any children of the current level; adds any others at that level
or in the children; then returns.

`SceneTimeRangeCompare` has been refactored to make use of this new
interface. I've also got a separate object which also implements
it which is working well including when both are enabled.

Relates to [this Slack thread](https://raintank-corp.slack.com/archives/C04753J0CH4/p1707464804258979).


packages/scenes/src/querying/SceneRequestAdder.ts
```diff
@@ -0,0 +1,31 @@
+import { DataQueryRequest, PanelData } from "@grafana/data";
+
+import { SceneObjectBase } from "../core/SceneObjectBase";
+import { SceneObjectState } from "../core/types";
+
+// A transformation function called by the query runner with responses
```
sd2k, line 6: naming: we already have 'transformations', maybe this should have a different name...

packages/scenes/src/querying/extraRequestProcessingOperator.ts
```diff
@@ -0,0 +1,33 @@
...
+// The responses must have length at least 2; the first is treated as the 'primary'
+// response and the rest as secondary responses.
+//
+// Each secondary response is transformed according to the transform function
+// identified by it's request ID. The transform function is passed the primary
+// response and the secondary response to be processed.
+//
+// The output is a single frame with the primary series and all transformed
+// secondary series combined.
+export const extraRequestProcessingOperator = (transforms: Map<string, TransformFunc>) =>
```
sd2k, line 19: same applies here re. naming of transforms

packages/scenes/src/querying/SceneRequestAdder.ts
```diff
@@ -0,0 +1,31 @@
...
+// transform that should be called with the response data.
+export interface ExtraRequest {
+  // The request.
+  req: DataQueryRequest;
+  // An optional transformation function.
+  transform?: TransformFunc;
+}
+
+// Indicates that this type wants to add extra requests to a query runner.
+export interface SceneRequestAdder<T extends SceneObjectState> extends SceneObjectBase<T> {
```
sd2k, line 22: I dislike the name `SceneRequestAdder` too, open to bikeshedding 🙂

torkelo:
@sd2k I think this looks interesting, sorry for late review. so much going on now with dashboard => scenes migration, and a customer project. We will try to review this in depth shortly, I think with some minor refactoring and name changes it looks overall promising / close, but I just did a very quick code scan

sd2k:
> @sd2k I think this looks interesting, sorry for late review. so much going on now with dashboard => scenes migration, and a customer project. We will try to review this in depth shortly, I think with some minor refactoring and name changes it looks overall promising / close, but I just did a very quick code scan

No worries, it's not urgent, just wanted to make sure it didn't fall too far off the radar 👍 

I'll try and add an example of how this can be used somewhere so it's a bit less abstract.

sd2k:
FYI I've added an example of using this to the [scenes-baseliner branch](https://github.com/grafana/scenes/compare/decouple-comparisons-from-queryrunner...scenes-baseliner) branch along with a demo, should be easy to run but you'll need to use a version of Grafana with https://github.com/grafana/grafana/pull/82299 merged in to load the plugin correctly since it depends on a WASM module.

I don't think that'd necessarily make it into the core repo since it's quite heavyweight but it's a decent example!

sd2k:
One thing I bumped into while using this in other projects was that I can't see a way to turn off the extra `SceneRequestAdder`s per-query runner. For example, in the live demo above the `ScenesBaseliner` runs for the various histogram panels as well as for the time series panels, but it can't really do much in those cases.

I imagine the solution would be to have a separate embedded scene with the baseline controls or something, but it might be nice to be able to say 'this `SceneRequestAdder` should only run for these queries' somehow 🤔

sd2k:
Hey @torkelo, do you think you (or someone else) will have time to take a look at this in the next week or so? Happy to walk through it if you like.

torkelo:
@sd2k sorry, it's a bit of a bad time now with G11 / GrafanaCon crunch.

sd2k:
I've tried to come up with some better names for the things I wasn't happy with ('adder' because it kinda sucks, and 'transform' because it's overused). The new names are maybe slightly better?

The best way to test this out, if anyone's interested, is to check out the Baselines demo in the demo app in the [downstream `scenes-baseliner` branch](https://github.com/grafana/scenes/tree/scenes-baseliner) which makes use of this PR to provide a second control similar to the time range comparison.

packages/scenes/src/querying/SceneRequestAdder.ts
```diff
@@ -0,0 +1,32 @@
...
+export interface ExtraRequest {
+  // The request.
+  req: DataQueryRequest;
+  // An optional function used to process the data before passing it
+  // to any transformations or visualizations.
+  processor?: ProcessorFunc;
+}
+
+// Indicates that this type wants to add extra requests to a query runner.
+export interface SceneRequestSupplementer<T extends SceneObjectState> extends SceneObjectBase<T> {
```
dprokop, line 23: I think naming this `SupplementaryRequestProvider` could be the path forward. WDYT?

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -593,26 +581,29 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
...
 
   /**
-   * Will walk up the scene graph and find the closest time range compare object
-   * It performs buttom-up search, including shallow search across object children for supporting controls/header actions
+   * Walk up the scene graph and find any request adders.
+   *
+   * This will return a map from id to the closest adder for each id.
    */
-  private getTimeCompare() {
+  private getClosestRequestAdders(): Map<string, SceneRequestSupplementer<any>> {
```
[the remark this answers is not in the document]
dprokop: Ignore this comment, I poroposed another name for this :)

dprokop (commented):
@sd2k  - I' really sorry for the late review, but your PR (https://github.com/grafana/scenes/pull/745) made me realise i forgot about this one. 

Overall - I'm in favor of proceeding with this one, as opposite to #745 which literally invites inheritance over composition.

I can't see of any objections why we would not accept this refeactor, it's a good idea that I've been thinking when first implementing the time range comparison. 

The minor thing I'm missing is generalised test for supplementary requests - all tests right now are performed in the context of time range comparison so they don't neessarily show the idea behind the supplementary requests when going through spec.

Also, please condsider my rename suggestion.

packages/scenes/src/querying/SceneRequestAdder.ts
sd2k: That sounds much better! I'll go through and rename any usages (I think there's a misnamed filename, too).

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -593,26 +581,29 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
...
+    getClosest(this.parent, (s) => {
+      if (isRequestAdder(s) && !found.has(s.constructor.name)) {
+        found.set(s.constructor.name, s);
       }
       s.forEachChild((child) => {
-        if (child instanceof SceneTimeRangeCompare) {
-          found = child;
+        if (isRequestAdder(child) && !found.has(child.constructor.name)) {
+          found.set(child.constructor.name, child);
         }
```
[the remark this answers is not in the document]
sd2k: I think this might be leftover from a previous commit, thanks for picking it up.

sd2k:
> @sd2k - I' really sorry for the late review, but your PR (#745) made me realise i forgot about this one.
> 
> Overall - I'm in favor of proceeding with this one, as opposite to #745 which literally invites inheritance over composition.

Agreed :+1: I think the main concerns I had were the introduction of even more interfaces and concepts to the library but I'll make sure they're documented.

> I can't see of any objections why we would not accept this refeactor, it's a good idea that I've been thinking when first implementing the time range comparison.
> 
> The minor thing I'm missing is generalised test for supplementary requests - all tests right now are performed in the context of time range comparison so they don't neessarily show the idea behind the supplementary requests when going through spec.

Understood, I'll try and add something to this effect.

> Also, please condsider my rename suggestion.

:100: - I much prefer yours!

---

One side note: in another branch I've been exploring actually using this functionality with some implementations of `SceneRequestSupplementer` that run some ML algorithms on the secondary requests. The controls for the components allow the user to adjust things like the confidence interval and other hyperparameters of the ML algorithms. When the user changes those, we need to rerun the _processors_, but we don't always need to rerun the _queries_.

My solution to this has been to change the `shouldRerun` method to return an object like `{ query: boolean; processor: boolean }`, which the implementation can use to tell the query runner whether it should rerun the query _and_ the processor or just the processor. This is pretty essential for some UXs like this:

https://github.com/grafana/scenes/assets/5464991/37039288-1d00-44c9-a64c-24c55f7ed8d0

since otherwise the query would need to be reissued every time the processor wanted to rerun.

I mention this because I don't want to lock us into a specific API in `SceneRequestSupplementer` if/when I try to upstream those changes (in a separate PR). The options I see are:

1. Keep the current signature (`shouldRerun(prev, next): boolean`) in this PR, and don't ever allow implementations to specify which of query/processor to run
2. Keep the current signature (`shouldRerun(prev, next): boolean`) in this PR, and if we allow queries/processors to run separately, release an API-incompatible change by changing the signature to `shouldRerun(prev, next): { query: boolean; processor: boolean; }`
3. Keep the current signature (`shouldRerun(prev, next): boolean`) in this PR, and if we allow queries/processors to run separately, release an API-**compatible** change by changing the signature to `shouldRerun(prev, next): boolean | { query: boolean; processor: boolean; }`, which is a bit more complex but possibly easier for basic use cases
4. Switch the signature to either `shouldRerun(prev, next): { query: boolean; processor: boolean; }` or `shouldRerun(prev, next): boolean | { query: boolean; processor: boolean; }` now.

Sorry for the long response, I've spent quite a bit of time working on this other branch the last few days! :smile:

packages/scenes/src/querying/SceneQueryRunner.ts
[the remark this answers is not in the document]
sd2k: Ah I remember now: it's because we only want to include the _closest_ of each type from this function, so by using a map we can easily check whether we already found a `SceneTimeRangeCompare` (for example) closer to the runner and ignore any that we find further up the scene graph).

I think we could change the return value of the function to an array since we don't care about the keys later though.

dprokop:
> One side note: in another branch I've been exploring actually using this functionality with some implementations of SceneRequestSupplementer that run some ML algorithms on the secondary requests. The controls for the components allow the user to adjust things like the confidence interval and other hyperparameters of the ML algorithms. When the user changes those, we need to rerun the processors, but we don't always need to rerun the queries.


I see. But I'm wondering if you really need a processor for this implementation. What if you implemented the processor as a custom transformation (https://grafana.com/developers/scenes/transformations#add-custom-transformations) and attach a behaviour to it, that would react to the controls changes and call `reprocessTransformations` ? It's a bit more complex, but it would keep the interface of the `SupplementaryRequestProvider` simpler and actually focused on query execution.

sd2k:
I added a [draft](https://github.com/grafana/scenes/pull/748) of how an alternate API for `shouldRerun` could look with the reprocessing handled by the query runner, in case it's interesting :slightly_smiling_face:

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -462,12 +470,9 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
   private prepareRequests = (
     timeRange: SceneTimeRangeLike,
     ds: DataSourceApi
-  ): [DataQueryRequest, DataQueryRequest | undefined] => {
-    const comparer = this.getTimeCompare();
+  ): { primary: DataQueryRequest, secondaries: DataQueryRequest[], processors: Map<string, ProcessorFunc> } => {
```
torkelo, line 473: Can we declare a type for this return type (also please switch prepareRequests to be a normal function instead of arrow function, no need for it to be an arrow function, it can live on prototype)

packages/scenes/src/querying/SupplementaryRequestProvider.ts
```diff
@@ -0,0 +1,46 @@
...
+  //
+  // When the provider's state changes this function will be passed both the previous and the
+  // next state. The implementation can use this to determine whether the change should trigger
+  // a rerun of the query or not.
+  shouldRerun(prev: T, next: T): boolean;
+}
+
+export function isSupplementaryRequestProvider(obj: any): obj is SupplementaryRequestProvider<any> {
+  return typeof obj === 'object' && 'getSupplementaryRequests' in obj;
+}
```
torkelo, line 46: Some renaming suggestions 

SupplementaryRequestProvider => ExtraQueryProvider   (or AdditionalQueryProvider)

ProcessorFunc => ExtraQueryProviderProcessor 
SupplementaryRequest ?> ExtraQueryDescriptor  (or similar) 
isSupplementaryRequestProvider => isExtraQueryProvider

sd2k: I think I prefer `Extra` to `Supplementary` or `Additional` too, the others are quite verbose. `ExtraQueryProviderProcessor` is pretty long though - would `ExtraQueryProcessor` work?

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -593,26 +590,29 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
...
     if (!this.parent) {
-      return null;
+      return [];
     }
-    return getClosest(this.parent, (s) => {
-      let found = null;
-      if (s instanceof SceneTimeRangeCompare) {
-        return s;
+    getClosest(this.parent, (s) => {
+      if (isSupplementaryRequestProvider(s) && !found.has(s.constructor.name)) {
```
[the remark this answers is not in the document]
sd2k: Yeah makes sense, TIL you can use functions as map keys! 🤯

packages/scenes/src/querying/SupplementaryRequestProvider.ts
torkelo: @sd2k  ExtraQueryDataProcessor maybe?

sd2k: Works for me, done :+1:

packages/scenes/src/querying/ExtraQueryProvider.ts
```diff
@@ -0,0 +1,46 @@
...
+// _primary_ query, and the data returned by the `ExtraQueryProvider`'s
+// _secondary_ query. It should return a new `PanelData` representing the processed output.
+// It should _not_ modify the primary PanelData.
+//
+// Examples of valid processing include alignment of data between primary and secondary
+// (see the `timeShiftAlignmentProcessor` returned by `SceneTimeRangeCompare`), or doing
+// some more advanced processing such as fitting a time series model on the secondary data.
+//
+// See the docs for `extraQueryProcessingOperator` for more information.
+export type ExtraQueryDataProcessor = (primary: PanelData, secondary: PanelData) => PanelData;
```
sd2k, line 19: Don't think it's suitable for this PR but if I want to offload my processing to a web worker this is going to need to return a Promise instead...

```suggestion
export type ExtraQueryDataProcessor = (primary: PanelData, secondary: PanelData) => Promise<PanelData>;
```

dprokop: If we don't introduce this now, it's going to be a breaking change in the future. I think it may be worth doing this straight away, wdyt @torkelo ?

dprokop (approved):
This is looking good now, the rename to `Extra...` is 👌

packages/scenes/src/querying/ExtraQueryProvider.ts
sd2k: I was thinking we _could_ change it to `export type ExtraQueryDataProcessor = (primary: PanelData, secondary: PanelData) => PanelData | Promise<PanelData>;` but it is clunky, I agree...

torkelo: Can it be an observable (like the current post processing) that will be much easier to slot in to the rxjs pipeline

dprokop: yeah, +1 for that!

sd2k: I've changed the type to `export type ExtraQueryDataProcessor = (primary: PanelData, secondary: PanelData) => Observable<PanelData>;` in bdab9f4d8dfdc413887d3a15ef1ba32210399603, is that what you meant? :thinking: Think I got the types right but I might have misunderstood.

dprokop: i think that's exactly it @sd2k

torkelo (approved):
Looks great, thanks for sticking with this slow PR review process, hopefully the next one will not take months to get approved! 

❤️

PR #589, opened 2024-02-10
AdHoc/GroupBy: Use variable system as way to subscribe to changes

torkelo:
Builds on https://github.com/grafana/scenes/pull/586 and fixes some bugs in main as well 

* Instead of subscribing to state of adhoc filters and group by (which also triggers when the variable goes into loading state) we now use the variable system. 

Bugfix 
* Currently we do not re-subscribe to adhoc filters nor group by when a query runner re-activates so nothing happens when coming back to a cached scene.


dprokop (approved):
This makes sense, thanks for picking up the adhoc fiter set refactor!

PR #595, opened 2024-02-13
Variables: Clear current value when no options are returned

torkelo:
Replicating the current logic of the old variable system: https://github.com/grafana/grafana/blob/6c42bd31c63d818fe74e50ea4ef684f849a544e0/public/app/features/variables/query/reducer.ts#L168 

Sets the value to `''` and text to 'None' , but I think setting both text and value to `''` is probably best 

# Release notes

All variables that extend from MultValueVariable (Query, DataSource, Custom) now clear the current value if no options / values are returned by query, clears to empty string or array depending on multi or not.

### DIFF
--- a/packages/scenes/src/variables/variants/MultiValueVariable.test.ts
+++ b/packages/scenes/src/variables/variants/MultiValueVariable.test.ts
@@ -61,6 +61,39 @@ describe('MultiValueVariable', () => {
       expect(variable.state.text).toBe('A');
     });
 
+    it('Should set to empty when no options are returned', async () => {
+      const variable = new TestVariable({
+        name: 'test',
+        options: [],
+        optionsToReturn: [],
+        value: 'A',
+        text: 'A',
+        delayMs: 0,
+      });
+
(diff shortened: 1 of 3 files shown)


packages/scenes/src/variables/variants/MultiValueVariable.test.ts
```diff
@@ -61,6 +61,39 @@ describe('MultiValueVariable', () => {
...
+        delayMs: 0,
+      });
+
+      await lastValueFrom(variable.validateAndUpdate());
+
+      expect(variable.state.value).toBe('');
+      expect(variable.state.text).toBe('');
+    });
+
+    it('Should set to empty arrauy when no options are returned and variable isMulti', async () => {
```
ivanortegaalba, line 80:
```suggestion
    it('Should set to empty array when no options are returned and variable isMulti', async () => {
```

Small typo 😄

ivanortegaalba (approved):
LGTM! 👍

PR #599, opened 2024-02-13
VariableDependencyConfig: Support `*` to extract dependencies from every state path

ivanortegaalba:
**Problem**
By default, when no `statePaths` was defined in `VariableDependencyConfig`, the whole state was parsed. This is not an expected behavior if no specific criteria to find variables is defined.

**Solution**
Enable wildcard for `statePaths` to try to find variables in every state path.

```ts
class TestObj extends SceneObjectBase<TestState> {
  public constructor() {
    super({
      query: 'query with ${queryVarA} ${queryVarB}',
      otherProp: 'string with ${otherPropA}',
      nested: {
        query: 'nested object with ${nestedVarA}',
      },
    });
  }
}

it('Should be able to extract dependencies from all state', () => {
    const sceneObj = new TestObj();
    const deps = new VariableDependencyConfig(sceneObj, { statePaths: ['*'] });

    expect(deps.getNames()).toEqual(new Set(['queryVarA', 'queryVarB', 'nestedVarA', 'otherPropA']));
  });
```

# Release notes

**Possible breaking change: `VariableDependencyConfig` default behavior**

- **Previously:** Using `VariableDependencyConfig` without options scanned the entire state.
- **Now:** Default behavior requires an explicit wildcard (`*`) to scan the whole state. This prevents unintended dependency resolution.

**Impact:** If you intentionally scanned the entire state, use `statePaths: ['*']`. Otherwise, specify desired `statePaths` or `variableNames`.

**Example:**

```diff
class TestObj extends SceneObjectBase<TestState> {
  public constructor() {
    super({
      query: 'query with ${queryVarA} ${queryVarB}',
      otherProp: 'string with ${otherPropA}',
      nested: {
        query: 'nested object with ${nestedVarA}',
      },
    });
  }
}

it('Should be able to extract dependencies from all state', () => {
    const sceneObj = new TestObj();
-    const deps = new VariableDependencyConfig(sceneObj, {});
+    const deps = new VariableDependencyConfig(sceneObj, { statePaths: ['*'] });

    expect(deps.getNames()).toEqual(new Set(['queryVarA', 'queryVarB', 'nestedVarA', 'otherPropA']));
  });
```

This mproves performance and avoids unexpected dependency resolution.

<details>
  <summary>📦 Published PR as canary version: <code>3.4.0--canary.599.7901015965.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@3.4.0--canary.599.7901015965.0
  # or 
  yarn add @grafana/scenes@3.4.0--canary.599.7901015965.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/VariableDependencyConfig.ts
+++ b/packages/scenes/src/variables/VariableDependencyConfig.ts
@@ -123,7 +123,7 @@ export class VariableDependencyConfig<TState extends SceneObjectState> implement
     if (newState !== prevState) {
       if (this._statePaths) {
         for (const path of this._statePaths) {
-          if (newState[path] !== prevState[path]) {
+          if (path === '*' || newState[path] !== prevState[path]) {
             this.scanStateForDependencies(newState);
             break;
           }
@@ -144,7 +144,7 @@ export class VariableDependencyConfig<TState extends SceneObjectState> implement
     this.scanStateForDependencies(this._state!);
   }
 
-  public setPaths(paths: Array<keyof TState>) {
+  public setPaths(paths: Array<keyof TState | '*'>) {
     this._statePaths = paths;
   }
 
--- a/packages/scenes/src/variables/VariableDependencyConfig.ts
+++ b/packages/scenes/src/variables/VariableDependencyConfig.ts
@@ -161,6 +161,7 @@ export class VariableDependencyConfig<TState extends SceneObjectState> implement
         for (const path of this._statePaths) {
           if (path === '*') {
             this.extractVariablesFrom(state);
+            break;
           } else {
             const value = state[path];
             if (value) {
--- a/packages/scenes/src/variables/VariableDependencyConfig.ts
+++ b/packages/scenes/src/variables/VariableDependencyConfig.ts
@@ -10,7 +10,7 @@ interface VariableDependencyConfigOptions<TState extends SceneObjectState> {
   /**
    * State paths to scan / extract variable dependencies from. Leave empty to scan all paths.
    */
-  statePaths?: Array<keyof TState>;
+  statePaths?: Array<keyof TState | '*'>;
 
   /**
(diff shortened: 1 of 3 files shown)


torkelo:
I have scanned through all our org's use of VariableDependencyConfig, and can't find any use where statePaths nor variableNames are defined, so I think this should be fine and very minor breaking change with possibly no impacts https://github.com/search?q=org%3Agrafana+VariableDependencyConfig&type=code

torkelo (commented):
but would be good add a release notes that describe the change

torkelo (approved):

packages/scenes/src/variables/VariableDependencyConfig.ts
```diff
@@ -159,13 +159,15 @@ export class VariableDependencyConfig<TState extends SceneObjectState> implement
     } else {
       if (this._statePaths) {
         for (const path of this._statePaths) {
-          const value = state[path];
-          if (value) {
-            this.extractVariablesFrom(value);
+          if (path === '*') {
+            this.extractVariablesFrom(state);
```
dprokop, line 163: Probably rare case, but should we break the loop here if someones defines the dependencies as `['somePath', '*', 'someOtherPath]`?

dprokop (approved):
Good adition, +1 to @torkelo comment about the release notes. 

Also, worth updating the description of the `statePaths` property in the advanced variables usage docs: https://grafana.com/developers/scenes/advanced-variables#step-4-add-variables-support-to-the-textinterpolator-object

ivanortegaalba:
Good catch! I just updated the release notes. I'm updating the docs too

packages/scenes/src/variables/VariableDependencyConfig.ts
ivanortegaalba: Yep, it would be weird, but I considered pushed changes to consider it

PR #610, opened 2024-02-21
SceneObjectBase: Minor optimization to setParent

torkelo:
Only needs to scan state props part of the state change


bfmatei (approved):
LGTM!

PR #614, opened 2024-02-22
AdHocFilters: Auto focus value when key is selected

torkelo:
It's not working for some reason. Some timing issue, the autoFocus + openMenuOnFocus is working for the key select but not the value select 

Update: It is working now.

* [x] Fixes blur issue that would just make it gain focus again 
* [x] Make auto focus value work
* [x] Change value placeholder to "Select value" 
  * Handled by #620

### DIFF
--- a/packages/scenes-app/src/demos/verticalControlsLayoutDemo.tsx
+++ b/packages/scenes-app/src/demos/verticalControlsLayoutDemo.tsx
@@ -67,7 +67,7 @@ export function getVerticalControlsLayoutDemo(defaults: SceneAppPageState) {
               hide: VariableHide.hideLabel,
               layout: 'vertical',
               applyMode: 'manual',
-              datasource: { uid: 'gdev-prometheus' },
+              datasource: { uid: 'se-demo' },
               filters: [{ key: 'job', operator: '=', value: 'grafana', condition: '' }],
             }),
           ],
--- a/packages/scenes-app/src/demos/verticalControlsLayoutDemo.tsx
+++ b/packages/scenes-app/src/demos/verticalControlsLayoutDemo.tsx
@@ -1,4 +1,4 @@
-import { VariableRefresh } from '@grafana/data';
+import { AdHocVariableFilter, VariableRefresh } from '@grafana/data';
 import {
   SceneFlexLayout,
   SceneVariableSet,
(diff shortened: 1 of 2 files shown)


packages/scenes-app/src/demos/verticalControlsLayoutDemo.tsx
```diff
@@ -67,7 +67,7 @@ export function getVerticalControlsLayoutDemo(defaults: SceneAppPageState) {
               hide: VariableHide.hideLabel,
               layout: 'vertical',
               applyMode: 'manual',
-              datasource: { uid: 'gdev-prometheus' },
+              datasource: { uid: 'se-demo' },
```
darrenjaneczek, line 70: Temporary?

I tried various datasources, but I wasn't able to get the demo to work until I added providers.
```
              getTagKeysProvider: async (variable: AdHocFiltersVariable, currentKey: string | null) => ({
                replace: true,
                values: [{text: 'job'}],
              }),
```

Maybe with delays, we can make these selector demos independent of a datasource.

darrenjaneczek:
```ts
    <Select
      disabled={model.state.readOnly}
      className={state.isKeysOpen ? styles.widthWhenOpen : undefined}
      width="auto"
      value={valueValue}
      placeholder={'value'}
      options={state.values}
      onChange={(v) => model._updateFilter(filter, 'value', v.value)}
      isOpen={state.isValuesOpen}
      isLoading={state.isValuesLoading}
      autoFocus={filter.key !== '' && filter.value === ''}
      openMenuOnFocus={true}
      onOpenMenu={async () => {
        setState({ ...state, isValuesLoading: true });
        const values = await model._getValuesFor(filter);
        setState({ ...state, isValuesLoading: false, isValuesOpen: true, values });
      }}
      onCloseMenu={() => {
        setState({ ...state, isValuesOpen: false });
      }}
    />
  );
```

I am wondering if we can get away with doing the fetching that occurs here:
```ts
      onOpenMenu={async () => {
        setState({ ...state, isValuesLoading: true });
        const values = await model._getValuesFor(filter);
        setState({ ...state, isValuesLoading: false, isValuesOpen: true, values });
      }}
```
exclusively after a key has been changed?

Issue #615, opened 2024-02-23
Legend sorting gets stuck


xforman2:
When sorting a legend it gets stuck after 3 clicks.

[Screencast from 2024-02-23 09-02-54.webm](https://github.com/grafana/scenes/assets/129223585/d76a29d3-d827-407d-8951-68f8a3a6e47f)

PR #616, opened 2024-02-23
Legend sorting fix

xforman2:
So the legend options get changed based on which sorting should be applied but in case of no sorting the values for `sortBy` and `sortDesc` are undefined. 
This then turns into a problem here:
https://github.com/grafana/scenes/blob/efa082d3c27452f4ad6504d8784cea72c2e62a31/packages/scenes/src/components/VizPanel/VizPanel.tsx#L236

since after  `mergeWIth`  the target properties are used.

I changed the `replace` to true, skipping the `mergeWith` because I did not want to break some other logic in `onOptionsChange`


torkelo (approved):

xforman2:
fixes #615

PR #618, opened 2024-02-23
Scenes: Implement "Live now" feature

kaydelaney:
First pass at migrating https://github.com/grafana/grafana/pull/37769 to scenes.
Not every nuance of the feature has been migrated, but it's a start.

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanel.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.tsx
@@ -225,15 +225,12 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
       );
     }
 
-    return this.getPanelTimeRange(timeRangeValue, this.getPlugin(), data);
-  }
-
-  private getPanelTimeRange = (timeRange: TimeRange, plugin: PanelPlugin | undefined, data?: PanelData) => {
+    const plugin = this.getPlugin();
     if (plugin && !plugin.meta.skipDataQuery && data && data.timeRange) {
       return data.timeRange;
     }
   
-    return timeRange;
+    return timeRangeValue;
   }
 
   public onTitleChange = (title: string) => {
--- a/packages/scenes/src/components/VizPanel/VizPanel.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.tsx
@@ -16,6 +16,7 @@ import {
   PluginType,
   renderMarkdown,
   PanelPluginDataSupport,
+  TimeRange,
 } from '@grafana/data';
 import { PanelContext, SeriesVisibilityChangeMode, VizLegendOptions } from '@grafana/ui';
 import { config, getAppEvents, getPluginImportUtils } from '@grafana/runtime';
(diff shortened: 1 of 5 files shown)


axelavargas (approved):
Awesome work, @kaydelaney 🙌🏾 

I've reviewed the code and it looks good to me. Implementing this as a behavior is a good approach in my opinion.

As you mentioned, there's currently missing logic for the live feature, specifically regarding determining the panel's refresh rate based on both the time range and width.

P.S. I did a small sync locally with the Grafana repository and works as expected.

ivanortegaalba (commented):
I like the approach of using a behavior too. We thought to have this only implemented in Dashboard, but I think this is a great addition to the library.

I'd add some tests, and probably documentation as part of the docs when the feature is ready to work as it works in Dashboard.

Great progress on this Kay! Let's add some tests, and LGTM!

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -215,6 +211,31 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
     });
   };
 
+  public getTimeRange = (data?: PanelData) => {
```
torkelo, line 214: should this really be an arrow function?

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -215,6 +211,31 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
...
+        sceneTimeRange.getTimeZone(),
+        sceneTimeRange.state.fiscalYearStartMonth,
+        sceneTimeRange.state.UNSAFE_nowDelay
+      );
+    }
+
+    return this.getPanelTimeRange(timeRangeValue, this.getPlugin(), data);
+  }
+
+  private getPanelTimeRange = (timeRange: TimeRange, plugin: PanelPlugin | undefined, data?: PanelData) => {
```
torkelo, line 231: feels a bit confusing to have getTimeRange and getPanelTimeRange functions, do we need both?

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -215,6 +211,31 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
     });
   };
 
+  public getTimeRange = (data?: PanelData) => {
+    const liveNowTimer = sceneGraph.findObject(this, (o) => o instanceof LiveNowTimer);
+    const sceneTimeRange = sceneGraph.getTimeRange(this);
+    let timeRangeValue = sceneTimeRange.state.value
+    if (liveNowTimer instanceof LiveNowTimer && liveNowTimer.isEnabled) {
```
torkelo, line 218: why two instanceof checks? (here and on ine 215)

torkelo (commented):
Look ok, would be nice to come up with a way where VizPanel would not need to know / check for a liveNowTimer but would require some way to separate time update events that should trigger query and those that should only cause viz-rerender

packages/scenes/src/components/VizPanel/VizPanel.tsx
dprokop: agree, think we can just put `getPanelTimeRange` implementation into the `getTimeRange`, it's like 3 lines of code and the function ain't used anywhere else.

packages/scenes/src/components/VizPanel/VizPanel.tsx
dprokop: @torkelo - `sceneGraph.findObject` ain't typed.

dprokop (approved):
Looks like a good start. Agree with @torkelo it would be great to make the viz panel independent from the live timer, but I don't think this is critical at the moment. We can address this as a separate improvement to this feature.

packages/scenes/src/components/VizPanel/VizPanel.tsx
torkelo: ah, I see.

PR #627, opened 2024-03-04
Scenes: Logic for dragging a row within another row

mdvictor:
This PR fixes the case where a row is dragged in the area of another row. Before this, dragging a row within another row would parent that row to the upper row, which caused the `SceneGridRow` to throw a `SceneGridRow must be a child of SceneGridLayout` error and crashed the dashboard. 

In the old arch, doing this would keep the row at the layout level and split the children panels in the new rows based on where the new row would be dropped (e.g.: if a row would be drag and dropped after the first child panel of another row that contained 3 panels, then the old row would have the one panel below it, until the new panel, and the new row would now hold the other 2 panels)

WIP, I haven't tested it extensively.

TODO: Write some tests for this.

https://github.com/grafana/scenes/assets/36818606/7a960d91-9832-42c2-9b30-9a349ff49fc4

### DIFF
--- a/packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
+++ b/packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
@@ -292,8 +292,6 @@ export class SceneGridLayout extends SceneObjectBase<SceneGridLayoutState> imple
 
     this.setState({ children: sortChildrenByPosition(newChildren) });
     this._skipOnLayoutChange = true;
-
-    console.log(this.state);
   };
 
   private toGridCell(child: SceneGridItemLike): ReactGridLayout.Layout {
--- a/packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
+++ b/packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
@@ -257,17 +257,43 @@ export class SceneGridLayout extends SceneObjectBase<SceneGridLayoutState> imple
       }
     }
 
-    // Update the parent if the child if it has moved to a row or back to the grid
     const indexOfUpdatedItem = gridLayout.findIndex((item) => item.i === updatedItem.i);
-    const newParent = this.findGridItemSceneParent(gridLayout, indexOfUpdatedItem - 1);
     let newChildren = this.state.children;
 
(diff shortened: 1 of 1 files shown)


mdvictor:
Bug when adding a row in the middle of another row:

https://github.com/grafana/scenes/assets/36818606/42c9df87-1f1b-4c47-9058-e88ba73e9892

torkelo:
It's kind of impossible to tell what is going on in the videos with all rows and panels named the same :) 

can you describe the bug/change?`

mdvictor:
Sorry @torkelo, I updated the description with more details on what has changed and why we need this for feature parity.

packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
```diff
@@ -257,13 +257,37 @@ export class SceneGridLayout extends SceneObjectBase<SceneGridLayoutState> imple
...
+    if (sceneChild instanceof SceneGridRow) {
+      let indexOfSeparation = 0;
+
+      // and we are drag and dropping this row inside another row, we need to
+      for (let i = indexOfUpdatedItem - 1; i >= 0; i--) {
+        const gridItem = gridLayout[i];
+        const aboveRow = this.getSceneLayoutChild(gridItem.i);
+  
+        // find if there is another row above in tree and
+        // get all children that are below this row from the row above and append them to this row
```
torkelo, line 273: why would we append children of the row above to this row?

mdvictor:
We talked about this in the sync and decided that this shouldn't be migrated because it's quite a weird behaviour even in the existing arch. I'll close this PR and open a new one that simply exits and does not crash the dashboard when trying to drag and drop a row in another row.

packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
mdvictor: This was mainly done for feature parity, albeit I agree it's weird. This currently can be done in the old arch, but after a discussion in the sync today we decided to not have that and just not allow a user to drag and drop a row under another row. Will open a new PR with that

PR #628, opened 2024-03-05
SceneObjectBase: Call self activation handlers before child data, time range and variable handlers

torkelo:
I noticed a bug in scenes dashboards with AnnotationDataLayer not working with data source variables, resulting in error `AnnotationsDataLayer.js:95 AnnotationsDataLayer error {message: 'Datasource ${ds} was not found'}` 

I tracked it down to an issue with scene $data, $timeRange and $variables being activated before the local (own) level activation handlers. Resulting in data layers on the top dashboard level being activated before the dashboard scene itself (that sets the window.__grafanaSceneContext without this the variable system is not working (because scopedVars with sceneObject was not passed in AnnotationsDataLayer)

Changes

* Call own local level activation handlers before calling $data, $timeRange, $variables and $behaviors activation handlers
* Pass scopedVars with context when getting data source in AnnotationsDataLayer 

# Release notes

Activation handlers are for a scene object is now called before any direct child activation handlers. Before this release the activation handlers of direction $data, $timeRange, $variables and $behaviors was called before the SceneObjects own activation handlers.

### DIFF
--- a/packages/scenes/src/core/SceneObjectBase.test.ts
+++ b/packages/scenes/src/core/SceneObjectBase.test.ts
@@ -125,6 +125,25 @@ describe('SceneObject', () => {
     });
   });
 
+  describe('Should activate itself before child objects', () => {
+    const dataChild = new SceneDataNode({});
+    const scene = new TestScene({
+      $data: dataChild,
+    });
+
+    let parentActive = false;
+
+    scene.addActivationHandler(() => {
+      parentActive = true;
+    });
+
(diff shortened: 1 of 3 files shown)


dprokop (commented):
@torkelo , I don't see this change in the PR: "Pass scopedVars with context when getting data source in AnnotationsDataLayer"

Is this something you missed?

torkelo:
@dprokop had not pushed it :)

PR #630, opened 2024-03-05
VariableDependencyConfig: Fixes support for explicit dependencies and scanned dependencies

### LINKED ISSUE #629: AdHocFilterVariable prevents SceneQueryRunner from reacting to variable changes
If there is an `auto`  `AdHocFilterVariable` up in the hierarchy, `ScenesQueryRunner` will not react to other variables changing, for example, data source.

This is because `ScenesQueryRunner` will add ad hoc filter and group by variable filter names to it's variable dependency https://github.com/grafana/scenes/blob/main/packages/scenes/src/querying/SceneQueryRunner.ts#L604

This prevents vairable dependency from infering variable names from it's state, and clears previously iferred ones
https://github.com/grafana/scenes/blob/main/packages/scenes/src/variables/VariableDependencyConfig.ts#L165 

Seems these PRs collided a bit:
https://github.com/grafana/scenes/pull/598
https://github.com/grafana/scenes/pull/596

Guessing intention of 596 was to register additional dependency, not override existing ones

torkelo:
Missed the if/else logic in https://github.com/grafana/scenes/pull/598 

We need to support both explicit and scanned dependencies 

Fixes #629

### DIFF
--- a/packages/scenes/src/variables/VariableDependencyConfig.test.ts
+++ b/packages/scenes/src/variables/VariableDependencyConfig.test.ts
@@ -85,7 +85,7 @@ describe('VariableDependencyConfig', () => {
     expect(deps.scanCount).toBe(2);
   });
 
-  it('Should not scan the state if variable name defined', () => {
+  it('Should not scan the state if no statePaths defined', () => {
     const sceneObj = new TestObj();
     sceneObj.setState({ query: 'new query with ${newVar}' });
     const deps = new VariableDependencyConfig(sceneObj, { variableNames: ['nonExistentVar'] });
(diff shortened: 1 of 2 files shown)


domasx2:
@ivanortegaalba added the `else` deliberately to fix some other issue https://github.com/grafana/scenes/pull/598

torkelo:
@domasx2 I do not think that is needed, was wrong way to fix it. We should not scan child scene objects. We changed to statePaths now need to be set for state to be scanned. So think https://github.com/grafana/scenes/pull/599 should fix that. and we also have https://github.com/grafana/scenes/pull/597 as a extra safety measure

domasx2 (approved):

PR #631, opened 2024-03-05
Fix bug where SceneQueryRunner ignores variable dependencies if AdHocFilter is present

### LINKED ISSUE #629: AdHocFilterVariable prevents SceneQueryRunner from reacting to variable changes

domasx2:
Fixes bug where `SceneQueryRunner` will not react to variables extracted from query / datasource props if `AdHocFilterVariable` is present in hierarchy. See bug for issue below for details

Fixes #629

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.ts
@@ -601,7 +601,7 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
       explicitDependencies.push(this._groupByVar.state.name);
     }
 
-    this._variableDependency.setVariableNames(explicitDependencies);
+    this._variableDependency.setExtraVariableNames(explicitDependencies);
   }
 }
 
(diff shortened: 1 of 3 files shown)


domasx2:
closing in favor of https://github.com/grafana/scenes/pull/630#pullrequestreview-1917028149

PR #632, opened 2024-03-05
Variables: Maintain custom (invalid) variable values set via URL on initial load

### LINKED ISSUE #387: Variables: Manually entered values (not present in the options list) are not preserved after a full page reload
The old variable system support custom (manually) entered variable values not present in the list of options. This works with new scene variable system as well but the value is load when doing a full page reload (even though it is present in the URL, but get's reset)

torkelo:
Fixes #387

The old variable system has this behavior of maintaining custom (invalid) values set via URL (say after full page reload). This was very tricky to implement in scenes as URL sync happens before validation (values are fetched). In the old system the url sync process updates the variable options, validates it and the changes the value to whatever came in from URL. 

To get the same behavior of ignoring validation just after initial URL sync but not say after a datalink is clicked (that changes a variable) I rely on the fact that initial URL sync happens before we activate the scene. When the variable changes value while in-active from URL sync we set skipNextValidation to true. 

* Also fixes URL compatibility with the old variable system that uses `ALL_VARIABLE_TEXT` as the URL value but scenes url system is using `ALL_VARIABLE_VALUE` 

Hope this fixes the Asserts EmbeddedDashboard issue as well 

<details>
  <summary>📦 Published PR as canary version: <code>3.10.0--canary.632.8169220578.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@3.10.0--canary.632.8169220578.0
  # or 
  yarn add @grafana/scenes@3.10.0--canary.632.8169220578.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/variants/MultiValueVariable.test.ts
+++ b/packages/scenes/src/variables/variants/MultiValueVariable.test.ts
@@ -497,7 +497,7 @@ describe('MultiValueVariable', () => {
       expect(variable.state.text).toEqual(['B', 'A']);
     });
 
-    it('updateFromUrl with old arch All value', async () => {
+    it('updateFromUrl with custom value should survive validation', async () => {
       const variable = new TestVariable({
         name: 'test',
         options: [],
@@ -514,7 +514,7 @@ describe('MultiValueVariable', () => {
       expect(variable.getValue()).toEqual('Custom value');
     });
 
-    it('updateFromUrl with custom value should survive validation', async () => {
+    it('updateFromUrl with old arch All value', async () => {
       const variable = new TestVariable({
         name: 'test',
         options: [
--- a/packages/scenes/src/variables/variants/MultiValueVariable.test.ts
+++ b/packages/scenes/src/variables/variants/MultiValueVariable.test.ts
@@ -461,7 +461,7 @@ describe('MultiValueVariable', () => {
       expect(variable.urlSync?.getUrlState()).toEqual({ ['var-test']: ['A'] });
     });
 
-    it('fromUrlState should update value for single value', async () => {
+    it('updateFromUrl should update value for single value', async () => {
       const variable = new TestVariable({
         name: 'test',
         options: [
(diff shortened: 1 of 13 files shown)


packages/scenes/src/variables/variants/MultiValueVariable.test.ts
```diff
@@ -497,6 +497,41 @@ describe('MultiValueVariable', () => {
...
+        optionsToReturn: [],
+        includeAll: true,
+        value: ALL_VARIABLE_VALUE,
+        text: ALL_VARIABLE_TEXT,
+        delayMs: 0,
+      });
+
+      variable.urlSync?.updateFromUrl({ ['var-test']: ALL_VARIABLE_TEXT });
+      expect(variable.getValue()).toEqual(['1', '2']);
+    });
```
dprokop, line 533: I think the titles of these cases are mixed up, should be the other way around.

packages/scenes/src/variables/variants/MultiValueVariable.ts
```diff
@@ -333,9 +347,22 @@ export class MultiValueUrlSyncHandler<TState extends MultiValueVariableState = M
...
+        urlValue = ALL_VARIABLE_VALUE;
+      }
+
+      /**
+       * Initial URL Sync happens before scene objects are activated.
+       * We need to skip validation in this case to make sure values set via URL are maintained.
+       */
+      if (!this._sceneObject.isActive) {
+        this._sceneObject.skipNextValidation = true;
+      }
```
dprokop, line 364: Won't this make the value an invalid one once i.e. variable is refreshed on time range? Feels like the `skipNextValidation` flag is only used for the initial state.

packages/scenes/src/variables/variants/MultiValueVariable.test.ts
torkelo: @dprokop good catch! updated it now

packages/scenes/src/variables/variants/MultiValueVariable.ts
dprokop: Aha, ok, thanks for explaining. I was not aware that it works like this in the old arch too.

dprokop (approved):

PR #635, opened 2024-03-06
Annotations: Don't always execute annotations on activate

torkelo:
Noticed that we always execute annotations on activate (even when we already have data, coming back to a scene). 

Changing this logic a bit so we do not always run on activate if we have data. We should probabaly add a check if time range is stale as well and run query then but leaving that for another time.


dprokop:
@torkelo  - wondering if this is still relevant given https://github.com/grafana/scenes/pull/640?

torkelo:
@dprokop not sure, will check after that one is merged

torkelo:
@dprokop think this is still relevant, still see annotation queries always being executed when you leave dashboard settings

kaydelaney (approved):

dprokop (approved):

PR #640, opened 2024-03-13
DataLayers: Unifiy DataLayers (group of data layers) and a single data layer 

torkelo:
This is a pretty big refactoring of the data layer system 

Addresses problems

* Not possible to update SceneDataLayers layers array and have scene query runners update as they currently subscribe to all the layers individually (and only set this subscription on activation). 
* New scene data layers added during "scene active" phase where are not activated , removed where not deactivated. 
* SceneDataLayerControls did not handle layers being added or removed 
* SceneDataLayers was implementing SceneDataProvider but was not really really implementing it (not acting as a layer or normal data provider) 

Changes
* Make SceneDataLayers and SceneDataLayerProvider all implement SceneDataProvider 
* SceneDataLayers now acts like a single layer (that handles the subscription and merging of results from multiple layers). Supports dynamic updates to layers array
* Simplify the "topic" handling, use the DataFrame meta property we already have. No need for layer topics. Simplifies merging layers and grouping layers. That is it simplifies handling a group of layers as s single layer, as now a single layer can provide annotations and alert states.  
* Simplify the logic of merging many layers (just a matter of merging all the series). 
* Simplify activation of new or removed layers. Now linked to subscription. 
* No need for SceneDataLayerProviderResult now, layers can use the default SceneDataProviderResult
* A pretty simple function that handles merging data layers, mergeMultipleDataLayers  (shared between SceneDataLayers and SceneQueryRunner) 
* A data layer can now be used a single layer (ie you can set $data to `new dataLayers.AnnotationsDataLayer`), you only need SceneDataLayers when you have multiple layers or want to add/remove the layer dynamically. 
* Rendering the individual data layer controls is now handled by the layer.Component  (makes it so SceneDataLayerControls does not care if the data layer is a SceneDataLayers or a single data layer). 

* I need to add a base class for SceneDataLayers (so we can extend and customize it a bit in core dashboards). But I don't like that we would then have SceneDataLayerBase and then SceneDataLayersBase so decided to rename `SceneDataLayers` to `SceneDataLayerSet` to follow the naming pattern we have for other "array" objects. 
* Export SceneDataLayerSetBase

* Update tests

Breaking changes

* A minor breaking change for custom data layers that extend SceneDataLayerBase  (no topic arg to publisResults, topic needs to be moved to frame meta property) 
* SceneDataLayers changed name to SceneDataLayerSet

# Release notes 

Breaking change to data layer system. `SceneDataLayers` has been renamed to `SceneDataLayerSet`. SceneDataLayerProvider now extends SceneDataProvider interface so SceneDataLayers can be used directly as data providers. 

<details>
  <summary>📦 Published PR as canary version: <code>4.0.0--canary.640.8344205533.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@4.0.0--canary.640.8344205533.0
  # or 
  yarn add @grafana/scenes@4.0.0--canary.640.8344205533.0
  ```
</details>


packages/scenes/src/core/types.ts
```diff
@@ -193,35 +192,31 @@ export type DeepPartial<T> = {
...
   isEnabled?: boolean;
   isHidden?: boolean;
 }
 
-export interface SceneDataLayerProvider extends SceneObject<SceneDataLayerProviderState> {
-  topic: DataTopic;
-  cancelQuery?: () => void;
-  getResultsStream(): Observable<SceneDataLayerProviderResult>;
+export interface SceneDataLayerProvider extends SceneDataProvider<SceneDataLayerProviderState> {
+  isDataLayer: true;
```
dprokop, line 215: +1

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -133,55 +133,31 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
...
-          })
-        )
-        .subscribe((result) => {
-          this._onLayersReceived(result);
-        });
+    if (dataLayers.length === 0) {
+      return;
     }
+
+    mergeMultipleDataLayers(dataLayers).subscribe(this._onLayersReceived.bind(this));
```
dprokop, line 140: This makes sure all layers are received together right? Hence no need for maping results and making sure layer results ain't duplicated?

packages/scenes/src/querying/layers/SceneDataLayerControls.tsx
```diff
@@ -43,35 +26,20 @@ function SceneDataLayerControlsRenderer({ model }: SceneComponentProps<SceneData
...
-            key={elementId}
-            layer={l}
-            onToggleLayer={() => model.toggleLayer(l)}
-            isEnabled={layersMap[l.state.key!]}
-          />
-        );
-      })}
+      {layers.map((layer) => (
+        <layer.Component model={layer} key={layer.state.key} />
+      ))}
```
dprokop, line 31: We need to handle hidden layers here, currently all layers will be rendered.

packages/scenes/src/querying/layers/annotations/AnnotationsDataLayer.tsx
```diff
@@ -20,9 +27,11 @@ export class AnnotationsDataLayer
   extends SceneDataLayerBase<AnnotationsDataLayerState>
   implements SceneDataLayerProvider
 {
+  static Component = AnnotationsDataLayerRenderer;
+
   private _scopedVars: ScopedVars = { __sceneObject: { value: this, text: '__sceneObject' } };
   private _timeRangeSub: Unsubscribable | undefined;
-  public topic = DataTopic.Annotations;
+  public isDataLayer: true = true;
```
dprokop, line 34: Think this is not needed given this class extends `SceneDataLayerBase`

packages/scenes/src/querying/layers/SceneDataLayerBase.ts
```diff
@@ -108,6 +106,7 @@ export abstract class SceneDataLayerBase<T extends SceneDataLayerProviderState>
 
     return () => {
       this.onDeactivate();
+      console.log('data layer deactivate');
```
dprokop, line 109: I can't see the layer deactivation handler being ever called.

In the dynamic layers demo:
 - adding callse activation handler
 - deactivating **doesn't** call deactivation handler

In the annotations demo:
- viewing demo shows activatipon logs,
- navigating away from the demo does not show layers deativation logs

dprokop (commented):
Really good refactor, simplifies some parts a bit, I agree! But I left some comments too after testing this PR.

packages/scenes/src/querying/SceneQueryRunner.ts
torkelo: @dprokop well it makes sure there is a single subscription to all layers, and all layer's data are received in a single array (iterable). But it does not currently batch the updates, so each layer data send will trigger an emit/data update on the receiver. So if you have 5 layers it will take 5 updates  ( calls to _onLayersReceived) for the array to include data from all layers. 

But this is the same as it was before. We can look at batching this in a later PR if it turns into performance problem .

packages/scenes/src/querying/layers/SceneDataLayerControls.tsx
torkelo: @dprokop ah, good catch! think I can do that in each layer.Component

packages/scenes/src/querying/layers/annotations/AnnotationsDataLayer.tsx
torkelo: ah, yes, forgot to remove this

packages/scenes/src/querying/layers/SceneDataLayerBase.ts
torkelo: @dprokop 

> deactivating doesn't call deactivation handler

Deactiving? you mean removing it or disabling it? 

If I remove it from the scene the layer is deactivated. But if I move to another scene the layers are not deactivated, will investigate. We might be missing an unsubscribe somewhere.

packages/scenes/src/querying/SceneQueryRunner.ts
dprokop: Yeah, so we have individual layer updates calling `_onLayersReceived`. In that method `annotations` is always a new array build from the indivudual layer updates.  Then I cannot see how the annotations from multiple layers got aggregated into a single array? Feels like I'm missing something.

packages/scenes/src/querying/layers/SceneDataLayerBase.ts
dprokop: > Deactiving? you mean removing it or disabling it?

I mean removing it, sorry

torkelo: @dprokop found the cause for this, was missing assigning this.dataLayerSub in SceneQueryRunner when calling mergeMultipleDataLayers 

Now data layers are deactivated when you move away from the scene

packages/scenes/src/querying/SceneQueryRunner.ts
torkelo: @dprokop it works the exact same way as before. 

mergeMultipleDataLayers has a map of layer key to result that it builds up as responses come in. 
https://github.com/grafana/scenes/pull/640/files#diff-98d2eec4dc7c1ce670e7cc1c0c7d6a178bc4d54b9c5387ef7e83b5c2251b49ebR21

All the values of that map are returned. Making it so that mergeMultipleDataLayers returns an observable with SceneDataProviderResult array from all layers. 

SceneDataLayerSet merges results from layers a bit differently in that it needs to actually merge an array of SceneDataProviderResult into a single SceneDataProvider result. it does this by just concatenating all data frames:
https://github.com/grafana/scenes/pull/640/files#diff-38baecad267059faf3e15f485b0524254cc227b868429b17e7834894e295933fR46

dprokop: ok, missed that results map in the `mergeMultipleDataLayers`, sorry!

packages/scenes-app/src/demos/annotations.tsx
```diff
@@ -88,7 +88,7 @@ export function getAnnotationsDemo(defaults: SceneAppPageState) {
           new SceneRefreshPicker({}),
         ],
         key: 'Multiple annotations layers',
-        $data: new SceneDataLayers({
```
dprokop, line 91: This rename will require breaking changes release notice

dprokop (commented):
@torkelo - i just went throu the code one more time and I think it looks good. Any chance we could make it releasable so that we can test core Grafana with a canary release of this branch?

packages/scenes/src/querying/layers/SceneDataLayerControls.tsx
```diff
@@ -7,34 +7,17 @@ import { SceneObjectBase } from '../../core/SceneObjectBase';
...
-
-  private _onActivate() {
-    const layers = sceneGraph.getDataLayers(this, true);
-    this.setState({ layersMap: layers.reduce((acc, l) => ({ ...acc, [l.state.key!]: l.state.isEnabled }), {}) });
-  }
-
-  public toggleLayer(l: SceneDataLayerProvider) {
-    this.setState({ layersMap: { ...this.state.layersMap, [l.state.key!]: !l.state.isEnabled } });
-    l.setState({ isEnabled: !l.state.isEnabled });
+    super({});
```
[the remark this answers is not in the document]
torkelo: No it's not, but added this one just to minimize the breaking change since the old constructor did not take any params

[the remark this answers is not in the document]
dprokop: right, goodie

dprokop (approved):
@torkelo - i've tested in core (https://github.com/grafana/grafana/pull/84804) and it seems to be working just fine, nice work!

torkelo:
@dprokop ah, great news! Thanks for helping test it!

PR #641, opened 2024-03-13
LiveNow: Fixes activation/deactivation issue and logic error with getTimeRange

kaydelaney:


packages/scenes/src/behaviors/LiveNowTimer.ts
```diff
@@ -6,34 +6,33 @@ import { SceneObjectState } from "../core/types";
...
-      if (!this.enabled) {
-        this.enabled = true;
-        this.timerId = window.setInterval(() => {
-          const panels = sceneGraph.findAllObjects(this.getRoot(), (obj) => obj instanceof VizPanel) as VizPanel[];
-          for (const panel of panels) {
-            panel.forceRender();
-          }
-        }, LiveNowTimer.REFRESH_RATE);
-      }
+      window.clearInterval(this.timerId);
```
torkelo, line 27: why window.clearInterval here, does not clearInterval work the same?

packages/scenes/src/behaviors/LiveNowTimer.ts
```diff
@@ -6,34 +6,33 @@ import { SceneObjectState } from "../core/types";
...
+      this.timerId = undefined;
+      this.enabled = true;
+      this.timerId = window.setInterval(() => {
```
torkelo, line 30: same ehre, why window.setInterval

torkelo (approved):

packages/scenes/src/behaviors/LiveNowTimer.ts
kaydelaney: Just a typing thing. `clearInterval` on its own has NodeJS types
image

whereas `window.clearInterval` doesn't:
image

PR #650, opened 2024-03-20
Add ability to opt out specifc queries from TWC

domasx2:
Hi!

App o11y has a use case for opting specifc queries out from time window comparison. We're adding anomaly thresholds which should not be compared to thresholds from previous period. 

This PR adds `timeRangeCompare` flag to data query, which can be set to `false` to prevent this query from being included in previous period request. Does that make sense?

image


dprokop (approved):
Think it makes sense @domasx2! At some point, when we have comparison available in core Grafana, we'll need to think about graduating `timeRangeCompare` to `DataQuery`.

PR #654, opened 2024-03-20
ControlsLabel: style: use info icons for description

darrenjaneczek:
There was a UX request to show an explicit info icon whenever a variable had a description.
This PR adds the icon when the description is present, and shows the tooltip when the new icon is on-hover.
The tooltip was removed from the surrounding field component and now will only appear when interacting with the new icon.

image_360

image_360


catherineymgui (approved):

torkelo (commented):
I think it's the wrong size , sound match the font size text

darrenjaneczek:
How is this:
image

image

torkelo:
@darrenjaneczek yes, that works

PR #658, opened 2024-03-22
Autofit: Fit widget in the screen depending on the height

ivanortegaalba:
**Problem**
When using Grafana on a TV, the users want to have all the widgets available on the screen. However, the screen size can vary between rooms, so they mark "auto size" when playing a playlist to have the dashboard fit vertically in the screen.

When migrating to the dashboard, this feature stopped working.

**Solution**
To be able to adapt the grid to the available height, we modify the RGL cells height to make sure they don't overflow the bottom of the grid view. 

https://github.com/grafana/scenes/assets/5699976/cdc7eb00-9a9e-417e-888c-eb2edb6414bb

<details>
  <summary>📦 Published PR as canary version: <code>4.2.0--canary.658.8522699461.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@4.2.0--canary.658.8522699461.0
  # or 
  yarn add @grafana/scenes@4.2.0--canary.658.8522699461.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
+++ b/packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
@@ -1,7 +1,7 @@
 import ReactGridLayout from 'react-grid-layout';
 
 import { SceneObjectBase } from '../../../core/SceneObjectBase';
-import { SceneLayout, SceneObjectState, SceneObjectUrlValues } from '../../../core/types';
+import { SceneLayout, SceneObjectState } from '../../../core/types';
 import { DEFAULT_PANEL_SPAN } from './constants';
 import { isSceneGridRow } from './SceneGridItem';
 import { SceneGridLayoutRenderer } from './SceneGridLayoutRenderer';
--- a/packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
+++ b/packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
@@ -19,7 +19,7 @@ interface SceneGridLayoutState extends SceneObjectState {
   isResizable?: boolean;
   isLazy?: boolean;
   /**
-   * fitPanels panels to height of the grid. This will scale down the panels vertically to fit available height.
+   * Fit panels to height of the grid. This will scale down the panels vertically to fit available height.
    * The row height is not changed, only the y position and height of the panels.
    */
   fitPanels?: boolean;
--- a/packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
+++ b/packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
@@ -21,8 +21,9 @@ interface SceneGridLayoutState extends SceneObjectState {
   /**
    * Fit panels to height of the grid. This will scale down the panels vertically to fit available height.
    * The row height is not changed, only the y position and height of the panels.
+   * UNSAFE: This feature is experimental and it might change in the future.
    */
-  fitPanels?: boolean;
+  UNSAFE_fitPanels?: boolean;
   children: SceneGridItemLike[];
 }
 
@@ -311,7 +312,7 @@ export class SceneGridLayout extends SceneObjectBase<SceneGridLayoutState> imple
     // Sort by position
     cells = sortGridLayout(cells);
 
-    if (this.state.fitPanels) {
+    if (this.state.UNSAFE_fitPanels) {
       cells = fitPanelsInHeight(cells, height);
     }
 
--- a/packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
+++ b/packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
@@ -8,7 +8,6 @@ import { SceneGridLayoutRenderer } from './SceneGridLayoutRenderer';
 
 import { SceneGridRow } from './SceneGridRow';
 import { SceneGridItemLike, SceneGridItemPlacement } from './types';
-import { SceneObjectUrlSyncConfig } from '../../../services/SceneObjectUrlSyncConfig';
 import { fitPanelsInHeight } from './utils';
 
 interface SceneGridLayoutState extends SceneObjectState {
@@ -19,13 +18,16 @@ interface SceneGridLayoutState extends SceneObjectState {
   /** Enable or disable item resizing */
   isResizable?: boolean;
   isLazy?: boolean;
-  autoFit?: boolean;
+  /**
+   * fitPanels panels to height of the grid. This will scale down the panels vertically to fit available height.
+   * The row height is not changed, only the y position and height of the panels.
+   */
+  fitPanels?: boolean;
   children: SceneGridItemLike[];
 }
 
 export class SceneGridLayout extends SceneObjectBase<SceneGridLayoutState> implements SceneLayout {
   public static Component = SceneGridLayoutRenderer;
-  protected _urlSync = new SceneObjectUrlSyncConfig(this, { keys: ['autofitpanels'] });
 
   private _skipOnLayoutChange = false;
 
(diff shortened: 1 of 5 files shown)


ivanortegaalba:
Uploading Screenshare - 2024-03-22 3_12_54 PM.mp4…

packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
```diff
@@ -32,6 +36,22 @@ export class SceneGridLayout extends SceneObjectBase<SceneGridLayoutState> imple
...
+
+  public updateFromUrl(values: SceneObjectUrlValues) {
+    const autoFit = values.autofitpanels;
+
+    if (typeof autoFit === 'string') {
+      this.setState({
+        autoFit: true,
+      });
+    }
+  }
```
dprokop, line 53: This kind of feels wrong to have the autofit as a part of the core layour implementation. I think we should solve this as an addon to the `SceneGridLayout`. Maybe we could have a "hook" that one could pass to the ScenegGidLayout via state (sth like `processGridLayout`) that would be called from `buildGridLayout` when you call `cells = fitPanelsInHeight(cells, height);`.

Then that code would become:

```
if (this.state.processGridLayout) {
      cells = processGridLayout(cells, height);
    }
```

And then the Autofit would be a behavior implemented in Grafana core that would provide that layout processin and URL sync. WDYT @ivanortegaalba ?

ivanortegaalba: Yep, that would work too. But I think it would be useful for a non-dashboard app that can be displayed in "TV Mode" or on a screen. So, having an option that says "fitContent" or "fitVertical" as a boolean would be easier since the size of the widget, the margins, the row height and everything else is managed by the core layout. Otherwise we need to export all those measures, and make them configurable, in order to properly configure the layout.

So, IMO, the simplest solution is to keep the manipulation at the core where all the params are set.

natellium:
Just coming in here to ask if this issue is still present in the new implementation https://github.com/grafana/grafana/issues/79313

packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
dprokop: Yes and no IMO - the manipulation is grid units only, building layout does not require any knowledge about measures like row height or gap. 

I see your point, tho I have not seen any asks from app plugin devs to add such an API. Not sure how realistic it is for the apps to be used in the TV mode, given they are way more interactive that a dashboard. 

Is there any way we can at least make the URL sync a core specific behavior? It feels very wrong to grids auto fit behavior to URL by default.

ivanortegaalba:
> Just coming in here to ask if this issue is still present in the new implementation [grafana/grafana#79313](https://github.com/grafana/grafana/issues/79313)

This issue is caused by not taking into account the height taken by the variables toolbar. This won't happen with this new approach, since the viewport is calculated as the size available on the screen, not fixing any height for the navbar or any other item out of the dashboard element. The old architecture, the calculation includes fixed values like the navbar.

packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
torkelo: We could also deprecate this legacy experimental feature and implement it more properly (with flex layout or some other more dynamic layout). 

But agree I do not think the core scenes lib needs to handle it

ivanortegaalba: I am happy to mark it as `UNSAFE_fitPanels` to not invite you to use it, as we did with `UNSAFE_nowDelay`.

But at least we need to port it somehow, don't we?

torkelo: > But I think it would be useful for a non-dashboard app that can be displayed in "TV Mode" or on a screen. 

apps do not use the absolute grid layout

torkelo: I think all use FlexLayout or CSSGridLayout which can easily be made "autofit" with simple CSS

torkelo: but looking closer at the implementation, it is really nice place for the logic to sit in between the runtime grid model and the model we pass to the react-grid-layout  component 

so maybe an unsafe prop is ok

packages/scenes/src/components/layout/grid/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export function fitPanelsInHeight(cells: ReactGridLayout.Layout[], height: number) {
+  // Take into account cell margint top + cell margin bottom + adding some marging at the bottom
+  const visibleHeight = height - GRID_CELL_VMARGIN * 4;
+  const currentGridHeight = Math.max(...cells.map((cell) => cell.h + cell.y));
+
+  const visibleGridHeight = Math.floor(visibleHeight / (GRID_CELL_HEIGHT + GRID_CELL_VMARGIN));
+  const scaleFactor = currentGridHeight / visibleGridHeight;
+
+  console.log('scaleFactor', scaleFactor);
```
mdvictor, line 11: I think we can clean these logs here, right?

packages/scenes/src/behaviors/LiveNowTimer.ts
```diff
@@ -1,51 +1,51 @@
...
-    public constructor(enabled = false) {
-      super({ enabled });
-      this.addActivationHandler(this._activationHandler);
-    }
-  
-    private _activationHandler = () => {
-      if (this.state.enabled) {
-        this.enable();
-      }
+  public constructor({ enabled = false }) {
```
mdvictor, line 14: I think the refactor here has created a bug. Testing this out I can't open grafana:

Screenshot 2024-04-01 at 17 28 59

I think the constructor should be:
`public constructor(enabled = false)` not
` public constructor({ enabled = false })`

mdvictor (changes requested):
I think there was an unintentional change maybe to the timer constructor?

mdvictor (approved):
LGTM! Just need to remove those console.logs in `utils.ts` and I think it's good to go!

packages/scenes/src/components/layout/grid/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+  console.log('y');
```
ivanortegaalba, line 12:
```suggestion
```

packages/scenes/src/components/layout/grid/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+    console.log('cell.h', cell.h);
+    console.log('Math.round(cell.y / scaleFactor)', Math.round(cell.y / scaleFactor));
+    console.log('Math.round(cell.h / scaleFactor)', Math.round(cell.h / scaleFactor));
+    return {
+      ...cell,
+      y: Math.round(cell.y / scaleFactor) || 0,
+      h: Math.round(cell.h / scaleFactor) || 1,
+    };
+  });
+}
```
ivanortegaalba, line 24:
```suggestion
export function fitPanelsInHeight(cells: ReactGridLayout.Layout[], height: number) {
  // Take into account cell margint top + cell margin bottom + adding some marging at the bottom
  const visibleHeight = height - GRID_CELL_VMARGIN * 4;
  const currentGridHeight = Math.max(...cells.map((cell) => cell.h + cell.y));

  const visibleGridHeight = Math.floor(visibleHeight / (GRID_CELL_HEIGHT + GRID_CELL_VMARGIN));
  const scaleFactor = currentGridHeight / visibleGridHeight;

  return cells.map((cell) => {
    return {
      ...cell,
      y: Math.round(cell.y / scaleFactor) || 0,
      h: Math.round(cell.h / scaleFactor) || 1,
    };
  });
}
```

PR #659, opened 2024-03-25
VizPanel: Do not apply the visualization's field config to annotation data frames

javiruiz01:
- Avoid adding default `fieldConfig` to annotations, follow what we do in grafana ([link](https://github.com/grafana/grafana/blob/2f5be54252b860925997249163b5f6c4f81acc1c/packages/grafana-data/src/field/fieldOverrides.ts#L600))

Why? When adding links to a panel like so:
```txt
fieldConfig: {
  defaults: {
    ...
    links: [
      {some link}
    ],
  },
},
```

The links were getting added not only to the series, but also to each field in the annotations, resulting in the annotation having repeated links that did not belong there.

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanel.test.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.test.tsx
@@ -496,6 +496,41 @@ describe('VizPanel', () => {
       expect(dataToRender.alertState).toBe(testData.alertState);
       expect(dataToRender.annotations).toBeDefined();
     });
+
+    it('should not add fieldConfig to annotations', async () => {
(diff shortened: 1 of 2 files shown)


torkelo (commented):
Wont this undo https://github.com/grafana/scenes/pull/540 ?

javiruiz01:
Actually it still works since we're still running the field overrides, it's just that we won't be applying the default config to the annotations, before we were not running `applyFieldOverrides` at all, that's why it was breaking

javiruiz01:
I updated the test by adding a link to the annotations test data to ensure we're not losing it after calling `applyFieldOverrides`

torkelo (approved):

PR #660, opened 2024-03-27
Check if target datasource is matching the panel ds, if not set the target ds to panel ds.

oscarkilhed:
In some circumstances using datasource variables the target datasource is not matched with the panel datasource. This causes issues for datasources using `DataSourceWithBackend`.

This makes sure we set the target datasource to the correct datasource.


dprokop:
@oscarkilhed - I've added release/patch labels.

PR #662, opened 2024-03-29
LiveNowTimer: The constructor doesn't receive a valid state

ivanortegaalba:
**Problem**
Using a boolean as a scene object state was causing errors where "enable" in the state was the state itself. This was discovered while creating tests for this PR. This state appears when calling `enable()` method even when using properly the initialization. This is because SceneObjectState expects to receive as parameter always the scene state

Captura de pantalla 2024-03-29 a las 17 27 01

**Solution**
The constructor parameters should be state timer. Also create some tests


packages/scenes/src/behaviors/LiveNowTimer.ts
ivanortegaalba, line 1: This is formatting the file, it was not formatted with Prettier.

PR #664, opened 2024-04-01
AdHocFiltersVariable: Pass scene queries to getTagValues calls

bfmatei:
Depends on https://github.com/grafana/grafana/pull/85436

This passes scene queries to getTagValues calls so that data source can decide which dimensions are applicable in a scene given provided queries context.

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -239,7 +239,9 @@ export class AdHocFiltersVariable
     const otherFilters = this.state.filters.filter((f) => f.key !== filter.key).concat(this.state.baseFilters ?? []);
 
     const timeRange = sceneGraph.getTimeRange(this).state.value;
-    let values = await ds.getTagValues({ key: filter.key, filters: otherFilters, timeRange });
+    const queries = this._getSceneQueries();
+    // @ts-expect-error TODO: remove this once 10.4.2 is released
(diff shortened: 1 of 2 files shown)


dprokop (approved):

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -239,7 +239,9 @@ export class AdHocFiltersVariable
     const otherFilters = this.state.filters.filter((f) => f.key !== filter.key).concat(this.state.baseFilters ?? []);
 
     const timeRange = sceneGraph.getTimeRange(this).state.value;
-    let values = await ds.getTagValues({ key: filter.key, filters: otherFilters, timeRange });
+    const queries = this._getSceneQueries();
+    // @ts-expect-error TODO: remove this once 10.4.2 is released
```
ashharrison90, line 243: currently the [other PR](https://github.com/grafana/grafana/pull/85436) isn't being backported, so that will actually release in 11.1.0. if you want it in 10.4.2, you'll have to add the `backport 10.4.x` label and request it as a critical backport. if you want it in 11, it should be enough to just add `backport 11.0.x` label.

ashharrison90 (approved):
lgtm 👍

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
bfmatei: Will update the description then. Thanks!

PR #666, opened 2024-04-02
Fix issue where curly braces are always added to IntervalMacro. 

oscarkilhed:
In the old architecture `$__interval` was passed to the backend as `$__interval`. With scenes this changed so that we're always adding curly braces e.g. `${__interval}`. The Influx db datasource does not handle this well. So for backwards compatibility we should probably keep the variable as is, without the braces.

<details>
  <summary>📦 Published PR as canary version: <code>4.1.1--canary.666.8521778570.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@4.1.1--canary.666.8521778570.0
  # or 
  yarn add @grafana/scenes@4.1.1--canary.666.8521778570.0
  ```
</details>


dprokop (approved):
Nice! added release labels so that we can update core - remember to backport the bump.

Other than that, there are some tests failing

PR #669, opened 2024-04-03
Disallow drag and drop a row inside another uncollapsed row

mdvictor:
https://github.com/grafana/scenes/assets/36818606/62a14da1-dd64-4a4f-87cc-05ce0405666a

This PR fixes the dashboard crashing when a row is dragged within the area of another uncollapsed row. It's not the most elegant solution, so I'm open to feedback if there is a better way to achieve this, but I think this is the simplest way to fix the bug without adding a lot of complexity, knowing that this will need to be revisited having invalidation in mind.

I think this needs to be further worked on, especially on UX side of things to let users know that they can't have nested rows (e.g.: highlight the uncollapsed row area, similar to how we do to repeated panels and disallow dropping if dragged element is a row). My knowledge of RGL is limited and I couldn't find a proper way to invalidate a drop, which makes me believe that the only way to do it would be to update the layout with the invalid state and then re-update with the old state.

The idea here is to return early if we detect after a drag and drop that we are in a row within a row scenario and force render the grid layout with the old layout. Since there is no actual layout change in the grid props, we need to force the render through the grid layout key prop.

TODO: tests

### DIFF
--- a/packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
+++ b/packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
@@ -11,7 +11,7 @@ import { SceneGridItemLike, SceneGridItemPlacement } from './types';
 
 interface SceneGridLayoutState extends SceneObjectState {
   /**
-   * Turn on or off dragging for all items. Indiviadual items can still disabled via isDraggable property
+   * Turn on or off dragging for all items. Individual items can still disabled via isDraggable property
    **/
   isDraggable?: boolean;
   /** Enable or disable item resizing */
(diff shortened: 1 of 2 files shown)


mdvictor:
I'm still working on an alternative where we would store the old children state `onDragStart` and return early in `onDragStop` if state is invalid which would call `onLayoutChange` and if an old state is present it would reset to that, which is still flickering, but I'm still looking into it.

mdvictor:
https://github.com/grafana/scenes/assets/36818606/71d106c3-89a0-486c-886b-f820c18b0199

Might've found a better way of doing this with what I mentioned above that doesn't flicker 🎉  Will push that to another branch as well

PR #670, opened 2024-04-03
AdHoc Filters: Allow custom value

ashharrison90:
- allow specifying a custom value as the adhoc filter value
- adds a unit test to ensure we preserve this behaviour

image

Fixes https://github.com/grafana/grafana/issues/85070

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
@@ -44,6 +44,7 @@ export function AdHocFilterRenderer({ filter, model }: Props) {
 
   const valueSelect = (
     <Select
+      allowCustomValue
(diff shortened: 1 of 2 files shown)


ashharrison90:
> Looks good. Question is if reloading the page with the custom value in URL will allow the value to be preserved? I think it's essential to make sure the custom value is persisted across reloads.

yep, reloading the page correctly restores the custom value 👍

dprokop (approved):
Works well for Prometheus :) 

image

image


I wonder - is there any way in the Select that would allow us to edit the manually entered filter value? I think it would be very helpful for scenarios when one makes a type.

ashharrison90:
@dprokop i don't think there's a prop on the component that would let us do that. i know exactly what you mean, that could get annoying. i'm sure we were discussing this in frontend platform the other day. i'll do some searching

PR #671, opened 2024-04-03
Disallow drag and drop rows within uncollapsed rows

mdvictor:
https://github.com/grafana/scenes/assets/36818606/7487da9c-bc39-4992-98bc-95c96dc9c7d8

This is a different approach to https://github.com/grafana/scenes/pull/669 that fixes https://github.com/grafana/grafana/issues/83916.  

It involves saving the initial layout when a drag action starts and checks on drag stop if the state is invalid (e.g.: a row is within a row). If layout is invalid we flag it so that when `onLayoutChange` is called we load the old layout and revert to it. I'm not entirely happy with this solution but think this is the best approach for now to get this to the finish line.

- In the future, I think we can try to wrap the entire uncollapsed row with the `GridItemWrapper` to clearly highlight it for row dragging, although I'm not sure how panels dragging would work in this case. It would need to resize and allow dropping a panel in the row layout, but not another row. Maybe a grid within a grid? I'll look into this when I have some spare time

- Another option would be similar to the old architecture where we always keep `SceneGridRow`s parent as the layout and move the children below it from the old row to the new one. In the old arch everything is at the same level so you can easily drag and drop a row anywhere.
For the following layout:
```
-- row-a
   -- panel-a
   -- panel-b
   -- panel-c
--row-b
```

If we drag `row-b` before `panel-c`:
```
-- row-a
   -- panel-a
   -- panel-b
-- row-b
   -- panel-c
```
When we uncollapse `row-b`, `panel-c` gets picked up as a child of `row-b`. Although this UX feels a bit confusing.

### DIFF
--- a/packages/scenes/src/components/layout/grid/SceneGridLayout.test.tsx
+++ b/packages/scenes/src/components/layout/grid/SceneGridLayout.test.tsx
@@ -197,7 +197,7 @@ describe('SceneGridLayout', () => {
     });
   });
 
-  describe('when moving a panel', () => {
+  describe('when moving a panel or row', () => {
     it('should update layout children placement and order ', () => {
       const layout = new SceneGridLayout({
         children: [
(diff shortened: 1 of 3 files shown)


kaydelaney (approved):

dprokop (approved):
I think this is good approach, the easies thing I can think of. 

Agree with you that eventually the entire uncollapsed row could be considered an individual grid item when dragging a row into it so that the UI wouldn't even suggest one can drop a row within a row. Let's tackle this as a separate improvement tho.

torkelo (approved):
I think this look ok, if we can it would be nice with a unit test for this scenario (As we already have unit tests for onDragStop) 

> In the future, I think we can try to wrap the entire uncollapsed row with the GridItemWrapper to clearly highlight it for row dragging, although I'm not sure how panels dragging would work in this case. It would need to resize and allow dropping a panel in the row layout, but not another row. Maybe a grid within a grid? I'll look into this when I have some spare time

We have tried a few times to do grids within grids but so far always failed with issues with drag an drop and resize etc, but this would obviously be the best solution if we can get it to work :) 

> Another option would be similar to the old architecture where we always keep SceneGridRows parent as the layout and move the children below it from the old row to the new one

Not sure what you mean?

PR #673, opened 2024-04-03
feat: function `clearSceneApp` to de-cache initialized scene apps so that they can be fully reinitialized

darrenjaneczek:
This supports a use case to completely refresh a scene app when in the event of an app plugin configuration change.
There are many ways to solve this problem, but this may be a useful option to speed up developer experience, to be used when a complete re-init of a scene app is acceptable UX.


darrenjaneczek:
Closed. I think we're going to try to encourage better use of SceneVariables so that these brute force approaches aren't necessary.

>I think it might be better to think how to invalidate this specific data / metric names., instead of the whole app

PR #682, opened 2024-04-08
Chore: Bump various dependencies

kaydelaney:
Brings frontend errors when linking locally down from 186 to 27


dprokop:
Nice, but seems like some tests are failing? Would be also good to ping some app teams (app o11y, k8s etc) so that they could test if those updated dependence bumps (especialy peer deps in scenes package) do not cause any issues in the app plugins.

Issue #214, opened 2023-05-25
Variables: Allow lazy loading of variable values


dprokop:
Set a default value that does not need to be validated on the dashboard load and load options lazily when select opens. 
Should speed up the dashboard load significantly.

PR #687, opened 2024-04-10
GroupBy: Fetch options when opening menu

### LINKED ISSUE #686: GroupBy variable: It may not receive the latest set of queries available in a scene
GroupBy variable, as other `MultiValueVariable`, provides available options (dimensions) on initialisation. 

Since the available values may depend on queries from the scene, the current implementatio. may be problematic. I.e. in the core Dashboards scenario adding a panel or editing queries in any panel does not invalidate available dimensions in the Group by variable picker. 

I think we need to change picker implementation to one that's similar to adhoc filters, whenever the picker is opened the options are resolved.

ashharrison90:
i think this does what we want, but leaving as draft for now to check the approach before adding tests etc

what this does:
- takes the rendering logic from `VariableValueSelectMulti`
- enhances with `isFetchingOptions` and `isOptionsOpen` logic controlled by `onOpenMenu`/`onCloseMenu`
- fetches and waits for new options when opening the menu via `model.validateAndUpdate()`

some questions:
- does this logic only apply to group by, or would we want it for any multi value variable?
  - if so, could move this new logic back into `VariableValueSelectMulti` 🤔 

Fixes #686 

<details>
  <summary>📦 Published PR as canary version: <code>4.11.2--canary.687.8737546972.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@4.11.2--canary.687.8737546972.0
  # or 
  yarn add @grafana/scenes@4.11.2--canary.687.8737546972.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/groupby/GroupByVariable.tsx
+++ b/packages/scenes/src/variables/groupby/GroupByVariable.tsx
@@ -1,3 +1,4 @@
+import React, { useEffect, useMemo, useState } from 'react';
 import { AdHocVariableFilter, DataSourceApi, MetricFindValue } from '@grafana/data';
 import { allActiveGroupByVariables } from './findActiveGroupByVariablesByUid';
 import { DataSourceRef, VariableType } from '@grafana/schema';
@@ -6,9 +7,10 @@ import { DataQueryExtended, SceneQueryRunner } from '../../querying/SceneQueryRu
 import { sceneGraph } from '../../core/sceneGraph';
 import { ValidateAndUpdateResult, VariableValueOption, VariableValueSingle } from '../types';
 import { MultiValueVariable, MultiValueVariableState, VariableGetOptionsArgs } from '../variants/MultiValueVariable';
-import { from, map, mergeMap, Observable, of, take } from 'rxjs';
+import { from, lastValueFrom, map, mergeMap, Observable, of, take } from 'rxjs';
 import { getDataSource } from '../../utils/getDataSource';
-import { renderSelectForVariable } from '../components/VariableValueSelect';
+import { InputActionMeta, MultiSelect } from '@grafana/ui';
+import { isArray } from 'lodash';
 
 export interface GroupByVariableState extends MultiValueVariableState {
   /** Defaults to "Group" */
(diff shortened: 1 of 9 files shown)


dprokop:
@ashharrison90 - I've made a couple of changes to this PR. In particular:

- I've added `isLazy` to `SceneVariableState`. Thanks to that and the change in SceneVariableSet the GroupBy values are not fetched when the picker renders. Overall this property needs to be added to other variables as well, but this is a bit bigger endeavour and won't address it here. We are tracking it in https://github.com/grafana/scenes/issues/214
-  I've added tests to the code.
- I've also included a followup fix to https://github.com/grafana/scenes/pull/685 as that one did not respect the fact, that there may be inactive runners that we still want the queries from- i.e. for panels below the fold. Overall, I have implemented sort of core dashboard specific code, that ignres inactive runners with the same key. This reflects a scenario when we clone a panel in panel edit - the key of the SceneQueryRunner is the same, but the original one is not active. It's probably not the most elegant solution, but nothing else came to my mind - I'm open for any other ideas.

@torkelo  - I would appreciate your eyes on this too, as I'm touching on the `isLazy` we discussed and introducing it in the interface for the future use.

packages/scenes/src/variables/types.ts
```diff
@@ -14,6 +14,7 @@ export interface SceneVariableState extends SceneObjectState {
   loading?: boolean;
   error?: any | null;
   description?: string | null;
+  isLazy?: boolean;
```
dprokop, line 17: Thinking about putting this in the `SceneVariable` interface rather than in the `SceneVariableState`, don't think that would ever be needed to dynamically modify the nature of a variable during runtime. Thoughts?

torkelo: @dprokop would we not need it in state (to make it an persisted & UI option for QueryVariable?

packages/scenes/src/variables/sets/SceneVariableSet.ts
```diff
@@ -67,7 +67,7 @@ export class SceneVariableSet extends SceneObjectBase<SceneVariableSetState> imp
 
     // Add all variables that need updating to queue
     for (const variable of this.state.variables) {
-      if (this._variableNeedsUpdate(variable)) {
+      if (!variable.state.isLazy && this._variableNeedsUpdate(variable)) {
```
torkelo, line 70: can we move this `!variable.state.isLazy` condition to _variableNeedsUpdate ?

packages/scenes/src/variables/groupby/GroupByVariable.tsx
```diff
@@ -178,5 +182,65 @@ export class GroupByVariable extends MultiValueVariable<GroupByVariableState> {
...
+    ? (value: string, meta: InputActionMeta) => {
+        if (meta.action === 'input-change') {
+          model.onSearchChange!(value);
+        }
+      }
+    : undefined;
+
+  const placeholder = 'Select value';
+
+  return (
```
torkelo, line 208: Any specific reason not doing this in the renderSelectForVariable? limit potential for bugs?

ashharrison90 (commented):
feels weird reviewing a PR where i'm marked as author 😂 

think this looks good 👍

packages/scenes/src/variables/types.ts
dprokop: Yes, it would, but... I think we can lift it to the public variable API (SceneVariableState) when we work on the general problem of variables lazy loading (ref https://github.com/grafana/scenes/issues/214). I'm afraid putting it in the state right now might be confusing for consumers, as this is now an exclusive feature of Group By. I.e. configuring Query variable as lazy now makes it not usable.

packages/scenes/src/variables/groupby/GroupByVariable.tsx
dprokop: Similar to my comment above - i think we need to tackle this with https://github.com/grafana/scenes/issues/214 as currently the selects rendered for variables do not have `onOpenMenu` handler implemented. I've tried to change it so thast it uses the `renderSelectForVariable` but this is a bigger effort that we need to address both in `SceneVariableSet` and the UI for variables.

packages/scenes/src/variables/sets/SceneVariableSet.ts
dprokop: Yes, I think we can. I initially thought this is called in update code paths, but it's called only for variables initialisation so it's OK to move it there. Addressed in 4a5f0db0578e4d6ad7173d2419ce0d412c9fecef

packages/scenes/src/variables/groupby/GroupByVariable.tsx
torkelo: > but this is a bigger effort that we need to address both in SceneVariableSet and the UI for variables.

what more changes are needed in SceneVariableSet and UI for variables?

torkelo (approved):

bfmatei (approved):
lgtm 🚀

ashharrison90:
delivered in core with https://github.com/grafana/grafana/pull/86671

PR #688, opened 2024-04-10
SceneObjectBase: useState add options that make the new activation behavior optional

torkelo:
Changes https://github.com/grafana/scenes/pull/679 and makes this new behavior optional & opt-in

### DIFF
--- a/packages/scenes/src/core/SceneObjectBase.tsx
+++ b/packages/scenes/src/core/SceneObjectBase.tsx
@@ -12,6 +12,7 @@ import {
   SceneDeactivationHandler,
   CancelActivationHandler,
   SceneObjectState,
+  UseStateHookOptions,
 } from './types';
 
 import { SceneComponentWrapper } from './SceneComponentWrapper';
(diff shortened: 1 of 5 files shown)


packages/scenes/src/core/SceneObjectBase.tsx
```diff
@@ -279,9 +280,9 @@ export abstract class SceneObjectBase<TState extends SceneObjectState = SceneObj
   /**
    * Utility hook to get and subscribe to state
    */
-  public useState() {
+  public useState(options?: UseStateHookOptions) {
     // eslint-disable-next-line react-hooks/rules-of-hooks
-    return useSceneObjectState(this);
+    return useSceneObjectState(this, options);
```
dprokop, line 285: @torkelo  Alternative  idea since we are talking an edge case here. How about NOT having `useState(options)` but instead use `useSceneObjectState` directly.

So in the VariableValueSelector this line:

```
  const state = variable.useState({ shouldActivateOrKeepAlive: true });
```

would be 

```
const state = useSceneObjectState(variable, { shouldActivateOrKeepAlive: true });
```

torkelo: @dprokop yea, agree actually started that way. But we do not export the standalone function. 

But let's do that (export it), and only add the options argument to the standalone function

dprokop: thanks <3

dprokop (approved):
Looks good and work fine.

@torkelo- let's update or remove the release notes, the are no longer correct.

PR #691, opened 2024-04-15
LocalValueVariable: Do not throw when missing parent variable

torkelo:
Alternative to https://github.com/grafana/grafana/pull/85934 

I think if we have to do that try catch there, might just be simpler to support this scenario in LocalValueVariable


mdvictor (approved):
Tested and LGTM

dprokop (approved):

PR #697, opened 2024-04-16
Remove sceneObject from scoped vars

bfmatei:
Remove `__sceneObject` from `scopedVars`. This is causing various issues with datasources that are stringifing the request object for various reasons.

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.test.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.test.ts
@@ -162,7 +162,6 @@ describe('SceneQueryRunner', () => {
 
       expect(Object.keys(scopedVars)).toMatchInlineSnapshot(`
         [
-          "__sceneObject",
           "__interval",
           "__interval_ms",
         ]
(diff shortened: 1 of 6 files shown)


torkelo (commented):
@bfmatei this would break scenes compatability with legacy template system (templateSrv). 

We rely on this scene object reference in calls to templateSrv.replace 

https://github.com/grafana/grafana/blob/main/public/app/features/templating/template_srv.ts#L251

Many data sources pass scopedVars to templateSrv.replace and this is the way we know the scene object context so we can lookup the right scene scoped vars. 

https://github.com/grafana/grafana/blob/main/packages/grafana-prometheus/src/variables.ts#L44

https://github.com/grafana/grafana/blob/main/packages/grafana-runtime/src/utils/DataSourceWithBackend.ts#L154 

countless others

dprokop:
@torkelo yeah, you are totally. right, I've missed the fact that `query` method can call `templateSrv.replace`, damn. 

We are trying to solve an issue with circular JSON serialisation in Zabbix plugin (they are building a string hash from the DataQueryRequest object by using JSON.stringify). We can of course fix it in Zabbix, but it feels like some other data sources may do a similar thing. 

But anyways, I am afraid in the end it's the Zabbix where we need to provide the fix for this issue. Sorry @bfmatei for confusion!

PR #699, opened 2024-04-17
Add method to get ancestor for localValueVariable

mdvictor:
In certain scenarios (e.g.: https://github.com/grafana/grafana/issues/84915) where we repeat a panel under a repeated row using the same variable, the `DashboardGridItem._performRepeat` looks up and finds the variable in the grid row, since they have the same name, and since that is a `LocalValueVariable` it cannot perform the repeat with it. 

With this method we can get the parent variable of the `LocalValueVariable` and use that to repeat.

Needed by https://github.com/grafana/grafana/pull/86425

<details>
  <summary>📦 Published PR as canary version: <code>4.11.1--canary.699.8720696583.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@4.11.1--canary.699.8720696583.0
  # or 
  yarn add @grafana/scenes@4.11.1--canary.699.8720696583.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/variants/LocalValueVariable.ts
+++ b/packages/scenes/src/variables/variants/LocalValueVariable.ts
@@ -60,4 +60,19 @@ export class LocalValueVariable
 
     return false;
   }
+
+  public getAncestorVariable(): SceneVariable | null {
+    // Parent (SceneVariableSet) -> Parent (The object that has our parent set) -> Parent (scope we need to access our sets ancestor)
+    const ancestorScope = this.parent?.parent?.parent;
+    if (!ancestorScope) {
+      throw new Error('LocalValueVariable requires a parent SceneVariableSet that has an ancestor SceneVariableSet');
+    }
+
+    const parentVar = sceneGraph.lookupVariable(this.state.name, ancestorScope);
+    if (!parentVar) {
+      return null;
+    }
+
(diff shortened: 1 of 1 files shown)


torkelo:
@mdvictor what would be the point of this scenario?

I do not think we should support it. 

If the row is repeated by variable  A, and you repeat a panel inside that row it should only see the local scoped value.

That is how repeating works.

mdvictor:
@torkelo I'm also wondering if this should be supported. It might be an edge case to use the same var for both panels and rows. 

My only argument for supporting this would be parity with the old arch, where you can have a dashboard with repeated panels inside repeated rows under the same variable and they will both update when that variable changes.

In the new arch the repeated panels will not update because the type is LocalValueVariable and it only accepts MultiValueVariable. 

Do we not want to support this specific implementation or the scenario at all?

mdvictor:
Okay, so my understanding is that we should support the scenario, but using the local value, not searching up the tree. I think that should work? I saw that LocalValueVars have an `isMulti` property, so maybe that should work. I'll look into it

torkelo:
> Just making sure I understand right: We do not want to support the scenario of having a repeated row and a repeated panel under it using the same variable?

@mdvictor , yea this scenarios does not make sense as the variable will only have single value under the row

PR #705, opened 2024-04-23
GroupByVariable: Sync label to URL

bfmatei:
This PR syncs labels of the values in the group by variable similarly to what we introduced for ad-hoc filters.

Unlike ad-hoc filters, group by is a little bit more of a special case hence the reason why some of the changes below. The solution might not be optimal, so please feel free to suggested changes and improvements.

Changes:
- introduced custom URL sync handler
   - it works similarly to the ad-hoc filters one
   - this is needed as currently we use the one multi-value variable one and we don't want this changed for all variables
- the state of group by variable received a new prop: `urlOptions` where we store the options that we received via the URL
   - when fetching the options, we concat the `urlOptions` with what we receive from the datasource (or with the `defaultOptions`)
   - this is needed because of `validateAndUpdate` and also because `changeValueTo` is sometimes called without the texts array
- we leverage the fact that the value of the `MultiSelect` can be a `SelectableValue` hence we can enforce the labels on values
- moved some common functions to variables utils

The gotcha: let's assume we pass a value via URL. This value is not returned by the datasource hence we consider it a custom value. We then remove the custom value using the UI button. When opening the dropdown again, the custom value can be found in the dropdown. Is this a behavior we want to keep? You can also see this behavior in the demo below.

PS: tests will come in a subsequent push. For the time being I'm trying to decide upon edge cases.

Demo:

https://github.com/grafana/scenes/assets/9215315/54ca86b9-9db9-4ac1-ac8e-144a8cd576f2

<details>
  <summary>📦 Published PR as canary version: <code>4.20.0--canary.705.9031064372.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@4.20.0--canary.705.9031064372.0
  # or 
  yarn add @grafana/scenes@4.20.0--canary.705.9031064372.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/groupby/GroupByVariable.tsx
+++ b/packages/scenes/src/variables/groupby/GroupByVariable.tsx
@@ -9,7 +9,7 @@ import { MultiValueVariable, MultiValueVariableState, VariableGetOptionsArgs } f
 import { from, lastValueFrom, map, mergeMap, Observable, of, take } from 'rxjs';
 import { getDataSource } from '../../utils/getDataSource';
 import { InputActionMeta, MultiSelect } from '@grafana/ui';
-import { isArray, unionBy } from 'lodash';
+import { isArray } from 'lodash';
 import { getQueriesForVariables } from '../utils';
 import { OptionWithCheckbox } from '../components/VariableValueSelect';
 import { GroupByVariableUrlSyncHandler } from './GroupByVariableUrlSyncHandler';
@@ -20,8 +20,6 @@ export interface GroupByVariableState extends MultiValueVariableState {
   /** The visible keys to group on */
   // TODO review this type and name (naming is hard)
   defaultOptions?: MetricFindValue[];
-  /** Options obtained through URL sync */
-  urlOptions?: MetricFindValue[];
   /** Base filters to always apply when looking up keys */
   baseFilters?: AdHocVariableFilter[];
   /** Datasource to use for getTagKeys and also controls which scene queries the group by should apply to */
@@ -85,12 +83,10 @@ export class GroupByVariable extends MultiValueVariable<GroupByVariableState> {
   }
 
   public getValueOptions(args: VariableGetOptionsArgs): Observable<VariableValueOption[]> {
-    const urlOptions = this.state.urlOptions ?? [];
-
     // When default dimensions are provided, return the static list
     if (this.state.defaultOptions) {
       return of(
-        unionBy(this.state.defaultOptions, urlOptions, 'value').map((o) => ({
+        this.state.defaultOptions.map((o) => ({
           label: o.text,
           value: String(o.value),
         }))
@@ -108,7 +104,7 @@ export class GroupByVariable extends MultiValueVariable<GroupByVariableState> {
         return from(this._getKeys(ds)).pipe(
           take(1),
           mergeMap((data: MetricFindValue[]) => {
-            const a: VariableValueOption[] = unionBy(data, urlOptions, 'value').map((i) => {
+            const a: VariableValueOption[] = data.map((i) => {
               return {
                 label: i.text,
                 value: i.value ? String(i.value) : i.text,
@@ -252,7 +248,10 @@ export function GroupByVariableRenderer({ model }: SceneComponentProps<MultiValu
       components={{ Option: OptionWithCheckbox }}
       onInputChange={onInputChange}
       onBlur={() => {
-        model.changeValueTo(uncommittedValue.map((x) => x.value!));
+        model.changeValueTo(
+          uncommittedValue.map((x) => x.value!),
+          uncommittedValue.map((x) => x.label!)
+        );
       }}
       onChange={(newValue, action) => {
         if (action.action === 'clear' && noValueOnClear) {
--- a/packages/scenes/src/variables/groupby/GroupByVariable.tsx
+++ b/packages/scenes/src/variables/groupby/GroupByVariable.tsx
@@ -252,7 +252,7 @@ export function GroupByVariableRenderer({ model }: SceneComponentProps<MultiValu
       components={{ Option: OptionWithCheckbox }}
       onInputChange={onInputChange}
       onBlur={() => {
-        model.changeValueTo(uncommittedValue.map((x) => x.value));
+        model.changeValueTo(uncommittedValue.map((x) => x.value!));
       }}
       onChange={(newValue, action) => {
         if (action.action === 'clear' && noValueOnClear) {
--- a/packages/scenes/src/variables/groupby/GroupByVariable.tsx
+++ b/packages/scenes/src/variables/groupby/GroupByVariable.tsx
@@ -1,17 +1,18 @@
 import React, { useEffect, useMemo, useState } from 'react';
-import { AdHocVariableFilter, DataSourceApi, MetricFindValue } from '@grafana/data';
+import { AdHocVariableFilter, DataSourceApi, MetricFindValue, SelectableValue } from '@grafana/data';
 import { allActiveGroupByVariables } from './findActiveGroupByVariablesByUid';
 import { DataSourceRef, VariableType } from '@grafana/schema';
-import { SceneComponentProps, ControlsLayout } from '../../core/types';
+import { SceneComponentProps, ControlsLayout, SceneObjectUrlSyncHandler } from '../../core/types';
 import { sceneGraph } from '../../core/sceneGraph';
 import { ValidateAndUpdateResult, VariableValueOption, VariableValueSingle } from '../types';
 import { MultiValueVariable, MultiValueVariableState, VariableGetOptionsArgs } from '../variants/MultiValueVariable';
 import { from, lastValueFrom, map, mergeMap, Observable, of, take } from 'rxjs';
 import { getDataSource } from '../../utils/getDataSource';
 import { InputActionMeta, MultiSelect } from '@grafana/ui';
-import { isArray } from 'lodash';
+import { isArray, unionBy } from 'lodash';
 import { getQueriesForVariables } from '../utils';
 import { OptionWithCheckbox } from '../components/VariableValueSelect';
+import { GroupByVariableUrlSyncHandler } from './GroupByVariableUrlSyncHandler';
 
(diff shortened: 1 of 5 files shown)


packages/scenes/src/variables/groupby/GroupByVariable.tsx
```diff
@@ -79,10 +84,12 @@ export class GroupByVariable extends MultiValueVariable<GroupByVariableState> {
   }
 
   public getValueOptions(args: VariableGetOptionsArgs): Observable<VariableValueOption[]> {
+    const urlOptions = this.state.urlOptions ?? [];
+
```
torkelo, line 88: confused, thought group by variables are lazy so we don't call this function unless when the dropdown opens. So why do we need URL options here? 

https://github.com/grafana/scenes/pull/687

packages/scenes/src/variables/groupby/GroupByVariableUrlSyncHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+      // Fail-safe as this cannot be type-checked properly
+      values = values ?? [];
+      labels = labels ?? [];
+
+      this._sceneObject.setState({
+        urlOptions: values.map((value, idx) => ({
+          value,
+          text: labels[idx],
+        })),
+      });
```
torkelo, line 60: unsure why we need this

packages/scenes/src/variables/groupby/GroupByVariableUrlSyncHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+      /**
+       * Initial URL Sync happens before scene objects are activated.
+       * We need to skip validation in this case to make sure values set via URL are maintained.
+       */
+      if (!this._sceneObject.isActive) {
+        this._sceneObject.skipNextValidation = true;
+      }
+
+      urlValue = Array.isArray(urlValue) ? urlValue : [urlValue];
+      let [values, labels] = unzip(urlValue.map((value) => (value ? value.split(',') : [value])));
```
torkelo, line 49: should there not be a kind unescaping here (when the label or value contains a comma (not as delimiter but the real value) )

packages/scenes/src/variables/groupby/GroupByVariableUrlSyncHandler.ts
```diff
@@ -0,0 +1,76 @@
+import { unzip, zip } from 'lodash';
+import { SceneObjectUrlSyncHandler, SceneObjectUrlValues } from '../../core/types';
+import { GroupByVariable } from './GroupByVariable';
+import { toUrlCommaDelimitedString } from '../utils';
+
+export class GroupByVariableUrlSyncHandler implements SceneObjectUrlSyncHandler {
```
torkelo, line 6: Maybe we can reuse MultiValueUrlSyncHandler but with an option to sync label? since they are so close to identical

packages/scenes/src/variables/groupby/GroupByVariable.tsx
bfmatei: There is also `validateAndUpdate` which calls `getValueOptions`. 

Sometimes `changeValueTo` is called without `texts` and `this.state.options` is used to determine the value of `texts`. Check this out: https://github.com/grafana/scenes/blob/main/packages/scenes/src/variables/variants/MultiValueVariable.ts#L257

packages/scenes/src/variables/groupby/GroupByVariableUrlSyncHandler.ts
bfmatei: Fixed!

packages/scenes/src/variables/groupby/GroupByVariableUrlSyncHandler.ts
bfmatei: Fixed!

packages/scenes/src/variables/groupby/GroupByVariableUrlSyncHandler.ts
bfmatei: Yup, I can merge them into one. I'll implement it in a subsequent commit.

packages/scenes/src/variables/groupby/GroupByVariableUrlSyncHandler.ts
bfmatei: We need the `urlOptions` in order not to lose the label of a custom value when we select an item from the available options.

Having them as `MetricFindValue` as opposed to a map or even `SelectableValue` directly makes it easier to reuse it without additional processing in the actual variable.

packages/scenes/src/variables/groupby/GroupByVariableUrlSyncHandler.ts
bfmatei: I started working on this but to be honest it kinda makes it weird to bring `urlOptions` to multi-value variable. WDYT?

torkelo: @bfmatei I think the whole concept of urlOptions is strange, need to figure out a way we can avoid it. 

the URL state should only be concerned with setting the current value/text not the options

packages/scenes/src/variables/groupby/GroupByVariableUrlSyncHandler.ts
torkelo: > We need the urlOptions in order not to lose the label of a custom value when we select an item from the available options.

not sure I follow, what do you mean "loose the label of a custom value" ? a custom value can never have different value/label

packages/scenes/src/variables/groupby/GroupByVariable.tsx
torkelo: but how does they relate to url sync?

packages/scenes/src/variables/groupby/GroupByVariableUrlSyncHandler.ts
```diff
@@ -0,0 +1,77 @@
...
+      /**
+       * Initial URL Sync happens before scene objects are activated.
+       * We need to skip validation in this case to make sure values set via URL are maintained.
+       */
+      if (!this._sceneObject.isActive) {
+        this._sceneObject.skipNextValidation = true;
+      }
+
+      urlValue = Array.isArray(urlValue) ? urlValue : [urlValue];
+      const valuesLabelsPairs = urlValue.map((value) => (value ? value.split(',') : [value]));
```
torkelo, line 50: Would like to change this so that we implement it for all variables with value/label (ie in default MultiValueVariable), and only sync label when value != label.  

I am worried though, this could be a breaking change, all current saved URLs with variable values with commas would break?  (I assume new URL values we escape commas, but not existing saved links)

packages/scenes/src/variables/groupby/GroupByVariable.tsx
dprokop: I'm confused by this as well, we never pass available options via URL.

bfmatei: Sorry, I should have given a more in-depth explanation for this.

So, the URL values are technically custom values until the options are loaded. After the options are loaded:
- some of these values are normal values (they can be found in the received options)
- some of these values are traditional custom values (the label is the same as the value and they are not found in the received options)
- and some of these values are a special case of custom values (the label is different from the value and they are not found in the received options).

`urlOptions` was needed because of the last type of values as there are situations where we can lose the label of a custom value. For example, when calling `changeValueTo` method without a `texts` parameter, the `text` is extracted from the available options. In order to overcome this, we can either require the `texts` parameter (which is a huge breaking change) or make the custom value/label pairs available to it.

In the case of `GroupByVariable` this can be easily solved because `changeValueTo` is called only from the renderer but for other types of `MultiValueVariable` this gets tricky to solve.

A neat side-effect of the `urlOptions` was that the 3rd type of custom values could also live as an independent option. For example, we receive a custom value with a label via the URL: value is `MyValue` and label is `MyLabel`. This custom value does not exist in the options that we load from the API. The user removes it by clicking the "X" button but realizes it should have stayed there. When opening the menu, the user could simply pick `MyLabel` from the list. Because they only saw `MyLabel`, they didn't know what was the actual value of it. Indeed, this is an edge case tho'.

I'll push an update to the PR in a couple of minutes with an implementation w/o `urlOptions`.

packages/scenes/src/variables/groupby/GroupByVariableUrlSyncHandler.ts
bfmatei: Provided more details here: https://github.com/grafana/scenes/pull/705#discussion_r1596579233

packages/scenes/src/variables/groupby/GroupByVariableUrlSyncHandler.ts
bfmatei: Pushing an update without it now.

packages/scenes/src/variables/groupby/GroupByVariableUrlSyncHandler.ts
bfmatei: Besides breaking the URLs, taking care of the texts is a problem that will need to be solved as well. I described it in the long comment above (i.e. require the texts parameter for `changeValueTo` method.

torkelo (approved):

dprokop (approved):
@bfmatei it looks and works good!

I've noticed a minor difference in the filters and group by rendering.

Let's assume both filter and group by use the same dimensions set:

```csv
label1, v1
label2, v2
label3, v3
```

Then let's assume an old url was preserved that did not contain label in the url. So we have url like this:

```?var-Filter=%20v1%7C%3D%7Caaa&var-Group%20by=%20v1```

Now, unless the select was opened, what's synced into the renderer in both filter and group by is `v1`. But then when openening the Group by `v1` is replaced with `label1`. Which is fine IMO. But it doesn't work like this for filters - the correct dimension is selected, but the value of the select does not change to `label1` when opened. See the video:

https://github.com/grafana/scenes/assets/2376619/8060ff1b-c896-41bb-a137-a347c45bf714




Not a deal breaker, but something that we can improve in the followup PR.
```

PR #714, opened 2024-05-02
AdHocFilters: clear filter value when key is changed

### LINKED ISSUE #694: Filters: When changing filter dimension, the value should be validated or cleared
Currently, when changing the filter dimension the value of that filter is kept intacted. This may produce wrong query results on refresh. 

To reproduce:

1. Apply a filter, e.g. foo = bar
2. Change the filter dimension from `foo` to `baz`
3. Filter value still shows `bar`, which doesn't make any sense at this point
4. On refresh, queries are sent with filter `baz` = `bar` (which is wrong and may not even be of the same type)

ashharrison90:
Fixes #694 
Fixes https://github.com/grafana/hyperion-planning/issues/20

<details>
  <summary>📦 Published PR as canary version: <code>4.17.2--canary.714.9000579577.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@4.17.2--canary.714.9000579577.0
  # or 
  yarn add @grafana/scenes@4.17.2--canary.714.9000579577.0
  ```
</details>


dprokop (commented):
I found one thing that I think is worth addressing in this PR:

1. Create a complete filter (label & value)
2. Change filter dimension (_observe, that the partial filter is synced to the URL_)
3. Refresh the page (observe that the filter value select is open).

I think we should treat the dimension change as new filter config transaction, meaning that the filter should not be commited to the URL until it's complete (dimension + value configured).

dprokop (approved):
Works great!

PR #716, opened 2024-05-03
VizPanelRenderer: Emit SetPanelAttention event

tskarhed:
Addresses this issue https://github.com/grafana/grafana/issues/47005#issuecomment-2089912829

Related PR in the Grafana repo: https://github.com/grafana/grafana/pull/87317

Is this a good place to add this? Also, are there some drawbacks that may cause performance issues?

<details>
  <summary>📦 Published PR as canary version: <code>4.24.4--canary.716.9299472214.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@4.24.4--canary.716.9299472214.0
  # or 
  yarn add @grafana/scenes@4.24.4--canary.716.9299472214.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
@@ -136,8 +136,8 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
       <div
         ref={ref as RefCallback<HTMLDivElement>}
         className={absoluteWrapper}
-        onFocus={() => panelAttentionService.setPanelWithAttention(model)}
-        onMouseMove={() => panelAttentionService.setPanelWithAttention(model)}
+        onFocus={() => panelAttentionService.setPanelWithAttention(model.state.key)}
+        onMouseMove={() => panelAttentionService.setPanelWithAttention(model.state.key)}
         data-viz-panel-key={model.state.key}
       >
         {width > 0 && height > 0 && (
--- a/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
@@ -1,4 +1,4 @@
-import React, { RefCallback } from 'react';
+import React, { RefCallback, useCallback, useMemo } from 'react';
 import { useMeasure } from 'react-use';
 
 // @ts-ignore
@@ -29,10 +29,13 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
     description,
   } = model.useState();
   const [ref, { width, height }] = useMeasure();
-  const appEvents = getAppEvents();
+  const appEvents = useMemo(() => getAppEvents(), []);
 
-  const setPanelAttention = () => appEvents.publish(new SetPanelAttentionEvent({ panelId: model.state.key }));
-  const debouncedMouseMove = debounce(() => setPanelAttention(), 100);
+  const setPanelAttention = useCallback(
+    () => appEvents.publish(new SetPanelAttentionEvent({ panelId: model.state.key })),
+    [model.state.key, appEvents]
+  );
+  const debouncedMouseMove = useMemo(() => debounce(setPanelAttention, 100), [setPanelAttention]);
 
   const plugin = model.getPlugin();
 
@@ -159,9 +162,8 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
             padding={plugin.noPadding ? 'none' : 'md'}
             menu={panelMenu}
             onCancelQuery={model.onCancelQuery}
-            //@ts-ignore
             onFocus={() => setPanelAttention()}
-            onMouseMove={debouncedMouseMove}
+            onMouseMove={() => debouncedMouseMove()}
           >
             {(innerWidth, innerHeight) => (
               <>
--- a/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
@@ -136,8 +136,8 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
       <div
         ref={ref as RefCallback<HTMLDivElement>}
         className={absoluteWrapper}
-        onFocus={() => panelAttentionService.setPanelWithAttention(model.state.key)}
-        onMouseMove={() => panelAttentionService.setPanelWithAttention(model.state.key)}
+        onFocus={() => panelAttentionService?.setPanelWithAttention(model.state.key)}
+        onMouseMove={() => panelAttentionService?.setPanelWithAttention(model.state.key)}
         data-viz-panel-key={model.state.key}
       >
         {width > 0 && height > 0 && (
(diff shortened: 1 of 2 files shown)


torkelo (commented):
@tskarhed is it not better to have this logic in PanelChrome? Scenes support many layout options (not only GridLayout), would be nice to share the logic between them

tskarhed:
> @tskarhed is it not better to have this logic in PanelChrome? Scenes support many layout options (not only GridLayout), would be nice to share the logic between them

True, it may be a bit more brittle though, causing multiple panels to have attention. But I can give it a go.

tskarhed:
Ah, yes, that was the reason. If you have one panel in focus and then move the mouse over another, they both have attention, so you need something above keeping track of the attention state.

torkelo:
@tskarhed maybe GrafanaContext can have that?

tskarhed:
@torkelo If used with PanelChrome, wouldn't that be a dependency on core? And PanelChrome no knowledge of panel id :/

torkelo:
@tskarhed grid items does not know panel id either , maybe VizPanelRenderer can set attention? Just a question where state should live

tskarhed:
Okay, for non-scenes I have it working with the state living in KeyboardSrv (which is accessible with the `useGrafana` hook). The the update still happens in the GridItem part (haven't figured out where else this would work), but it could easily be added to any component in core and be extended.

For Scenes, it seems like VizPanelRenderer is a good place. It can emit an event when it gets attention (preferably it would also be able to read the current attention state to determine if it has attention in order to avoid a ton of events being emitted on MouseMove). Just not sure how to read this possible parent state. At the moment I'm looking at storing the state in `DashboardScene`, and have it subscribe to the event. The nice thing about the scenes approach is that it is easy to pass around the ScenesObject instead of ids.

torkelo:
@tskarhed why not use `useGrafana` and the KeyboardSrv to set state for scenes as well inside VizPanelRenderer? 

Ah, plugins platform has not yet made useGrafana accessible to plugins :(  

Then I don't know, maybe we need a new context that plugins (ie scenes) can interact with?

tskarhed:
@torkelo Good point. At the moment there is also duplicate functionality for the keybindings, which I could consolidate. `setupKeyboardShortcuts` for Scenes and `KeybingingSrv` for core. These could probably be handled the same way (although withFocusedPanel passing the actual Scene object is super useful, compared to just the panel id).

Maybe I'll look into exposing either `useGrafana` or just the keybindings if there may be unintended consequences.

tskarhed:
I created a service this time around. There are still some typing issues with `VizPanel`, and onMouseMove should probably be debounced, or something of the sort.

tskarhed:
@torkelo I have updated the approach using a service instead. Please tell me what you think.

torkelo (commented):
Sorry had this unpublished review comment from 1-2 days ago (done on mobile)

Will take a look more closely next week (traveling today)

packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
```diff
@@ -130,7 +133,13 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
 
   return (
     <div className={relativeWrapper}>
-      <div ref={ref as RefCallback<HTMLDivElement>} className={absoluteWrapper} data-viz-panel-key={model.state.key}>
+      <div
+        ref={ref as RefCallback<HTMLDivElement>}
+        className={absoluteWrapper}
+        onFocus={() => panelAttentionService.setPanelWithAttention(model)}
+        onMouseMove={() => panelAttentionService.setPanelWithAttention(model)}
```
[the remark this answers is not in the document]
tskarhed: I updated using panelIds instead now. I tend to agree with you. I think the approach you mention makes sense.

- Add onFocus and onMouseMover handlers to PanelChrome
- Emit an event wherever PanelChrome is used (no appEvents in grafana ui)
- Listen to the event in KeybindingSrv and traverse the DOM to find panelId

packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
```diff
@@ -150,6 +159,9 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
             padding={plugin.noPadding ? 'none' : 'md'}
             menu={panelMenu}
             onCancelQuery={model.onCancelQuery}
+            //@ts-ignore
+            onFocus={() => setPanelAttention()}
+            onMouseMove={debouncedMouseMove}
```
tskarhed, line 164: I can't get these to actually run. PanelChrome is updated with these in the core PR

torkelo: @tskarhed we have an issue with yarn link, when you yarn link scenes get's it's own bundled version of grafana/ui (so many theme and contexts issues arise). So your code change to PanelChrome is not going to be present as scenes will be compiled with what ever version of grafana/ui is defined in package.json.

Not sure how to fix this yarn linking / building issue.

tskarhed: @torkelo Makes sense. If you think the general approach is good, I could revert the changes I did in https://github.com/grafana/grafana/pull/87317 `keyboardShortcuts`, and retry it once Scenes has the correct version.

packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
```diff
@@ -26,10 +29,16 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
     description,
   } = model.useState();
   const [ref, { width, height }] = useMeasure();
+  const appEvents = getAppEvents();
+
+  const setPanelAttention = () => appEvents.publish(new SetPanelAttentionEvent({ panelId: model.state.key }));
+  const debouncedMouseMove = debounce(() => setPanelAttention(), 100);
```
torkelo, line 35: think this should be memoized or use some hook

packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
```diff
@@ -150,6 +159,9 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
             padding={plugin.noPadding ? 'none' : 'md'}
             menu={panelMenu}
             onCancelQuery={model.onCancelQuery}
+            //@ts-ignore
+            onFocus={() => setPanelAttention()}
```
torkelo, line 163: why ts ignore?

tskarhed: Because of the grafana-ui import issue. It doesn't recognize the added prop. It just doesn't work at all either. Maybe I screwed up with the set up, I'll give it another try in a bit.

torkelo: ah, of course

tskarhed:
It still has the `@ts-ignore` comments (which are needed at the moment), but it works using the canary grafana/ui

packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
```diff
@@ -150,6 +162,9 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
             padding={plugin.noPadding ? 'none' : 'md'}
             menu={panelMenu}
             onCancelQuery={model.onCancelQuery}
+            // @ts-ignore
+            onFocus={() => setPanelAttention()}
+            onMouseMove={() => debouncedMouseMove()}
```
torkelo, line 167:
```suggestion
            onFocus={setPanelAttention}
            onMouseMove={debouncedMouseMove}
```

torkelo (approved):

PR #724, opened 2024-05-08
Performance: Limit data layer state updates

torkelo:
Various performance fixes 

* [x] We were using infinity replay subjects so when coming back to a scene data layers emitted all their previous emitted results 
* [x] Ignore replayed data layer results that has already been processed (this happens when a data provider get's re-activated, we subscribe the the SceneDataLayerSet this causes the ReplaySubject to emit last data, even if that has already been incorporating into the current state). Created a new DataLayersMerger that can detect when a result has already been seen.  
* [x] In SceneQueryRunner ignore updating state when data layers are received if there are no annotations (And the previous result also had no annotations)

<details>
  <summary>📦 Published PR as canary version: <code>4.22.0--canary.724.9079775010.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@4.22.0--canary.724.9079775010.0
  # or 
  yarn add @grafana/scenes@4.22.0--canary.724.9079775010.0
  ```
</details>


packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -182,9 +185,19 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
       }
     }
 
-    const baseStateUpdate = this.state.data ? this.state.data : { ...emptyPanelData, timeRange: timeRange.state.value };
+    // Skip unnessary state updates
+    if (
+      alertState === this.state.data?.alertState &&
```
dprokop, line 190: This condition is quite hard to reason about, as it assumes that one knows that the alert state is a reference to an state taken from the alert states results. I wonder if it wouldn't be easier if we didn't set that particular alert state reference in the state but always create a new object, and here instead rely on object equality rather than reference equalit? WDYT?

torkelo: @dprokop pushed update that switched to isEqual instead, for some reason I thought alertState was an enum :)

PR #726, opened 2024-05-09
Variables: Do not update value when value and text are the same

torkelo:
Fixes an annoying issue where a multi-value variable would trigger a cascading update even when the value did not change, for example when just opening the variable multi select and closing it without changing anything


dprokop (approved):
Tested against main an works as expected. Nice catch with the query being executed after no changes are made.

PR #728, opened 2024-05-09
Typescript: Enable strict mode

torkelo:
Should fix some of the errors that show up when yarn linking into main repo


dprokop:
hm, i see 94 errors when linking main and 94 errors when linking this branch. Are you sure this brings any improvement?

Sergej-Vlasov:
same result here 94 errors before and after. Strict mode can be annoying with extra errors but no additional errors here so would be for enabling strict mode

torkelo:
@dprokop well it fixes a few strict errors (see the ones fixed in PR) so it does something, not sure what else to enable

torkelo:
@dprokop is there any argument to not extend tsconfig (Which has the important strict: true option)

torkelo:
>  Strict mode can be annoying with extra errors but no additional errors here so would be for enabling strict mode
@Sergej-Vlasov not sure what your mean :)

Sergej-Vlasov:
> > Strict mode can be annoying with extra errors but no additional errors here so would be for enabling strict mode
> 
> @Sergej-Vlasov not sure what your mean :)

Sorry, meant to say that I support enabling strict mode in short 😅

Sergej-Vlasov (approved):

PR #729, opened 2024-05-09
SceneVariableSet: Do not propagate variable value changes when a local variable has the same name

torkelo:
Part of row repeat fixes https://github.com/grafana/grafana/pull/87539

So we have this problem in row repeats that when the variable changes (that you repeat by). All the panel query runners inside all the repeated rows were notified that their variable dependency changed and started issuing a new queries (even when their local value did not change) . 

So figured that this notification (via variableDependency) should not happen if there is a local SceneVariableSet with a local variable of the same name. 

Created a test case for this scenario all is fine. This logic make sense, the variable has not changed for the local scope (as there is a local variable with the same name that has not changed).

The only really hard puzzle piece to solve is that his completely breaks the RowRepeatBehavior in core as it relies on this notification to process repeats. 

So the problem here is that the first row (which we use as template / source of repeats) also get's a locally scoped value (the first variable value), so after the first repeat process this local value on the first row will block any updates the variable, which logically is correct the local value is not changing. ARHHHHfasFASDSDASFA S FQ"!° going to go for a walk. 

Think the only solution is to the notification of variable changes from the behavior itself to the dashboard scene, this is now done in https://github.com/grafana/grafana/pull/87539

### DIFF
--- a/packages/scenes/src/variables/sets/SceneVariableSet.ts
+++ b/packages/scenes/src/variables/sets/SceneVariableSet.ts
@@ -324,6 +324,14 @@ export class SceneVariableSet extends SceneObjectBase<SceneVariableSetState> imp
       return;
     }
 
+    // If we find a nested SceneVariableSet that has a variable with the same name we stop the traversal
+    if (sceneObject.state.$variables && sceneObject.state.$variables !== this) {
+      const localVar = sceneObject.state.$variables.getByName(variable.state.name);
+      if (localVar) {
+        return;
+      }
+    }
+
     if (sceneObject.variableDependency) {
       sceneObject.variableDependency.variableUpdateCompleted(variable, hasChanged);
     }
--- a/packages/scenes/src/variables/sets/SceneVariableSet.test.tsx
+++ b/packages/scenes/src/variables/sets/SceneVariableSet.test.tsx
@@ -15,6 +15,7 @@ import { sceneGraph } from '../../core/sceneGraph';
 import { SceneTimeRange } from '../../core/SceneTimeRange';
 import { LocalValueVariable } from '../variants/LocalValueVariable';
 import { TestObjectWithVariableDependency, TestScene } from '../TestScene';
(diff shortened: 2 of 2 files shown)


packages/scenes/src/variables/sets/SceneVariableSet.ts
```diff
@@ -324,6 +324,14 @@ export class SceneVariableSet extends SceneObjectBase<SceneVariableSetState> imp
       return;
     }
 
+    // If we find a nested SceneVariableSet that has a variable with the same name we stop the traversal
+    if (sceneObject.state.$variables && sceneObject.state.$variables !== this) {
+      const localVar = sceneObject.state.$variables.getByName(variable.state.name);
+      if (localVar) {
+        return;
+      }
+    }
```
dprokop, line 333: @torkelo wouldn't this only work for s single nested SceneVariableSet? I mean, shouldn't be continue up the hierarchy and make sure other objects also ignore updates in such scenario?

dprokop: wait, sorry, nvm. We are traversing down in here to notify objects. So we exit early if there's a local var, so technically variables down the hierarchy will not be affected by the outer scope variable changes, as they now depend on the local variable which does not change. I was thinking about a scenario that the nested object would depend on the global variable, but that's not really possible, as the local variable scope will take precedence.

dprokop (approved):

dprokop:
Think we need to wait with merging this and coordinate with https://github.com/grafana/grafana/pull/87539 ?

packages/scenes/src/variables/sets/SceneVariableSet.ts
torkelo: @dprokop yea, there is maybe a scenario where the local variable scope (that has the local variable) should be notified, but we can add that later if we need it

torkelo:
@dprokop yes, good point. Will wait merging this until that one is approved and ready to merge

PR #733, opened 2024-05-10
SceneGridLayout: Prevent panels from moving on mount

torkelo:
Noticed panels moving about when coming back to dashboard from dashboard settings and sometimes from panel edit. 

This fixes that problem, I am using a ref here and updating the class using dom classList to avoid having to always re-render the full grid after mount

### DIFF
--- a/packages/scenes/src/components/layout/grid/SceneGridLayoutRenderer.tsx
+++ b/packages/scenes/src/components/layout/grid/SceneGridLayoutRenderer.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useRef } from 'react';
 import ReactGridLayout from 'react-grid-layout';
 import AutoSizer from 'react-virtualized-auto-sizer';
 import { SceneComponentProps } from '../../../core/types';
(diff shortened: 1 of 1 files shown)


packages/scenes/src/components/layout/grid/SceneGridLayoutRenderer.tsx
```diff
@@ -138,6 +147,18 @@ function validateChildrenSize(children: SceneGridItemLike[]) {
   }
 }
 
+function updateAnimationClass(ref: React.MutableRefObject<HTMLDivElement>, isDraggable: boolean, retry?: boolean) {
```
ivanortegaalba, line 150: It works! Question out of curiosity. Could just use the dependencies array from `useEffect` to rerun the class adjustment when the `ref` or `ref.current` changes?

ivanortegaalba (approved):
@torkelo I pushed a small commit to fix the types errors, caused by a known React type issue. I hope you don't mind 😄

packages/scenes/src/components/layout/grid/SceneGridLayoutRenderer.tsx
torkelo: @ivanortegaalba useEffects are not triggerred when a useRef changes, for that we would have to store the ref in a useState (which would re-render the whole grid and loose the point of using the ref to set class)

PR #734, opened 2024-05-11
PlainReact: Expose scene features through contexts and hooks and normal react components 

torkelo:
**Problem**

It's been clear for a while that scenes have issues when going beyond simple dashboard-like declarative views.

* Steep learning curve 
* Hard to mix with familiar React patterns & state management & hooks 
* The clonable & serializable state model that can be walked is very rarely needed in scene apps (beyond internal scene mechanics) 
* Requiring all data, variables, queries and visualizations be added to a state tree makes rendering transient views unnecessarily cumbersome 

These issues have troubled me greatly the 6+ months, I wish we had encountered these issues earlier in the process, we might have been able to rethink the architecture. 

I did try a quick react context approach very early on but failed to find a way to nest contexts while controlling re-renders (and a few other issues, like how to cache state, and complex state logic among others). I wish I had pursued it more but was too focused on solving core/future persisted dynamic dashboard needs. 

It's too late to start from scratch. So in this PR, I am trying my best to expose pure react components, contexts and hooks that re-use as much as possible of the existing scene fundamentals (variable system, query system, time range, VizPanel, URL sync).

Started this late yesterday, so still very early but looks more promising than I expected. 

**For a quick sense take a look at the usage examples**

* [Home](https://github.com/grafana/scenes/blob/react-context-poc/packages/scenes-app/src/react-demo/Home.tsx) (has the top level context & routes and env variable)
* [Repeat by variable](https://github.com/grafana/scenes/blob/react-context-poc/packages/scenes-app/src/react-demo/RepeatByVariablePage.tsx)
* [Repeat by series](https://github.com/grafana/scenes/blob/react-context-poc/packages/scenes-app/src/react-demo/RepeatBySeriesPage.tsx)
* [Dynamic queries](https://github.com/grafana/scenes/blob/react-context-poc/packages/scenes-app/src/react-demo/DynamicQueriesPage.tsx)
* [Dynamic visualization](https://github.com/grafana/scenes/blob/react-context-poc/packages/scenes-app/src/react-demo/DynamicVisualizationPage.tsx)
* [Dynamic variables](https://github.com/grafana/scenes/blob/react-context-poc/packages/scenes-app/src/react-demo/DynamicVariablesPage.tsx)
* [Nested contexts](https://github.com/grafana/scenes/blob/react-context-poc/packages/scenes-app/src/react-demo/NestedContextPage.tsx)
* [Interpolation hook](https://github.com/grafana/scenes/blob/react-context-poc/packages/scenes-app/src/react-demo/InterpolationHookPage.tsx)
* [Interoperability](https://github.com/grafana/scenes/blob/react-context-poc/packages/scenes-app/src/demos/interopDemo.tsx) test (this is under the existing demo pages) 
* [Drilldown demo](https://github.com/grafana/scenes/blob/react-context-poc/packages/scenes-app/src/react-demo/DrilldownDemoPage.tsx) (nested routes and breadcrumb with maintained variable and time range state)

For an explanation of the approach, [there is a design doc](https://docs.google.com/document/d/1ir7k6ISRzSdJ9b0ymaLjhlO3H7Eh2aZDkGdN-Y3tZWg/edit)

## Components

* [x] SceneContextProvider that represents a scene, accepts defining an optional time range scope. This also acts as a scope for SceneVariableSet. Can be nested
* [x] CustomVariable. Adds a CustomVariable to closest scene context. 
* [x] VariableSelect. Renders a variable value selector (no matter what level the variable is defined on)  
* [x] TimeRangePicker. Render the time range picker using closest scene time range 
* [x] VizPanel renders panels & visualizations with provided data provider. Has a very nice separation between panel props (title, etc) and visualization definition 
* [x] RefreshPicker renders the SceneRefreshPicker. 
* [x] BreadcrumbContext & BreadcrumbContextProvider & Breadcrumb (Helps with maintaining state in breadcrumb URLs) 

## Hooks 
* [x] useTimeRange
* [x] useVariables
* [x] useVariableValues returns the values of a specific multi-value variable (and updates/re-renders the calling component when values change)
* [x] useQueryRunner. Adds a SceneQueryRunner to the closest scene context. Returns the query runner / handles updates to queries.
* [x] useUpdateWhenSceneChanges (Re-renders components when specified variables or time range changes)
* [x] useVariableInterpolator. Hook that returns an interpolation function, also re-renders the calling component when variables or time range change. 

Also a new abstraction to define visualizations for the new react VizPanel component 
* [x] VizConfigBuilders. A new form of builder that only defines a visualization  (pluginId, options, fieldConfig) 

## Solved issues

* [x] Pattern for dynamically adding/removing scene objects to a scene (context) as they are rendered
* [x] Dynamically add variables via rendering a react component 
* [x] URL sync works for dynamically added variables & time ranges (needs more work to handle key conflicts) 
* [x] Figure out how we can use these new hooks & components from inside a plain (ie current) scene. To make it possible to mix the approaches (to aid migration & maybe other reasons). EmbeddedScene can now optionally wrap the scene in a SceneContext. 
* [x] Support query controller 

## Unsolves issues / todo

* [ ] No caching of RVizPanel state.  For example if you toggle a legend in a graph and then go to another page and come back.
* [ ] No caching of data (returning to view re-queries data).
     * [I did a first stab / test in this branch ](https://github.com/grafana/scenes/compare/react-context-poc...react-context-poc-caching)
* [ ] URL sync for dynamically added objects whose key conflicts with existing objects. For example 2 time ranges. 
* [ ] URL sync when changing path (currently UrlSyncManager skips this expecting a full sync will happened when path changes and a new scene is initialized) 
* [ ] Lots of missing props on VizPanel
* [ ] Some missing props on useQueryRunner 
* [ ] No layouts that work with this model (SceneFlexLayout & CSSGridLayout) 
* [ ] Transformations (should be easy) 
* [ ] Data layers (annotations) (should be relatively easy) 
* [ ] Minor issue with RefreshPicker flickering on the first mount (need to refactor this to be pure react component rather than proxy scene object). very low prio issue, but annoying.  

## Thoughts 

***Contexts / state outside the routes / pages*** 

By letting go of the "scene" concept as these serializable declarative app views (which scenes was primarily designed for) it's easier to have a state outside the pages/views (time range and variables that live above page routes or drilldowns). This would be possible in the plain scene model as well with some small changes. It's nice to be able to define the shared time range & variables that all child pages should have once. And they never need to be re-activated and get state sharing (form of caching) for free here. 

***Separation between panel and visualization*** 

I really love the separation between the visualization definition (pluginId, pluginVersion, options, fieldConfig) and the panel options that RVizPanel give. Something I have been trying to do before with VizPanel but failed to come up with the right model/abstraction. I was trying to break it up into two components one dealing with PanelChrome and the other rendering the visualization plugin, but as the level that renders PanelChrome needs to handle info from both this approach failed in various ways. But by just defining one object that fully encapsulates the visualization we get the same effect (something we could do to VizPanel). 

This separation enables you to define different visualization definitions and easily reuse them in your scene app. We can even ship some good defaults one with scenes lib. 

**Next steps** 

I think this looks too promising to no continue work on and get help from scene app devs who are interested. The dashboard / scene squad is a bit thinly stretched with dashboard migration & big customer POC work so any help is appreciated. 

Things needed before merge / making it real

* [x] Where do these new components live (Moved to new scenes-react package)
* [ ] Agree on naming of components. 
* [x] Reverted URL sync changes, leaving them for follow-issue 
* [x] Some initial tests of the context & react components & hooks (could also be done in a follow-up issue if carefully defined and followed-up) 

<details>
  <summary>📦 Published PR as canary version: <code>4.26.1--canary.734.9385486259.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@4.26.1--canary.734.9385486259.0
  npm install @grafana/scenes@4.26.1--canary.734.9385486259.0
  # or 
  yarn add @grafana/scenes-react@4.26.1--canary.734.9385486259.0
  yarn add @grafana/scenes@4.26.1--canary.734.9385486259.0
  ```
</details>


packages/scenes-app/src/react-demo/visualizations.tsx
```diff
@@ -0,0 +1,16 @@
...
+  .setCustomFieldConfig('drawStyle', GraphDrawStyle.Bars)
+  .setCustomFieldConfig('fillOpacity', 6)
+  .build();
+
+export const graphWithGrapdientColor = RVisualizationBuilders.timeseries()
+  .setCustomFieldConfig('fillOpacity', 10)
+  .setCustomFieldConfig('lineWidth', 3)
+  .setCustomFieldConfig('gradientMode', GraphGradientMode.Scheme)
+  .setColor({ mode: FieldColorModeId.ContinuousGrYlRd })
+  .build();
```
torkelo, line 16: I really really like this separation between panel and visualization, here it's super easy to just define visualization definitions that you can share / reuse in your scene app.

This separation was something I tried really hard to do with VizPanel (breaking it up into two abstractions), but failed to figure out how, as whatever renders PanelChrome needs data, and for data we need the visualization

But with RVizPanel the separation between panel and visualization is very clean.

torkelo: One thing it's missing is extending a defined visualization with new overrides. Normally the overrides (like data links and naming overrides) will be data / panel specific. 

so we could have something like. 

```ts
const vizWithOverrides = RVisualizationBuilders.timeseries()
   .extend(graphWithGrapdientColor)
  .setOverrides((b) =>
    b.matchFieldsWithName('Library').overrideLinks([
      {
        title: 'Go to library details',
        url: '${__url.path}/lib/${__value.text}',
      },
    ])
  )
  .build();
```

Elfo404:
This looks very promising, will take a look asap!

packages/scenes/src/react-context/hooks.ts
```diff
@@ -0,0 +1,132 @@
...
+ * Only returns the variables on the closest context level.
+ * We could modify it to extract all variables from the full context tree.
+ */
+export function useVariables(): SceneVariable[] {
+  const scene = useSceneContext();
+  const variables = sceneGraph.getVariables(scene);
+  return variables.useState().variables;
+}
+
+export function useVariableValues(name: string): [VariableValueSingle[] | undefined, boolean] {
```
matyax, line 42: This wrapper around `useSceneContext()` and `sceneGraph` is great for us application developers. If I may make a suggestion, the following additions could be very useful:

- singular `useVariableValue` when the value is not an array
- support for providing a type to `useVariable(s)Value<T>`, so we don't have to type check in every component where we use these hooks. I think something similar should have to be added to the places where we write to variables, to be sure we use the same types when writing and reading.

The main reason for this last suggestion comes from a recurring pattern that can be found, for example, in the Explore Logs app: https://github.com/search?q=repo%3Agrafana%2Fexplore-logs+%22variable+instanceof%22&type=code

Apologies in advance for interrupting in a PR where I'm not a reviewer.

torkelo: @matyax no apologies needed! I shared this PR widely to get well-needed feedback from as many as possible! 

 yea, I agree with need of a similar function for single-value variables (and making it generic), the reason I only have this multi-value one was that it was all I needed for the demo. 

Will be sure to add this as a todo in an epic I will be creating once this first PR is merged. 

got any feedback on where these components should live (and how to deal with the naming conflict with existing scene objects). 

See proposals
https://raintank-corp.slack.com/archives/C04FRQ2S0AV/p1715782850562119?thread_ts=1715594100.226169&cid=C04FRQ2S0AV 

Current top contender I think is a new scenes-react package (means there will be some components with same as scene objects, for example QueryVariable)

oscarkilhed:
Started implementing some demo stuff to try this out, ran into a build issue and made this pr to this branch: https://github.com/grafana/scenes/pull/741

ivanortegaalba:
@torkelo "This looks too promising to stop working on it". I completely agree.

This is kind of the expected API a dev will expect to create UI and reuse dashboard components. I want to keep playing with it, but my first feeling is "this should be the default" to use Scenes. We don't even need to mention scenes anywhere, instead, we should only use dashboard domain names and components. Scenes will be the engine under the hood, but IMO, the way to develop Dashboards on plugins should be something like what we see in this PR.

About how to expose them, I tried to answer in the previous paragraph. The default components should be the react components, and the ones namespace should be the Scenes ones. Is not any of the proposals you mentioned before, but once this approach is validated, I'd advocate for a breaking change to make the React components the default ones.

matyax:
+1 @ivanortegaalba 

> we should only use dashboard domain names and components

I'm not sure if I'm misreading your comment there, but I wanted to add that this will also be used to build applications (like Explore Logs), so I would advocate for not using/exposing dashboard-specific terms. For example, just today I had to work with a behavior within Explore Logs using an enum called `DashboardCursorSync` https://github.com/grafana/explore-logs/blob/main/src/Components/ServiceScene/ServiceScene.tsx#L440

packages/scenes-app/src/react-demo/RepeatBySeriesPage.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+export function RepeatBySeriesPage() {
+  return (
+    <PageWrapper title="Repeat by series" subTitle="Repeats visualizations returned by a single query">
+      <RCustomVariable name="series" label="Series count" query="1,2,5,10,20,30" initialValue={'3'}>
+        <Stack direction={'column'}>
+          <RVariableSelect name="series" />
+          <RepeatPanelBySeries />
+        </Stack>
+      </RCustomVariable>
```
dprokop, line 25: Didn't see a demo for having multiple variables in a scene, is this currently possible without deep nesting those components?

dprokop (commented):
This is looking really good @torkelo - seems like this exploration is addressing most, if not all of the issues people were struggling with when using scenes API directly.

I'm a bit worried there's no architectural explanation in tis PR. While some of us know Scenes in depth enough, I think the implementation details in the contexts implementations may make it hard to review this PR. 


On the API side it looks good to me, with two comments:
- Not sure why nesting time range scope requires SceneContextProvider, I would expect something less generic, like simply TimeRangeContextProvider (that under the hood can compose SceneContextProvider of course).
- Can't see a demo with mulitple variables at the same level. IIUC. current proposal is to nest variables, but this comlicates the tree and reasoning about the variables IMO. I would expect some react abstraction similar to `SceneVaraiablesSet`, so that multiple variables can be placed in the react tree on the same level easily. But this may be a details.


Regarding where the API should live I think it deserves a separate package, living in the scenes monorepo. I would expect scenes-react or sth like this, that lives next to scenes core API, and is released as a new npm library dependant on scenes API. We should probably version this independently, with the current code marked as 0.0.1. So to summarise, this is the structure of the repo I would consider:

- scenes (core lib, released to npm with the current versioning)
- scenes-react (react abstraction API, released to npm with new versioning starting with 0.0.1)
- scenes-app (demo app,  not released to npm).

domasx2 (commented):
Very cool! There goes my plan for summer hackathon :) Would love to try refactoring app o11y at some point

My 2ct on where the components should live: should optimize for avoiding confusion between scene objects and react component counterparts and help IDE code intelligence suggest the right things as much as possible. Separate package sounds best, maybe even with scene object names having "scene" in them, and react not 

> It's too late to start from scratch. 

I'm not enitrely convinced. If we can figure out a better solution for the long term, maybe it's worth making a break earlier rather than later, not be bound to a sunk cost. 

Also perhaps internal dashboard needs and app developer needs could be considered as distinct problems requiring distinct solutions with different trade offs. Not necessarily requiring a shared layer. Eg, scene variables are heavily tied to dashboard semantics and can be awkward to use in apps

packages/scenes-app/src/react-demo/RepeatBySeriesPage.tsx
torkelo: @dprokop this demo has 2 variables (added to the top-level context), The env variable defined in the Home.tsx file and the series variable defined here are both added to the closest (ie top) scene context. 

The reason RCustomVariable is nesting children under is to guarantee that the variable is added to the state tree before children are rendered.

packages/scenes/src/react-context/useSceneQuery.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * Maintains the state of SceneQueryRunner in the scene
+ *
+ * @example // To access query results do
+ * const query = useSceneQuery(...);
+ * const {data} = query.useState();
+ */
+export function useSceneQuery(options: UseQueryOptions): SceneDataProvider {
```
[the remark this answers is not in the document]
torkelo: @domasx2 yea, I had it like that at first, but VizPanel (panel chrome) needs to cancel queries and this is done via the SceneDataProvider interface, so by returning the data provider here I can reuse so much (basically 100%) of VizPanel. There is also the need for max data points from width, etc, and other parts of the data provider interface we visualization needs. So that is why I opted to return the data provider instead of just data. It's just a single line `const {data} = query.useState()` to get the data (which contains loading state, streaming state, error state etc)

ivanahuckova (commented):
I really like this a lot! Looking at the code in the demo app makes it much clearer to me what is happening and how to build apps using these components. Great job, and I look forward to seeing this move forward.

packages/scenes-app/src/react-demo/RepeatBySeriesPage.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+import { PageWrapper } from './PageWrapper';
+import { getFrameDisplayName } from '@grafana/data';
+import { plainGraph } from './visualizations';
+import { DemoVizLayout } from './utils';
+
+export function RepeatBySeriesPage() {
+  return (
+    <PageWrapper title="Repeat by series" subTitle="Repeats visualizations returned by a single query">
+      <RCustomVariable name="series" label="Series count" query="1,2,5,10,20,30" initialValue={'3'}>
```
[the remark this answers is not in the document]
torkelo: @ivanahuckova yea, this looks just like a demo mistake I think

packages/scenes/src/react-context/RCustomVariable.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { useEffect, useState } from 'react';
+import { CustomVariable } from '../variables/variants/CustomVariable';
+import { useSceneContext } from './hooks';
+import { VariableValue } from '../variables/types';
+
+export interface RCustomVariableProps {
+  query: string;
```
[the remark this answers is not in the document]
torkelo: @ivanahuckova yes, this was a design mistake we did with CustomVariable where we based on the current persisted model which is a csv string, we corrected many similar mistakes (made csv string propper arrays) but this one we missed. 

Not sure what to do now, it would be a breaking change. For react components strings (and in general) are also a lot easier to diff vs arrays which needs deep equality checks

packages/scenes/src/react-context/BreadcrumbContext.tsx
```diff
@@ -0,0 +1,113 @@
...
+        if (v.urlSync && !v.state.skipUrlSync) {
+          const state = v.urlSync.getUrlState();
+          Object.assign(params, state);
+        }
+      }
+
+      if (timeRange.urlSync) {
+        const state = timeRange.urlSync.getUrlState();
+        Object.assign(params, state);
+      }
```
yduartep, line 92: Some days ago I reported an un-expected behaviour when zooming into a graph. I wanted to come back to the previous time range when clicking on the Back browser button instead of coming back to the previous page. I wonder if this peace of code resolve that.

packages/scenes/src/react-context/hooks.ts
yduartep: Agree with @matyax, it would be nice to have singular **useVariableValue** with typing. I think it would cover also the case for multiple values. Example: 

```
useVariableValue<(Something extends VariableValueSingle)[]>
```

packages/scenes/src/react-context/useSceneQuery.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { SceneDataProvider } from '../core/types';
+import { DataQueryExtended, SceneQueryRunner } from '../querying/SceneQueryRunner';
+import { DataSourceRef } from '@grafana/schema';
+import { isEqual } from 'lodash';
+import { useSceneContext } from './hooks';
+
+export interface UseQueryOptions {
+  queries: DataQueryExtended[];
+  maxDataPoints?: number;
+  datasource?: DataSourceRef;
```
yduartep: What about if I have **behaviours** associated to my QueryRunner?

packages/scenes/src/react-context/useSceneQuery.tsx
```diff
@@ -0,0 +1,55 @@
...
+    };
+  }, [queryRunner, scene]);
+
+  // Update queries when they change
+  useEffect(() => {
+    if (!isEqual(queryRunner.state.queries, options.queries)) {
+      queryRunner.setState({ queries: options.queries });
+      queryRunner.runQueries();
+    }
+  }, [queryRunner, options]);
```
yduartep: This is amazing. What about if I want to attach a behaviour that re-run the queries when a variable change? Can I pass it as an option?

packages/scenes/src/services/UniqueUrlKeyMapper.ts
```diff
@@ -11,7 +11,8 @@ export class UniqueUrlKeyMapper {
   public getUniqueKey(key: string, obj: SceneObject) {
     const objectsWithKey = this.index.get(key);
     if (!objectsWithKey) {
-      throw new Error("Cannot find any scene object that uses the key '" + key + "'");
+      //      throw new Error("Cannot find any scene object that uses the key '" + key + "'");
+      return key;
```
yduartep, line 15: What happen if I try to get a scene object from this key? would I receive an **undefined** object or is it possible that any exception in other place will. be thrown?

packages/scenes/src/react-context/BreadcrumbContext.tsx
torkelo: @yduartep not really, we fix that by changing this line to https://github.com/grafana/scenes/blob/main/packages/scenes/src/services/UrlSyncManager.ts#L120 to not pass true there (replace), that will create a new browser history entry everytime a scene state change updates the URL.

Been on my todo to change that for some time

packages/scenes/src/react-context/useSceneQuery.tsx
torkelo: @yduartep for most scenarios I do not think behaviors will be needed, but if they really are we can always add them

packages/scenes/src/react-context/useSceneQuery.tsx
torkelo: @yduartep SceneQueryRunner will already do that automatically if the variable is used in any of its queries.  

for manual control of query execution when a unrelated variable change I think there should be some other way to do that (like calling queryRunner.runQueries from a useEffect by the calling (consumer) code).

packages/scenes/src/services/UniqueUrlKeyMapper.ts
torkelo: @yduartep not sure what you mean

yduartep:
This is really amazing. I think will make our dev live easier if we can define scenes applications like a normal React app using components, contexts and hooks. Every time I try to review or build a scene page for complicated views, become hard to understand and follow the whole tree of objects. Great work here...!! PLEASEEEEE continue working on this. It's really promising.

packages/scenes-react/package.json
```diff
@@ -0,0 +1,118 @@
+{
+  "name": "@grafana/scenes-react",
+  "version": "4.24.0",
```
jackw, line 3: Should this package versioning be fixed or independent of the scenes package?

packages/scenes-react/package.json
```diff
@@ -0,0 +1,118 @@
...
+    "@grafana/e2e-selectors": "^10.4.1",
+    "react-grid-layout": "1.3.4",
+    "react-use": "17.4.0",
+    "react-virtualized-auto-sizer": "1.0.7",
+    "uuid": "^9.0.0"
+  },
+  "peerDependencies": {
+    "@grafana/data": "^10.4.1",
+    "@grafana/runtime": "^10.4.1",
+    "@grafana/scenes": "workspace:*",
```
jackw, line 49: I think it might be worth checking the version of lerna in this repo is compatible with the `workspace:` protocol before merging. Looking at the canary release attached to this PR it's publishing without replacing it. https://unpkg.com/browse/@grafana/scenes-react@4.24.3--canary.734.9287358327.0/package.json

packages/scenes-react/package.json
torkelo: @jackw should be same as scenes package, think lerna enforce this

packages/scenes-react/package.json
torkelo: @jackw looks like lerna is not updating this, any ideas?

packages/scenes-react/package.json
jackw: Lerna defaults to fixed but you can switch to independent strategy you wanted to separate them. Maybe it doesn't make so much sense if they're tightly coupled though.

packages/scenes-react/package.json
jackw: Ah I'd not realised this is peerDeps! There's an [open issue](https://github.com/lerna/lerna/issues/3671) about it in the lerna repo. I think you might need to match it to the version in the repo so lerna understands to update it when versioning the packages. `"@grafana/scenes": "4.24.2"` or `"@grafana/scenes": "^4.24.2"` would be better if consumers want to rely on patch / minor updates to scenes package?

torkelo: @jackw what do you think is best here a peer dependency or a normal dependency? Switched to normal dependency and now lerna updates it correctly , but not sure what is best

torkelo:
Created some initial unit test, ready for review & merge. 

Here is a epic issue with follow-up tasks: https://github.com/grafana/scenes/issues/761 

The biggest thing I do not like right now is the name conflicts. While the new package now allows for some components to use whatever name they want (and conflict with scene objects exported from main lib).
I don’t really like having two things with the same name (bad for code navigation, auto import, etc). Especially for VizPanel, having a react component named VizPanel and scene object named the feels super bad.
Options

A) Ignore for now and rename either scene object or react version later
B) Rename new react versions
VizPanel  => VizPanelView or PanelViz or Panel
 CustomVariable => DefineCustomVariable (since these components don’t render anything, they just add a scene object to closest context)

L2D2Grafana (commented):
LGTM 👍 For a anyone's trying to run the demo locally I had to add this to the demo.sh
```
scenesreact=$(pwd)/packages/scenes-react

yarn install

cd $scenesreact
yarn install
yarn build
```

packages/scenes-react/package.json
jackw: We can see in the current canary release that the [scenes-react package.json](https://unpkg.com/browse/@grafana/scenes-react@4.24.5--canary.734.9315492006.0/package.json) has a pinned version of scenes. There is a risk with a pinned dependency that a project which also depends directly on the scenes library will have a node_modules structure like:

```
.
└── node_modules/
    └── @grafana/
        ├── scenes // <-- the version used whenever the project imports @grafana/scenes directly
        └── scenes-react/
            └── node_modules/
                └── scenes // <-- the version used by @grafana/scenes-react
```

If multiple versions of scenes in a single project could create issues, but you'd like consumers to continue using scenes directly, then maybe we try the dependency value as a semver range (`"workspace:^"`) and see if Lerna can update it and keep the `^`? This would mean that as long as a consumer is referencing the same major version of the scenes lib in their project they should only ever get a single "hoisted" version of the package. WDYT? Does that make sense?

dprokop:
Quick find: after navigating away from `Dynamic queries` demo, there's a runtime error happening:


```ts
Uncaught TypeError: Cannot read properties of undefined (reading 'setState')
    at eval (SceneContextObject.js:68:11)
    at safelyCallDestroy (react-dom.development.js:22932:1)
    at commitHookEffectListUnmount (react-dom.development.js:23100:1)
    at commitPassiveUnmountInsideDeletedTreeOnFiber (react-dom.development.js:25098:1)
    at commitPassiveUnmountEffectsInsideOfDeletedTree_begin (react-dom.development.js:25048:1)
    at commitPassiveUnmountEffects_begin (react-dom.development.js:24956:1)
    at commitPassiveUnmountEffects (react-dom.development.js:24941:1)
    at flushPassiveEffectsImpl (react-dom.development.js:27038:1)
    at flushPassiveEffects (react-dom.development.js:26984:1)
    at commitRootImpl (react-dom.development.js:26935:1)
```

torkelo:
@dprokop yea, noticed that yesterday as well. bug introduced in my simplification commit, was a simple bug/mistake in addVariable clean up function

packages/scenes-app/src/demos/interopDemo.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { useTimeRange } from '@grafana/scenes-react';
+
+export function getInteropDemo(defaults: SceneAppPageState) {
+  return new SceneAppPage({
+    ...defaults,
+    subTitle: 'Testing using the hooks and plain react components from normal scene',
+    getScene: () => {
+      return new EmbeddedScene({
+        ...getEmbeddedSceneDefaults(),
+        //        context: new SceneContextObject({}),
```
dprokop, line 23: This demo does not work, the hooks cannot resolve the context as it doesn't exist.

I was wondering, since EmbeddedScene may be considered a context provider in most cases, what if it had a possibility to set it as a react context for interoperability? Something like `exposeContext: boolean` (not sure about the name), that would under the hood, in the `EmbeddedScene` renderer, setup the correct `SceneContext.Provider`.

dprokop: Technically, the `exposeContext` could be a top level `SceneObjectState` API. But I'm not 100% given that the `SceneContext` allows children manipulation and heavily depends on it. This could easily lead to broken scenes.

packages/scenes-app/src/demos/interopDemo.tsx
```diff
@@ -0,0 +1,51 @@
...
+  });
+}
+
+class CustomSceneObject extends SceneObjectBase<SceneObjectState> {
+  static Component = ({ model }: SceneComponentProps<CustomSceneObject>) => {
+    const [timeRange, _] = useTimeRange();
+
+    return <div>Time hook: {timeRange.from.toString()}</div>;
+  };
+}
```
dprokop, line 51: This is a great, I think the hooks could be easily used in `SceneReactObject` as well to allow some apps to start using the scenes-react APIs. 

Yesterday I talked with @juanicabanas about a certain case where he wants to access `DashboardScene` down deep in the hierarchy that's rendering a dummy react component which needs an access to the root `DashboardScene`. Today, this requires props drilling or manually setting up a react context, that provides the required scene object . If the `DashboardScene` for example was able to be set up itself as a context provider, then this interoperability would be very easy to achieve. But this will always yield a problem of a making it type safe- how would the consumer down the hierarchy know what is the type of the resolved context object if the hook is generic :(

packages/scenes-app/src/react-demo/DrilldownDemoPage.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { PlainGraphWithRandomWalk } from './PlainGraphWithRandomWalk';
+
+export function DrilldownDemoPage() {
+  /**
+   * The breadcrumb provider should really be wrapping all routes, but placing it here just to show the concept and so far this is the only pages that use it.
+   * It needs to be above the PageWrapper as it's the the component that uses the BreadcrumbContext
+   */
+  return (
+    <BreadcrumbProvider>
+      <Breadcrumb text="Drilldown demo" path={`${urlBase}/drilldown`} />
```
dprokop, line 17: I'm a bit struggling to understand how the `DrilldownLibraryPage` provides the breadcrumb. Where is the `text` taken from?

packages/scenes-app/src/react-demo/DrilldownDemoPage.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <Switch>
+        <Route path={`${urlBase}/drilldown`} component={DrilldownHome} exact />
+        <Route path={`${urlBase}/drilldown/lib/:lib`} component={DrilldownLibraryPage} />
+      </Switch>
```
dprokop, line 21: This is my fav demo, the simplicity if building drill down pages is just 💯 compared to SceneAppPage.

mdvictor (commented):
This looks great! I think this makes things a lot simpler and we should continue this work.

I had trouble running the demos through `demo.sh`, got some weird react errors, eventually got it working locally, not through docker.

packages/scenes-app/src/react-demo/DrilldownDemoPage.tsx
torkelo: @dprokop if you take a look at PageWrapper: https://github.com/grafana/scenes/pull/734/files/126e7e5bc2c3bb6105489a0fdeb92e9235024901#diff-a6c93be673ac22a956e62a0dd1fa818089796cf0ceb6321e2f3f075d3020e12bR14 

it will use the BreadcrumbContext to access any breadcrumbs defined, DrilldownLibraryPage does not add any breadcrumbs, that is added here on the parent level so that all child drilldowns get this parent level breadcrumb

packages/scenes-app/src/demos/interopDemo.tsx
torkelo: @dprokop this demo is not working as we moved the new contexts to new library so the core scenes library EmbededScene can not support this scenario any more. 

We have to add a new EmbeddedSceneWithContext scene object exported from `scenes-react` (Another argument maybe that a separate package is not great). 

Plan to restore this demo to working condition after this is merged and work on this in a separate PR (it's in the Epic)

packages/scenes-app/src/demos/interopDemo.tsx
torkelo: @dprokop Sadly not every SceneObject can act as a context right now, but DashboardScene can easily wrap its content in a SceneContextProvider (with itself as the context's parent). this would enable hook & react components usage. But this context would only get the global time range and variables, so it would not work for local contexts, say on panel or row level (if we ever add time range to them) this would need nested SceneContextProviders. 

Maybe some way to simplify this and make any scene object act as a context value but would require some new universal props on SceneObjectState and functions on SceneObject interface. Think we can revisit this after we see how the interoperability turns out

torkelo:
@mdvictor demo.sh ? not sure what that does. 

the simplest way is just `yarn dev` this will build the two libs and the demo app (in parallel watchers). 

then in your local grafana dev folder conf/custom.ini 

```ini
[plugin.grafana-scenes-app]
path = /Users/torkelodegaard/dev/scenes/packages/scenes-app
```

update to what ever path you have to the scenes repo

packages/scenes-app/src/react-demo/DrilldownDemoPage.tsx
dprokop: I see that now, yeah, thought PageWrapper is wrapping the BradcrumbProvider. Thanks

packages/scenes-app/src/demos/interopDemo.tsx
torkelo: @dprokop 

> I was wondering, since EmbeddedScene may be considered a context provider in most cases, what if it had a possibility to set it as a react context for interoperability? Something like exposeContext: boolean (not sure about the name), that would under the hood, in the EmbeddedScene renderer, setup the correct SceneContext.Provider.

This is exactly what it did in a previous commit, but removed that code as scenes lib cannot use the context and context provider from scenes-react

packages/scenes-react/src/components/VizPanel.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { DataProxyProvider } from '../DataProxyProvider';
+import { usePrevious } from 'react-use';
+import { getPanelOptionsWithDefaults } from '@grafana/data';
+import { writeSceneLog } from '../utils';
+import { useSceneContext } from '../hooks/hooks';
+
+export interface VizPanelProps {
+  title: string;
+  dataProvider?: SceneDataProvider;
+  viz: VizConfig;
```
dprokop, line 18: a nit, but i would call this `config`, the `viz` property name says very little about what's expected here (and what it actually gets)

packages/scenes/src/components/EmbeddedScene.tsx
```diff
@@ -65,6 +65,8 @@ function EmbeddedSceneRenderer({ model }: SceneComponentProps<EmbeddedScene>) {
       </div>
     </div>
   );
+
+  return inner;
```
dprokop: Unrelated/cosmetic change?

packages/scenes/src/services/UrlSyncManager.ts
```diff
@@ -37,8 +38,11 @@ export class UrlSyncManager implements UrlSyncManagerLike {
       this._stateSub.unsubscribe();
     }
 
+    const location = locationService.getLocation();
+
     this._sceneRoot = root;
-    this._lastPath = locationService.getLocation().pathname;
+    this._lastPath = location.pathname;
+    this._urlParams = new URLSearchParams(location.search);
```
dprokop, line 45: I assume this is to capture the url params when the scene rendered, so that later on, when syncing, we can get the actual params, not the ones that got updated on navigation?

dprokop (approved):
Not much comments from my side @torkelo , apart of a few observations. I think work is an excellent start to this react-first API.

packages/scenes-react/src/components/VizPanel.tsx
torkelo: @dprokop maybe `vizConfig` to be extra clear?

packages/scenes/src/components/EmbeddedScene.tsx
torkelo: @dprokop yea, remnant from where I conditionally wrapped the inner content in a SceneContext, reverted this change now so this file should be unchanged

mdvictor (approved):

packages/scenes-react/src/components/VizPanel.tsx
torkelo: @dprokop not sure I like it after testing, we would have  `config.options`, and `config.fieldConfig` etc, compared to ´viz.options`, `viz.fieldConfig` which I think read much better 

alternative maybe is `visualization` , but since we have VizConfigBuilders I and type of viz is VizConfig, think `viz´ is enough

packages/scenes/src/services/UrlSyncManager.ts
torkelo: @dprokop will actually revert this change as the reason I added it is no longer part of this PR (it was to access latest urlParams in a function syncNewObj, but this function is removed in this PR, doing all the URL sync changes in a separate PR

matyax:
What would be the best way to keep track on the development of this Scenes flavor to consider porting apps to it?

torkelo:
@matyax keep an eye on https://github.com/grafana/scenes/issues/761 and for this to be usable any time soon I hope many scene app devs will help with issues in the epic :)

PR #735, opened 2024-05-13
NestedScene: Don't show `cursor: pointer;` for everything

ashharrison90:
- removes `cursor: pointer;` from the `NestedScene` wrapper
- might be missing something, but i don't think this is necessary
- most obvious place to spot the problems this causes is the [alerting landing page in ops](https://ops.grafana-ops.net/alerting)
- everything under `Insights` shows a pointer cursor 🤯 
- this means it can look like something is clickable when it's not (e.g. try clicking in the gap to the right of `Grafana-managed alert rules`)


torkelo (approved):

dprokop (approved):
I guess that was a mistake :)

PR #736, opened 2024-05-13
Group by: Do not allow custom options

dprokop:
Custom options do not make sense in the group by variable which is supposed to provide source of aggregations based on the data source provided dimensions.

Ref https://github.com/grafana/hyperion-planning/issues/46


ashharrison90 (approved):
lgtm 👍

PR #744, opened 2024-05-23
demo: alternative time range selection on timeseries panel

darrenjaneczek:
This demo uses a custom implementation of `SceneTimeRangeLike` to overrides the default time range selection behavior to capture the selection, store it in a different time range, and display it on the panel as an annotation.

[override-scenes-timerange-behavior-to-set-annotation-for-alt-timerange.webm](https://github.com/grafana/scenes/assets/38694490/113353ed-d558-469d-9fdd-4c93f10c628c)

### DIFF
--- a/packages/scenes-app/src/demos/panelTimeRangeHandler.tsx
+++ b/packages/scenes-app/src/demos/panelTimeRangeHandler.tsx
@@ -1,0 +1,245 @@
+import {
+  SceneFlexLayout,
+  SceneFlexItem,
+  SceneQueryRunner,
+  SceneAppPage,
+  EmbeddedScene,
+  SceneAppPageState,
+  PanelBuilders,
+  VizPanel,
+  SceneObjectBase,
+  SceneTimeRangeLike,
+  sceneGraph,
+  SceneTimeRangeState,
+  SceneDataState,
+  SceneTimeRange,
+} from '@grafana/scenes';
+import { DATASOURCE_REF } from '../constants';
+import { getEmbeddedSceneDefaults } from './utils';
+import { FieldType, MutableDataFrame, TimeRange, getDefaultTimeRange } from '@grafana/data';
+
+export function getPanelTimeRangeHandlerDemoScene(defaults: SceneAppPageState): SceneAppPage {
+  return new SceneAppPage({
+    ...defaults,
+    subTitle: 'This timeseries panel has the ability to select an alternative time range. ',
+    getScene: () => {
+      return new EmbeddedScene({
+        ...getEmbeddedSceneDefaults(),
+        body: new SceneFlexLayout({
+          direction: 'column',
+          children: [
+            new SceneFlexItem({
+              height: 400,
+              body: PanelBuilders.timeseries()
+                .setData(getQueryRunnerFor3SeriesWithLabels())
+                .setTitle('Change selection by mouse-dragging a time range')
+                .setDisplayName('${__field.labels.cluster}')
+                .setBehaviors(
+                  [
+                    (vizPanel: VizPanel) => {
+                      patchPanelContext(vizPanel);
+                      const altTimeRangeScene = sceneGraph.findObject(vizPanel, (scene) => {
+                        return scene.state.key === 'altTimeRangeScene';
+                      });
+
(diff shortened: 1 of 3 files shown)


packages/scenes/src/components/SceneRefreshPicker.tsx
```diff
@@ -152,7 +152,7 @@ export function SceneRefreshPickerRenderer({ model }: SceneComponentProps<SceneR
   const { refresh, intervals, autoEnabled, autoValue, isOnCanvas, primary, withText } = model.useState();
   const isRunning = useQueryControllerState(model);
 
-  let text = refresh === RefreshPicker.autoOption.value ? autoValue : withText ? 'Refresh' : undefined;
+  let text = refresh === RefreshPicker.autoOption?.value ? autoValue : withText ? 'Refresh' : undefined;
```
darrenjaneczek, line 155: I noticed that `autoOption` was undefined, causing a few demos to fail.

leeoniya:
perhaps loosely related: https://github.com/grafana/grafana/pull/88107

darrenjaneczek:
> perhaps loosely related: [grafana/grafana#88107](https://github.com/grafana/grafana/pull/88107)
> 
> considering adding `onSelectRange()` callback to panelContext. maybe also `onClickDataPoint()` callback in near future.
> 
> feels like we maybe should introduce some kind of point annotations (besides exemplars), and x+y rectangular annotations? 🤔

Yes, quite related.
This one is more tied to time range selection. The other can be two-dimensional.

If panels could all generally have alternative behaviors for range selection, I think that could simplify this greatly. 

For this specific implementation, it would also be really interesting to be able to modify the range of the annotation by dragging the left or right edges. Can that be achieved simply by manipulating the panel context?

leeoniya:
> Can that be achieved simply by manipulating the panel context?

that would be super expensive to do on mousemove, since requires redraw. but maybe via eventBus/signals or smth. cc @dprokop

packages/scenes-app/src/demos/panelTimeRangeHandler.tsx
```diff
@@ -0,0 +1,245 @@
...
+              height: 400,
+              body: PanelBuilders.timeseries()
+                .setData(getQueryRunnerFor3SeriesWithLabels())
+                .setTitle('Change selection by mouse-dragging a time range')
+                .setDisplayName('${__field.labels.cluster}')
+                .setBehaviors(
+                  [
+                    (vizPanel: VizPanel) => {
+                      patchPanelContext(vizPanel);
+                      const altTimeRangeScene = sceneGraph.findObject(vizPanel, (scene) => {
```
gtk-grafana, line 41: might be nice to have a sceneGraph.findKey(s: string) function

packages/scenes-app/src/demos/panelTimeRangeHandler.tsx
```diff
@@ -0,0 +1,245 @@
...
+                  ])
+                .build(),
+            }),
+            new SceneFlexItem({
+              height: 30,
+              body: PanelBuilders.text().setOption('content', '').setTitle('Root time range from: ${__from:date:iso} to: ${__to:date:iso}').build(),
+            }),
+            new SceneFlexItem({
+              key: "altTimeRangeScene",
+              $timeRange: new SceneTimeRange(),
```
gtk-grafana, line 111: also be nice if we could take an array of SceneTimeRanges somehow

packages/scenes/src/components/SceneRefreshPicker.tsx
darrenjaneczek: Moved to 
- https://github.com/grafana/scenes/pull/751

packages/scenes-app/src/demos/panelTimeRangeHandler.tsx
darrenjaneczek: https://github.com/grafana/scenes/pull/753

leeoniya:
@darrenjaneczek https://github.com/grafana/grafana/pull/88107 was merged, it uses a different type of annotation that doesnt have the bottom hoverable region. maybe take a look.

PR #745, opened 2024-05-24
SceneQueryRunner: make some fields and methods protected

sd2k:
PR #587 is quite a large change and introduces more concepts to scenes, so isn't easy to justify merging right now. As a workaround, I'd like to be able to subclass and override some methods of the SceneQueryRunner to implement some similar behaviour in a different library, but I need various fields and methods to be accessible to the subclass, which this PR does by making them protected rather than private.

Note: I can _almost_ just completely copy/paste the class into my own code rather than extending it, but the [`getQueriesForVariables`] function looks for all _instances_ of SceneQueryRunner, which my copy/pasted class would not match (unlike a subclass, which should), so I think some ad-hoc variable support would break :(

I'm hoping this is a less controversial change than #587 which will unblock me - let me know what you think!

[`getQueriesForVariables`]: https://github.com/grafana/scenes/blob/07c5dc66746d36208cd121b938883db8ef2243f7/packages/scenes/src/variables/utils.ts#L88-L94

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.ts
@@ -11,6 +11,7 @@ import {
   DataSourceApi,
   DataTopic,
   PanelData,
+  ScopedVars,
   preProcessPanelData,
   rangeUtil,
 } from '@grafana/data';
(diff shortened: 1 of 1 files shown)


dprokop (commented):
@sd2k - please have a look at my review of @sd2k - please see my review of https://github.com/grafana/scenes/pull/587. I would prefer we push the other PR forward rather than introducing this change

sd2k:
Closing, not needed now that #587 has been merged :tada:

PR #750, opened 2024-05-27
Variables: Prioritize showing adhoc variable key and operator

oscarkilhed:
When the value of the adhoc filter gets long, you can't see the key and operator, which are arguably more important than showing a few more characters of the already truncated value.

Before: 
image

After:
image

Fixes: https://github.com/grafana/grafana/issues/88328

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
@@ -196,7 +196,8 @@ const getStyles = (theme: GrafanaTheme2) => ({
     flexShrink: 1,
   }),
   key: css({
-    flexShrink: 0,
+    minWidth: '90px',
+    flexShrink: 1,
   }),
   operator: css({
     flexShrink: 0,
@@ -205,6 +206,9 @@ const getStyles = (theme: GrafanaTheme2) => ({
     paddingLeft: theme.spacing(3 / 2),
     paddingRight: theme.spacing(3 / 2),
     borderLeft: 'none',
+    width: theme.spacing(3),
+    marginRight: theme.spacing(1),
+    boxSizing: 'border-box',
     // To not have button background and last select border intersect
     position: 'relative',
     left: '1px',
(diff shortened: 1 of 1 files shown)


dprokop:
@oscarkilhed - can we also add some max width constraint to the value input? While editing it (entering custom value) it can push the time controls outside of the container:

Screenshot 2024-05-27 at 13 08 02

packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
```diff
@@ -191,6 +192,15 @@ const getStyles = (theme: GrafanaTheme2) => ({
   widthWhenOpen: css({
     minWidth: theme.spacing(16),
   }),
+  value: css({
+    flexShrink: 1,
+  }),
+  key: css({
+    flexShrink: 0,
```
torkelo, line 199: why happens when the key is really long, will this stop ellipsis on the key? I would expect both to be ellipised with some min width.

oscarkilhed: Hmm. You're right. Trying some stuff out I think having it set to flex-shrink: 1 for the key as well looks best. The problem I run into with min-width is that for short keys it takes up unnecessary space.

oscarkilhed:
> @oscarkilhed - can we also add some max width constraint to the value input? While editing it (entering custom value) it can push the time controls outside of the container:
> 
> Screenshot 2024-05-27 at 13 08 02

Working on this. Apparently it's not that easy to solve. Not sure what's causing the element to grow outside of the parent element.

oscarkilhed:
Managed to fix the overlap of the remove button and time controls. Seems to have been because the remove button had 0 width.

Will continue working on the controls being pushed outside of the viewport.

oscarkilhed:
@dprokop I suggest we adress the growing issues in a different PR. I've got a draft PR [here](https://github.com/grafana/grafana/pull/88372) that solves the issue.

PR #757, opened 2024-05-28
MultiValueVariable: Fixes issue when value is all value but all value is not enabled

torkelo:
Adds logic for a special case when current value is the ALL_VALUE but the variable has not enabled all value, in this case we should just pick the first option. 

Refactored the validation function a bit so we can simplify the long nested if/else statements and leverage early return pattern instead. 

Fixes https://github.com/grafana/grafana/issues/88391

<details>
  <summary>📦 Published PR as canary version: <code>4.24.2--canary.757.9283024412.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@4.24.2--canary.757.9283024412.0
  # or 
  yarn add @grafana/scenes@4.24.2--canary.757.9283024412.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/layout/grid/SceneGridLayout.test.tsx
+++ b/packages/scenes/src/components/layout/grid/SceneGridLayout.test.tsx
@@ -319,27 +319,27 @@ describe('SceneGridLayout', () => {
       layout.onDragStop(
         [
           {
-              w: 12,
-              h: 8,
-              x: 0,
-              y: 2,
-              i: 'a',
+            w: 12,
+            h: 8,
+            x: 0,
+            y: 2,
+            i: 'a',
           },
           {
-              w: 24,
-              h: 1,
-              x: 0,
-              y: 0,
-              i: 'row-a',
+            w: 24,
+            h: 1,
+            x: 0,
+            y: 0,
+            i: 'row-a',
           },
           {
-              w: 12,
-              h: 8,
-              x: 0,
-              y: 10,
-              i: 'b',
-          }
-      ],
+            w: 12,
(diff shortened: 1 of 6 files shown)


ivanortegaalba (approved):
💯 Thanks Torkel for the quick fix

ivanortegaalba:
I tested locally, and the issue was solved 👍

packages/scenes/src/components/layout/grid/SceneGridLayout.test.tsx
```diff
@@ -319,27 +319,27 @@ describe('SceneGridLayout', () => {
...
-              i: 'b',
-          }
-      ],
+            w: 12,
+            h: 8,
+            x: 0,
+            y: 10,
+            i: 'b',
+          },
+        ],
```
torkelo, line 342: @ivanortegaalba what version of prettier are you on?

ivanortegaalba: Using the one from `node_modules` 😅  it only happens to me with this file. Does the format for you differently? I assumed this file was unformatted

ivanortegaalba: What worries me is why this is not formatted with the pre-commit hook?

ivanortegaalba: So, I think it is an issue with only this file that was unformatted. You didn't touch it in your PR, I removed a console log from there that I saw when reviewing your tests

PR #763, opened 2024-06-03
Variables: Search / typing performance improvements

torkelo:
The search / typing performance is terrible for the core Select component when we get over 10K items (with 100K options we are talking 1-3 seconds per keystroke).  

In this PR 

* I am moving the option filtering to be outside the Select component, and limiting the options we pass to the Select component to 10K.  
* Using uFuzzy for the actual filtering. 
* Adding a new demo to test a variable with many options 
* in order to properly memoize the uFuzzy instance, allOptions, haystack etc I created a new function called getOptionSearcher which replaces some of the logic we had in getOptionsForSelect. Added the same tests we have for getOptionsForSelect to the getOptionSearcher (plus a search test). The GroupBy variable is still using getOptionsForSelect (and I think it might be used by scene app plugins also), so not sure we can remove it. I can refactor a bit so that getOptionsForSelect i using getOptionSearch under the hood.  

Anyway this PR radically improves search / typing performance (and menu open perf) going from 1-3 seconds per keystroke to ~50-100ms

### DIFF
--- a/packages/scenes/src/variables/components/VariableValueSelect.tsx
+++ b/packages/scenes/src/variables/components/VariableValueSelect.tsx
@@ -9,17 +9,31 @@ import { VariableValue, VariableValueSingle } from '../types';
 import { selectors } from '@grafana/e2e-selectors';
 import { GrafanaTheme2, SelectableValue } from '@grafana/data';
 import { css, cx } from '@emotion/css';
+import { getOptionSearcher } from './getOptionSearcher';
+
+const filterNoOp = () => true;
 
 export function VariableValueSelect({ model }: SceneComponentProps<MultiValueVariable>) {
-  const { value, key } = model.useState();
+  const { value, text, key, options, includeAll } = model.useState();
+  const [inputValue, setInputValue] = useState('');
+  const optionSearcher = useMemo(
+    () => getOptionSearcher(options, includeAll, value, text),
+    [options, includeAll, value, text]
+  );
 
-  const onInputChange = model.onSearchChange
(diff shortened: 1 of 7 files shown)


packages/scenes/src/variables/components/VariableValueSelect.tsx
```diff
@@ -9,17 +9,31 @@ import { VariableValue, VariableValueSingle } from '../types';
...
+        model.onSearchChange!(value);
       }
-    : undefined;
+      return value;
+    }
+
+    return value;
+  };
+
+  const filteredOptions = optionSearcher(inputValue);
```
dprokop, line 36: @torkelo any chance we could apply this to filters and group by selects as well? I think that would be extremely helpful. I can help with that later this week if needed

dprokop (approved):
That's a massive improvement. Comparing the demo with and without the optimisation is 🤯

packages/scenes/src/variables/components/VariableValueSelect.tsx
torkelo: @dprokop I updated GroupByVariable to use it as well, but for AdhocFilters it requires a bit more work as it's not using the same types so not as easy to re-use, requires some refactoring

ivanortegaalba (approved):
The improvement is massive 💯 

Not critical: If the user want to scroll to bottom, it won't find the results if they are more than 10k. Shall we add a non-selectable option at the bottom if we crop the results saying "X more..." to indicate we're not displaying all results?. WDYT?

PR #764, opened 2024-06-03
Now works with custom value preserved

torkelo:
Merges into https://github.com/grafana/scenes/pull/763

Adds the behavior to preserve custom value input value when opening the Select after adding a custom value. 

With this change we can completely revert https://github.com/grafana/grafana/pull/87843   (and do not need https://github.com/grafana/grafana/pull/88601)

### DIFF
--- a/packages/scenes-app/src/demos/variables.tsx
+++ b/packages/scenes-app/src/demos/variables.tsx
@@ -211,7 +211,7 @@ export function getVariablesDemo(defaults: SceneAppPageState) {
                     .setTitle('Description')
                     .setOption(
                       'content',
-                      'This tab is mainly to test a variable with 100 000 options, to test search / typing performance'
+                      'This tab is mainly to test a variable with 100 000 options, to test search / typing performance. manyOptions=$manyOptions'
                     )
                     .build(),
                 }),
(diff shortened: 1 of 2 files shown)


ashharrison90 (commented):
@torkelo nice, not opposed to this at all and is definitely cleaner 🥳 

just to check my understanding, this won't work for the initial state though right? 

e.g. if you add a custom adhoc value, save the dashboard with that variable value, refresh the page then click the adhoc value select this won't restore the input text. are we concerned about that case? 🤔

torkelo:
> e.g. if you add a custom adhoc value, save the dashboard with that variable value, refresh the page then click the adhoc value select this won't restore the input text. are we concerned about that case?
@ashharrison90 I don't think so, I am more concerned with the use case of taking a long value say `My very long region name and sub name` and selecting that and wanting to change it, to so say. `My very long region name .*` 

But can't think of how to solve that one (other than keeping the current value selected and focused so if you start typing it will be cleared, but for that we need deep access to the input element)

ashharrison90 (approved):
ok let's go with this for now 👍 we can iterate later if needed

PR #766, opened 2024-06-04
AdHocFiltersVariable: Performance improvements

ashharrison90:
- marks variable value select `virtualized`
- makes the options and input state controlled outside of select for better performance
  - extension of https://github.com/grafana/scenes/pull/763 for adhoc filters
  - also means we can revert https://github.com/grafana/grafana/pull/87843 as this is cleaner 
- adds a demo page

<details>
  <summary>📦 Published PR as canary version: <code>4.27.0--canary.766.9398430196.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@4.27.0--canary.766.9398430196.0
  npm install @grafana/scenes@4.27.0--canary.766.9398430196.0
  # or 
  yarn add @grafana/scenes-react@4.27.0--canary.766.9398430196.0
  yarn add @grafana/scenes@4.27.0--canary.766.9398430196.0
  ```
</details>

### DIFF
--- a/packages/scenes-app/src/demos/variables.tsx
+++ b/packages/scenes-app/src/demos/variables.tsx
@@ -20,6 +20,7 @@ import {
   TextBoxVariable,
   QueryVariable,
   CustomVariable,
+  AdHocFiltersVariable,
 } from '@grafana/scenes';
 import { getQueryRunnerWithRandomWalkQuery } from './utils';
 
@@ -212,6 +213,58 @@ export function getVariablesDemo(defaults: SceneAppPageState) {
                     .setOption(
                       'content',
                       'This tab is mainly to test a variable with 100 000 options, to test search / typing performance. manyOptions=$manyOptions'
+                    )
+                    .build(),
+                }),
+              ],
+            }),
+          });
+        },
+      }),
+      new SceneAppPage({
+        title: 'Many adhoc variable values',
+        url: `${defaults.url}/many-adhoc-values`,
+        getScene: () => {
+          return new EmbeddedScene({
+            controls: [new VariableValueSelectors({})],
+            $variables: new SceneVariableSet({
+              variables: [
+                new AdHocFiltersVariable({
+                  name: 'manyAdhocOptions',
+                  getTagKeysProvider: async () => ({
+                    replace: true,
+                    values: [{
+                      value: 'a',
+                      text: 'A'
+                    }, {
+                      value: 'b',
+                      text: 'B'
+                    }, {
+                      value: 'c',
+                      text: 'C'
+                    }]
+                  }),
+                  getTagValuesProvider: async () => {
+                    return {
+                      replace: true,
+                      values: getRandomOptions(100000).map(({ value, label }) => ({
+                        value,
+                        text: label,
+                      }))
+                    };
+                  },
+                }),
+              ],
+            }),
+            body: new SceneFlexLayout({
+              direction: 'column',
+              children: [
+                new SceneFlexItem({
+                  body: PanelBuilders.text()
+                    .setTitle('Description')
+                    .setOption(
+                      'content',
+                      'This tab is mainly to test an adhoc variable with 100 000 options, to test search / typing performance. manyAdhocOptions=$manyAdhocOptions'
                     )
                     .build(),
                 }),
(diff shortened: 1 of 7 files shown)


dprokop:
hahah, was just talking about this :D

dprokop:
Also, another thing I see (in the provided demo) is perceivable performance difference between thevariables many options demo vs filters many options demo (turn the sound on for commentary):

https://github.com/grafana/scenes/assets/2376619/b72e3f6e-cabc-49d9-819f-209fbd5a77aa

ashharrison90:
> @ashharrison90 , i've noticed a visual regression with this PR:
> 
> Before:
> 
> image image
> After:
> 
> image image
> Mind the difference of the options being passed to the Select component, seems like this implementation adds an empty option.

ok this is due to `getOptionSearcher` adding the value to the options array if it doesn't exist in the options. maybe @torkelo knows if there's a reason why this is necessary? my guess is that without it nothing would show in the dropdown, but i think it's better to provide a custom `isValidNewOption={(inputValue) => inputValue.trim().length > 0}`. i've made that change anyway, can always undo if necessary.

> Also, another thing I see (in the provided demo) is perceivable performance difference between thevariables many options demo vs filters many options demo (turn the sound on for commentary):
> 
>  Screen.Recording.2024-06-04.at.16.30.19.mov

this isn't that surprising tbh 😅 the adhoc filter has to do more work. it _always_ fetches options `onMenuOpen`, and it has to loop over the entire options array to convert from a `SelectableValue` to a `VariableValueOption`. i think that conversion is pretty fast, but when there's 10,000 options it adds up! 

i've made the limit in `getOptionSearcher` configurable and lowered it to 1000 in `AdHocFilterRenderer` which seems to have helped a fair bit. think 1000 options in a select is still more than enough 😅 i've left it at 10,000 elsewhere for now, but we could probably make that change across the board

ashharrison90:
> @ashharrison90
> 
> > ok this is due to getOptionSearcher adding the value to the options array if it doesn't exist in the options. maybe @torkelo knows if there's a reason why this is necessary? my guess is that without it nothing would show in the dropdown, but i think it's better to provide a custom isValidNewOption={(inputValue) => inputValue.trim().length > 0}. i've made that change anyway, can always undo if necessary.
> 
> I am not familiar with isValidNewOption, what does it do? no sure how it relates to showing the current value to the options? Reading the docs it seems unrelated.
> 
> But yes, adding the current value to the options I think is to make it show the current value

@torkelo the default `isValidNewOption` provided by `react-select` will not show the `Use custom value: <customValue>` option **for the current custom value**. but if you override it and just say any non-empty string is fine, it will show that option which i think negates the need for manually adding that custom option to the options array.

ashharrison90:
it's working for me on both the `Many variable options` and `Many adhoc variable values` demo pages on this branch:
image

torkelo:
@ashharrison90 I tested i again on this branch and it's still not working, it does not show custom values set from URL. it seems to work just after creating a value, but it does not accept an existing value that set before mount. 

Try this URL 
http://localhost:3000/a/grafana-scenes-app/demos/variables/many-values?var-interval=1d&var-server=AA&var-keyValue=1&var-queryVar=A&var-pod=AAA&var-handler=AAAA&var-manyOptions=custom

Screenshot 2024-06-05 at 21 00 06

dprokop:
>@ashharrison90 I tested i again on this branch and it's still not working, it does not show custom values set from URL. it seems to work just after creating a value, but it does not accept an existing value that set before mount.

I observe this as well, but it works fine for the AdHocFilters (see this link http://localhost:3000/a/grafana-scenes-app/demos/variables/many-adhoc-values?var-manyAdhocOptions=a,A%7C%3D%7CusingACustom)

ashharrison90:
@torkelo @dprokop yeah gotcha, working on it now 👍

ashharrison90:
@torkelo @dprokop ok seems like if you pass a `SelectableValue` instead of just a plain string it displays correctly 😕 haven't dug into why exactly yet.

think i've also fixed the performance with the adhoc filter now as well - i was missing the no op `filterOption`. with that the performance seems much more in line with the normal `VariableValueSelect`

torkelo:
@ashharrison90 nice!

torkelo (approved):

dprokop:
@ashharrison90 awesome, testing now

dprokop (approved):

PR #768, opened 2024-06-04
SceneQueryRunner: Act as if we're loading when waiting for variables to load.

oscarkilhed:
When a scene loads and VizPanels with variable dependencies are waiting to load, there's no indication that the panel will eventually load.

This PR sets the state of the PanelData to loading when we're waiting for the variable dependencies.

Fixes: https://github.com/grafana/grafana/issues/84574

One could argue that setting the LoadingState to "NotStarted" or introducing a new state called "WaitingForDependency" would make more sense. But it would also introduce a lot more changes to grafana as a whole.

Before (notice the loading indicator on the panel chrome is not running before the variable is loaded):
loading-before

After:
loading-after

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.test.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.test.ts
@@ -607,7 +607,8 @@ describe('SceneQueryRunner', () => {
       await new Promise((r) => setTimeout(r, 1));
 
       expect(variable.state.loading).toBe(true);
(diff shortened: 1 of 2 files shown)


torkelo:
been thinking about doing this change for the last couple of months, I think it's the right thing to do. Maybe add a unit test?

oscarkilhed:
@torkelo yeah, I need to change the way the unit tests detect if we're making a query as well. Will add tests for this as part of that

torkelo (approved):

PR #769, opened 2024-06-04
provide an option to hide the label from the variable value controller

eskirk:
when using the `VariableValueControl` there is not an option to hide the label in the variable value controller, this change provides an option to work alongside the variable-level hide config so users can hide the filters in the main control bar, and also hide the label in the places where the variable value controller is used. 

image

### DIFF
--- a/packages/scenes/src/variables/components/VariableValueControl.tsx
+++ b/packages/scenes/src/variables/components/VariableValueControl.tsx
@@ -10,7 +10,7 @@ export interface VariableValueControlState extends SceneObjectState {
   /** Render the specific select control for a variable */
   variableName: string;
   /** Hide the label in the variable value controller */
-  hideControllerLabel?: boolean;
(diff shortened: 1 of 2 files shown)


packages/scenes/src/variables/components/VariableValueControl.tsx
```diff
@@ -9,6 +9,8 @@ export interface VariableValueControlState extends SceneObjectState {
   layout?: ControlsLayout;
   /** Render the specific select control for a variable */
   variableName: string;
+  /** Hide the label in the variable value controller */
+  hideControllerLabel?: boolean;
```
torkelo, line 13:
```suggestion
  hideLabel?: boolean;
```

think it makes sense to use the same name / term as on the variable level

torkelo (approved):

dprokop (approved):
+1 for @torkelo comment about property name.

PR #770, opened 2024-06-05
VariableValueSelect: Adjust data-testid in OptionWithCheckbox

Sergej-Vlasov:
adjust `data-testid` in `OptionWithCheckbox` to align with https://github.com/grafana/grafana/pull/88118

<details>
  <summary>📦 Published PR as canary version: <code>4.26.2--canary.770.9388861265.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@4.26.2--canary.770.9388861265.0
  npm install @grafana/scenes@4.26.2--canary.770.9388861265.0
  # or 
  yarn add @grafana/scenes-react@4.26.2--canary.770.9388861265.0
  yarn add @grafana/scenes@4.26.2--canary.770.9388861265.0
  ```
</details>


packages/scenes/src/variables/components/VariableValueSelect.tsx
```diff
@@ -176,16 +176,18 @@ export const OptionWithCheckbox = ({
...
       title={data.title}
     >
       <div className={optionStyles.checkbox}>
         <Checkbox value={isSelected} />
       </div>
-      <div className={selectStyles.optionBody}>
+      <div
+        className={selectStyles.optionBody}
+        data-testid={selectors.pages.Dashboard.SubMenu.submenuItemValueDropDownOptionTexts(
+          data.label || String(data.value)
```
torkelo, line 188: never use || like this (as it would be false with "0"), better to do data.label ?? String(data.value). 

But curious why this is needed, I think for variables the label should always be set?

Sergej-Vlasov: mostly because typescript complains because `data.label?: string | undefined`.
Good catch with `||` I will change to `??`. I believe it would only evaluate to false if `0` is integer and not string `"0"`, but better to use `??`.

torkelo (approved):

PR #775, opened 2024-06-06
React scenes: Add useDataTransformer hook

oscarkilhed:
Decided to take a look at implementing a react hook for using transformations. Based on the useQueryRunner hook. 

https://github.com/grafana/scenes/assets/468940/9d3bf4e7-9247-4386-862d-96ad478880a6


torkelo:
looks good!

packages/scenes-react/src/hooks/useDataTransformer.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export function useDataTransformer(options: UseDataTransformerOptions) {
+  const scene = useSceneContext();
+  const key = useId();
+
+  let dataTransformer = scene.findByKey<SceneDataTransformer>(key);
+
+  if (!dataTransformer) {
+    dataTransformer = new SceneDataTransformer({ $data: options.data, transformations: options.transformations });
```
torkelo, line 19: So one problem that is a bit messy here, as useQueryRunner is built to be able to be used standalone (without VizPanel etc), it adds the SceneQueryRunner to the scene (via addToScene), which sets the parent. 

So using it here like this actually resets the parent (and you get a warning in console). In VizPanel I solved this by wrapping the data provider in a new DataProxyProvider, but this adds a level of "proxying" and feels in-efficient. 

Not sure what the correct solution is though, change SceneObjectBase to somehow allow sharing references to some scene object instance (without affecting parent) ?

oscarkilhed: Ah. Missed that bit, but makes sense. I really liked this way of passing the data provider to the hook, so would be nice if we could get it working. 

I don't have any intuition about the implications of sharing references unfortunately.

I could implement this using the proxy for now, it should be fairly simple to replace once we figure out a better way.

packages/scenes-react/src/hooks/useDataTransformer.ts
```diff
@@ -0,0 +1,37 @@
...
+  transformations: Array<DataTransformerConfig | CustomTransformerDefinition>;
+  data: SceneDataProvider;
+}
+
+export function useDataTransformer(options: UseDataTransformerOptions) {
+  const scene = useSceneContext();
+  const key = useId();
+
+  let dataTransformer = scene.findByKey<SceneDataTransformer>(key);
+  console.log('asd')
```
mdvictor, line 18: sneaky log :D

oscarkilhed: What the hell, I thought I removed this before comitting :)

packages/scenes-app/src/react-demo/GraphWithTransformation.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { plainGraph } from './visualizations';
+import { DataTransformerID, ReducerID } from '@grafana/data';
+
+interface Props {
+  maxDataPoints?: number;
+  title: string;
+  reducer: ReducerID;
+}
+
+export function GraphWithWindowTransformation({ maxDataPoints, title, reducer}: Props) {
```
torkelo, line 12: since this is only used from the TranasformationsDemo page (for now), it might be easier to have the content of this file in there as well to make that demo a bit more easy to follow without having to jump to a different file

torkelo (approved):
Looks good!

PR #781, opened 2024-06-10
Variables: Add support to read only variables  and expose missing types for custom variables in core grafana

axelavargas:
**What is this PR solving?**

For `snapshots` (a feature of core grafana), we only store the `current` variable value, and don't store `options` for variables such as `queries, datasources, ...` .
The problem was that the Scenes implementation does validate `options` against the `current` variable value, if the `current` value did not exist, it cleared up that value.  

**What is this PR doing ?**

I added some tweaks to support read-only variables, and also exposed missing types to support a new a`SnapshotVariable` in core grafana ([PR](https://github.com/grafana/grafana/pull/88967) ), which has a similar implementation to `CustomVariable`, but it's read-only and does not validate `options` with the current variable value.  
SnapshotVariables

## Todo: 
- [x] Double-check approach 
- [x] Add unit test

<details>
  <summary>📦 Published PR as canary version: <code>4.30.0--canary.781.9463613592.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@4.30.0--canary.781.9463613592.0
  npm install @grafana/scenes@4.30.0--canary.781.9463613592.0
  # or 
  yarn add @grafana/scenes-react@4.30.0--canary.781.9463613592.0
  yarn add @grafana/scenes@4.30.0--canary.781.9463613592.0
  ```
</details>


packages/scenes/src/variables/variants/SnapshotVariable.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+import { MultiValueVariable, MultiValueVariableState, VariableGetOptionsArgs } from './MultiValueVariable';
+import { sceneGraph } from '../../core/sceneGraph';
+
+export interface SnapshotVariableState extends MultiValueVariableState {
+  query: string;
+  isReadOnly: boolean;
+}
+
+export class SnapshotVariable extends MultiValueVariable<SnapshotVariableState> {
```
dprokop, line 16: This is a peculiar type of variable, that I think won't neccessarily be used anywhere else but core dashboards. Maybe we could move it to core, rather than exposing from the library?

axelavargas: Good point, I will move the variable to core, and keep the read-only implementation here :)

torkelo (approved):

PR #782, opened 2024-06-10
React scenes: Annotations support

mdvictor:
Screenshot 2024-06-12 at 14 48 45

This PR adds annotation support through the `AnnotationLayer` component. These layers can be controlled by the `AnnotationPicker` component.

It also changes the SceneContextProvider to hold an array of context children instead of a single child.

TODO:

- [x] Tests
- [x] Refactor
- [x] Add annotation controls


packages/scenes-react/src/DataLayerProxyProvider.ts
```diff
@@ -0,0 +1,18 @@
+import { SceneDataLayerProvider, SceneDataState, SceneObjectRef } from "@grafana/scenes";
+import { DataProxyProvider, DataProxyProviderState } from "./DataProxyProvider";
+
+export interface DataLayerProxyProviderState extends SceneDataState {
+  source: SceneObjectRef<SceneDataLayerProvider>;
+}
+
+// :/
+export class DataLayerProxyProvider extends DataProxyProvider implements SceneDataLayerProvider {
```
mdvictor, line 9: Not very happy with this and didn't spend much on it. Problem is that the queryRunner checks if the `isDataLayer` property exists otherwise it won't use the data.

packages/scenes-app/src/react-demo/AnnotationsDemoPage.tsx
```diff
@@ -0,0 +1,56 @@
...
+        lines: 15,
+        refId: 'Anno',
+        scenarioId: 'annotations',
+      },
+    };
+
+    return (
+        <PageWrapper title="Annotations" subTitle="Annotation demo page">
+            <AnnotationLayer name="Global Anno 1" query={query1} />
+            <AnnotationLayer name="Global Anno 2" query={query2} />
```
mdvictor, line 45: I think these layers should (also?) hold children and only apply to them so we can have different annotations on different panels or groups of panels. Maybe something like:
```
<AnnotationLayer name="global" />
<AnnotationLayer name="local1" >
  <Panel />
  <Panel />
</AnnotationLayer />
<AnnotationLayer name="local2" >
  <Panel />
</AnnotationLayer />
```

But I'm not sure what to tie the `SceneDataLayerProvider` to

mdvictor: And with this I think the `useDataLayers` hook wouldn't matter that much since we can just do

```
<AnnotationLayer name="local2" >
  <Panel />
</AnnotationLayer />
```

to apply an annotation to a panel, but it would probably still be helpful to have the hook and apply annotation directly to the query runner.

torkelo: yea, I don't think we need the hook. 

I think AnnotationLayer should always take children (so wrap content)

packages/scenes-react/src/contexts/SceneContextObject.tsx
```diff
@@ -79,4 +81,31 @@ export class SceneContextObject extends SceneObjectBase<SceneContextObjectState>
...
+        this.setState({ $data: layerSet });
+      }
+
+      writeSceneLog('SceneContext', `Adding annotation data layer: ${layer.state.name} key: ${layer.state.key}`);
+
+      return () => {
+          layerSet.setState({ layers: layerSet.state.layers.filter((x) => x !== layer) });
+          writeSceneLog('SceneContext', `Removing annotation data layer: ${layer.state.name} key: ${layer.state.key}`);
+      }
+  }
```
torkelo, line 110: Because these functions will only be called from AnnotationLayer I would just move them to be util functions in that file

packages/scenes-react/src/components/AnnotationPicker.tsx
```diff
@@ -0,0 +1,20 @@
...
+  let picker = scene.findByKey<SceneDataLayerControls>(key);
+
+  if (!picker) {
+    picker = new SceneDataLayerControls();
+  }
+
+  useEffect(() => scene.addToScene(picker), [picker, scene]);
+
+  return <picker.Component model={picker} />;
+}
```
torkelo, line 20: Instead of doing this, I think we can model this in the same way as VariableControl  (name it DataLayerControl).

and have it take name as prop, then lookup the closest data layer (in all data sets walking up) and when find render it

packages/scenes-react/src/components/DataLayerControl.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const layer = getLayer(layerSets, name);
+
+  if (!layer) {
+    return <div>Annotation {name} not found</div>;
+  }
+
+  return <DataLayerControlSwitch layer={layer} />;
+}
+
+function getLayer(layers: SceneDataLayerProvider[], name: string): SceneDataLayerProvider | undefined {
```
mdvictor, line 23: need to write some tests for this one

packages/scenes-react/src/contexts/SceneContextObject.tsx
```diff
@@ -9,7 +9,7 @@ import {
 import { writeSceneLog } from '../utils';
 
 export interface SceneContextObjectState extends SceneObjectState {
-  childContext?: SceneContextObject;
+  contextChildren?: SceneContextObject[];
```
torkelo, line 12: this naming is a bit awkward (as there already is a property named children), maybe
* childContexts or 
* nestedContexts ?

packages/scenes-react/src/contexts/SceneContextProvider.tsx
```diff
@@ -52,7 +52,7 @@ export function SceneContextProvider({ children, timeRange, withQueryController
 
     if (parentContext) {
       getUrlSyncManager().handleNewObject(childContext);
-      parentContext.setState({ childContext });
+      parentContext.setState({ contextChildren: [...(parentContext.state.contextChildren ?? []), childContext] });
```
torkelo, line 55:
```suggestion
      parentContext.addChildContext(childContext)
```

would be nice, just for readability to move that to a function on the context class

packages/scenes-react/src/contexts/SceneContextProvider.tsx
```diff
@@ -62,7 +62,9 @@ export function SceneContextProvider({ children, timeRange, withQueryController
       deactivate();
 
       if (parentContext) {
-        parentContext.setState({ childContext: undefined });
+        parentContext.setState({
+          contextChildren: parentContext.state.contextChildren?.filter((context) => childContext !== context),
+        });
       }
```
torkelo, line 68:
```suggestion
        parentContext.removeChildContext(childContext);
```

packages/scenes-react/src/utils/testUtils.tsx
```diff
@@ -25,7 +25,7 @@ export function TestContextProvider({ children, value }: TestContextProviderProp
 
   useEffect(() => {
     if (parentContext) {
-      parentContext.setState({ childContext: value });
+      parentContext.setState({ contextChildren: [...(parentContext.state.contextChildren ?? []), value] });
```
torkelo, line 28:
```suggestion
      parentContext.addChildContext(value)
```

packages/scenes-react/src/utils/testUtils.tsx
```diff
@@ -35,7 +35,7 @@ export function TestContextProvider({ children, value }: TestContextProviderProp
       deactivate();
 
       if (parentContext) {
-        parentContext.setState({ childContext: undefined });
+        parentContext.setState({ contextChildren: [...(parentContext.state.contextChildren ?? []), value] });
```
torkelo, line 38:
```suggestion
        parentContext.removeChildContext(value)
```

torkelo (approved):
Looks good, just had some tweak suggestions

packages/scenes-react/src/contexts/SceneContextObject.tsx
mdvictor: I think `childContexts` sounds good!

packages/scenes-react/src/contexts/SceneContextProvider.tsx
mdvictor: Agreed!

PR #784, opened 2024-06-11
SceneRefreshPicker: Fixes url state issue

torkelo:
Fixes a minor issue with `refresh=` url state appearing in `scenes-react` demos, there is was a mismatch URL state and updateFromUrl

It's a bit messy, the "off state" for auto refresh is represented by an empty string (both in internal design system RefreshPicker and in state), so leaving it like that as changing that could be messy


axelavargas (approved):
looks good to me :)

ivanortegaalba (approved):
👍

PR #785, opened 2024-06-11
Variables: Fixes sceneInterpolator when string contains variables with object prototype function names

torkelo:
Fixes issues interpolating a string like `"$toString"` , if there is no variable named `toString` this should leave the original string unchanged but due to using object property lookup we matched object prototype functions (Both on scopedVars and macros) 

<details>
  <summary>📦 Published PR as canary version: <code>4.30.0--canary.785.9463381206.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@4.30.0--canary.785.9463381206.0
  npm install @grafana/scenes@4.30.0--canary.785.9463381206.0
  # or 
  yarn add @grafana/scenes-react@4.30.0--canary.785.9463381206.0
  yarn add @grafana/scenes@4.30.0--canary.785.9463381206.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/interpolation/sceneInterpolator.ts
+++ b/packages/scenes/src/variables/interpolation/sceneInterpolator.ts
@@ -60,10 +60,12 @@ function lookupFormatVariable(
   scopedVars: ScopedVars | undefined,
   sceneObject: SceneObject
 ): FormatVariable | null {
-  const scopedVar = scopedVars?.[name];
+  if (scopedVars && scopedVars.hasOwnProperty(name)) {
+    const scopedVar = scopedVars[name];
 
-  if (scopedVar) {
-    return getSceneVariableForScopedVar(name, scopedVar);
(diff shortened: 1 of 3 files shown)


packages/scenes/src/variables/interpolation/sceneInterpolator.ts
```diff
@@ -60,19 +60,22 @@ function lookupFormatVariable(
   scopedVars: ScopedVars | undefined,
   sceneObject: SceneObject
 ): FormatVariable | null {
-  const scopedVar = scopedVars?.[name];
+  if (scopedVars && scopedVars.hasOwnProperty(name)) {
```
oscarkilhed, line 63: Not sure exactly how the scopedVars work, but could there not be an issue if the variable name is __dataContext which is a property of scopedVars? Maybe this is just a problem in theory.

oscarkilhed (approved):

PR #786, opened 2024-06-11
chore: update create-plugin version and config for demo app

sd2k:
I ran `yarn create @grafana/create-plugin@latest update` then updated the plugin's custom tsconfig.json to include the docusaurus docs, and the docker-compose.yaml to use the grafana-dev image, but otherwise everything ran pretty smoothly.

As per the [updating docs](https://grafana.com/developers/plugin-tools/migration-guides/update-create-plugin-versions?current-package-manager=npm).

PR #787, opened 2024-06-11
Add Machine Learning section to docs & example page to demo app

sd2k:
This PR:
- adds a section on using `@grafana/scenes-ml` with Scenes to the website docs
- adds a page to the demo app which shows off those features
- updates the demo app's webpack config to work with `@grafana/scenes-ml` (note: this change avoids a dependency on #786 but won't be necessary once we update the demo app's plugin config using a similar method to that PR)

Keeping this as draft for now since it depends on #786 and I want to add some more detailed suggestions on when to use each component to the docs.


docusaurus/website/sidebars.js
```diff
@@ -60,6 +60,18 @@ const sidebars = {
         'advanced-time-range-comparison',
       ],
     },
+    {
+      type: 'category',
+      label: 'Machine Learning',
```
dprokop, line 65: Can you call this section `@grafana/scenes-ml`? The machine learning header indicates scenes provides some functionalities around it which is not true.

dprokop (approved):

PR #788, opened 2024-06-12
ScenesReact: Cache SceneQueryRunners and other scene object by a key / hashing string

torkelo:
With  `@grafana/scenes-react` creating scene object on the fly as react components are rendered (or hooks called) we need a different way to cache query results, query variables and viz panel state. 

This PR explores scene object cache and key hashing mechanism. 

Example, simple string key:  

```tsx
  const dataProvider = useQueryRunner({
    queries: [{ uid: 'gdev-testdata', refId: 'A', scenarioId: 'random_walk', alias: queryAlias ?? 'env = $env' }],
    maxDataPoints: maxDataPoints,
    cacheKey: "My unique cache key string",
  });
```

Example: Object as key 

```tsx
  const queries = [{ uid: 'gdev-testdata', refId: 'A', scenarioId: 'random_walk', alias: queryAlias ?? 'env = $env' }];

  const dataProvider = useQueryRunner({
    queries,
    maxDataPoints: maxDataPoints,
    cacheKey: queries,
  });
```

By default the system will get the hash key by doing `JSON.stringify` on objects (but with sorted properties so property order does not matter), same as react-query. 

cacheKey can also be an array so you can combine multiple strings (or numbers) or object values 

```tsx
  const queries = [{ uid: 'gdev-testdata', refId: 'A', scenarioId: 'random_walk', alias: queryAlias ?? 'env = $env' }];

  const dataProvider = useQueryRunner({
    queries,
    maxDataPoints: maxDataPoints,
    cacheKey: [queries, maxDataPoints],
  });
```

If you have really large query objects and want faster hashing by reference there is a util function `cacheByRef` which returns a unique number for a given object reference. 

The cache key hashing will also always append the scene object constructor name to the key to avoid conflicts when same cache key is used for different SceneObject types.


leeoniya:
> If you have really large query objects and want faster hashing by reference

i have a lot of local dashboards that embed the data in testData csv content, or raw frames. also the snapshots from Get help embed the data, so would be good to make sure that does not get re-stringified on each render.

torkelo:
@leeoniya yes, the hashing by reference is for those cases. But this method described here is not to be used for core dashboards (at least not initially, as there we cache the full scene), only for custom scene apps.

packages/scenes-react/src/hooks/useSceneObject.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Also handles the lookup of the object from the cache if a cache key is provided.
+ */
+export function useSceneObject<T extends SceneObject>(options: UseSceneObjectProps<T>) {
+  const scene = useSceneContext();
+  const key = useId();
+  const cache = getSceneObjectCache();
+  let cacheKeyHash = options.cacheKey ? cache.getHashKey(options.cacheKey, options.objectConstructor) : undefined;
+
+  let obj = scene.findByKey<T>(key);
+
```
matyax, line 23: Thinking out loud. If during runtime I purposely change the `cacheKey` to get a new `SceneObject`, I would still get the same `SceneObject` because it was first looked using `key`, right?

matyax (commented):
Looks great! Left a couple of comments.

packages/scenes-react/src/caching/SceneObjectCache.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export class SceneObjectCache {
+  #cache: LRUCache<string, SceneObject>;
+  #objectRefIds = new WeakMap<any, number>();
+  #objectRefIdCounter = 0;
+
+  public constructor() {
+    this.#cache = new LRUCache({
+      max: 500,
+      ttl: 1000 * 60 * 5,
```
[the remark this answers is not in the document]
torkelo: @matyax implementing retries would be quite simple to do, backend_srv (underlying fetch abstraction) supports it I think

packages/scenes-react/src/hooks/useSceneObject.ts
torkelo: @matyax ? not sure I understand what you mean, if the cacheKey changes you would get a new scene object

packages/scenes-react/src/hooks/useSceneObject.ts
```diff
@@ -0,0 +1,52 @@
...
+  if (!obj && cacheKeyHash) {
+    obj = cache.get<T>(cacheKeyHash);
+
+    if (obj) {
+      if (obj.parent !== scene) {
```
mdvictor, line 28: Nit: can just be if(obj && obj.parent !== scnee)

mdvictor (approved):
This LGTM, I think we should go forward with it.

PR #799, opened 2024-06-19
TimeRangeCompare: Do not re-run queries if all have opted out of comparison

domasx2:
:wave: 

We found an issue in app o11y where changing time window comparison value causes query runner to re-run queries even if all queries have opted out of TWC. This causes unnecessary querying and side effects for us.

This PR fixes makes twc not trigger re-run if all queries are opted-out

Thanks!

### DIFF
--- a/packages/scenes/src/components/SceneTimeRangeCompare.tsx
+++ b/packages/scenes/src/components/SceneTimeRangeCompare.tsx
@@ -124,7 +124,7 @@ export class SceneTimeRangeCompare
     queries: SceneDataQuery[]
   ): boolean {
     return (
-      prev.compareWith !== next.compareWith && queries.filter((query) => query.timeRangeCompare !== false).length > 0
+      prev.compareWith !== next.compareWith && queries.find((query) => query.timeRangeCompare !== false) !== undefined
     );
   }
 
--- a/packages/scenes/src/components/SceneTimeRangeCompare.tsx
+++ b/packages/scenes/src/components/SceneTimeRangeCompare.tsx
@@ -4,7 +4,7 @@ import { ButtonGroup, ButtonSelect, Checkbox, ToolbarButton, useStyles2 } from '
 import React from 'react';
 import { sceneGraph } from '../core/sceneGraph';
 import { SceneObjectBase } from '../core/SceneObjectBase';
-import { SceneComponentProps, SceneObjectState, SceneObjectUrlValues } from '../core/types';
+import { SceneComponentProps, SceneDataQuery, SceneObjectState, SceneObjectUrlValues } from '../core/types';
 import { DataQueryExtended } from '../querying/SceneQueryRunner';
 import { ExtraQueryDescriptor, ExtraQueryDataProcessor, ExtraQueryProvider } from '../querying/ExtraQueryProvider';
 import { SceneObjectUrlSyncConfig } from '../services/SceneObjectUrlSyncConfig';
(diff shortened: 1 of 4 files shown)


packages/scenes/src/components/SceneTimeRangeCompare.tsx
```diff
@@ -117,9 +117,15 @@ export class SceneTimeRangeCompare
...
-  public shouldRerun(prev: SceneTimeRangeCompareState, next: SceneTimeRangeCompareState): boolean {
-    return prev.compareWith !== next.compareWith;
+  // The query runner should rerun the comparison query if the compareWith value has changed and there are queries that haven't opted out of TWC
+  public shouldRerun(
+    prev: SceneTimeRangeCompareState,
+    next: SceneTimeRangeCompareState,
+    queries: SceneDataQuery[]
+  ): boolean {
+    return (
+      prev.compareWith !== next.compareWith && queries.filter((query) => query.timeRangeCompare !== false).length > 0
```
sd2k, line 127: tiny nit: can use find and exit early if we find any

```suggestion
      prev.compareWith !== next.compareWith && queries.find((query) => query.timeRangeCompare !== false) !== undefined
```

sd2k (commented):
It took me a few minutes to figure out why this was happening... but it makes sense now. The state of the TRC is changing which makes it think queries should be rerun, and even though there are no _additional_ queries to rerun, the primary query still gets rerun.

I wonder if it's possible to do this without changing the `ExtraQueryProvider` interface, by setting a property like `foundExtraQueries` in the `TimeRangeCompare`'s `getExtraQueries` method and checking it in `shouldRerun` 🤔 I'm not sure that's better, though...

domasx2:
> I wonder if it's possible to do this without changing the ExtraQueryProvider interface, by setting a property like foundExtraQueries in the TimeRangeCompare's getExtraQueries method and checking it in shouldRerun 🤔 I'm not sure that's better, though...

domasx2:
> I wonder if it's possible to do this without changing the ExtraQueryProvider interface, by setting a property like foundExtraQueries in the TimeRangeCompare's getExtraQueries method and checking it in shouldRerun 🤔 I'm not sure that's better, though...

There can be many query runners for a single `TimeRangeCompare`, so I think this kind of approach may get sticky with complex state :thinking:

sd2k:
Ah yeah, good point.

This will work but I'm still a little annoyed by it (my fault, not yours :grinning:) - I feel like the query runner should be smart enough to realise that the provider didn't return any extra queries from `getExtraQueries` and skip calling `provider.shouldRerun` if the query hasn't changed; that way each provider wouldn't have to deal with properties like `timeRangeCompare` in more than one place. But getting that to work properly sounds complicated so perhaps it isn't worth it.

I'll wait for @dprokop to take a look before approving. Thanks @domasx2!

dprokop (approved):
Looks good @domasx2! Sorry me and @sd2k missed that when merging the refactor.

Agree with @sd2k this ideally was handled in the run query, but that seems like a way bigger effort to get in place. However, this `shouldRerun` interface change is OK anyways, I can imagine a situation when certain extra query provider can opt in/out based on a query bein relevant or not - regardles of any additional query setup as we have with `timeRangeCompare` for time comparison.


@domasx2  - minor feedback regarding this PR overall, using `TWC` abbreviation makes this PR not accessible to anyone who's not familiar with the meaning. I believe we do not use this anywhere, so it's probably only the 3 of us involved here who got it right of the bat. :)

sd2k (approved):

PR #801, opened 2024-06-20
VizPanel: Clear options value when the option is undefined

ivanortegaalba:
Fixes [#89270](https://github.com/grafana/grafana/issues/89270)

**Problem**
[_.mergeWith](https://lodash.info/doc/mergeWith) ignores `undefined` values from the source object. When clearing a `VizPanel` option, we set it to `undefined` so it was not working.

To be able to set an option as `undefined`, we have to treat that case in the customizer cb passed to `_.mergeWith`.

|Before|After|
|-|-|
<video src="https://github.com/grafana/scenes/assets/5699976/76ba9bfb-c0e7-4068-9394-bff9a3bafd14"></video>|<video src="https://github.com/grafana/scenes/assets/5699976/f5654061-c535-439c-a960-32732a3c140d"></video>

<details>
  <summary>📦 Published PR as canary version: <code>5.2.1--canary.801.9610977729.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.2.1--canary.801.9610977729.0
  npm install @grafana/scenes@5.2.1--canary.801.9610977729.0
  # or 
  yarn add @grafana/scenes-react@5.2.1--canary.801.9610977729.0
  yarn add @grafana/scenes@5.2.1--canary.801.9610977729.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanel.test.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.test.tsx
@@ -254,6 +254,18 @@ describe('VizPanel', () => {
       expect(panel.state.options.option2).toBe('updated option');
     });
 
+    test('should update partial options values to undefined value', () => {
+      panel.onOptionsChange({
+        option2: 'updated option',
+      });
+      panel.onOptionsChange({
+        option2: undefined,
+      });
+
+      expect(panel.state.options.showThresholds).toBe(true);
+      expect(panel.state.options.option2).toBe(undefined);
+    });
+
     test('should always allow overriding array values', () => {
       panel.onOptionsChange({ sortBy: ['asc'] });
       expect(panel.state.options.sortBy).toEqual(['asc']);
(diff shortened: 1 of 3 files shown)


torkelo:
I thought we set replace=true when changing options? We do that when we change field options

ivanortegaalba:
> I thought we set replace=true when changing options? We do that when we change field options

If I'm not wrong, we're not 
https://github.com/grafana/grafana/blob/main/public/app/features/dashboard/components/PanelEditor/getVisualizationOptions.tsx#L223

Anyways, the merge for options, even if it doesn't replace, should be considered undefined as a valid value. Otherwise, we will be always forced to replace if we want to remove the value of a property from an object.

axelavargas (approved):
Great work, @ivanortegaalba 🥇 , I did a local test and indeed this fixes the issue. I think the approach makes sense, as you said, we don't want to replace the whole object, but partial options.  

Approving the PR, however, it seems a file was added to the PR by error?

`filename.txt`

ivanortegaalba:
> Great work, @ivanortegaalba 🥇 , I did a local test and indeed this fixes the issue. I think the approach makes sense, as you said, we don't want to replace the whole object, but partial options.
> 
> Approving the PR, however, it seems a file was added to the PR by error?
> 
> `filename.txt`

Ups! OMG, sorry! 🤦 Removed!

PR #802, opened 2024-06-21
Dependencies: Bump grafana packages to v11

Sergej-Vlasov:
Bump grafana packages to v11

<details>
  <summary>📦 Published PR as canary version: <code>5.3.3--canary.802.9757517518.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.3.3--canary.802.9757517518.0
  npm install @grafana/scenes@5.3.3--canary.802.9757517518.0
  # or 
  yarn add @grafana/scenes-react@5.3.3--canary.802.9757517518.0
  yarn add @grafana/scenes@5.3.3--canary.802.9757517518.0
  ```
</details>


Sergej-Vlasov:
Can be tested in this [draft PR on main](https://github.com/grafana/grafana/pull/89600)

packages/scenes/package.json
```diff
@@ -45,10 +45,10 @@
     "uuid": "^9.0.0"
   },
   "peerDependencies": {
-    "@grafana/data": "^10.4.1",
-    "@grafana/runtime": "^10.4.1",
-    "@grafana/schema": "^10.4.1",
-    "@grafana/ui": "^10.4.1",
+    "@grafana/data": "^11.0.0",
```
jackw, line 48: Does this mean scenes is incompatible with 10.4.x? If not I would suggest using wider semver ranges than tilde and major bumping otherwise plugin devs will need to update all their Grafana dependencies too.

`>=10.4` or `>=10.4 <12` could be options. The latter if you'd like to make sure devs are pre-warned before the next major arrives.

Sergej-Vlasov: I believe it will force update to v11, so changing to `>=10.4` to be less aggressive. Good spot!

mdvictor (approved):
LGTM

ashharrison90 (commented):
are we able to bump to 11.1.0? 🤔 it's the latest minor, and will allow us to unskip [this test](https://github.com/grafana/scenes/blob/main/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.test.tsx#L120) which would have caught a regression 😭

Sergej-Vlasov:
> are we able to bump to 11.1.0? 🤔 it's the latest minor, and will allow us to unskip [this test](https://github.com/grafana/scenes/blob/main/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.test.tsx#L120) which would have caught a regression 😭

I think we should do that soon in a major release version so that we do not suddenly break projects that rely on `v10.4`

ashharrison90:
> > are we able to bump to 11.1.0? 🤔 it's the latest minor, and will allow us to unskip [this test](https://github.com/grafana/scenes/blob/main/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.test.tsx#L120) which would have caught a regression 😭
> 
> I think we should do that soon in a major release version so that we do not suddenly break projects that rely on `v10.4`

i more meant could we change the version in `scenes-app` from `^11.0.0` to `^11.1.0`. cause i think right now, when you `yarn install` it's getting that version and using it in the tests in `scenes`. or can we add a separate `devDependency` for `@grafana/ui`? 🤔

Sergej-Vlasov:
> > > are we able to bump to 11.1.0? 🤔 it's the latest minor, and will allow us to unskip [this test](https://github.com/grafana/scenes/blob/main/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.test.tsx#L120) which would have caught a regression 😭
> > 
> > 
> > I think we should do that soon in a major release version so that we do not suddenly break projects that rely on `v10.4`
> 
> i more meant could we change the version in `scenes-app` from `^11.0.0` to `^11.1.0`. cause i think right now, when you `yarn install` it's getting that version and using it in the tests in `scenes`. or can we add a separate `devDependency` for `@grafana/ui`? 🤔

@ashharrison90 I tried to update `scenes-app` and `scenes-react` to `v11.1.0` and it produced quite a few typescript and jest errors in dev environment. I am currently short on time to fix that so I could merge this update to `v11.0.0` and we could do update to `v11.1.0` in a separate PR that will require more adjustments. Does this work for you?

ashharrison90:
@Sergej-Vlasov sure 👍

PR #810, opened 2024-07-02
Fix variable interpolation in query options interval

mdvictor:
This PR adds interpolation to the query options interval in the SceneQueryRunner.


axelavargas (approved):
🥳 the code looks good to me! Great job @mdvictor 🙌🏾

PR #811, opened 2024-07-02
SceneRefreshPicker: Pass up to date state into setupIntervalTimer during constructor

Sergej-Vlasov:
setupIntervalTimer was getting `refresh: ""` and not initialising auto-refresh. Passing initial state manually. 

<details>
  <summary>📦 Published PR as canary version: <code>5.3.5--canary.811.9767690637.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.3.5--canary.811.9767690637.0
  npm install @grafana/scenes@5.3.5--canary.811.9767690637.0
  # or 
  yarn add @grafana/scenes-react@5.3.5--canary.811.9767690637.0
  yarn add @grafana/scenes@5.3.5--canary.811.9767690637.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/SceneRefreshPicker.tsx
+++ b/packages/scenes/src/components/SceneRefreshPicker.tsx
@@ -31,19 +31,20 @@ export class SceneRefreshPicker extends SceneObjectBase<SceneRefreshPickerState>
   private _autoTimeRangeListener: Unsubscribable | undefined;
 
   public constructor(state: Partial<SceneRefreshPickerState>) {
-    const initialState = {
+    super({
       refresh: '',
       ...state,
       autoValue: undefined,
       autoEnabled: state.autoEnabled ?? true,
       autoMinInterval: state.autoMinInterval ?? config.minRefreshInterval,
       intervals: state.intervals ?? DEFAULT_INTERVALS,
-    };
-
-    super(initialState);
+    });
 
     this.addActivationHandler(() => {
-      this.setupIntervalTimer(initialState);
+      this.setupIntervalTimer({
+        refresh: state.refresh ?? '',
+        intervals: state.intervals ?? DEFAULT_INTERVALS,
+      });
 
       return () => {
         if (this._intervalTimer) {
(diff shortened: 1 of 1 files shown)


packages/scenes/src/components/SceneRefreshPicker.tsx
```diff
@@ -31,17 +31,19 @@ export class SceneRefreshPicker extends SceneObjectBase<SceneRefreshPickerState>
...
       autoMinInterval: state.autoMinInterval ?? config.minRefreshInterval,
       intervals: state.intervals ?? DEFAULT_INTERVALS,
-    });
+    };
+
+    super(initialState);
 
     this.addActivationHandler(() => {
-      this.setupIntervalTimer();
+      this.setupIntervalTimer(initialState);
```
torkelo, line 46: this is not up to date state, the state can change, this would freeze it when the scene object was created

Sergej-Vlasov: refactored a bit. the main issue is that when we call `setupIntervalTimer()` in `this.addActivationHandler` inside constructor, `this.state` inside `setupIntervalTimer` does not have the same values as state inside constructor. `refresh` value inside setupIntervalTimer is `""` while its defined to `"5s"` inside the constructor. Thats the whole idea of passing it manually to `setupIntervalTimer`. This solution does feel dirty so interested if there is a better quick fix here.

packages/scenes/src/components/SceneRefreshPicker.tsx
```diff
@@ -106,9 +109,11 @@ export class SceneRefreshPicker extends SceneObjectBase<SceneRefreshPickerState>
     return rangeUtil.calculateInterval(timeRange.state.value, resolution, this.state.autoMinInterval);
   };
 
-  private setupIntervalTimer = () => {
+  private setupIntervalTimer = (partialState?: Pick<SceneRefreshPickerState, 'refresh' | 'intervals'>) => {
     const timeRange = sceneGraph.getTimeRange(this);
-    const { refresh, intervals } = this.state;
+    // when calling setupIntervalTimer inside constructor this.state is out of sync with constructor state,
+    //  therefore passing partialState from constructor manually
+    const { refresh, intervals } = partialState || this.state;
```
ivanortegaalba, line 116: If `this.setupIntervalTimer` is set when activated, the `this.state` should be the one set when created, unless other components are modifying the state of this object. So we shouldn't need to pass a `partialState`.

We can check any state change using `.subscribeToState((n, p) => {`. I'd do this to understand why the initial state is not the same state that you get when activating

ivanortegaalba:
I think, with the new approach, we can close this one, right?

Sergej-Vlasov:
Closing in favour of https://github.com/grafana/scenes/pull/814

PR #815, opened 2024-07-03
Adhoc filters: Fix regression with displaying groups

ashharrison90:
- fixes the display of groups in adhoc filters
- we initially regressed this with https://github.com/grafana/scenes/pull/766
- this PR does a couple of things:
  - creates a separate `getAdhocOptionSearcher` instead of trying to reuse the generic `getOptionSearcher` which is typed for `VariableOptions`
    - ~will add some tests for this as well~ ✅ 
  - moves the `handleOptionGroups` logic to `AdHocFilterRenderer`
  - so the flow is something like:
    - opening the option menu calls `model._getValuesFor`
    - this in turn calls `ds.getTagValues`, which returns a `MetricFindValue[]` array containing an optional group parameter
    - `_getValuesFor` then maps this to `SelectableValue[]` but keeps the group parameter
    - this is set in state as `values` in `AdHocFilterRenderer`
    - `AdHocFilterRenderer` then uses the new option searcher to filter these options down, and calls `handleOptionGroups` to group them all together as before
- you can test this locally before/after by adjusting the adhoc filters demo and providing a custom `getTagValuesProvider`:
```
                  getTagValuesProvider: async () => ({
                    replace: true,
                    values: [{
                      text: 'Alice',
                      value: 'alice',
                      group: 'People',
                    }, {
                      text: 'Bar',
                      value: 'bar'
                    }, {
                      text: 'Cat',
                      value: 'cat',
                      group: 'Animals',
                    }, {
                      text: 'Bob',
                      value: 'bob',
                      group: 'People',
                    }, {
                      text: 'Dog',
                      value: 'dog',
                      group: 'Animals',
                    }, {
                      text: 'Foo',
                      value: 'foo',
                    }]
                  })
```
image

- note: there is a minor styling issue where we're not showing boundaries between groups for virtualized `Select`s. this needs fixing separately in core.
  - you can see it in the [canary storybook](https://developers.grafana.com/ui/canary/index.html?path=/story/forms-select--multi-select-with-option-groups-virtualized) as well
- note 2: ~working on a test to prevent future regressions now 🤓~ there is already a test for this in `AdhocFiltersVariable.test.ts`, but it's currently skipped and can't be enabled until scenes moves to `@grafana/ui@11.1.0`

<details>
  <summary>📦 Published PR as canary version: <code>5.3.6--canary.815.9779024969.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.3.6--canary.815.9779024969.0
  npm install @grafana/scenes@5.3.6--canary.815.9779024969.0
  # or 
  yarn add @grafana/scenes-react@5.3.6--canary.815.9779024969.0
  yarn add @grafana/scenes@5.3.6--canary.815.9779024969.0
  ```
</details>

### DIFF
--- a/packages/scenes/jest.config.ts
+++ b/packages/scenes/jest.config.ts
@@ -4,6 +4,7 @@ module.exports = {
   moduleNameMapper: {
     // D3 exposes ESModules. To avoid issues with Jest, we need to point them to the CJS version.
     d3: '<rootDir>/../../node_modules/d3/dist/d3.min.js',
+    '\\.css$': '<rootDir>/utils/test/__mocks__/style.ts',
   },
   testEnvironment: 'jsdom',
   setupFilesAfterEnv: ['./utils/setupTests.ts'],
--- a/packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
@@ -5,19 +5,36 @@ import { GrafanaTheme2, SelectableValue } from '@grafana/data';
 import { Button, Field, InputActionMeta, Select, useStyles2 } from '@grafana/ui';
 import { css, cx } from '@emotion/css';
 import { ControlsLabel } from '../../utils/ControlsLabel';
-import { getOptionSearcher } from '../components/getOptionSearcher';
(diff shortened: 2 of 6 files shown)


dprokop (approved):
Works OK now!

PR #816, opened 2024-07-04
Adhoc filters/Group by: Support groups

ashharrison90:
- supports displaying groups in adhoc filter keys 
- supports displaying groups in groupby
- adds (skipped) unit tests for both
  - can't enable the tests until the repo moves to use `@grafana/ui@11.1.0` 

<details>
  <summary>📦 Published PR as canary version: <code>5.4.0--canary.816.9794480649.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.4.0--canary.816.9794480649.0
  npm install @grafana/scenes@5.4.0--canary.816.9794480649.0
  # or 
  yarn add @grafana/scenes-react@5.4.0--canary.816.9794480649.0
  yarn add @grafana/scenes@5.4.0--canary.816.9794480649.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/types.ts
+++ b/packages/scenes/src/variables/types.ts
@@ -70,6 +70,7 @@ export interface ValidateAndUpdateResult {}
 export interface VariableValueOption {
   label: string;
   value: VariableValueSingle;
+  group?: string;
 }
 
 export interface SceneVariableSetState extends SceneObjectState {
--- a/packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
@@ -6,35 +6,11 @@ import { Button, Field, InputActionMeta, Select, useStyles2 } from '@grafana/ui'
 import { css, cx } from '@emotion/css';
 import { ControlsLabel } from '../../utils/ControlsLabel';
 import { getAdhocOptionSearcher } from './getAdhocOptionSearcher';
+import { handleOptionGroups } from '../utils';
 
 interface Props {
   filter: AdHocFilterWithLabels;
   model: AdHocFiltersVariable;
-}
-
-// Collect a flat list of SelectableValues with a `group` property into a hierarchical list with groups
-function handleOptionGroups(values: SelectableValue[]): Array<SelectableValue<string>> {
-  const result: Array<SelectableValue<string>> = [];
-  const groupedResults = new Map<string, Array<SelectableValue<string>>>();
-
-  for (const value of values) {
-    const groupLabel = value.group;
-    if (groupLabel) {
-      let group = groupedResults.get(groupLabel);
-
-      if (!group) {
-        group = [];
-        groupedResults.set(groupLabel, group);
-        result.push({ label: groupLabel, options: group });
-      }
-
-      group.push(value);
-    } else {
-      result.push(value);
-    }
-  }
-
-  return result;
 }
 
 function keyLabelToOption(key: string, label?: string): SelectableValue | null {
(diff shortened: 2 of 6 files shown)


packages/scenes/src/variables/types.ts
```diff
@@ -70,6 +70,7 @@ export interface ValidateAndUpdateResult {}
 export interface VariableValueOption {
   label: string;
   value: VariableValueSingle;
+  group?: string;
```
torkelo, line 73: group??

torkelo:
some doc somewhere that describes the problem here?

packages/scenes/src/variables/types.ts
ashharrison90: yeah so, the `MetricFindValue` interface has an optional `group` property that that allows you to group/categorise options together. 

as an example, your datasource might return something like:
```
const metricFindValues: MetricFindValue[] = [{
  text: 'Alice',
  value: 'alice',
  group: 'People',
}, {
  text: 'Bar',
  value: 'bar'
}, {
  text: 'Cat',
  value: 'cat',
  group: 'Animals',
}, {
  text: 'Bob',
  value: 'bob',
  group: 'People',
}, {
  text: 'Dog',
  value: 'dog',
  group: 'Animals',
}, {
  text: 'Foo',
  value: 'foo',
}]
```

and in that case you want to group the `People` and `Animals` together in their respective categories.

with groupby, the problem is we map to a `VariableValueOption` and lose the `group` information. so this extends the `VariableValueOption` interface with the optional group parameter, which is then passed to our `Select` and renders the groups correctly. but maybe we should just not be using the `VariableValueOption` for groupby... wdyt? 🤔

dprokop:
@torkelo  - this is the original PR that introduced possibility to provide grouped values for `getTagKeys` / `getTagValues` responses.

packages/scenes/src/variables/types.ts
dprokop: @ashharrison90 - i think we may use another type, but this would limit us in the future, i.e. if we wanted to use it in other variable types or if we wanted the group names to be searchable (today they are not)

ashharrison90: @dprokop yeah i mean i'm fine with using this type as long as it makes sense 👍

torkelo: So another option to consider is metricfindvalue that is an object (with many properties/attributes) and a way to say one of those properties should be used a a group / category. 

One big limitation right now with a variable value is that it only can contain a single value and text representation, and not many values.

dprokop: > So another option to consider is metricfindvalue that is an object (with many properties/attributes) and a way to say one of those properties should be used a a group / category.

@torkelo  - i'm not sure I understand a benefit. `MetricFindValue` is already an object, fairly well structured for most if not all the needs that we have seen so far. I guess what you suggest to consider is the following shape of `MetricFindValue`:

```ts
interface MetricFindValue {
    text: string;
    value?: string | number;
    groupByKey: string;
    [key: string]: any;
}
```


I think i see the point, but also this increases slightly the Grafana responsibility for transforming the results into groups. I don't think this added complexity is bringing that much of a value, and also loosens the interface quite a bit.

ashharrison90: i guess another option would be to have a completely separate bit of state in `GroupByVariable` that maintains a map of values to groups. then when we convert them to `SelectableValue`s, we can reapply the group information using the map? using the example above, we'd have something like:
```tsx
this.state.groupMap = {
  alice: 'People',
  cat: 'Animals',
  bob: 'People',
  dog: 'Animals'
}
```

then we'd have a tiny bit of code in `toSelectableValue` like:
```tsx
  if (this.state.groupMap[value]) {
    result.group = this.state.groupMap[value];
  }
```

advantage is we don't have to modify the `VariableValueOption` type. disadvantage is it's a bit messier. wdyt @dprokop @torkelo ? 🤔

dprokop: I'm actually in favor of the original approach, which has minimal footprint and sets us well for supporting grouped options in variables other than group by and filters.

dprokop (approved):

PR #818, opened 2024-07-05
Add query variable in react-scenes

mdvictor:
Adds the Query Variable type to react-scenes

Depends on https://github.com/grafana/scenes/pull/827


packages/scenes-app/src/react-demo/DynamicVariablesPage.tsx
```diff
@@ -18,6 +24,48 @@ export function DynamicVariablesPage() {
...
+        hide={hide}
+        refresh={VariableRefresh.onTimeRangeChanged}
+        includeAll={includeAll}
+        isMulti={isMulti}
+        sort={sort}
+      >
+        <Stack direction="column">
+          <Stack>
+            <VariableControl name="job2" />
+            <Button variant="secondary" onClick={() => setHide(VariableHide.hideLabel)}>
```
mdvictor, line 42: Added a lot of buttons here to toy around with state here but will ultimately remove most of them for the demo.

packages/scenes-react/src/variables/QueryVariable.tsx
```diff
@@ -0,0 +1,81 @@
...
+      regex, 
+      value: initialValue, 
+      isMulti,
+      hide, 
+      includeAll,
+     });
+  }
+
+  useEffect(() => {
+    if (variableAdded) {
```
mdvictor, line 54: @torkelo I'm not sure how to best tackle this. We need a way to refresh the variable when state changes, especially around value changes e.g: sort, regex which should trigger `getValueOptions`. I didn't find a way to easily do this other than re-adding the variable and letting it run it's activation, but that might not be the best approach. Simply setting the variable state won't work. Any ideas?

mdvictor:
Also, maybe we should add a prefix to these variables to delimit them from their scenes counterparts.

torkelo:
> Also, maybe we should add a prefix to these variables to delimit them from their scenes counterparts.

@mdvictor it's own of the reasons we have separate package to allow some name conflicts, personally I really dislike the name conflicts, tried to get help with naming suggestions (around renaming or prefixing / suffixing) but did not get much feedback on options.

packages/scenes-react/src/variables/QueryVariable.tsx
torkelo: @mdvictor yea, we might need some updates to QueryVariable to be able to re-evaluate options when sort / regex changes or re-issue query when query changes

packages/scenes-react/src/variables/QueryVariable.tsx
```diff
@@ -0,0 +1,81 @@
...
+        datasource, 
+        refresh, 
+        sort, 
+        regex, 
+        value: initialValue, 
+        isMulti,
+        hide, 
+        includeAll,
+      })
+    }
```
torkelo, line 67: better to handle this state update in a separate useEffect that has some deep quality checks as this will not always trigger remove and re-add whenever this re-renders unless parent is very careful to memorize all props.

see useQueryRunner for reference

torkelo (commented):
Looks good!

packages/scenes-react/src/variables/QueryVariable.tsx
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  useEffect(() => {
+    const removeFn = scene.addVariable(variable);
+    setVariableAdded(true);
+    return removeFn;
+  }, [variable, scene, name]);
+
+  useEffect(() => {
+    variable?.setState({
```
torkelo, line 60: need a check here so this is not run on first mount

packages/scenes/src/variables/sets/SceneVariableSet.ts
```diff
@@ -255,6 +260,12 @@ export class SceneVariableSet extends SceneObjectBase<SceneVariableSetState> imp
     writeVariableTraceLog(variable, 'updateAndValidate error', err);
   }
 
+  private _handleVariableUpdate(variable: SceneVariable) {
+    this._variablesToUpdate.add(variable);
+
+    this._updateNextBatch();
+  }
```
torkelo, line 267: I think it's better to do this in a separate PR and not sure we can do it like this, not all state updates should cause a full update like this (would likely cause infinite update loops)

packages/scenes/src/variables/variants/query/QueryVariable.tsx
```diff
@@ -150,6 +158,12 @@ export class QueryVariable extends MultiValueVariable<QueryVariableState> {
     this.setState({ options: result, loading: false });
   }, 400);
 
+  private _updateOptionsOnStateChange = (newState: QueryVariableState, oldState: QueryVariableState) => {
+    if (!isEqual(pick(newState, this._stateChangePropsToCompare), pick(oldState, this._stateChangePropsToCompare))) {
+      this.publishEvent(new SceneVariableStateChangeEvent(this), true);
+    }
+  };
```
torkelo, line 165: I think it's better to give consumer manual control over which state updates should trigger a re-eval of options

packages/scenes/src/variables/sets/SceneVariableSet.ts
mdvictor: Sounds good, I'll move this to another PR. Yeah, it's risky.  I think it worked because of [this](https://github.com/grafana/scenes/blob/a6c1ab6b4fc731627740a86263a47eaa2e497bcd/packages/scenes/src/variables/sets/SceneVariableSet.ts#L197) but I might be missing something. In any case, I changed this to be called manually.

packages/scenes/src/variables/variants/MultiValueVariable.ts
```diff
@@ -329,6 +330,10 @@ export abstract class MultiValueVariable<TState extends MultiValueVariableState
     return options;
   }
 
+  public refreshOptions() {
```
mdvictor, line 333: Added this func to update options after the state has changed.

packages/scenes/src/variables/variants/query/QueryVariable.tsx
mdvictor: I agree. Added a method in MultiValueVariable as it will need to be called in other variables as well (custom/datasource). Will make a new PR with these changes.

packages/scenes-react/src/variables/QueryVariable.tsx
```diff
@@ -0,0 +1,86 @@
...
+      query,
+      datasource, 
+      refresh, 
+      sort, 
+      regex, 
+      hide, 
+      includeAll,
+    });
+
+    variable.refreshOptions();
```
torkelo, line 75: because query and data source are objects, I would do a deep equals diff here to make sure thing have actually changed

mdvictor:
@torkelo added explicit comparison to check changes both here and in the datasource variable [PR](https://github.com/grafana/scenes/pull/820)

torkelo (approved):

PR #820, opened 2024-07-09
ScenesReact: Add datasource variable

mdvictor:
Adds datasource variable in scenes react.

Depends on https://github.com/grafana/scenes/pull/827


packages/scenes-react/src/variables/DataSourceVariable.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+    variable?.setState({
+      pluginId,
+      regex,
+      label,
+      hide, 
+      includeAll,
+    })
+
+    variable.refreshOptions();
```
torkelo, line 65: wont this be called once on first mount? (after variable is added), this should only happen when something changed after the initial mount/render, but looks like this will be called just after the variable is added (even when all the props are the same as they where when the variable was added)

mdvictor: @torkelo You are right. I've changed this now

packages/scenes-react/src/variables/DataSourceVariable.tsx
```diff
@@ -0,0 +1,75 @@
...
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [hide, includeAll, label, pluginId, regex, variable])
```
torkelo, line 67: instead of using eslint-disable-next-line react-hooks/exhaustive-deps and not including variableAdded, another alternative is just checking if pluginId, regex, label, hide etc has changed , not sure which I prefer but feels a bit more explicit / clear that way

torkelo (approved):

PR #821, opened 2024-07-09
ScenesReact: Add useVariableValue for single values

mdvictor:
Add useVariableValue hook for single value use


packages/scenes-react/src/hooks/useVariableValue.ts
```diff
@@ -0,0 +1,23 @@
+import { MultiValueVariable, SceneVariables, VariableValueSingle, sceneGraph } from '@grafana/scenes';
+import { useSceneContext } from './hooks';
+
+export function useVariableValue(name: string): [VariableValueSingle | undefined, boolean] {
```
torkelo, line 4: can you try making this generic? with a generic param like `<T = VariableValueSingle>`

torkelo (approved):

PR #822, opened 2024-07-09
ScenesReact: Add useQueryVariable hook

mdvictor:
Adds a `useQueryVariable` hook that allows a user to define a QueryVariable and manipulate it.


packages/scenes-react/src/hooks/useVariableQuery.ts
```diff
@@ -0,0 +1,26 @@
+import { MultiValueVariable, VariableValueOption, sceneGraph } from '@grafana/scenes';
+import { useSceneContext } from './hooks';
+import { useEffect, useState } from 'react';
+import { lastValueFrom } from 'rxjs';
+
+export function useVariableQuery(name: string): VariableValueOption[] {
```
torkelo, line 6: not really sure what the purpose of this is when we have useVariableValue 

I was thinking this would be an alternative way of both defining (adding/updating) a QueryVariable and getting the options. for scenarios where you don't really care about the variable as a variable but only the options (query results).

But it's more of a "convenience" thing than something really needed. Say you want to repeat some layout by the result of a query variable, but you don't really use the that query variable in any expression you only just care about the results of that "query".  I was thinking of this scenario

mdvictor:
I renamed this to `useQueryVariable` as it seems more intune to the specifications - a hook that will create a queryVar and can be used to pull options for various scenarios.

packages/scenes-app/src/react-demo/RepeatByVariablePage.tsx
```diff
@@ -23,6 +23,11 @@ export function RepeatByVariablePage() {
 function RepeatPanelByVariable() {
   const [values, loading] = useVariableValues<string>('panels');
 
+  //move this into a different demo
+  const queryVar = useQueryVariable({ name: 'test', datasource: 'gdev-testdata', query: '*' });
+  const state = queryVar?.useState();
+  console.log(state?.options);
```
mdvictor, line 29: need to create a demo for this

packages/scenes-react/src/hooks/useVariableQuery.ts
mdvictor: I've modified things a bit but was wondering if we really need/want to also define a QueryVariable and attach it to the scene context, or if it would be better to just have a hook that accepts a datasource ref and query, build a QueryVariable, fetch the options and return them, without polluting the scene context with a var that might never be used.

packages/scenes-react/src/hooks/useQueryVariable.ts
```diff
@@ -0,0 +1,45 @@
...
+
+interface QueryVariableOptions {
+  name: string;
+  datasource: string;
+  query: string | SceneDataQuery;
+  regex?: string;
+}
+
+// adds or updates a query variable in the scene and returns it or null if
+// the variable is not a QueryVariable
```
torkelo, line 14: I would replace this with a prop jsdoc comment and one that is accurate :)

PR #823, opened 2024-07-09
LazyLoader: Hide wrapper if child is hidden 

svennergr:
When using a `SceneCSSGridLayout`, the `isLazy` property and items setting their `isHidden` property dynamically, the `LazyLoader` would still be visible.

This PR just subscribes to the item's state and also hides the `LazyLoader`.

<details>
  <summary>📦 Published PR as canary version: <code>5.3.8--canary.823.9872224365.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.3.8--canary.823.9872224365.0
  npm install @grafana/scenes@5.3.8--canary.823.9872224365.0
  # or 
  yarn add @grafana/scenes-react@5.3.8--canary.823.9872224365.0
  yarn add @grafana/scenes@5.3.8--canary.823.9872224365.0
  ```
</details>

### DIFF
--- a/packages/scenes-app/src/demos/cssGridLayoutDemo.tsx
+++ b/packages/scenes-app/src/demos/cssGridLayoutDemo.tsx
@@ -5,6 +5,7 @@ import {
   PanelBuilders,
   SceneAppPage,
   SceneAppPageState,
+  SceneCSSGridItem,
   SceneCSSGridLayout,
   SceneComponentProps,
   SceneObjectBase,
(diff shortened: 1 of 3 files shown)


torkelo (commented):
not sure this is the way to solve it, it duplicates the logic found in the item renderer in a kind of messy way (and adds subscriptions to objects that might not even be SceneCSSGridItem) 

https://github.com/grafana/scenes/blob/main/packages/scenes/src/components/layout/CSSGrid/SceneCSSGridLayout.tsx#L126

not 100% sure what the best way to solve it is though 🤔

svennergr:
@torkelo I pushed another approach of just hiding empty divs with css when loaded. PTAL.

dprokop:
@svennergr  - I think the css only approach is better compared to the first approach. Just gonna ask @kaydelaney for look on this too.

I've tested with the lazy demos that we have for grid layout and I have not observed any regression. Gonna have a look in core grafana dashboards as soon as the canary is released.

PR #825, opened 2024-07-10
LazyLoader: add use of `useStyles2` and improve name

svennergr:
Makes use of the memoized `useStyles2` hook and improves names of the used CSS class.

### DIFF
--- a/packages/scenes/src/components/layout/LazyLoader.tsx
+++ b/packages/scenes/src/components/layout/LazyLoader.tsx
@@ -3,6 +3,7 @@ import { useEffectOnce } from 'react-use';
 
 import { uniqueId } from 'lodash';
 import { css } from '@emotion/css';
+import { useStyles2 } from '@grafana/ui';
 
 export function useUniqueId(): string {
   const idRefLazy = useRef<string | undefined>(undefined);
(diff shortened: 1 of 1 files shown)


packages/scenes/src/components/layout/LazyLoader.tsx
```diff
@@ -59,20 +60,28 @@ export const LazyLoader: LazyLoaderType = React.forwardRef<HTMLDivElement, Props
       };
     });
 
+    // If the element was loaded, we add the `hideEmpty` class to potentially
+    // hide the LazyLoader if it does not have any children. This is the case
+    // when children have the `isHidden` property set. 
+    // We always include the `className` class, as this is coming from the 
+    // caller of the `LazyLoader` component.
+    const classes = `${loaded ? hideEmpty : ''} ${className}`;
```
torkelo, line 68: just curius why the loaded check is needed can't we always have the hideEmpty class?

svennergr: The `LazyLoader` is empty if it is not in view, as the `children` only loads if `loaded` is true:
https://github.com/grafana/scenes/pull/825/files#diff-b9c3d730e8841e5699e757b9a2c84d1b10f6e609af246185a0936d8dfeaa1a81R71

And then, obviously, if the `LazyLoader` has `display: none` the `IntersectionObserver` would never trigger.

torkelo (approved):

PR #827, opened 2024-07-11
Possibility to refresh variable options based on state changes

mdvictor:
Add possibility to manually trigger a variable options refresh after a state change. Needed for various variable components in react scenes. E.g.: https://github.com/grafana/scenes/pull/820

### DIFF
--- a/packages/scenes/src/variables/sets/SceneVariableSet.ts
+++ b/packages/scenes/src/variables/sets/SceneVariableSet.ts
@@ -8,6 +8,7 @@ import { writeSceneLog } from '../../utils/writeSceneLog';
 import {
   SceneVariable,
   SceneVariableDependencyConfigLike,
+  SceneVariableOptionsRefreshEvent,
   SceneVariables,
   SceneVariableSetState,
   SceneVariableValueChangedEvent,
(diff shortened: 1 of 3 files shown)


packages/scenes/src/variables/sets/SceneVariableSet.ts
```diff
@@ -255,6 +261,12 @@ export class SceneVariableSet extends SceneObjectBase<SceneVariableSetState> imp
     writeVariableTraceLog(variable, 'updateAndValidate error', err);
   }
 
+  private _handleVariableOptionsUpdate(variable: SceneVariable) {
+    this._variablesToUpdate.add(variable);
+
+    this._updateNextBatch();
+  }
```
torkelo, line 268: don't think this is needed, The variable can handle this and if the value changes due to options changes then it should emit the event for variable value change

torkelo: but if the current variable value is a single value then updating options does not change anything (no need to update other variables).

torkelo: @mdvictor I just don't understand why the set needs to handle this, this concern of updating options based on sorting etc is fully the responsibility of the variable. 

the set is only concerned if the value changes and if that should trigger cascading updates, and for that we have the SceneVariableValueChangedEvent event

mdvictor: ah, I understand what you mean now, yeah that makes sense, I'll move this for the variable itself to handle.

mdvictor: @torkelo I modified this to be handled by the variable itself, definitely better

torkelo (approved):

PR #824, opened 2024-07-10
Experiment with a proxy for solving JSON.stringify issue

dprokop:
Just experimenting with an approach to solve scopedVars `JSON.stringify` issue upstream. Publishing this only to get a canary release out to test in core.

fyi @kaydelaney @torkelo @joshhunt @ivanortegaalba 

<details>
  <summary>📦 Published PR as canary version: <code>5.3.9--canary.824.9873640037.0</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install @grafana/scenes-react@5.3.9--canary.824.9873640037.0
  npm install @grafana/scenes@5.3.9--canary.824.9873640037.0
  # or 
  yarn add @grafana/scenes-react@5.3.9--canary.824.9873640037.0
  yarn add @grafana/scenes@5.3.9--canary.824.9873640037.0
  ```
</details>

PR #828, opened 2024-07-11
Wrap __sceneObject scoped var in an serialisable wrapper

dprokop:
Alt to https://github.com/grafana/scenes/pull/824

This is a simpler approach that wraps `scopedVars.__sceneObject` in a serializable wrapper. This seems to be working just fine, can be tested against https://github.com/grafana/grafana/pull/90272

<details>
  <summary>📦 Published PR as canary version: <code>5.5.0--canary.828.9973016839.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.5.0--canary.828.9973016839.0
  npm install @grafana/scenes@5.5.0--canary.828.9973016839.0
  # or 
  yarn add @grafana/scenes-react@5.5.0--canary.828.9973016839.0
  yarn add @grafana/scenes@5.5.0--canary.828.9973016839.0
  ```
</details>


packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -108,7 +109,7 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
   private _containerWidth?: number;
   private _variableValueRecorder = new VariableValueRecorder();
   private _results = new ReplaySubject<SceneDataProviderResult>(1);
-  private _scopedVars = { __sceneObject: { value: this, text: '__sceneObject' } };
+  private _scopedVars = { __sceneObject: { value: new SafeSerializableSceneObject(this), text: '__sceneObject' } };
```
torkelo, line 112:
```suggestion
  private _scopedVars = { __sceneObject: new SafeSerializableSceneObject(this) };
```

@dprokop I have an idea that could make this a tiny bit cleaner and reduce some code duplication. We just have to make  SafeSerializableSceneObject have two public properties (value & text), value could just be an empty string, and text would be "__sceneObject". 

Then it becomes a bit cleaner to use and we do not have to duplicate ` { __sceneObject: { value: new SafeSerializableSceneObject(this), text: '__sceneObject' } };` everywhere.

torkelo (commented):
I think this looks good!

packages/scenes/src/utils/SafeSerializableSceneObject.ts
```diff
@@ -0,0 +1,17 @@
+import { SceneObject } from '../core/types';
+
+export class SafeSerializableSceneObject {
```
ivanortegaalba, line 3: @dprokop can we have this `toString` directly declared in [SceneObject](https://github.com/grafana/scenes/blob/a62cb4f43956045b175379d59cf0972566f098bd/packages/scenes/src/core/types.ts#L57)?

We could remove the circular reference to the parent, and then stringify normally. 

We could do this through `toJSON`:
```
toJSON() {
        const { _parent, ...rest } = this;
        return rest;
    }
```
Nowadays, the tests break when a test case breaks, because it tries to strigify the SceneObject too. So we could avoid stringifying on a scene object including the parent.

dprokop: > We could remove the circular reference to the parent, and then stringify normally.

the circular reference is not only about parent. It can be also other parts of the state

packages/scenes/src/querying/SceneQueryRunner.ts
dprokop: `value` cannot be an empty string- in core we actually need an access to this object itself for interpolation :) But i've simplified the code a bit, value now returns identity, and there's `text` available as well so indeed less duplication overall.

kaydelaney (approved):

ivanortegaalba (approved):
I hardly tested against:
✅ Datalinks scoped variables
✅ Repeated panels variables

Captura de pantalla 2024-07-19 a las 13 16 07

Evertything works as expected! Great work Dominik!

gtk-grafana:
@dprokop This PR breaks Explore logs.

```
SceneQueryRunner.ts:467 PanelQueryRunner Error TypeError: Cannot read properties of undefined (reading '$variables')
    at lookupVariable (lookupVariable.js:2:1)
    at lookupFormatVariable (sceneInterpolator.js:38:34)
    at sceneInterpolator.js:17:1
    at String.replace (<anonymous>)
    at sceneInterpolator (sceneInterpolator.js:14:1)
    at Object.interpolate (sceneGraph.js:30:27)
    at TemplateSrv.replace (template_srv.ts:252:25)
    at DatasourceSrv.get (datasource_srv.ts:134:34)
    at eval (getDataSource.ts:21:35)
    at Generator.next (<anonymous>)
```

gtk-grafana:
Looks like the error is thrown against grafana 11.1, but seems to work fine with latest main

PR #830, opened 2024-07-12
AdHocFiltersVariable - new ad hoc filters UI 

Sergej-Vlasov:
This is a POC work on the new UI (combobox style) for the AdHoc Filters. 

`package.json` additions:
- `@floating-ui/react` - for combobox dropdown position/interaction (same version as in grafana) [npm](https://www.npmjs.com/package/@floating-ui/react)
- `@tanstack/react-virtual ` - for dropdown list virtualisation [npm](https://www.npmjs.com/package/@tanstack/react-virtual)

To try locally: follow the steps in the [Demo App README](https://github.com/grafana/scenes/blob/main/packages/scenes-app/README.md), enable Scenes Test App in plugins and navigate to http://localhost:3000/a/grafana-scenes-app/demos/adhoc-filters/new-filters

The designs for this can be found in this [figma link](https://www.figma.com/design/LtJFa2HK2n0IgXlUDoLxuP/Scenes---Ad-hoc-Filters?node-id=119-43391&t=iNcHsN2sMkdKU1OD-0)

<details>
  <summary>📦 Published PR as canary version: <code>5.11.1--canary.830.10631142943.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.11.1--canary.830.10631142943.0
  npm install @grafana/scenes@5.11.1--canary.830.10631142943.0
  # or 
  yarn add @grafana/scenes-react@5.11.1--canary.830.10631142943.0
  yarn add @grafana/scenes@5.11.1--canary.830.10631142943.0
  ```
</details>


packages/scenes-app/src/demos/adhocFiltersDemo.tsx
```diff
@@ -194,6 +195,57 @@ export function getAdhocFiltersDemo(defaults: SceneAppPageState) {
...
       }),
+      new SceneAppPage({
+        title: 'New Filters UI',
+        url: `${defaults.url}/new-filters`,
+        getScene: () => {
+          return new EmbeddedScene({
+            ...getEmbeddedSceneDefaults(),
+            $variables: new SceneVariableSet({
+              variables: [
+                new AdHocFiltersVariableV2({
```
dprokop, line 206: Instead of introducing new variable type (which makes us have to handle such variable in core and apps) let's introduce a new layout type for ad hoc variable. Similar to how we do it already with horizontal and vertical layout.

dprokop:
Wanted to try it out and was confused it doesn't work... Because my data source was non existent, the underlying error has not been shown. Then i have configured my ds and now it returns dimensions and values, but... there's no indication an async action is taking place.

@Ijin08 and @catherineymgui - we need some designs for these two scenarios, seems like those are not included in the Figma file.

---

@Sergej-Vlasov  - i've tried this out and it works surprisingly smooth as for the first POC. Things I think will make sense

1. **Backwards key should remove the preceding filter** - now it feels an awkward key combo to Shift+tab to get to the preceding filter.


2. **Flaky edit behavior** - after Shift+tab and pressing enter, and then pressing Tab again creates filter in a wip-ish state:
image

3. **X button on filter is inconsistent**. When clicked, it removes the entire filter, but when Shift+Tab and Enter, it removes the filter value only (see 2 above as well).
 
4. **In multi-filter scenario, I need to be precise to activate the next filter creation**. I need to click precisely on the input placegholder, otherwise the filters input will not get focus. I think clicking anywhere in inside the filters combobox should focus the underlying input. Of course apart from clicking on the individual filter pill.
image

5. Thing to consider (cc @Ijin08 & @catherineymgui) - should the individual filter pill become editable after a single or double click? My personal take is it should be "forced" to be edited by double clicking on it.

Sergej-Vlasov:
@catherineymgui @Ijin08 as we discussed the possible error cases to consider:
1. API error - something wrong with api , data or any other X factor such as network error
   - solution could be the red/yellow warning icon at the end of input with tooltip
2. Options fetch error - could not retrieve the options for the dropdown (this can be network/api related)
   - solution could probably be a dropdown option that says "Failed to fetch. Try again"

catherineymgui:
@Sergej-Vlasov I think your solutions for errors make sense. Let me know if I need to mock up anything. 

@dprokop I agree with most of your points except editing on clicks. I was thinking of editing on one click (because what happens otherwise, if someone clicks once)? What do you mean by ‘forcing’ editing?

Additional thoughts / feedback. Would love discussion:

- When someone clicks out of an incomplete label / value, the pill stays. Should the pill just be removed (because otherwise it is incomplete, but also no way to remove an incomplete pill)? This is what happens right now. 

https://github.com/user-attachments/assets/faf8e281-8e46-4b7c-a906-c9dde915610e

- Can we add a hover state to the pill (with the action/hover color theme)?
- On edit - the pill switches to text. I think technically this makes sense but I think because the value is truncated as well, it is a bit weird? Instead, what if in edit mode is a pill in an “active / selected” state? With ability to “type” in the pill to change. That way we can show the whole value still, without any additional truncation or jumping between lines 

Mocked up example of what I mean for hover and edit state: 

https://github.com/user-attachments/assets/7442ea1b-ad9c-4641-889d-b139d7cc56bd

- Should we add some sort of inference if there are no more label / value options? We can remove the "Filter by label values" text and don't change the cursor. And add a tooltip? I'm not sure if this is the best approach but was thinking an icon/tooltip like this:

https://github.com/user-attachments/assets/0f3fae0d-e906-47bd-83e9-8b65e36404c1

- When someone is in the process of creating label values and dropdown appears - I think the 'Filter by label values' text can disappear? This is what shows up right now:

Screenshot 2024-07-16 at 17 32 38

--- 

General question, what will be part of this POC? Here are some additional functionality we talked about but not sure if this is after POC:
- Typing to filter / select value
- Truncated pills if it is too long (we talked about it briefly in our meeting, but I know logs is asking for this because they may have very long filters)
- Custom values
- Collapse / expand rows

Sergej-Vlasov:
@dprokop @catherineymgui @Ijin08  pushed latest improvements. Here is what has been done:

@dprokop - addressed all of your feedback except for:
   - double click (I would agree with @catherineymgui that its more intuitive with single click)
   - error states (that's in todo list)
   
@catherineymgui - addressed the following:
   - > When someone clicks out of an incomplete label / value, the pill stays. Should the pill just be removed (because otherwise it is incomplete, but also no way to remove an incomplete pill)? This is what happens right now.
   - > Can we add a hover state to the pill (with the action/hover color theme)?
   
@catherineymgui we will discuss the following points during sync:
   - > On edit - the pill switches to text. I think technically this makes sense but I think because the value is truncated as well, it is a bit weird? Instead, what if in edit mode is a pill in an “active / selected” state? With ability to “type” in the pill to change. That way we can show the whole value still, without any additional truncation or jumping between lines
   - > Should we add some sort of inference if there are no more label / value options? We can remove the "Filter by label values" text and don't change the cursor. And add a tooltip? I'm not sure if this is the best approach but was thinking an icon/tooltip like this:
   - > When someone is in the process of creating label values and dropdown appears - I think the 'Filter by label values' text can disappear? This is what shows up right now:

@dprokop would be great to discuss where POC scope ends during our sync to have clarity on the @catherineymgui comment:
>General question, what will be part of this POC? Here are some additional functionality we talked about but not sure if this is after POC:

Typing to filter / select value
Truncated pills if it is too long (we talked about it briefly in our meeting, but I know logs is asking for this because they may have very long filters)
Custom values
Collapse / expand rows
   
### Current TO-DO list:
- [ ] make input span the whole width
- [ ] clean up dropdown and extract into separate component
- [ ] add loading/error/empty states to dropdown component
- [ ] add ability to commit custom value if we have no matches in dropdown
- [ ] virtualise dropdown component

adamquan:
This is going to be very helpful for folks at Palantir! They are very excited about this. Thank you, team!

torkelo (commented):
This is looking great. 

One problem I found is that the value dropdown cuts of values so it's very hard to see values longer than a few chars 

Screenshot 2024-08-15 at 10 10 57

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersComboboxRenderer.tsx
```diff
@@ -0,0 +1,65 @@
...
+}) {
+  const { filters, readOnly } = model.useState();
+  const styles = useStyles2(getStyles);
+  const focusOnInputRef = useRef<() => void>();
+
+  return (
+    <div
+      className={`${styles.comboboxWrapper} ${!readOnly ? styles.comboboxFocusOutline : ''}`}
+      onClick={() => {
+        focusOnInputRef.current?.();
```
Sergej-Vlasov, line 22: `focusOnInputRef.current?.();` is achieved via useImperativeHandle by passing ref through 2 forward refs into:
```typescript
// refs.domReference - comes from floating UI and points at input
useImperativeHandle(parentRef, () => () => refs.domReference.current?.focus(), [refs.domReference]);
```

to avoid `forwardRef` madness alternative would use something like this:
```typescript
  const divRef = useRef<HTMLDivElement>(null);
  const wipInputRef = useRef<HTMLInputElement | null>(null);

  useEffect(() => {
    if (divRef.current) {
      const inputEl = divRef.current.querySelector('input'); // or could be static id to avoid accidentally targeting other inputs in the tree
      if (inputEl) {
        wipInputRef.current = inputEl;
      }
    }
  }, []);
  ...
   onClick={() => {
        wipInputRef.current?.focus();  // instead of focusOnInputRef.current?.();
      }}
```

I do like alternative to `forwardRef` and `useImperativeHandle` a bit more, so looking for comments

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersCombobox.tsx
```diff
@@ -0,0 +1,506 @@
...
+    loop: true,
+  });
+
+  const { getReferenceProps, getFloatingProps, getItemProps } = useInteractions([role, dismiss, listNav]);
+
+  // pass ability to focus on input element back to parent
+  useImperativeHandle(parentRef, () => () => refs.domReference.current?.focus(), [refs.domReference]);
+
+  function onChange(event: React.ChangeEvent<HTMLInputElement>) {
+    // // part of POC for seamless filter parser
```
Sergej-Vlasov, line 112: uncomment below to test seamless filter parser (this feature definitely work in progress)

dprokop:
UI/UX feedback (mainly for @Ijin08 and @catherineymgui)
1. **Loading indicator** - i don't think the proposed solution works well, it's barely visible and static, gives an impression like nothing is going on actually:

image

I think the current filter's loading indicator (spinner) works a bit better, and is aligned with design system(https://developers.grafana.com/ui/latest/index.html?path=/story/forms-input--simple&args=loading:!true)

image

2. **Error indicator** 
Similar comment as with loading indicator, i don't think the proposed solution works well:

image

It's hard to read and is not in line with the design system: https://developers.grafana.com/ui/latest/index.html?path=/story/forms-input--with-field-validation&args=loading:!true

3. Same comment that @torkelo made, the dropdown cuts off values so it's very hard to see values longer than a few chars

image


Feature patiry / functionality:
1. Filteing dimensions and values: It filters by exact match, opposite to the current ad hoc filters that are more fuzzy.

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersComboboxRenderer.tsx
```diff
@@ -0,0 +1,65 @@
...
+      }}
+    >
+      <Icon name="filter" className={styles.filterIcon} size="lg" />
+
+      {filters.map((filter, index) => (
+        <AdHocFiltersComboboxEditSwitch key={index} filter={filter} model={model} readOnly={readOnly} />
```
dprokop, line 28: Nit: This name is a bit confusing, alt proposal: `AdHocFilterPill`

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersComboboxRenderer.tsx
dprokop: I agree the `useImperativeHandle` is simpler here. The drawback (for me personally) is that it's a bit harder to follow. I think it's ok to use it, but please leave a verbose comment about this :)

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersCombobox.tsx
```diff
@@ -0,0 +1,506 @@
...
+interface AdHocComboboxProps {
+  filter?: AdHocFilterWithLabels;
+  isAlwaysWip?: boolean;
+  model: AdHocFiltersVariable;
+  handleChangeViewMode?: () => void;
+}
+
+type AdHocInputType = 'key' | 'operator' | 'value';
+
+export const AdHocCombobox = forwardRef(function AdHocCombobox(
```
dprokop, line 33: I suggest refactoring this component a little bit into smaller chunks, it's quite a challenge to read through the code and understand it.

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersComboboxEditSwitch.tsx
```diff
@@ -0,0 +1,120 @@
...
+            handleChangeViewMode();
+          }
+        }}
+        role="button"
+        aria-label={`Edit filter with key ${filter.key}`}
+        tabIndex={0}
+        ref={pillWrapperRef}
+      >
+        <span>
+          {filter.key} {filter.operator} {filter.value}
```
dprokop, line 58: This is not entirely correct. The underlying structure contains key/value and keyLabel/valueLabel. The latter are human readable names while the former are machine readable. They often are going to be the same, but some data sources may implement the differentiation between human and machine readable names by providing different `text` and `value` properties of the `MetricFindValue` that represent filter dimensions and values. @ashharrison90 implemented that human readable rendering in https://github.com/grafana/scenes/pull/690 and we need to make sure this exists in this new UI as well :) 

To give you an example, given the available filter dimensions are: `[{text: 'Human readable dimension name', value: 'dimension-id'}, ...]` the UI filter pill should show "Human readable dimension name = ..."

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersCombobox.tsx
```diff
@@ -0,0 +1,506 @@
...
+                <OptionsErrorPlaceholder handleFetchOptions={() => handleFetchOptions(filterInputType)} />
+              ) : !filteredDropDownItems.length && filterInputType !== 'value' ? (
+                <NoOptionsPlaceholder />
+              ) : (
+                <>
+                  {rowVirtualizer.getVirtualItems().map((virtualItem) => {
+                    const item = filteredDropDownItems[virtualItem.index];
+                    const index = virtualItem.index;
+
+                    return (
```
dprokop, line 378: We need to support options groups introduced in https://github.com/grafana/scenes/pull/816

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersCombobox.tsx
```diff
@@ -0,0 +1,506 @@
...
+    // }
+
+    const value = event.target.value;
+    setInputValue(value);
+    setActiveIndex(0);
+  }
+
+  const filteredDropDownItems = options.filter((item) =>
+    (item.label ?? item.value)?.toLocaleLowerCase().startsWith(inputValue.toLowerCase())
+  );
```
dprokop, line 155: This ain't sufficient. We need to use performant fuzzy search, pretty much the same that was implemented in https://github.com/grafana/scenes/pull/763

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersCombobox.tsx
dprokop: To play with groups in demo app you can configure your adhoc filter scene object by providing `getTagValuesProvider`:

```ts
getTagValuesProvider: async () => ({
  replace: true,
  values: [{
    text: 'Alice',
    value: 'alice',
    group: 'People',
  }, {
    text: 'Bar',
    value: 'bar'
  }, {
    text: 'Cat',
    value: 'cat',
    group: 'Animals',
  }, {
    text: 'Bob',
    value: 'bob',
    group: 'People',
  }, {
    text: 'Dog',
    value: 'dog',
    group: 'Animals',
  }, {
    text: 'Foo',
    value: 'foo',
  }]
})
```

catherineymgui:
I agree with most of @dprokop's points except:

Error on dropdown - is this a fetch error that only shows when you go into the dropdown? I think it's okay to keep in it, as it's the same pattern as when there are no options found in Explore (for whatever reason).

Screenshot 2024-08-20 at 10 23 58

I would just update the text? "There is an error fetching the labels. Try again."

dprokop:
> I would just update the text? "There is an error fetching the labels. Try again."

How does this align with the design system @catherineymgui? Also `Try again` is not very intuitive here as the user needs to blur and focus the input in order to "try again".

catherineymgui:
@dprokop I don't think we have any comprehensive error handling guidelines in the design system? :cry:

I know what you showed is more for missing user input (like in a form - oh please fill this out, we are missing some additional content or you need to correct this). 

https://grafana.com/developers/saga/components/input/#states

Re: Text

I am struggling to think of a way to tell user what to do that isn't super awkward... how about this
"There is an error fetching labels. Click into the input field again to retry." ??

Sergej-Vlasov:
What has been done and what is remaining for later

### New filters UI -  [design](https://www.figma.com/design/LtJFa2HK2n0IgXlUDoLxuP/Scenes---Ad-hoc-Filters?node-id=119-43391\&t=GlbJPpx9JX62C2Ue-0)

- [x] implement new ad hoc filter combobox design
- [x] Create filter by selecting label, operator and value in a single input
- [x] Clear filter if its incomplete and focused away during creation
- [x] After creation filter should appear in a unified pill
- [x] Ability to remove created filter using `x` icon and backspace
- [x] ability to navigate using keyboard
- [x] Dropdown select contains standard items without checkbox
- [x] loading/error/empty states in dropdown
- [x] ability to commit custom value if no matches in dropdown
- [x] error indicator on the input
- [x] virtualise dropdown component
- [x] support groups in dropdown
- [x] dynamic width of dropdown
- [x] filters fuzzy search
- [ ]  NEXT STEPS in separate PRs
- [ ] Truncated pills if it is too long
- [ ] Collapse / expand rows
- [ ] support multi select for filter value

ashharrison90 (commented):
mostly nits, but i think they make a meaningful difference to the readability 🤓 

couple of other things:
- i think the current behaviour allows you to set a custom value for key, have we intentionally decided to change that? 🤔 
- filtering on operators isn't working currently

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersCombobox.tsx
```diff
@@ -0,0 +1,502 @@
...
+import { DropdownItem, LoadingOptionsPlaceholder, NoOptionsPlaceholder, OptionsErrorPlaceholder } from './DropdownItem';
+import { flattenOptionGroups, flushSyncInputType, fuzzySearchOptions, setupDropdownAccessibility } from './utils';
+import { handleOptionGroups } from '../../utils';
+import { useFloatingInteractions } from './useFloatingInteractions';
+
+interface AdHocComboboxProps {
+  filter?: AdHocFilterWithLabels;
+  isAlwaysWip?: boolean;
+  model: AdHocFiltersVariable;
+  handleChangeViewMode?: () => void;
```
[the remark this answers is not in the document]
Sergej-Vlasov: Yes, it does not apply to combobox in `AdHocFiltersAlwaysWipCombobox` because once it completes it resets and no need to handle view/edit mode (its always edit mode)

Sergej-Vlasov:
@ashharrison90 addressed all points except for operator filtering. uFuzzy setup needs adjustment, but cannot quickly figure out how. Will pick it up first thing tomorrow morning 👍

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/utils.ts
```diff
@@ -11,31 +11,44 @@ export const VIRTUAL_LIST_OVERSCAN = 5;
 export const VIRTUAL_LIST_ITEM_HEIGHT = 38;
 
 export function fuzzySearchOptions(options: Array<SelectableValue<string>>) {
-  const ufuzzy = new uFuzzy();
+  const ufuzzy = new uFuzzy({
+    alpha: '[=|>|<|!|~|a-z]',
```
leeoniya, line 15: this probably should be `alpha: '=><!~a-z'`

the alpha setting should exclude the wrapping `[` and `]` wrapping brackets (uFuzzy will add them automatically). and we can ditch the `|` since square brackets already have OR semantics.

leeoniya: still, i think it would be better to leave the default `a-z` alpha setting and instead do a second-pass filter for any special chars using regular `String.includes` or `String.indexOf`.

leeoniya: it may also be better for performance to do the post-filtering and not to abuse `alpha`, since special chars in alpha will compile to bigger regexps because symbols don't change case:

image

you can use the more explicit long-hand to avoid this problem, but imo it's more :exploding_head: 

https://github.com/leeoniya/uFuzzy?tab=readme-ov-file#charsets-alphabets-diacritics

Sergej-Vlasov: @leeoniya Considering the amount of operators is usually quite low I decided to implement custom search for operators and not modify uFuzzy to meet all cases.

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const pillWrapperRef = useRef<HTMLDivElement>(null);
+
+  const handleChangeViewMode = useCallback(
+    (event?: React.MouseEvent) => {
+      event?.stopPropagation();
+      if (readOnly) {
+        return;
+      }
+      let viewMode = false;
+      flushSync(() => {
```
bfmatei, line 27: Is there any way to avoid `flushSync`? It contains a couple of side effects which can lead to a bad user experience.

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const [viewMode, setViewMode] = useState(true);
+  const pillWrapperRef = useRef<HTMLDivElement>(null);
+
+  const handleChangeViewMode = useCallback(
+    (event?: React.MouseEvent) => {
+      event?.stopPropagation();
+      if (readOnly) {
+        return;
+      }
+      let viewMode = false;
```
bfmatei, line 26: Nit: Any reason why this exists and you don't just `pillWrapperRef.current?.focus();` inside the flush?

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
```diff
@@ -0,0 +1,118 @@
...
+      <div
+        className={cx(styles.combinedFilterPill, { [styles.readOnlyCombinedFilter]: readOnly })}
+        onClick={handleChangeViewMode}
+        onKeyDown={(e) => {
+          if (e.key === 'Enter') {
+            handleChangeViewMode();
+          }
+        }}
+        role="button"
+        aria-label={`Edit filter with key ${filter.keyLabel ?? filter.key}`}
```
bfmatei, line 51: Nit: I think you can extract both `filter.keyLabel ?? filter.key` and `filter.valueLabel ?? filter.value` into separate variables for readability purposes.

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersCombobox.tsx
```diff
@@ -0,0 +1,501 @@
...
+                            listRef.current[index] = node;
+                          },
+                          onClick(event) {
+                            if (filterInputType !== 'value') {
+                              event.stopPropagation();
+                            }
+                            model._updateFilter(filter!, filterInputType, item);
+                            setInputValue('');
+
+                            if (filterInputType === 'key') {
```
bfmatei, line 398: Just a nit but this can be easily be reduced to:
```
flushSyncInputType(filterInputType, setInputType);

if (filterInputType === 'value') {
  handleChangeViewMode?.();
}
```

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersCombobox.tsx
```diff
@@ -0,0 +1,501 @@
...
+  basePill: css({
+    display: 'flex',
+    alignItems: 'center',
+    background: theme.colors.action.disabledBackground,
+    border: `1px solid ${theme.colors.border.weak}`,
+    padding: theme.spacing(0.125, 1, 0.125, 1),
+    color: theme.colors.text.primary,
+    overflow: 'hidden',
+    whiteSpace: 'nowrap',
+    minHeight: '22px',
```
bfmatei, line 453: I would really avoid hardcoded pixels in favor of `theme.spacing(x)`

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
```diff
@@ -0,0 +1,118 @@
...
+    display: 'flex',
+    alignItems: 'center',
+    background: theme.colors.action.selected,
+    borderRadius: theme.shape.radius.default,
+    border: `1px solid ${theme.colors.border.weak}`,
+    padding: theme.spacing(0.125, 0, 0.125, 1),
+    color: theme.colors.text.primary,
+    overflow: 'hidden',
+    whiteSpace: 'nowrap',
+    minHeight: '22px',
```
bfmatei, line 95: I would really avoid hardcoded pixels in favor of `theme.spacing(x)`

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/DropdownItem.tsx
```diff
@@ -0,0 +1,85 @@
...
+);
+
+const getStyles = (theme: GrafanaTheme2) => ({
+  option: css({
+    label: 'grafana-select-option',
+    top: 0,
+    left: 0,
+    width: '100%',
+    position: 'absolute',
+    padding: '8px',
```
bfmatei, line 40: I would really avoid hardcoded pixels in favor of `theme.spacing(x)`

bfmatei (commented):
First look over this PR.

Besides the comments, I suggest fixing the responsive layout.

Screenshot 2024-08-29 at 14 37 28

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
Sergej-Vlasov: The reason for the `flushSync` is related to `focus()`:
We need to focus on the element AFTER we exit view mode. This is why I use flushSync to update state concurrently and after I trigger `focus()`.
 Here how it looks with `flushSync` (focus is kept on the pill after edit):
 

https://github.com/user-attachments/assets/65f9e48f-3eaa-401b-8166-53f9a6726c7e


and without `flushSync` focus disappears and you need extra Tab click to focus back again:

https://github.com/user-attachments/assets/7a0f9a74-22ca-4db8-afbb-dcad90b0fe00

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersCombobox.tsx
Sergej-Vlasov: exported logic into a separate function and replace spaghetti code

Sergej-Vlasov:
@bfmatei addressed all comments 👍 

I would rather fix responsive layout in a separate PR after this, because it requires adjustments to `VariableValueSelectWrapper` and that will need extra testing. So in order not to drag this PR out I will address that separately

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
Sergej-Vlasov: in our case I would prefer `flushSync` approach because `setTimeout` affects `Tab` and `Shift+Tab` movement by focusing on the pill wrapper instead of next (with tab) / previous (with shift+tab) elements. 
Here is visual example with `flushSync`:

https://github.com/user-attachments/assets/e1456150-e90d-4126-b1fc-caf6a99a5f1b

And here is example with `setTimeout`:

https://github.com/user-attachments/assets/e00006ea-d491-4fb3-b2da-5106167683f9

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersCombobox.tsx
```diff
@@ -0,0 +1,487 @@
...
+  }, [open, filterInputType]);
+
+  // when not in wip mode this is the point of switching from view to edit mode
+  //    and in this case we default to 'value' input type and focus input
+  useEffect(() => {
+    if (!isAlwaysWip && refs.domReference.current) {
+      setInputType('value');
+      setInputValue('');
+
+      refs.domReference.current.focus();
```
bfmatei, line 230: Shouldn't this be:
```suggestion
      refs.domReference.current?.focus();
```

bfmatei (approved):
The redesign itself seems to work fine. I'm not particularly thrilled about the code as it's pretty hard to follow although I know we will address it as a technical debt in a subsequent PR.

With the removal of `flushSync`, I'm OK-ish with the PR.

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
```diff
@@ -0,0 +1,121 @@
...
+export function AdHocFilterPill({ filter, model, readOnly }: Props) {
+  const styles = useStyles2(getStyles);
+  const [viewMode, setViewMode] = useState(true);
+  const [shouldFocus, setShouldFocus] = useState(false);
+  const pillWrapperRef = useRef<HTMLDivElement>(null);
+
+  const keyLabel = filter.keyLabel ?? filter.key;
+  const valueLabel = filter.valueLabels?.[0] ?? filter.value;
+
+  const handleChangeViewMode = useCallback(
```
ashharrison90, line 23: maybe `toggleViewMode` (or `toggleEditMode`?)

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+interface Props {
+  filter: AdHocFilterWithLabels;
+  model: AdHocFiltersVariable;
+  readOnly?: boolean;
+}
+
+export function AdHocFilterPill({ filter, model, readOnly }: Props) {
+  const styles = useStyles2(getStyles);
+  const [viewMode, setViewMode] = useState(true);
```
ashharrison90, line 16: maybe this is too much opinion, but i think the whole code reads better if it's flipped to `const [editMode, setEditMode] = useState(false);`

ashharrison90 (approved):
lgtm 👍 some nits/opinions but they can be tackled in follow up PRs / not at all 😄 

couple of things:
- agree with the other comments, think `AdHocFiltersCombobox` needs splitting into smaller components in a follow up PR
- ignoring the exhaustive-deps hook is a code smell. we should also tackle that in a follow up PR

having said all that - the _behaviour_ is great. well done! 🥳 this is perfect for a first pass behind a feature toggle. now we can refactor and make the code more maintainable whilst preserving behaviour 🙌

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
Sergej-Vlasov: will address these in a follow-up PRs

PR #834, opened 2024-07-16
Scenes: Set correct default options on plugin load

mdvictor:
Adds a new state flag to allow getting default options after plugin type change.

Fixes https://github.com/grafana/grafana/issues/89836

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanel.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.tsx
@@ -258,7 +258,7 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
     this.setState({ displayMode });
   };
 
-  public onOptionsChange = async (optionsUpdate: DeepPartial<TOptions>, replace = false, isAfterPluginChange = false) => {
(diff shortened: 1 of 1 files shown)


packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -180,7 +181,7 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
       plugin,
       currentOptions: panel.options,
       currentFieldConfig: panel.fieldConfig,
-      isAfterPluginChange: false,
+      isAfterPluginChange: this.state.isAfterPluginChange ?? false,
```
ivanortegaalba, line 184: An alternative to this is tackling it from the VizPanel.

The problem is when we create a VizPanel, if a VizPanel is created it cannot be after a plugin change. If the VizPanel needs to know when the plugin changes, my approach would be to implement in the `VizPanel` a method to change the plugin that manages all the logic when the plugin changes in the VizPanel.

All this logic is nowadays managed by VizPanelManager but I'd have it in the VizPanel instead.

PR #835, opened 2024-07-16
SceneTimeRange: Allow time range refresh on activation

gtk-grafana:
Add optional state to SceneTimeRange allowing invalidation and refresh of time range using percent or absolute interval.

Adds optional `invalidateAfterMs` and `invalidateAfterPercent` state props to `SceneTimeRangeState`.
If set, on activation we will check to see if enough time has elapsed to meet either criteria. If both values are set in the time range, the shortest duration value will be used.

The values can be set together, or independently. Providing both allows the scenes developer to specify a minimum elapsed interval before the time range is refreshed, which would be helpful when users are looking at very long durations (e.g. weeks), but update more often when users are looking at very short durations (minutes).

10% and 5m seem to be reasonable: 
5 minute interval will refresh after 30s
15minute interval will refresh after 1.5 minutes
30 minute interval will refresh after 3 minutes
1 hour intervals and greater will refresh after 5 minutes

<details>
  <summary>📦 Published PR as canary version: <code>5.6.1--canary.835.10040331799.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.6.1--canary.835.10040331799.0
  npm install @grafana/scenes@5.6.1--canary.835.10040331799.0
  # or 
  yarn add @grafana/scenes-react@5.6.1--canary.835.10040331799.0
  yarn add @grafana/scenes@5.6.1--canary.835.10040331799.0
  ```
</details>


gtk-grafana:
I'd like to get this rolled out to Explore Logs, do we typically include a version bump in the feature PR, or open a separate PR to bump the version?

packages/scenes/src/core/types.ts
```diff
@@ -151,6 +151,15 @@ export interface SceneTimeRangeState extends SceneObjectState {
...
+  /**
+   * When set, the time range will invalidate relative ranges after the specified interval has elapsed
+   */
+  invalidateAfterMs?: number
+  /**
+   * When set, the time range will invalidate relative ranges after the specified percentage of the current interval has elapsed.
+   * If both invalidate values are set, the smaller value will be used for the given interval.
+   */
+  invalidateAfterPercent?: number
 }
```
torkelo, line 163: do we need both? 

alt names
* autoRefreshAfterPercentDiff 
* refreshAfterPercentDiff

thinking refresh is better name as that is what we call re-evaluating the time range and updating it

since it only happens on activation, maybe
refreshOnActivate: { percent?: number, afterMs?: number }

allows us to add more options in the future. 

Another idea I had was making this a behavior you add to $behaviors but I think this is so universal that all time ranges should have this. Wonder if we should set a default value of 10% as well.

torkelo (commented):
thin this looks very good and close to mergable, just some naming / defaults to decide. 

thoughts @dprokop @axelavargas @ivanortegaalba ? 

This is trying solve the problem of auto refreshing view when you come back to a cached scene with a relative time range after some period of time and want to auto refresh time range (and data)

packages/scenes/src/core/types.ts
gtk-grafana: > do we need both?

I can see use-cases for both, if I have a fast query or is otherwise critical that users are always seeing most up-to-date data as possible I could see users wanting to set the millisecond threshold as the intervalMs from the request, but the percentage alone makes a good default that works well in most situations, and will likely be the most popular option. However I liked the idea of using them both together so users can set a maximum duration of how stale the time range can be, independent of how large the time range is, but the interval would shorten as a percentage for shorter time ranges where the percentage of the time range interval is less then the static refresh interval.
i.e. `refreshIntervalMs = floor(timeRangeIntervalMs * percentage, refreshIntervalMs)`
 
But to answer the question, we don't _need_ them both as developers could implement this functionality in the calling application themselves by checking the current time range and setting either the percentage or duration depending on their needs, but the ability to add either or both adds a good mix of flexibility and simplicity for the implementor.

> refreshOnActivate: { percent?: number, afterMs?: number }

Good call, I've refactored and renamed things in the [latest commit](https://github.com/grafana/scenes/pull/835/commits/3c6eb07f50a9c45063c1d4997496215c712bfd75)

> Another idea I had was making this a behavior you add to $behaviors but I think this is so universal that all time ranges should have this. Wonder if we should set a default value of 10% as well.

I'll defer to you experts on this, personally I wouldn't mind a default getting set, I avoided it in this PR because I didn't want the potential of regression to hold up merge.

gtk-grafana: Setting the refresh interval as the `intervalMs` from the query might be the a perma-fresh approach for a default, but the implementation of that isn't apparent to me at this time.

packages/scenes/src/core/SceneTimeRange.tsx
```diff
@@ -193,4 +237,9 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
 
     this.setState(update);
   }
+
+  public isRelative(): boolean {
+    return this.state.to === NOW
+  }
 }
+const NOW = 'now'
```
torkelo, line 245: this does not work, as to can be expressions like `now-5m`

gtk-grafana: Good catch, updated!

packages/scenes/src/core/SceneTimeRange.tsx
```diff
@@ -56,6 +56,10 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
       setWeekStart(this.state.weekStart);
     }
 
+    if(this.isRelative()){
```
torkelo, line 59: rangeUtil.isRelativeTimeRange

dprokop (approved):
@gtk-grafana  - sorry for a late review. This looks like a great addition and the code is 👍.

 I think we could actually default to 10% as @torkelo suggests. Even core grafana dashboards need this ATM.

PR #836, opened 2024-07-17
Scenes: Move change panel plugin logic in VizPanel

mdvictor:
This PR moves panel plugin logic from grafana, from `VizPanelManager` (related [PR](https://github.com/grafana/grafana/pull/90535) ) into the `VizPanel`. This is needed because the `VizPanel` is aware of when the plugin is updated and we need the new plugin to fetch default configs, to be able to merge existing options/fieldConfigs, if they exist, with default ones.

This also changes things a bit in the sense that we no longed create a new `VizPanel` but reuse the existing one by updating it's state with the updated options/fieldConfigs. In the `changePluginType` function we receive overwrite options and fieldConfigs which represent, in the context of core Grafana, configs that are cached when switching panels in panel edit. These configurations, if present, are passed to the VizPanel to properly set both existing configs and default panel plugin configs after loading the new plugin. If we do not have any cached configs then we will load the new panel with default values.

TODO:
- [x] Add a demo [here](http://localhost:3000/a/grafana-scenes-app/demos/dynamic-panel-options-and-field-config) for dynamically changing a plugin panel.

<details>
  <summary>📦 Published PR as canary version: <code>5.5.0--canary.836.9987320115.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.5.0--canary.836.9987320115.0
  npm install @grafana/scenes@5.5.0--canary.836.9987320115.0
  # or 
  yarn add @grafana/scenes-react@5.5.0--canary.836.9987320115.0
  yarn add @grafana/scenes@5.5.0--canary.836.9987320115.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanel.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.tsx
@@ -255,25 +255,17 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
   };
 
   public async changePluginType(pluginId: string, options?: DeepPartial<{}>, newFieldConfig?: FieldConfigSource) {
-    //we need old state for newOptions
     const {
       options: prevOptions,
       fieldConfig: prevFieldConfig,
       pluginId: prevPluginId,
     } = this.state;
 
-    //clear field config cache
+    //clear field config cache to update it later
     this._dataWithFieldConfig = undefined;
 
-    //we need options/fieldConfig to be set in loadPlugin to handle scenario where we 
-    //need the default options. e.g.: we change plugin type to new plugin that
-    //has no cached options, so we need to get default options merged with {}, instead
-    //of old plugin options, so we do not pollute new plugin options.
-    // but we cannot just set the state to {} because things will crash on rerender(e.g.: PanelOptions)
-    // because it expects options to have properties.
     await this._loadPlugin(pluginId, options ?? {}, newFieldConfig, true);
 
-    //update options if a migration is needed
     const panel: PanelModel = {
       title: this.state.title,
       options: this.state.options,
@@ -284,7 +276,6 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
 
     const newOptions = this._plugin?.onPanelTypeChanged?.(panel, prevPluginId, prevOptions, prevFieldConfig);
 
-    //newOptions returns empty when changing to timeseries, we need isEmpty, otherwise it loses cachedOptions
     if (newOptions && !isEmpty(newOptions)) {
       this.onOptionsChange(newOptions, true, true);
     }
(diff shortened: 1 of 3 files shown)


packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -245,6 +254,33 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
     return sceneTimeRange.state.value;
   };
 
+  public async changePluginType(pluginId: string, options?: DeepPartial<{}>, newFieldConfig?: FieldConfigSource) {
```
ivanortegaalba, line 257:
```suggestion
  public async changePluginType(pluginId: string, newOptions?: DeepPartial<{}>, newFieldConfig?: FieldConfigSource) {
```
I understand, when changing the plugin type, we want to apply new options and field config to the new plugin, right? I'd try to be kind of specific for people using the changePluginType in their apps :D

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -245,6 +254,33 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
...
+    const {
+      options: prevOptions,
+      fieldConfig: prevFieldConfig,
+      pluginId: prevPluginId,
+    } = this.state;
+
+    //clear field config cache to update it later
+    this._dataWithFieldConfig = undefined;
```
ivanortegaalba, line 265: I think this is not used, right? 🤔

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -245,6 +254,33 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
...
+
+    const panel: PanelModel = {
+      title: this.state.title,
+      options: this.state.options,
+      fieldConfig: this.state.fieldConfig,
+      id: 1,
+      type: pluginId,
+    };
+
+    const newOptions = this._plugin?.onPanelTypeChanged?.(panel, prevPluginId, prevOptions, prevFieldConfig);
```
ivanortegaalba, line 277:
```suggestion
    // onPanelTypeChanged is mainly used by plugins to migrate from Angular to React. 
    // For example, this will migrate options from 'graph' to 'timeseries' if the previous and new plugin ID matches. 
    const newOptions = this._plugin?.onPanelTypeChanged?.(panel, prevPluginId, prevOptions, prevFieldConfig);
```

ivanortegaalba: What do you think about adding some comments to explain why this exists? It was hard for us to find 😓

ivanortegaalba (approved):
I hardly tested in Grafana, and it works as expected. Amazing job Victor!!!! 💯 

✅ When changing type, the options are preserved
✅ When changing type, the new options are applied to the viz
✅ When changing type, the color palette is properly computed
✅ When changing type, the overrides are preserved
✅ When changing type, and changing again, the options are remembered between plugins, including overrides.

Really great work Victor!

packages/scenes/src/components/VizPanel/VizPanel.tsx
torkelo: it's heavily used

torkelo (commented):
Looks good

dprokop (approved):

ivanortegaalba:
LGTM!

PR #840, opened 2024-07-22
Update timerange in templateSrv for queryVars

mdvictor:
We have an [escalation](https://github.com/grafana/support-escalations/issues/11540) around the ADS DS, where the timerange is not being picked on query. The time range is being pulled from templateSrv [here](https://github.com/grafana/azure-data-explorer-datasource/blob/4e52da60758a51f898040bb9fd4829e1aab6d6a7/src/datasource.ts#L501), but with the new scenes architecture this range is null/never updated in templateSrv. This PR updates the timerange using the request range in `QueryVariable.tsx`, before querying in the plugin itself.

In the old architecture there were events dispatched on various actions to update the timerange in the templateSrv. It seems in scenes we aren't updating it at all so the templateSrv timerange is always null. Updating templateSrv in SceneTimeRange on state update should cover most cases where range in templateSrv is needed, I think.

<details>
  <summary>📦 Published PR as canary version: <code>5.6.1--canary.840.10038144895.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.6.1--canary.840.10038144895.0
  npm install @grafana/scenes@5.6.1--canary.840.10038144895.0
  # or 
  yarn add @grafana/scenes-react@5.6.1--canary.840.10038144895.0
  yarn add @grafana/scenes@5.6.1--canary.840.10038144895.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/core/SceneTimeRange.tsx
+++ b/packages/scenes/src/core/SceneTimeRange.tsx
@@ -8,7 +8,7 @@ import { SceneTimeRangeLike, SceneTimeRangeState, SceneObjectUrlValues } from '.
 import { getClosest } from './sceneGraph/utils';
 import { parseUrlParam } from '../utils/parseUrlParam';
 import { evaluateTimeRange } from '../utils/evaluateTimeRange';
-import { config } from '@grafana/runtime';
+import { config, getTemplateSrv } from '@grafana/runtime';
 
 export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> implements SceneTimeRangeLike {
   protected _urlSync = new SceneObjectUrlSyncConfig(this, { keys: ['from', 'to', 'timezone'] });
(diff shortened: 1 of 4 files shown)


packages/scenes/src/core/SceneTimeRange.tsx
```diff
@@ -37,15 +39,19 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
...
-                  this.state.from,
-                  this.state.to,
-                  timeZoneSource.getTimeZone(),
-                  this.state.fiscalYearStartMonth,
-                  this.state.UNSAFE_nowDelay
-                ),
+                value: timeRange,
               });
+
+              getTemplateSrv()?.updateTimeRange(timeRange);
```
dprokop, line 54: @mdvictor - i'm not a big fan of doing this. The ADX is built on top of a private API (`timeRange` in template srv is private https://github.com/grafana/grafana/blob/main/public/app/features/templating/template_srv.ts#L54).


I think we should implement a getter for a timeRange in [core templateSrv](https://github.com/grafana/grafana/blob/main/public/app/features/templating/template_srv.ts) that would figure out the time range out from `window.__grafanaSceneContext`, similarly [how we switch to scenes string interpolation](https://github.com/grafana/grafana/blob/main/public/app/features/templating/template_srv.ts#L263). If grafana scene context wouldn't be there we simply return this.timeRange

PR #841, opened 2024-07-22
Make UrlSyncManager non singleton

aocenas:
Fixes: https://github.com/grafana/grafana/issues/93225
This changes the UrlSyncManager to be used as non singleton.

This relies on this grafana PR: https://github.com/grafana/grafana/pull/90759. Also see the whole demo branch for the Sidecar project here https://github.com/grafana/grafana/pull/88797.

This isn't much changes but need some advice on how to approach any possible backward compatibility issues and testing. As there was already some code created with the non singleton use in mind the changes to apps seem minimal. See this draft PR for the Explore Traces app: https://github.com/grafana/explore-traces/pull/102

### DIFF
--- a/packages/scenes-react/src/contexts/SceneContextProvider.tsx
+++ b/packages/scenes-react/src/contexts/SceneContextProvider.tsx
@@ -1,11 +1,12 @@
-import React, { createContext, useContext, useEffect, useState } from 'react';
+import React, { createContext, useContext, useEffect, useMemo, useState } from 'react';
 import {
   SceneTimeRangeState,
   SceneTimeRange,
   behaviors,
   UrlSyncContextProvider,
-  getUrlSyncManager,
+  UrlSyncManager,
 } from '@grafana/scenes';
+import { useLocationService } from '@grafana/runtime';
 
 import { SceneContextObject, SceneContextObjectState } from './SceneContextObject';
 
@@ -31,6 +32,9 @@ export interface SceneContextProviderProps {
  * Wrapps the react children in a SceneContext
  */
 export function SceneContextProvider({ children, timeRange, withQueryController }: SceneContextProviderProps) {
+  const locationService = useLocationService()
+  const urlSyncManager = useMemo(() => new UrlSyncManager(locationService), [locationService]);
+
   const parentContext = useContext(SceneContext);
   const [childContext, setChildContext] = useState<SceneContextObject | undefined>();
 
@@ -38,7 +42,7 @@ export function SceneContextProvider({ children, timeRange, withQueryController 
   const initialTimeRange = timeRange;
 
   useEffect(() => {
-    const state: SceneContextObjectState = { children: [] };
+    const state: SceneContextObjectState = { children: [], urlSyncManager };
 
     if (withQueryController) {
       state.$behaviors = [new behaviors.SceneQueryController()];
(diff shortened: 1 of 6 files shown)


packages/scenes/src/components/SceneApp/SceneAppPageView.tsx
```diff
@@ -23,6 +23,7 @@ export function SceneAppPageView({ page, routeProps }: Props) {
   const scene = page.getScene(routeProps.match);
   const isInitialized = containerState.initializedScene === scene;
   const {layout} = page.state;
+  const locationService = useLocationService();
```
torkelo, line 26: originally I created the GrafanaContext for this purpose to expose scoped services, 
https://github.com/grafana/grafana/pull/52128 

I even had a special interface for the context for plugins exposed via grafana/runtime but plugins platform felt unsure about exposing it to plugins.

packages/scenes-react/src/contexts/SceneContextProvider.tsx
```diff
@@ -31,14 +32,17 @@ export interface SceneContextProviderProps {
  * Wrapps the react children in a SceneContext
  */
 export function SceneContextProvider({ children, timeRange, withQueryController }: SceneContextProviderProps) {
+  const locationService = useLocationService()
```
torkelo, line 35: to be able to merge this we would need a fallback for older versions of Grafana that does not export this

packages/scenes/src/components/SceneApp/SceneAppPageView.tsx
aocenas: Ah ok missed that, cool so I guess I can just use that context instead. Do you know some context why plugins platform felt unsure about exposing it to plugins?

packages/scenes-react/src/contexts/SceneContextProvider.tsx
aocenas: So how would that work? I mean I guess I have to make the package.json deps to be the newest grafana/runtime so it compiles without errors but the does it make sense to just do something like:
```
if (useLocationService) {
   useLocationService
} else {
  // do something else?
}
``` 
Or is there some better way to do some detection of this?

Also in general I do intend for the singleton access to still work. In the sense that I think there should be some opt in API or config setting for apps that say they are safe to use the sidecar so this can be rolled out safely and so any app that does not update would just be opened full screen still.

packages/scenes/src/components/SceneApp/SceneAppPageView.tsx
torkelo: think it was just that they where unsure if it was a good pattern at that point, or if was going to be useful 

@jackw do you recall?

packages/scenes-react/src/contexts/SceneContextProvider.tsx
torkelo: > So how would that work? I mean I guess I have to make the package.json deps to be the newest grafana/runtime so it compiles without errors but then does it make sense to just do something like:

previously we have imported it with tsignore so we do not need to bump dependency (as that makes the version unusable for plugins for a few months, and remove the tsignore later after the grafana dependency upgrade. Not a very great solution :(

packages/scenes/src/components/SceneApp/SceneAppPageView.tsx
jackw: >think it was just that they where unsure if it was a good pattern at that point, or if was going to be useful

From what I remember it was about getting that direction agreed on for plugins then doing the necessary to deprecate the old singletons, document it all, and iron out any kinks before it was exposed so we don't go through the bother of "breaking changes" due to exposing it at the initial time of usage. I felt the name of the hook was a little too all encompassing and might mean people put whatever they want in there without realising plugins get access but that's probably just me being overly concerned about over exposing things. 🤷

aocenas:
Created a new PR for this as there were changes that conflict: https://github.com/grafana/scenes/pull/899

PR #842, opened 2024-07-22
SceneQueryRunner: Limit concurrent requests 

gtk-grafana:
Why: We want to limit the number of concurrent requests being made to loki in the Explore Logs app.

What: Adding a singleton to the SceneQueryRunner which inits the p-limit concurrency queue, for instances of SceneQueryRunner that have the concurrency options defined.

How: Allow passing in of new `concurrency` options to the SceneQueryRunner.

Todo: How does it work with query splitting?


gtk-grafana:
Converting a observable to a promise will prevent any state change in the observable besides the final one.

torkelo:
Grafana backendSrv already limits requests to 5 concurrent I think

PR #845, opened 2024-07-24
WIP: Add support for systemPanelFilterVar/systemDynamicRowSizeVar

kaydelaney:
Putting this up as a draft so @axelavargas can take a look.
Still needs to be put behind a feature flag.

### DIFF
--- a/packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
+++ b/packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
@@ -3,7 +3,7 @@ import ReactGridLayout from 'react-grid-layout';
 import { SceneObjectBase } from '../../../core/SceneObjectBase';
 import { SceneLayout, SceneObjectState } from '../../../core/types';
 import { DEFAULT_PANEL_SPAN, GRID_COLUMN_COUNT } from './constants';
-import { isSceneGridRow, SceneGridItem } from './SceneGridItem';
+import { isSceneGridRow } from './SceneGridItem';
 import { SceneGridLayoutRenderer } from './SceneGridLayoutRenderer';
 
 import { SceneGridRow } from './SceneGridRow';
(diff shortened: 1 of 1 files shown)


torkelo (commented):
Is there a way to build this as a behavior or / pluggable feature for the grid? (not sure if that is the case or if it's a good idea, but feels like a non core feature of the grid itself

kaydelaney:
> Is there a way to build this as a behavior or / pluggable feature for the grid? (not sure if that is the case or if it's a good idea, but feels like a non core feature of the grid itself

I agree it really isn't a core feature. I'll try and see if I can think of a way to make it more "modular"

packages/scenes/src/components/layout/grid/SceneGridLayout.tsx
```diff
@@ -355,8 +364,43 @@ export class SceneGridLayout extends SceneObjectBase<SceneGridLayoutState> imple
...
+      if (this.savedLayout === undefined) {
+        this.savedLayout = this.state.children.map((v) => v.clone());
+      }
+
+      const panelFilterInterpolated = interpolate(this, searchString).toLowerCase();
+      const filteredChildren = this.state.children.filter(
+        (v) =>
+          typeof (v.state as any).body?.state?.title === 'string' &&
+          (v.state as any).body.state.title.toLowerCase().includes(panelFilterInterpolated)
+      );
```
axelavargas, line 381: We can also interpolate the `title` , we could match variables that are in the title, something like: 

```suggestion
      const filteredChildren = this.state.children.filter(
        (v) => {
          const panelTitleInterpolated = interpolate(v, (v.state as any).body?.state?.title).toLowerCase();
          return typeof (v.state as any).body?.state?.title === 'string' &&
            (panelTitleInterpolated.includes(panelFilterInterpolated));
        }
      );

```

axelavargas (commented):
Hey @kaydelaney, I think this approach works better for our use case 🌟 . I gave it a local test with a dummy scenes app and the Grafana core repository. 

### Dummy scenes app
https://github.com/user-attachments/assets/0a0ceba0-f8e7-49d2-80e9-17f56af61256

### Grafana core

https://github.com/user-attachments/assets/7bc4fbda-c6da-4e0a-93e1-2ca9348c34e1

## UX considerations:

Including "search by title" and "resize per row" in the layout canvas might not give the best user experience. However, we're planning to hide this feature behind a toggle in core Grafana, so it'll only be available to a specific customer and won't affect the wider user base much. Also, not many Grafana applications use the  `SceneGridLayout`.

What do you think, @torkelo? Do you have strong opinions on this or a better idea for solving the problem? 🙃

PR #859, opened 2024-08-02
SceneQueryRunner: run queries when datasource variable updates

gtk-grafana:
Runtime datasources don't update the UID of the datasource string that is passed into the datasource property when the datasource variable is updated.

As such we should allow developers to pass in a datasource variable into the $variables, and any update to the datasource variable should cause any queryRunner to re-query.

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.test.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.test.ts
@@ -37,6 +37,7 @@ import { LocalValueVariable } from '../variables/variants/LocalValueVariable';
 import { SceneObjectBase } from '../core/SceneObjectBase';
 import { ExtraQueryDescriptor, ExtraQueryProvider } from './ExtraQueryProvider';
 import { SafeSerializableSceneObject } from '../utils/SafeSerializableSceneObject';
+import { DataSourceVariable } from '../variables/variants/DataSourceVariable';
 
 const getDataSourceMock = jest.fn().mockReturnValue({
   uid: 'test-uid',
@@ -741,6 +742,35 @@ describe('SceneQueryRunner', () => {
       expect(queryRunner.state.data?.state).toBe(LoadingState.Done);
 
       variable.changeValueTo('AB');
+
+      await new Promise((r) => setTimeout(r, 1));
+
+      expect(runRequestMock.mock.calls.length).toBe(2);
+    });
+
+    it('Should execute query datasource when variable updates', async () => {
+      const variable = new TestVariable({ name: 'A', value: '', query: 'A.*' });
+      const datasourceVariable = new DataSourceVariable({ name: 'ds', label: 'ds', value: 'uid-1', pluginId: '', options: [], });
+      const queryRunner = new SceneQueryRunner({
+        queries: [{ refId: 'A', query: '$A' }],
+      });
+
+      const timeRange = new SceneTimeRange();
+
+      const scene = new SceneFlexLayout({
+        $variables: new SceneVariableSet({ variables: [variable, datasourceVariable] }),
+        $timeRange: timeRange,
+        $data: queryRunner,
+        children: [],
+      });
+
+      scene.activate();
+      // should execute query when variable completes update
+      variable.signalUpdateCompleted();
+      await new Promise((r) => setTimeout(r, 1));
+      expect(queryRunner.state.data?.state).toBe(LoadingState.Done);
+
+      datasourceVariable.changeValueTo('uid-2')
 
       await new Promise((r) => setTimeout(r, 1));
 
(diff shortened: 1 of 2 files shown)


packages/scenes/src/querying/SceneQueryRunner.test.ts
```diff
@@ -747,6 +748,35 @@ describe('SceneQueryRunner', () => {
...
+      });
+
+      const timeRange = new SceneTimeRange();
+
+      const scene = new SceneFlexLayout({
+        $variables: new SceneVariableSet({ variables: [variable, datasourceVariable] }),
+        $timeRange: timeRange,
+        $data: queryRunner,
+        children: [],
+      });
```
torkelo, line 765: don't see anywhere where SceneQueryRunner depends on $ds

Not sure I understand the PR / problem description

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -267,6 +268,10 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
     if (variable instanceof GroupByVariable && this._isRelevantAutoVariable(variable)) {
       this.runQueries();
     }
+
+    if (variable instanceof DataSourceVariable) {
+      this.runQueries();
+    }
```
torkelo, line 274: nothing here that checks that this data source variable is in any way used / referenced by this SceneQueryRunner (and if it was it would automatically re-run queries when it changed)

packages/scenes/src/querying/SceneQueryRunner.test.ts
gtk-grafana: Sorry about that, the specific problem we're experiencing is that in Explore Logs we have a query runner that calls a resource call ([index/volume](https://grafana.com/docs/loki/latest/reference/loki-http-api/#query-log-volume)) using a custom runtime datasource. This query runner has a simple query expression, it outputs a regex query for the value of a single custom variable. ```{${SERVICE_NAME}=~\`${VAR_SERVICE_EXPR}.+\`}``` The custom variable is a free text search by the user so they can search for services matching the text that they enter in the variable input.

The datasource isn't being interpolated anywhere in the expression, and as such if the user changes the datasource in another route in the application, and then navigates to the scene with this variable, the query results against the wrong datasource are shown to the user, even though the datasource variable correctly shows the updated datasource.

Currently, only scene query runners with adHocFilterVariables have logic for updating when the datasource changes, however this logic is broken for runtime datasources, as the `_isRelevantAutoVariable` function checks against the datasource uid, which in the context of the scene query runner is always constant for runtime datasources, and as such `_isRelevantAutoVariable` will always return false.

Additionally, an AdHocFilterVariable is not appropriate for this situation, as this particular variable state is driven by a free text search input that the user types in (i.e. we need `changeValueTo`).

This PR is attempting to allow developers to define more explicit dependency relationships between variables, in this case whenever a datasource variable is changed we can re-run the resource query only when certain variables have changed, so we can still enjoy the benefits of caching query results (i.e. not querying every time a particular scene is activated), and prevent from needing to add logic to all possible scenes that route to a specific scene that would manually run a query, or set a state indicating that a particular query would need to be re-run on activation, as right now the only way to catch this bug is to subscribe to the state of the variable, which won't help for inactive scenes that have previously queried against the previous datasource.

To further illustrate the problem, when I first came across this bug, I came up with a hack that gave us the desired functionality without any scene core changes by updating the expression of the query to include a no-op expression containing the value of the datasource variable, i.e. `__invalid_label__!="${DATASOURCE_VAR}"`. This PR is an attempt to allow us to maintain that same relationship between a query runner and a variable, without needing to resort to such hackery.

TL;DR We want to bust the cache whenever a particular variable changes. The proposed solution in this PR, is that if a developer explicitly passes a datasource variable to the $variables state of a scene query runner, that query would be re-executed whenever the dependency variable has changed, as its last execution was against a datasource that is no longer relevant to the application context.

packages/scenes/src/querying/SceneQueryRunner.ts
svennergr: Yea, this use case would only be for cases where the `SceneQueryRunner` is used in combination with a `RuntimeDatasource`, but we still want to re-query when the datasource was changed. WDYT about adding another condition to https://github.com/grafana/scenes/pull/859/files#diff-3a9d2ec7e70ac7ecd766a266947161ad66de667317aad3e793f8e494f0b1d619R272 if the current datasource of the `SceneQueryRunner` is a `RuntimeDatasource`? Or should we solve that in a different way?

packages/scenes/src/querying/SceneQueryRunner.test.ts
torkelo: @gtk-grafana 

> The datasource isn't being interpolated anywhere in the expression, and as such if the user changes the datasource in another route in the application, and then navigates to the scene with this variable, the query results against the wrong datasource are shown to the user, even though the datasource variable correctly shows the updated datasource.

if the datasource variable is not being interpolated, then where is it used? and why would this scene query runner care about it? 

> This PR is an attempt to allow us to maintain that same relationship between a query runner and a variable, without needing to resort to such hackery.

not sure that is true, it looks to always re-run all scene query runners when a data source variable changes no matter if it's relevant or not. 

Would this not be solved if you actually used the data source variable in the query definition (for the runtime data source)

gtk-grafana: Thanks Torkel, yes it looks like using the datasource prop of the query to contain the un-interpolated datasource variable as the uid was exactly what we were looking for. Closing!

PR #861, opened 2024-08-06
Add backwards compatibility for query variables referencing themselves

oscarkilhed:
There are some old dashboards that have query variables that reference themself in the query, this is preventing the query runner from executing the variable query as it is waiting for the variable to load. 

I'm not confident this is the best solution to this issue. A different approach would be to remove the reference to itself when we build the scene, however this could potentially be quite messy. We should probably also look into if this is an issue with other variable types.

https://github.com/grafana/support-escalations/issues/11766

### DIFF
--- a/packages/scenes/src/core/sceneGraph/sceneGraph.test.tsx
+++ b/packages/scenes/src/core/sceneGraph/sceneGraph.test.tsx
@@ -1,13 +1,21 @@
-import React from 'react';
 import { render } from '@testing-library/react';
-import { TestAnnotationsDataLayer } from '../../querying/layers/TestDataLayer';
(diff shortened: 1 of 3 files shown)


ivanortegaalba (commented):
IMO, this is good to avoid hard braking and have the dashboard invalid. At least we perform with no errors. As the next step, we could show a warning to know this variable will never be resolved.

Let's add some tests for it @oscarkilhed! Great work! 💯

ivanortegaalba:
Tests are breaking too

torkelo (approved):

ivanortegaalba (approved):
I added some tests! Merging!

PR #865, opened 2024-08-09
GroupByVariable: Support custom filter keys

### LINKED ISSUE #847: GroupByVariable allow custom input
Allow users to type in custom values into a `GroupByVariable` if the value isn't listed in the dropdown. For example, `CustomVariable` and `QueryVariable` already allow this.

andreaalopez:
Allow custom input for GroupByVariable
Closes #847

### DIFF
--- a/packages/scenes/src/variables/groupby/GroupByVariable.tsx
+++ b/packages/scenes/src/variables/groupby/GroupByVariable.tsx
@@ -264,6 +264,7 @@ export function GroupByVariableRenderer({ model }: SceneComponentProps<MultiValu
       id={key}
       placeholder={'Select value'}
       width="auto"
+      allowCustomValue={true}
       inputValue={inputValue}
       value={uncommittedValue}
       noMultiValueWrap={true}


torkelo (approved):

torkelo:
@andreaalopez can you sign CLA?

dprokop:
@ashharrison90 wasn't there a reason why we didn't enable custom value creation for Group By?

ashharrison90:
> @ashharrison90 wasn't there a reason why we didn't enable custom value creation for Group By?

@dprokop sorry completely missed this. it used to support it, and we reverted it [here](https://github.com/grafana/scenes/pull/736). i don't think there's a huge issue with supporting it tbh 🤷 we support custom values on both sides of an adhoc filters which is very similar tbh

ch-plattner:
Hi, if there are no blockers, we'd love to merge this. Thanks for engaging on this contribution!

ashharrison90:
yeah i think we can just merge 👍

PR #868, opened 2024-08-13
AdHocFilters: Add support for new `isOneOf` multi value operator

ashharrison90:
- adds new `supportsMultiValueOperator` property to `AdHocFiltersVariable`
- adds new `isMultiValueOperator` method
- adjusts `_getOperators` to conditionally return the new multi value operators depending on the value of `supportsMultiValueOperator`
- modifies `_updateFilter` to take in a generic partial filter update
  - this allows for updating multiple keys at once! 🥳 
- adds supports for rendering a `<MultiSelect>` when a multi value operator is selected in `AdHocFilterRenderer`
- adjusts the default `renderFilter` expression (which should produce a prometheus-compatible expression) to handle `=|` and `!=|` by converting to regex and joining with `|`
- extends `AdHocFiltersVariableUrlSyncHandler` to handle multiple value/label pairs
  - modifies `valueLabel` to `valueLabels` so it can hold multiple labels
  - this bit feels a little messy/brittle imo 😕 any suggestions? 

For https://github.com/grafana/hyperion-planning/issues/1
For https://github.com/grafana/grafana/issues/85074

<details>
  <summary>📦 Published PR as canary version: <code>5.11.0--canary.868.10613389161.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.11.0--canary.868.10613389161.0
  npm install @grafana/scenes@5.11.0--canary.868.10613389161.0
  # or 
  yarn add @grafana/scenes-react@5.11.0--canary.868.10613389161.0
  yarn add @grafana/scenes@5.11.0--canary.868.10613389161.0
  ```
</details>

### DIFF
--- a/packages/scenes-app/src/demos/adhocFiltersDemo.tsx
+++ b/packages/scenes-app/src/demos/adhocFiltersDemo.tsx
@@ -40,6 +40,7 @@ export function getAdhocFiltersDemo(defaults: SceneAppPageState) {
                   // Only want keys for this series
                   baseFilters: [{ key: '__name__', operator: '=', value: 'ALERTS', condition: '' }],
                   datasource: { uid: 'gdev-prometheus' },
+                  supportsMultiValueOperators: true,
                 }),
               ],
             }),
@@ -77,6 +78,7 @@ export function getAdhocFiltersDemo(defaults: SceneAppPageState) {
             applyMode: 'manual',
             datasource: { uid: 'gdev-prometheus' },
             filters: [{ key: 'job', operator: '=', value: 'grafana', condition: '' }],
+            supportsMultiValueOperators: true,
           });
 
           return new EmbeddedScene({
@@ -138,6 +140,7 @@ export function getAdhocFiltersDemo(defaults: SceneAppPageState) {
                   hide: VariableHide.hideLabel,
                   datasource: { uid: 'gdev-prometheus' },
                   filters: [{ key: 'job', operator: '=', value: 'has no text', condition: '' }],
+                  supportsMultiValueOperators: true,
                 }),
                 new AdHocFiltersVariable({
                   name: 'button-text',
@@ -147,6 +150,7 @@ export function getAdhocFiltersDemo(defaults: SceneAppPageState) {
                   addFilterButtonText: 'Add a filter',
                   datasource: { uid: 'gdev-prometheus' },
                   filters: [{ key: 'job', operator: '=', value: 'has text on add button', condition: '' }],
+                  supportsMultiValueOperators: true,
                 }),
 
                 new AdHocFiltersVariable({
@@ -157,6 +161,7 @@ export function getAdhocFiltersDemo(defaults: SceneAppPageState) {
                   addFilterButtonText: 'Filter',
                   datasource: { uid: 'gdev-prometheus' },
                   filters: [{ key: 'job', operator: '=', value: 'also has text on add button', condition: '' }],
+                  supportsMultiValueOperators: true,
                 }),
               ],
             }),
--- a/packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
@@ -40,7 +40,7 @@ export function AdHocFilterRenderer({ filter, model }: Props) {
   // To not trigger queries on every selection we store this state locally here and only update the variable onBlur
   // TODO remove expect-error when we're on the latest version of @grafana/data
   // @ts-expect-error
-  const [uncommittedValue, setUncommittedValue] = useState(filter.values ? filter.values.map((value, index) => keyLabelToOption(value, filter.valueLabels?.[index])) : []);
+  const [uncommittedValue, setUncommittedValue] = useState<SelectableValue>(filter.values ? filter.values.map((value, index) => keyLabelToOption(value, filter.valueLabels?.[index])) : []);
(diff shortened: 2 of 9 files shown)


packages/scenes-app/src/demos/adhocFiltersDemo.tsx
```diff
@@ -40,6 +40,7 @@ export function getAdhocFiltersDemo(defaults: SceneAppPageState) {
                   // Only want keys for this series
                   baseFilters: [{ key: '__name__', operator: '=', value: 'ALERTS', condition: '' }],
                   datasource: { uid: 'gdev-prometheus' },
+                  supportsMultiValueOperators: true,
```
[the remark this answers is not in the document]
ashharrison90: hmmmmm, good point. i think we were trying to avoid putting something like that in the datasource since then we'd have to wait for the datasource request to return. 

we could just put all this logic behind the feature toggle, and implement support in the other datasources before enabling the toggle? 🤔 

looking at the docs, it looks like adhoc filters are only supported in Prometheus, Loki, InfluxDB, and Elasticsearch. 
prometheus support is added in [grafana#91837](https://github.com/grafana/grafana/pull/91837), can't imagine it would be much work to add support in loki/influx/elastic.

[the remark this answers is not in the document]
dprokop: Given that we are going to introduce this behind a feature toggle I think we could assume it's only Prometheus that supports multi value filters by default from day one. 

IMO whether or not data source should allow this operator would eventually depend on a different API (_type-aware filters_) where the availability of operators would be driven by `getTagKeys` call results, so that each dimension would define what type it is (string, number, enum) and whether or not it supports multi value selection. 

WDYT?

packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
```diff
@@ -48,11 +51,57 @@ export function AdHocFilterRenderer({ filter, model }: Props) {
...
+    isMulti: true,
+    // TODO remove expect-error when we're on the latest version of @grafana/data
+    // @ts-expect-error
+    value: filter.values ?? null,
+    components: {
+      Option: OptionWithCheckbox,
+    },
+    closeMenuOnSelect: false,
+    onChange: (v: SelectableValue) => {
+      const updatedValue = v.map((option: SelectableValue<string>) => option.value).join('__gfp__');
```
dprokop, line 90: What do you think about keeping the value as a regex string rather than joining individual values with a special set of characters? I think that would be more intuitive.

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -105,6 +109,36 @@ export type getTagValuesProvider = (
...
+  value: '<'
+}, {
+  value: '>'
+}, {
+  value: '=~',
+  description: 'Matches regex',
+}, {
+  value: '!~',
+  description: 'Does not match regex',
+}];
```
dprokop, line 140: For consideration:

```suggestion
const OPERATORS: operatorDefinition[] = [{
  value: '=',
}, {
  value: '!='
}, {
  value: '=|',
  description: 'Is one of. Use to filter on multiple values.',
  isMulti: true,
}, {
  value: '!=|',
  description: 'Is not one of. Use to exclude multiple values.',
  isMulti: true
}, {
  value: '=~',
  description: 'Matches regex',
}, {
  value: '!~',
  description: 'Does not match regex',
}, {
  value: '<'
}, {
  value: '>'
}];
```

dprokop (commented):
@ashharrison90 - really great progress on this. I've left couple of comments in the code, and here are some more UX/UI related things:

1. With multi value filter the filter is _applied_ as soon as you choose an individual value currently. This results in queries being performed as you build the filter. I think we need to align this behavior with multi value variable so that the queries are executed when the user selected all desired values. That would be on filter value input blur.
2. Removing individual value by clicking (x) on the pill results in the filter value popover being shown - i think we need to stop the events propagation.

@Sergej-Vlasov  - please keep an eyer and review this PR as we need to integrate this work with the work you are doing in https://github.com/grafana/scenes/pull/830

packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
ashharrison90: tbf i don't think this even really gets used at all, since the datasource looks for the presence of `values`... 

we could just set it to the first value in the `values` array? 🤔 

we could also make it the regex string. i guess that would mean we have to join the values with `|` but also regex escape the individual values

packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
```diff
@@ -65,12 +121,16 @@ export function AdHocFilterRenderer({ filter, model }: Props) {
       width="auto"
       value={valueValue}
       filterOption={filterNoOp}
+      hideSelectedOptions={false}
```
bfmatei, line 124: Should this be set to `false` only if `isMultiValue`?

ashharrison90: it doesn't actually seem to do anything in the single select case 😅 

but you're right, for clarity this should only apply to the multi value props

bfmatei (approved):
I tested this through https://github.com/grafana/grafana/pull/91837 and it seems to work fine

mdvictor (approved):
LGTM!

PR #870, opened 2024-08-14
MultiValueVariable: Fix issue where url update would not take options into account

mdvictor:
This PR fixes an issue where in a variable with key-value pairs([esc](https://github.com/grafana/support-escalations/issues/11759)), setting the key in the url param would not update to it's value but instead create a new custom option with the key as a value. 

Before:

https://github.com/user-attachments/assets/cf3aca6a-cfc6-4c7f-825d-7f1bf1fc6ff3

After:

https://github.com/user-attachments/assets/575498fe-a385-43c0-a293-bd1d180e5d17

### DIFF
--- a/packages/scenes/src/variables/variants/MultiValueVariable.test.ts
+++ b/packages/scenes/src/variables/variants/MultiValueVariable.test.ts
@@ -7,6 +7,7 @@ import { SceneVariableValueChangedEvent } from '../types';
 import { CustomAllValue } from '../variants/MultiValueVariable';
 import { TestVariable } from './TestVariable';
 import { subscribeToStateUpdates } from '../../../utils/test/utils';
+import { CustomVariable } from './CustomVariable';
 
 describe('MultiValueVariable', () => {
   describe('When validateAndUpdate is called', () => {
@@ -519,6 +520,34 @@ describe('MultiValueVariable', () => {
       });
 
       variable.urlSync?.updateFromUrl({ ['var-test']: '2' });
+      expect(variable.state.value).toEqual('2');
+      expect(variable.state.text).toEqual('B');
+    });
+
+    it('updateFromUrl should update value from label in the case of key/value custom variable', async () => {
+      const variable = new CustomVariable({
+        name: 'test',
+        options: [],
+        value: '',
+        text: '',
+        query: 'A : 1,B : 2',
+      });
+
+      await variable.urlSync?.updateFromUrl({ ['var-test']: 'B' });
+      expect(variable.state.value).toEqual('2');
+      expect(variable.state.text).toEqual('B');
+    });
+
+    it('updateFromUrl should update value from value in the case of key/value custom variable', async () => {
+      const variable = new CustomVariable({
+        name: 'test',
+        options: [],
+        value: '',
+        text: '',
+        query: 'A : 1,B : 2',
+      });
+
+      await variable.urlSync?.updateFromUrl({ ['var-test']: '2' });
       expect(variable.state.value).toEqual('2');
       expect(variable.state.text).toEqual('B');
     });
--- a/packages/scenes/src/variables/variants/MultiValueVariable.ts
+++ b/packages/scenes/src/variables/variants/MultiValueVariable.ts
@@ -1,5 +1,5 @@
 import { isArray, isEqual } from 'lodash';
-import { map, Observable } from 'rxjs';
+import { lastValueFrom, map, Observable } from 'rxjs';
 
 import { ALL_VARIABLE_TEXT, ALL_VARIABLE_VALUE } from '../constants';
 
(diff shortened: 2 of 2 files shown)


packages/scenes/src/variables/variants/MultiValueVariable.test.ts
```diff
@@ -523,6 +524,34 @@ describe('MultiValueVariable', () => {
...
+    it('updateFromUrl should update value from label in the case of key/value custom variable', async () => {
+      const variable = new CustomVariable({
+        name: 'test',
+        options: [],
+        value: '',
+        text: '',
+        query: 'A : 1,B : 2',
+      });
+
+      await variable.urlSync?.updateFromUrl({ ['var-test']: 'B' });
```
torkelo, line 536: url should not have the text / display value, it has the the value part I think

torkelo (commented):
I think this is likely a good solution, hm... wonder how it works in the old arch, maybe it works because URL sync happens so late, after variable has been fully initiated and options fetched/built

packages/scenes/src/variables/variants/MultiValueVariable.ts
```diff
@@ -419,6 +419,23 @@ export class MultiValueUrlSyncHandler<TState extends MultiValueVariableState = M
         this._sceneObject.skipNextValidation = true;
       }
 
+      if (!this._sceneObject.state.options || !this._sceneObject.state.options.length) {
+        const options = await lastValueFrom(this._sceneObject.getValueOptions({}));
```
torkelo, line 423: could this lead to double queries being issued for QueryVariables ?

packages/scenes/src/variables/variants/MultiValueVariable.test.ts
mdvictor: This is the scenario where the issue stemmed from. In the old arch you can set the key/text/display value and it will use it's correct value, but the URL shows the key. So we need something like this for backwards comp.

packages/scenes/src/variables/variants/MultiValueVariable.ts
mdvictor: Yes, it will :/ once here and once on activation itself

mdvictor:
@torkelo As far as I can tell, we also load the options before setting the value from the URL in the old arch [here](https://github.com/grafana/grafana/blob/main/public/app/features/variables/state/actions.ts#L396), [here](https://github.com/grafana/grafana/blob/main/public/app/features/variables/custom/adapter.ts#L31)

packages/scenes/src/variables/variants/MultiValueVariable.ts
torkelo: that will not do :(

torkelo: the variable will be validated later (when SceneVariableSet calls validateAndUpdate), can't we do this lookup at that point?

torkelo:
@mdvictor not sure I understand the issue, looked again, I thought it was related to setting the correct key in the select after a full page reload, b

but see now that it's about setting the variable to the display value via URL. Both the new and old variable system sets the variable value via url using the value not the text representation

torkelo:
wonder where/why they have urls with the text representation, grafana only generates URLs with the value

torkelo:
@mdvictor so the problem is this if statement here: https://github.com/grafana/scenes/blob/main/packages/scenes/src/variables/variants/MultiValueVariable.ts#L196

that code there makes sure you can add a custom value (enter anything) in the select, that value get's set in the URL and if you reload the page we need to preserve that custom value even if it's not among the options returned by the query (or specified in the variable csv string). 

We might be able to modify that logic a bit and ignore the if statement if current value === stateUpdate.text 

like this

```
if (
      this.skipNextValidation &&
      stateUpdate.value !== this.state.value &&
      stateUpdate.text !== this.state.text &&
      !isAllValueFix
    ) {
      stateUpdate.value = this.state.value;
      stateUpdate.text = this.state.text;
    }
```

mdvictor:
Closing this in favor of https://github.com/grafana/scenes/pull/874

mdvictor:
> wonder where/why they have urls with the text representation, grafana only generates URLs with the value

I don't know hey that happened either

> @mdvictor so the problem is this if statement here: https://github.com/grafana/scenes/blob/main/packages/scenes/src/variables/variants/MultiValueVariable.ts#L196
> 
> that code there makes sure you can add a custom value (enter anything) in the select, that value get's set in the URL and if you reload the page we need to preserve that custom value even if it's not among the options returned by the query (or specified in the variable csv string).
> 
> We might be able to modify that logic a bit and ignore the if statement if current value === stateUpdate.text
> 
> like this
> 
> ```
> if (
>       this.skipNextValidation &&
>       stateUpdate.value !== this.state.value &&
>       stateUpdate.text !== this.state.text &&
>       !isAllValueFix
>     ) {
>       stateUpdate.value = this.state.value;
>       stateUpdate.text = this.state.text;
>     }
> ```

Thank you! Indeed that fixes the problem, I've opened a new PR with these changes and closed this one

PR #872, opened 2024-08-14
SceneGridLayout: Change to useMeasure and non absolute div wrapper

torkelo:
In order to make the sticky controls work we cannot have an absolute wrapper around the grid (as this makes the div there the sticky lives not get the needed hight for sticky to work).

But in changing this to useMeasure and same CSS as in 
https://github.com/grafana/grafana/blob/main/public/app/features/dashboard/dashgrid/DashboardGrid.tsx#L327 

but I am running into the same problem I have always had in that when the wrapping div is not absolute I cannot reduce the size, the resize observer is not register any change to reduced viewport as the inner grid div current width is blocking a reduction in size. 

Have no idea why it works in DashboardGrid.tsx#L327


ashharrison90 (approved):
lgtm 👍

PR #874, opened 2024-08-19
MultiValueVariable: Fix issue where url update would not take options into account

mdvictor:
This PR fixes an issue where in a variable with key-value pairs([esc](https://github.com/grafana/support-escalations/issues/11759)), setting the key in the url param would not update to it's value but instead create a new custom option with the key as a value. 

Before:

https://github.com/user-attachments/assets/cf3aca6a-cfc6-4c7f-825d-7f1bf1fc6ff3

After:

https://github.com/user-attachments/assets/575498fe-a385-43c0-a293-bd1d180e5d17


torkelo (approved):
I think this is ok, there is a big downside to this and that is if one key (text) is the same as another's value then it will break after a full page reload. 

Example
A: 1
B: 2
C: A

can you just a quick test to see what happens in a situation like that? maybe the order in which we check things will save that situation

axelavargas (approved):
🏅

PR #877, opened 2024-08-28
Make SceneRefreshPicker respect config.minRefreshInterval

oscarkilhed:
https://github.com/grafana/support-escalations/issues/12157

The refresh picker in scenes does not respect the min_refresh_interval from the config.

This PR makes sure we filter the list of intervals to only include intervals longer than the min config. Also sets the interval to the minimum interval in the list if updated from the URL with a bad interval.

### DIFF
--- a/packages/scenes/src/components/SceneRefreshPicker.tsx
+++ b/packages/scenes/src/components/SceneRefreshPicker.tsx
@@ -11,6 +11,10 @@ import { SceneObjectUrlSyncConfig } from '../services/SceneObjectUrlSyncConfig';
 
 export const DEFAULT_INTERVALS = ['5s', '10s', '30s', '1m', '5m', '15m', '30m', '1h', '2h', '1d'];
 
+function filterDissalowedIntervals(i: string){
+  return config.minRefreshInterval ? rangeUtil.intervalToMs(i) >= rangeUtil.intervalToMs(config.minRefreshInterval) : true
+}
+
 export interface SceneRefreshPickerState extends SceneObjectState {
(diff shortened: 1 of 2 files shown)


torkelo (commented):
tricky wonder if this option should also impact scene apps vs just dashboards

oscarkilhed:
@torkelo Hmm that's a good point. There's a few options I guess. Either we enforce it for all apps and dashboards, or we default to the config value for apps and dashboards but let apps override it, or we let the dashboards and apps handle it independently. I think the safes would probably be option 2, to let apps override it but default to the config.

torkelo:
@oscarkilhed that is a good option, by default filter based on this setting (for both dashboards and scene apps) , but let scene apps override it

packages/scenes/src/components/SceneRefreshPicker.tsx
```diff
@@ -11,6 +11,10 @@ import { SceneObjectUrlSyncConfig } from '../services/SceneObjectUrlSyncConfig';
 
 export const DEFAULT_INTERVALS = ['5s', '10s', '30s', '1m', '5m', '15m', '30m', '1h', '2h', '1d'];
 
+function filterDissalowedIntervals(i: string){
```
bfmatei, line 14:
```suggestion
function filterDissalowedIntervals(i: string) {
```

Nit but don't forget to run prettier before merging.

oscarkilhed:
Updated to allow overriding the minRefreshInterval

torkelo (approved):

mdvictor (approved):

PR #886, opened 2024-09-02
SceneTimeRange: Support initialize time range with time and time.window

### LINKED ISSUE #883: SceneTimeRange: Support the window url parameter
In the old time range system we supported initiating the initial time range from a `time` and `time.window` parameters that would set the from and to values (then be ignored, best cleared). 

https://github.com/grafana/grafana/blob/main/public/app/features/dashboard/services/TimeSrv.ts#L159

Implemented originally in this PR 
https://github.com/grafana/grafana/pull/18636 

Need to re-implement this in SceneTimeRange

ivanortegaalba:
Fixes #883 

TODO: 
* Tests

### DIFF
--- a/packages/scenes/src/core/SceneTimeRange.tsx
+++ b/packages/scenes/src/core/SceneTimeRange.tsx
@@ -8,7 +8,7 @@ import { SceneTimeRangeLike, SceneTimeRangeState, SceneObjectUrlValues } from '.
 import { getClosest } from './sceneGraph/utils';
 import { parseUrlParam } from '../utils/parseUrlParam';
 import { evaluateTimeRange } from '../utils/evaluateTimeRange';
-import { config, RefreshEvent } from '@grafana/runtime';
+import { config, locationService, RefreshEvent } from '@grafana/runtime';
 
 export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> implements SceneTimeRangeLike {
   protected _urlSync = new SceneObjectUrlSyncConfig(this, { keys: ['from', 'to', 'timezone', 'time', 'time.window'] });
@@ -200,13 +200,15 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
   };
 
   public getUrlState() {
+    const params = locationService.getSearchObject();
     const urlValues: SceneObjectUrlValues = { from: this.state.from, to: this.state.to };
 
     if (this.state.timeZone) {
       urlValues.timezone = this.state.timeZone;
     }
 
-    if (urlValues.from && urlValues.to) {
+    // Clear time and time.window once they are converted to from and to
+    if (params.time && params['time.window']) {
       urlValues.time = null;
       urlValues['time.window'] = null;
     }
@@ -217,19 +219,15 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
   public updateFromUrl(values: SceneObjectUrlValues) {
     const update: Partial<SceneTimeRangeState> = {};
 
-    let from = values.from;
-    let to = values.to;
+    let from = parseUrlParam(values.from);
+    let to = parseUrlParam(values.to);
 
     if (values.time && values['time.window']) {
       const time = Array.isArray(values.time) ? values.time[0] : values.time;
       const timeWindow = Array.isArray(values['time.window']) ? values['time.window'][0] : values['time.window'];
       const timeRange = getTimeWindow(time, timeWindow);
-
       from = timeRange.from;
       to = timeRange.to;
-    } else {
-      from = parseUrlParam(values.from);
-      to = parseUrlParam(values.to);
     }
 
     if (!from && !to) {
@@ -266,11 +264,13 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
  * @param timeWindow - time window in ms or interval string
  */
 function getTimeWindow(time: string, timeWindow: string) {
-  const valueTime = parseInt(time, 10);
+  // Parse the time, assuming it could be an ISO string or a number in milliseconds
+  const valueTime = isNaN(Date.parse(time)) ? parseInt(time, 10) : Date.parse(time);
+
   let timeWindowMs;
 
   if (timeWindow.match(/^\d+$/) && parseInt(timeWindow, 10)) {
-    // when time window specified in ms
+    // when time window is specified in ms
     timeWindowMs = parseInt(timeWindow, 10);
   } else {
     timeWindowMs = rangeUtil.intervalToMs(timeWindow);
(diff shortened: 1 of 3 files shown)


packages/scenes/src/core/types.ts
```diff
@@ -151,18 +151,26 @@ export interface SceneTimeRangeState extends SceneObjectState {
...
    * */
   UNSAFE_nowDelay?: string;
+  /**
+   * Defines the point in time to define the window time-time.window/2 to time+time.window/2. Should be specified in ms. For example ?time=1500000000000&time.window=10000 results in a 10-second time range from 1499999995000 to 1500000005000`.
+   */
+  time?: string;
+  /**
+   * Defines the duration of the time range from time-time.window/2 to time+time.window/2. Should be specified in ms. For example ?time=1500000000000&time.window=10000 results in a 10-second time range from 1499999995000 to 1500000005000`.
+   */
+  timeWindow?: string;
```
torkelo, line 162: we do not need these here, these are just URL params used to init the initial time we do not need them in state

packages/scenes/src/core/SceneTimeRange.tsx
```diff
@@ -11,12 +11,19 @@ import { evaluateTimeRange } from '../utils/evaluateTimeRange';
...
-    const to = state.to ?? 'now';
+    let from = state.from ?? 'now-6h';
+    let to = state.to ?? 'now';
     const timeZone = state.timeZone;
+
+    if (state.time && state.timeWindow) {
+      const timeWindow = getTimeWindow(state.time, state.timeWindow);
+      from = timeWindow.from.toISOString();
+      to = timeWindow.to.toISOString();
+    }
```
torkelo, line 25: do not think we need to handle this here, as time & time.window are just initial init values

packages/scenes/src/core/SceneTimeRange.tsx
```diff
@@ -200,28 +207,55 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
...
+      urlValues.time = this.state.time;
+      urlValues['time.window'] = this.state.timeWindow;
+
+      // Remove from and to if time and time window are set
+      urlValues.from = null;
+      urlValues.to = null;
+    } else {
+      urlValues.from = this.state.from;
+      urlValues.to = this.state.to;
+    }
```
torkelo, line 225: not sure this is needed

packages/scenes/src/core/SceneTimeRange.tsx
```diff
@@ -200,28 +207,55 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
...
+    let from = values.from;
+    let to = values.to;
+
+    if (values.time && values['time.window']) {
+      const time = Array.isArray(values.time) ? values.time[0] : values.time;
+      const timeWindow = Array.isArray(values['time.window']) ? values['time.window'][0] : values['time.window'];
+      const timeRange = getTimeWindow(time, timeWindow);
+
+      from = timeRange.from.toISOString();
+      to = timeRange.to.toISOString();
```
torkelo, line 242: this is the only part we need I think

ivanortegaalba:
@torkelo actually this was a question, if we wanted to support this for the object or only from URL. I see your answer now. Thank you! I'll make some adjustments.

torkelo:
@ivanortegaalba I think it mainly makes sense from URL, at least as far as I understand the feature but it's been a while since it was introduced :)

ivanortegaalba:
@torkelo I found a limitation when trying to implement the desired behavior.

If I understood correctly, the desired behavior would be:
1. Given the following URL `http://localhost:3000/d/edgxrth3yeo74f/change-plugin-type?time=1500000000000&time.window=10000`
2. The dashboard calculates the `from` and `to` for the state, but doesn't change the URL or clean them if they are present too
3. When the time range changes because of user interaction those parameters are gone

I found it hard to do this behavior because [we don't receive the current URL values in `getUrlState`](https://github.com/grafana/scenes/pull/886/files#diff-69e780754ebc20ef38ccfb7677e152440327a22b4cf115ccb6f83d9cf1eeb830R202).

This PR implements other ways to do it:
1. Given the following URL `http://localhost:3000/d/edgxrth3yeo74f/change-plugin-type?time=1500000000000&time.window=10000`
2. `from` and `to` are calculated and used in the URL, removing `time` and `time.window`

Do you have any tip how can I proceed?

packages/scenes/src/core/SceneTimeRange.tsx
```diff
@@ -201,27 +201,45 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
...
-    if (!values.to && !values.from) {
-      return;
+    const update: Partial<SceneTimeRangeState> = {};
+
+    let from = values.from;
+    let to = values.to;
+
+    if (values.time && values['time.window']) {
+      const time = Array.isArray(values.time) ? values.time[0] : values.time;
+      const timeWindow = Array.isArray(values['time.window']) ? values['time.window'][0] : values['time.window'];
```
torkelo, line 225: you don't need to check for isArray, just check if they are a string

torkelo:
@ivanortegaalba 
> Given the following URL http://localhost:3000/d/edgxrth3yeo74f/change-plugin-type?time=1500000000000&time.window=10000
> from and to are calculated and used in the URL, removing time and time.window

I think this is sounds good, and is what I would have implemented. What do you think? Is there a problem with that approach?

packages/scenes/src/core/SceneTimeRange.tsx
```diff
@@ -241,3 +259,25 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
...
   }
 }
+
+/**
+ * Calculates the duration of the time range from time-time.window/2 to time+time.window/2. Both be specified in ms. For example ?time=1500000000000&time.window=10000 results in a 10-second time range from 1499999995000 to 1500000005000`.
+ * @param time - time in ms
+ * @param timeWindow - time window in ms or interval string
+ */
+function getTimeWindow(time: string, timeWindow: string) {
+  const valueTime = parseInt(time, 10);
```
torkelo, line 269: can we make this support ISO string times as well (so it's more consistent with the format of from/to which support both epoc and iso strings)

ivanortegaalba:
> @ivanortegaalba
> 
> > Given the following URL http://localhost:3000/d/edgxrth3yeo74f/change-plugin-type?time=1500000000000&time.window=10000
> > from and to are calculated and used in the URL, removing time and time.window
> 
> I think this is sounds good, and is what I would have implemented. What do you think? Is there a problem with that approach?

No, this is the trivial one :)

packages/scenes/src/core/SceneTimeRange.tsx
ivanortegaalba: What if we have more than one element in the URL? TS tells you there is the possibility to have more than one. In case it is a string, we will do nothing then?

packages/scenes/src/core/SceneTimeRange.tsx
ivanortegaalba: Done!

torkelo (approved):

PR #889, opened 2024-09-05
AdHocFilters - multi value support in new filters UI

Sergej-Vlasov:
Code is rough, but creating a PR so that we can start the review while I do code clean up

Introduce the multi value ad hoc filters support in the new filters UI

<details>
  <summary>📦 Published PR as canary version: <code>5.14.2--canary.889.10771581250.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.14.2--canary.889.10771581250.0
  npm install @grafana/scenes@5.14.2--canary.889.10771581250.0
  # or 
  yarn add @grafana/scenes-react@5.14.2--canary.889.10771581250.0
  yarn add @grafana/scenes@5.14.2--canary.889.10771581250.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
@@ -18,7 +18,8 @@ export function AdHocFilterPill({ filter, model, readOnly }: Props) {
   const pillWrapperRef = useRef<HTMLDivElement>(null);
 
   const keyLabel = filter.keyLabel ?? filter.key;
-  const valueLabel = filter.valueLabels?.[0] ?? filter.value;
+  //@ts-expect-error
+  const valueLabel = filter.valueLabels?.join(', ') || filter.values?.join(', ') || filter.value;
 
   const handleChangeViewMode = useCallback(
     (event?: React.MouseEvent) => {
@@ -55,7 +56,7 @@ export function AdHocFilterPill({ filter, model, readOnly }: Props) {
         tabIndex={0}
         ref={pillWrapperRef}
       >
-        <span>
+        <span className={styles.pillText}>
           {keyLabel} {filter.operator} {valueLabel}
         </span>
         {!readOnly ? (
@@ -117,5 +118,8 @@ const getStyles = (theme: GrafanaTheme2) => ({
     '&:hover': {
       color: theme.colors.text.primary,
     },
+  }),
+  pillText: css({
+    whiteSpace: 'break-spaces',
   }),
 });
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
@@ -44,6 +44,11 @@ export function AdHocFilterPill({ filter, model, readOnly, focusOnInputRef }: Pr
   }, [shouldFocus]);
 
   if (viewMode) {
+    const pillText = (
+      <span className={styles.pillText}>
+        {keyLabel} {filter.operator} {valueLabel}
+      </span>
+    );
     return (
       <div
         className={cx(styles.combinedFilterPill, { [styles.readOnlyCombinedFilter]: readOnly })}
@@ -58,15 +63,14 @@ export function AdHocFilterPill({ filter, model, readOnly, focusOnInputRef }: Pr
         tabIndex={0}
         ref={pillWrapperRef}
       >
-        <Tooltip
-          show={valueLabel.length < 20 ? false : undefined}
-          content={<div className={styles.tooltipText}>{valueLabel}</div>}
-          placement="top"
-        >
-          <span className={styles.pillText}>
-            {keyLabel} {filter.operator} {valueLabel}
-          </span>
-        </Tooltip>
+        {valueLabel.length < 20 ? (
+          pillText
+        ) : (
+          <Tooltip content={<div className={styles.tooltipText}>{valueLabel}</div>} placement="top">
+            {pillText}
+          </Tooltip>
+        )}
+
         {!readOnly ? (
           <IconButton
             onClick={(e) => {
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
@@ -18,6 +18,7 @@ export function AdHocFilterPill({ filter, model, readOnly }: Props) {
   const pillWrapperRef = useRef<HTMLDivElement>(null);
 
   const keyLabel = filter.keyLabel ?? filter.key;
+  // TODO remove when we're on the latest version of @grafana/data
   //@ts-expect-error
   const valueLabel = filter.valueLabels?.join(', ') || filter.values?.join(', ') || filter.value;
 
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
@@ -1,6 +1,6 @@
 import { css, cx } from '@emotion/css';
 import { GrafanaTheme2 } from '@grafana/data';
-import { useStyles2, IconButton } from '@grafana/ui';
+import { useStyles2, IconButton, Tooltip } from '@grafana/ui';
 import React, { useState, useRef, useCallback, useEffect } from 'react';
 import { AdHocCombobox } from './AdHocFiltersCombobox';
 import { AdHocFilterWithLabels, AdHocFiltersVariable } from '../AdHocFiltersVariable';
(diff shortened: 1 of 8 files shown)


packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersCombobox.tsx
```diff
@@ -321,111 +480,169 @@ export const AdHocCombobox = forwardRef(function AdHocCombobox(
...
+            handleRemoveMultiValue(item);
+          }
+        }}
+        fill="text"
+        size="sm"
+        variant="secondary"
+        className={styles.removeButton}
+        tooltip={`Remove filter value - ${item.label ?? item.value}`}
+      >
+        <Icon name="times" size="md" id={`${item.value}-${index}`} />
```
Sergej-Vlasov, line 631: I could not use `IconButton` because it would not pass an `id` to svg (which is needed to prevent outside click)

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/utils.ts
```diff
@@ -86,6 +87,9 @@ export const setupDropdownAccessibility = (
       disabledIndices.push(i);
     }
     let label = options[i].label ?? options[i].value ?? '';
+    if (label.length < (options[i].description?.length || 0)) {
+      label = options[i].description!;
+    }
```
ashharrison90, line 92: ah, i guess this is using the description to calculate the width of the dropdown? think this might be overcalculating the width slightly since the description uses a smaller font size.

again, definitely don't need to fix this immediately. just surfacing it 👍

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/utils.ts
```diff
@@ -8,6 +8,7 @@ const VIRTUAL_LIST_WIDTH_ESTIMATE_MULTIPLIER = 8;
 const VIRTUAL_LIST_PADDING = 8;
 export const VIRTUAL_LIST_OVERSCAN = 5;
 export const VIRTUAL_LIST_ITEM_HEIGHT = 38;
+export const VIRTUAL_LIST_ITEM_HEIGHT_WITH_DESCRIPTION = 60;
```
ashharrison90, line 11: curious question - what happens if the description is very long? does it just stay on one line and truncate, or does it wrap and break this magic number? 🤔 i don't believe we have this problem currently, just curious 😄

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/useFloatingInteractions.ts
```diff
@@ -56,8 +56,13 @@ export const useFloatingInteractions = ({
...
     outsidePress: (event) => {
-      // if outside click lands on operator pill, then ignore outside click
-      if (event.currentTarget instanceof HTMLElement && event.currentTarget.id === operatorIdentifier) {
+      const target = (event.currentTarget || event.target) as HTMLElement | null;
+      let idToCompare = target?.id || '';
+      if (target?.nodeName === 'path') {
+        idToCompare = target.parentElement?.id || '';
+      }
+
+      if (outsidePressIdsToIgnore.includes(idToCompare)) {
```
ashharrison90, line 65: do we need `event.target` here? 🤔 

if you're wanting to catch `SVGElement`s as well, think we could probably do something like:
```tsx
if (event.currentTarget instanceof Element) {
  ...
}
```

ashharrison90 (commented):
first pass, this looks great! 🙌 

- there's a bug where if you click inside the actual checkbox it won't select/deselect the option

nits that i don't think we need to fix immediately:
- popover moving with the cursor on selection/deselection of multiple items is a bit annoying if you're using a mouse
- when removing a pill with the keyboard, focus reverts back to body. this happens with both single and multi value pills.

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/utils.ts
Sergej-Vlasov: adjusted calculation to take into account smaller description font

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/utils.ts
Sergej-Vlasov: currently if description is very long it will expand the dropdown width until it hit available width and then becomes scrollable. We could come up with magical number for maxWidth if we would want that.

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/useFloatingInteractions.ts
Sergej-Vlasov: probably not really, just defaulting to it as a fallback. 
Also `instanceof Element` does not work here unfortunately because instances can be `HTMLElement`, `SVGElement`, `SVGPathElement` and those type are not really important, the important bit is `id` and we can just type assert to `HTMLElement` to not complicate things

Sergej-Vlasov:
> there's a bug where if you click inside the actual checkbox it won't select/deselect the option

adjusted so that clicking checkbox now functional

> when removing a pill with the keyboard, focus reverts back to body. this happens with both single and multi value pills.

it did not focus on body, but instead would just loose focus and clicking `Tab` would refocus on input. But still added some logic to conveniently focus on input when removing multi value or whole filter 

> popover moving with the cursor on selection/deselection of multiple items is a bit annoying if you're using a mouse

this one is quite tricky, I tried passing ref to the multi value pill wrapper in order to keep position, but the `apply` middleware that we pass in `useFloating` is memoized and does not get updated values to that of that wrapper. Therefore I tried to overwrite `transform` value on the dropdown itself and it kinda works but quite wonky I would say. Looking for second opinions on the dropdown fixation in multi value edit mode

dprokop:
Awesome work! Initial UI/UX feedback from testing:

1. Spacing of the "chip" changes when starting to edit a filter:

https://github.com/user-attachments/assets/845712f3-a32b-477c-9ec8-c5bab859fded

2. Label is visible twice:

Screenshot 2024-09-06 at 10 54 33

3. Can we align the design of the Apply button with the design of filters in the table panel:

image

dprokop:
Also, can end up in this a bit awkward state when editing filter in a middle:

image

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
```diff
@@ -55,20 +58,28 @@ export function AdHocFilterPill({ filter, model, readOnly }: Props) {
...
+          <span className={styles.pillText}>
+            {keyLabel} {filter.operator} {valueLabel}
+          </span>
+        </Tooltip>
         {!readOnly ? (
           <IconButton
             onClick={(e) => {
               e.stopPropagation();
               model._removeFilter(filter);
+              setTimeout(() => focusOnInputRef?.());
```
dprokop, line 75: Shouldn't we use `requestAnimationFrame` instead here?

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
```diff
@@ -55,20 +58,28 @@ export function AdHocFilterPill({ filter, model, readOnly }: Props) {
...
             }}
             onKeyDownCapture={(e) => {
               if (e.key === 'Enter') {
                 e.preventDefault();
                 e.stopPropagation();
                 model._removeFilter(filter);
+                setTimeout(() => focusOnInputRef?.());
```
dprokop, line 82: Shouldn't we use `requestAnimationFrame` instead here?

dprokop (approved):
Other than my nit comments about code:
1. When selecting from long list of options, the list scrolls back up when value is selected:

https://github.com/user-attachments/assets/670d49ce-b3e9-4bb0-b275-7b5a99c62a4e


2. The value placeholder is unnecessary and confusing in my opinion, especially visible if you edit an individual filter. I think the input should be either narrower or autosized:

image


I think we can address those issues independently as a followup PR.

ashharrison90 (approved):
lgtm 👍 agree with dominik, can handle nits in follow up PRs

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/useFloatingInteractions.ts
ashharrison90: `HTMLElement`/`SVGElement` etc all extend from `Element`, so they will all work fine. agreed, `id` is the important bit which is why i think you want to check that `currentTarget` is `instanceof Element`, since that's the interface that defines the `id` property.

remember, any time you type assert you're lying to yourself (and the compiler) about what that element is. we should really really avoid doing this 😅 have pushed update 👍

axelavargas (approved):
I also gave it a quick local test and found minors a11y issue that can be addressed in a different PR. Overall it looks good :) 

### Apply Button is not reachable via keyboard:

https://github.com/user-attachments/assets/5912a04a-bf01-4ab5-9c69-86735e34e1d5

### The focus ring is not correct when editing values 

image

PR #896, opened 2024-09-11
MultiValueVariabe: Change when value changed event is published

mdvictor:
This PR changes then the `SceneVariableValueChangedEvent` is published. It should only publish, in the case of `All` value, only when options actually change. 

This will help us in various scenarios that are currently giving us headaches like [here](https://github.com/grafana/grafana/pull/93168) where the variable potentially `hasChanged`, after a dashboard refresh, for example, but in truth there is no change in the options, and that var is set only because it sees the `All` value selected.

Another example is in the case of variables that depend on themselves, case that is supported in the new arch due to backwards compat with the old arch. If the value is `All` then the variable is constantly considered as changed and thus publishes the event which will trigger an infinite query in the variable.

In both scenarios above the issues happen only with the `All` value.

<details>
  <summary>📦 Published PR as canary version: <code>5.14.4--canary.896.10848324548.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.14.4--canary.896.10848324548.0
  npm install @grafana/scenes@5.14.4--canary.896.10848324548.0
  # or 
  yarn add @grafana/scenes-react@5.14.4--canary.896.10848324548.0
  yarn add @grafana/scenes@5.14.4--canary.896.10848324548.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/variants/MultiValueVariable.test.ts
+++ b/packages/scenes/src/variables/variants/MultiValueVariable.test.ts
@@ -231,7 +231,33 @@ describe('MultiValueVariable', () => {
       expect(variable.state.text).toEqual([ALL_VARIABLE_TEXT]);
     });
 
-    it('Should handle $__all value and send change event even when value is still $__all', async () => {
+    it('Should handle $__all value and send change event when value is still $__all, but options change', async () => {
+      const variable = new TestVariable({
+        name: 'test',
+        options: [],
+        optionsToReturn: [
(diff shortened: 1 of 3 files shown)


ivanortegaalba (approved):
I couldn't find any issue with this. I'd like for @torkelo to have a quick look to see if any behavior could break if we don't publish that event when the options don't change.

Looks great on my side!

torkelo (approved):

PR #899, opened 2024-09-12
Use locationService from context in UrlSyncManager

aocenas:
Fixes: https://github.com/grafana/grafana/issues/93225
This changes the UrlSyncManager to use the `locationsService` from context making it possible to render 2 scenes app without them clashing on URL handling, which is needed for the sidecar project https://github.com/grafana/grafana/pull/91648.

See this draft PR for the Explore Traces app: https://github.com/grafana/explore-traces/pull/102 for usage.

### DIFF
--- a/packages/scenes/src/services/useUrlSync.ts
+++ b/packages/scenes/src/services/useUrlSync.ts
@@ -1,13 +1,14 @@
 import { SceneObject, SceneUrlSyncOptions } from '../core/types';
 import { useEffect, useState } from 'react';
 import { useLocation } from 'react-router-dom';
+// @ts-ignore
 import { useLocationService } from '@grafana/runtime';
 import { writeSceneLog } from '../utils/writeSceneLog';
 import { useUrlSyncManager } from './UrlSyncManager';
 
 export function useUrlSync(sceneRoot: SceneObject, options: SceneUrlSyncOptions = {}): boolean {
   const location = useLocation();
-  const locationService = useLocationService()
+  const locationService = useLocationService();
   const [isInitialized, setIsInitialized] = useState(false);
   const urlSyncManager = useUrlSyncManager(options, locationService);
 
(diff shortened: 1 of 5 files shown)


packages/scenes/src/services/useUrlSync.ts
```diff
@@ -1,14 +1,16 @@
 import { SceneObject, SceneUrlSyncOptions } from '../core/types';
 import { useEffect, useState } from 'react';
 import { useLocation } from 'react-router-dom';
-import { locationService } from '@grafana/runtime';
+// @ts-ignore
+import { useLocationService } from '@grafana/runtime';
```
aocenas, line 5: From out previous conversation https://github.com/grafana/scenes/pull/841#discussion_r1689710633 I am not sure I get this right.

You mention using ts-ignore to prevent ts complaining while keeping the old version of `runtime` as a dependency but I assume that still means basically doing runtime checks to see if the export is there right?

packages/scenes/src/services/useUrlSync.ts
```diff
@@ -1,14 +1,19 @@
...
+import { locationService as locationServiceRuntime, useLocationService } from '@grafana/runtime';
 import { writeSceneLog } from '../utils/writeSceneLog';
 import { useUrlSyncManager } from './UrlSyncManager';
 
 export function useUrlSync(sceneRoot: SceneObject, options: SceneUrlSyncOptions = {}): boolean {
   const location = useLocation();
+  // As we this is basically a version/feature check for grafana/runtime this 'if' should be stable (ie for one instance
+  // of grafana this will always be true or false) so it should be safe to ignore the hook rule here
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const locationService = useLocationService ? useLocationService() : locationServiceRuntime;
```
torkelo, line 14: maybe we can have a util function that does this (an scenes internal useLocationServiceWrapper) so we can share this messy temporary hack in one place

aocenas: Good point moved it.

torkelo (approved):

dprokop (approved):
Looks good!

PR #902, opened 2024-09-13
VizPanel: Harden plugin change handling

torkelo:
This is not needed for https://github.com/grafana/grafana/pull/93045 as I was able to work around it. 

But felt that VizPanel should handle pluginId changes a bit better, so making this change for that reason.

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanel.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.tsx
@@ -120,12 +120,17 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
   }
 
   private _onActivate() {
-    if (!this._plugin) {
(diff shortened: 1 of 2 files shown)


ivanortegaalba (approved):
The code looks good to me! What issues were causing the not detecting plugin changes on activation?

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -120,12 +120,17 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
   }
 
   private _onActivate() {
-    if (!this._plugin) {
+    if (!this._plugin || this.state.pluginId !== this._plugin.meta.id) {
```
torkelo, line 123: @ivanortegaalba this code did not check if the plugin id had changed

Issue #457, opened 2023-11-14
[BUG] - Date picker in Observability is switching to UTC when changing the date, instead of keeping the selected time zone


marinnedea:
Under Observability I can set to Sweden UTC+1
![image](https://github.com/grafana/scenes/assets/7581631/388bec86-1328-4925-8c26-001d6ca0c513)
![image](https://github.com/grafana/scenes/assets/7581631/0837abd2-5be1-422a-9dd4-22d02c5b74a0)

But when I try changing to Nov 2
![image](https://github.com/grafana/scenes/assets/7581631/e2066add-468d-48eb-9da7-ad4d27ac363b)

it changes the timezone back to UTC
![image](https://github.com/grafana/scenes/assets/7581631/6f0a9ca5-5c27-42c9-b997-6e7a9de85035)

dprokop:
@domasx2  - this should be fixed with https://github.com/grafana/scenes/pull/420 - is App Observability on the latest scenes?

marinnedea:
@domasx2 tested this in Grafana Play: https://play.grafana.org/a/grafana-k8s-app/navigation/cluster?promName=grafanacloud-play-prom&from=2023-10-31T21:00:00.000Z&to=2023-11-01T20:59:59.000Z - I guess this one is not up to date. 

I'm not able to replicate on my own instance ( Grafana v10.3.0-63317 (e9e7e0b628bb7ac5f657c630375279b7df30826f)

domasx2:
App o11y is on `1.19.0`. Haven't updated yet, currently on freeze for ObsCon.  But this is k8s app :)

marinnedea:
Closing this as it is definitely fixed in 10.3.x . Thank you!

PR #907, opened 2024-09-19
SceneTimeRange: Fixes inconsistent representation of time range

torkelo:
Not sure I fully understand this change (https://github.com/grafana/scenes/pull/420/ ) so this PR is reverting it.

Opening revert PR just to remember to have a discussion on this on next scenes sync. 

The original PR was fixing 
https://github.com/grafana/scenes/issues/457 

but after investing a similar issue (https://github.com/grafana/support-escalations/issues/12532) we have right now in main I think the problem is not in scenes but in the TimePicker and how it handle string raw values (it does handle them as UTC iso strings but just as plain strings) 

In my view the bug in https://github.com/grafana/scenes/issues/457  was not that the timezone was lost but the fact that we showed an iso string in those inputs to begin with. Just try setting absolute time range using the calendar dates and the time range is formatted normally, I think the same should be true when selecting an absolute time range via graph. We should not show an iso string there (with or without timezone offset).

A problem with this fix is also how it creates inconsistency, the TimeRange raw (from/to) becomes formatted differently after a call to onTimeRangeChange vs initial value evaluated in constructor or via updateFromUrl 

We have instead two ways to fix this, update TimeRangePicker so that it handles ISO date strings in from/to, PR that tries to to do that https://github.com/grafana/grafana/pull/93479 

another possible fix is to change in scenes how we store raw from/to (go back to let raw/from to be moment instances for absolute ranges)

### DIFF
--- a/packages/scenes/src/core/SceneTimeRange.tsx
+++ b/packages/scenes/src/core/SceneTimeRange.tsx
@@ -1,4 +1,4 @@
-import { getTimeZone, rangeUtil, setWeekStart, TimeRange, toUtc } from '@grafana/data';
+import { getTimeZone, rangeUtil, setWeekStart, TimeRange } from '@grafana/data';
 import { TimeZone } from '@grafana/schema';
 
 import { SceneObjectUrlSyncConfig } from '../services/SceneObjectUrlSyncConfig';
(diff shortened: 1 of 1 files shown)


kaydelaney:
Tested out locally and it seems to work, though tests are failing

torkelo:
@kaydelaney had accidentially based the PR on an older version of SceneTimeRange (must have had a modified SceneTimeRange file and overwritten some newer changes). Restored the file from main and redid the changes (so much smaller diff now), and tests are green

kaydelaney (approved):

PR #910, opened 2024-09-23
SceneTimeRange: Use new weekStart prop on TimeRangePicker

torkelo:
Apparently, the setWeekStart runtime function has stopped working some time ago (maybe due a browser update), not sure. 

Frontend platform added a weekStart prop to TimeRangePicker recently (but did not notify us about this) 

This updates SceneTimePicker to pass the weekStart state prop to TimeRangePicker 

I also have a a bit related PR that fixes weekStart for all scene app plugins (so TimeRangePicker uses the system or user weekStart by default) 

https://github.com/grafana/grafana/pull/93464


axelavargas (approved):
The code looks good to me 🥇

PR #912, opened 2024-09-24
VizPanelBuilder: Allow limiting number of frames initially rendered

gtk-grafana:
The goal of this PR is to bring the UI we show in explore when there are more then N frames being rendered in a panel. This was implemented in Explore Logs, but I think this is something that all Scene apps could benefit from to prevent breakdown scenes with many timeseries from eating up browser resources. 

This PR:
image

image

Hover header:
image

Current UI in explore:
image

Example usage:

```
PanelBuilders.timeseries()
.setData($data)
.setTitleItems([new LimitFramesTitleItemScene({
   frameLimit: 20, // limits the default number of frames that are rendered in the viz
 })])
```
If the LimitFramesTitleItemScene titleItem is not defined the panels will (attempt to) render everything.

Future work:
* I mostly copied the UI from the existing feature in Explore, but it still uses quite a bit of panel header real estate, would be good to iterate on the UI in a future PR and find something a bit cleaner.

<details>
  <summary>📦 Published PR as canary version: <code>5.27.0--canary.912.12031759221.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.27.0--canary.912.12031759221.0
  npm install @grafana/scenes@5.27.0--canary.912.12031759221.0
  # or 
  yarn add @grafana/scenes-react@5.27.0--canary.912.12031759221.0
  yarn add @grafana/scenes@5.27.0--canary.912.12031759221.0
  ```
</details>


packages/scenes/src/components/LimitSeriesTitleItem.tsx
```diff
@@ -0,0 +1,146 @@
...
+    this.addActivationHandler(this.onActivate.bind(this));
+  }
+
+  private onActivate() {
+    const panel = sceneGraph.getAncestor(this, VizPanel);
+    const $data = panel.state.$data;
+
+    // @todo how to fix scene object already has parent
+    // Cloning breaks query subscriptions
+    const transformer = new SceneDataTransformer({
```
dprokop, line 35: @gtk-grafana - scenes now have `clearParent` method on scene object that may be useful for this problem, ref https://github.com/grafana/scenes/pull/892. 

I wonder about `Cloning breaks query subscriptions` - how do you observe this? Maybe we have some bug in the `SceneQueryRunner` cloning.

gtk-grafana: Ok, I was able to narrow down the problem a bit, if I update this PR to clone $data, I can get the desired functionality in the consuming application, but I have to do some hacky subscription nesting whenever the key of the data provider changes, i.e.

```
    this._subs.add(
        panel.subscribeToState((newState, prevState) => {
          if(newState.$data?.state.key !== prevState.$data?.state.key){
            this._subs.add(
                panel?.state.$data?.getResultsStream().subscribe((result) => {
                  if (result.data.errors && result.data.errors.length > 0) {
                    // DO WORK ON ERROR
                  }
                })
            )
          }
        })
      )
```

as opposed to before, I would just call getResultsStream on the panel $data, i.e.
```
    this._subs.add(
        panel.state.$data?.getResultsStream().subscribe((result) => {
          if (result.data.errors && result.data.errors.length > 0) {
           // DO WORK ON ERROR
          }
        })
      );
```

It looks like when setting the $data state of the panel, any existing subscriptions to the old data provider are cancelled and cleaned up.

When we don't clone $data, we get the invalid parent warning, but since we're not instantiating a new data provider the existing subscriptions are not cancelled.

Not sure if this is a cloning bug, or just a problem with setting the $data state of a panel?

gtk-grafana: Although adding `clone['_results'] = this._results` to the SceneQueryRunner clone method seems to pass along the subscriptions to the cloned object, which seems to fix the original `getResultsStream` implementation, but I'm not sure about the implications of that change at this time.

gtk-grafana: @dprokop I went ahead and made the update to the clone method, almost everything works as expected, but it looks it creates a duplicate query that is immediately cancelled when the data transformer is set to the panel state. Looks like this is a consequence of setting the $data state of the viz though and not the updates to the cloning method.

gtk-grafana: So, finally getting some time to come back to this, looks like my above idea of cloning the data provider is never going to work within scenes without cancelling a query, as the LimitSeriesTitleItem scene is a child of the VizPanel, which will activate its data provider before activation of the LimitSeriesTitleItem scene. 

TL;DR there is no† way for a child of a query provider to set the $data state without interrupting an ongoing query.
† Caveat being if the query runner runQueriesMode is manual, but that's not much of a solution for this problem.

Paths forward: 
1. export a transformation helper (like the `limitFramesTransformation`) that the implementer must add as a parent to the query provider as a `SceneDataTransformer`. This means the caller must add the title item component and add a transformation to their data provider, and we'll need the same `seriesLimit` number passed into both places.
Not a good solution for downstream developers.
2. (???) New prop on query runners that can find the parent VizPanel and automatically add the UI to the title items. -  I feel like query runners setting VizPanel UI is the wrong direction. Not going to spend time on this.
3. (???) Do the work in VizPanelBuilder, if a `TimeSeriesLimitSeriesTitleItemScene` exists in the titleItems (or maybe we need to create a new VizPanel state prop), we wrap the existing dataProvider in our transformer. - Will shift this PoC to attempt to flush this out.

gtk-grafana: 3 seems to work great 🎉

packages/scenes/src/core/sceneGraph/sceneGraph.ts
```diff
@@ -193,6 +193,15 @@ export function findAllObjects(scene: SceneObject, check: (obj: SceneObject) =>
   return found;
 }
 
+export function findDescendent<T extends SceneObject>(scene: SceneObject, descendentType: SceneType<T>) {
```
gtk-grafana, line 196: Not really necessary, but slightly more performant when only looking for a single descendent

torkelo (commented):
Thanks for contributing this! think this could be super useful to in so many scene apps and even in dashboards. 

think it's pretty close, some issues.

packages/scenes/src/core/PanelBuilders/VizPanelBuilder.ts
```diff
@@ -235,4 +244,39 @@ export class VizPanelBuilder<TOptions extends {}, TFieldConfig extends {}>
 
     return panel;
   }
+
+  /**
+   * If titleItems contains TimeSeriesLimitSeriesTitleItemScene, replace queryProvider with SceneDataTransformer and limit the number of series initially rendered
+   */
+  private limitPanelSeries() {
```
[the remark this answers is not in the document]
gtk-grafana: Good catch, I renamed things to align a bit better

packages/scenes/src/core/PanelBuilders/VizPanelBuilder.ts
```diff
@@ -235,4 +244,39 @@ export class VizPanelBuilder<TOptions extends {}, TFieldConfig extends {}>
...
+    if(this._state.$data){
+      const limitSeriesTitleItem = this.getLimitSeriesTitleItem()
+
+      if (limitSeriesTitleItem && limitSeriesTitleItem.state.seriesLimit) {
+        this.setData(new SceneDataTransformer({
+          $data: this._state.$data,
+          transformations: [() => limitFramesTransformation(limitSeriesTitleItem.state.seriesLimit)],
+        }))
+      }
```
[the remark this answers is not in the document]
gtk-grafana: I pushed an update which will grab the transformation by key (instead of grabbing the top-level SceneDataTransformer), and added/updated the test cases in LimitFramesTitleItemScene.test.tsx to make sure nested SceneDataTransformers still work as expected.

packages/scenes/src/components/LimitSeriesTitleItem.tsx
```diff
@@ -0,0 +1,128 @@
...
+  public constructor(state: TimeSeriesLimitSeriesTitleItemSceneState) {
+    super(state);
+    this.addActivationHandler(this.onActivate.bind(this));
+  }
+
+  private onActivate() {
+    const panel = sceneGraph.getAncestor(this, VizPanel);
+
+    const $transformedData = sceneGraph.getData(panel);
+    const untransformedQueryRunner = sceneGraph.findDescendent(panel, SceneQueryRunner)
```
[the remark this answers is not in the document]
gtk-grafana: I pushed an update here to grab the first SceneDataTransformer child, and if undefined we grab the first SceneQueryRunner child. If there's a better way to match the `SceneDataProvider` interface please let me know and I'll get this updated.

gtk-grafana:
After the latest commit, `findDescendent` is no longer used in this PR, although it is still being used in the tests. We could remove that sceneGraph function in this PR if reviewers think that is out of scope, or unnecessary in this PR.

packages/scenes/src/core/PanelBuilders/VizPanelBuilder.ts
```diff
@@ -227,6 +238,8 @@ export class VizPanelBuilder<TOptions extends {}, TFieldConfig extends {}>
    * Build the panel.
    */
   public build() {
+    this.limitFrames();
```
torkelo, line 241: think this should be opt-in by default (at least to begin with)

gtk-grafana: Agreed and it already is, but the limitFrames function name doesn't give much affordance to its behavior. I renamed it to `limitFramesIfLimitSceneExists`, which still isn't great, but perhaps better.

torkelo:
@gtk-grafana I really like this feature, much needed. While this solution works it feels unnecessarily complex (blaming scenes arch for here) , Tried to explore a different solution to this feature in this PR https://github.com/grafana/scenes/pull/978

PR #918, opened 2024-09-26
Combobox: Fix "Apply" button positioning when scrolling

ashharrison90:
- fix "Apply" button positioning when scrolling
- believe this was caused by the enablement of `bodyScrolling` - need to use a "fixed" positioning strategy instead now

# before 

https://github.com/user-attachments/assets/6452ea78-5601-4e5e-9a61-74bf3b9da174

# after

https://github.com/user-attachments/assets/1ee53e97-b03b-43fc-96fb-f8f3ad912caf

Fixes https://github.com/grafana/hyperion-planning/issues/124


ashharrison90:
ok with the new changes it should work in all cases 👌

Sergej-Vlasov (approved):
LGTM! Much better apply positioning than before 👍

bfmatei (approved):
lgtm to as well 🚀

PR #919, opened 2024-09-27
Fix showing all value in datalinks when linking to another db and including vars

mdvictor:
Fixes an issue where setting a datalink to another dashboard that included template vars and selecting the `All` value in the initial dashboard would generate the URL for the second dashboard using all the values concatenated instead of the special `All` value.

This PR addresses that and bring back the functionality from the old arch

Before:

https://github.com/user-attachments/assets/5493db27-27f0-4e15-8ae8-782f213620f8

After:

https://github.com/user-attachments/assets/86f92cc2-cb48-48ea-a1ef-d14a22abe5f6

Closes https://github.com/grafana/grafana/issues/90180

### DIFF
--- a/packages/scenes/src/variables/macros/AllVariablesMacro.ts
+++ b/packages/scenes/src/variables/macros/AllVariablesMacro.ts
@@ -3,6 +3,8 @@ import { isCustomVariableValue, SceneVariable } from '../types';
 import { formatRegistry, FormatVariable } from '../interpolation/formatRegistry';
 import { SkipFormattingValue } from './types';
 import { VariableFormatID } from '@grafana/schema';
+import { MultiValueVariable } from '../variants/MultiValueVariable';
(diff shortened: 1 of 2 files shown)


packages/scenes/src/variables/macros/AllVariablesMacro.ts
```diff
@@ -20,6 +22,12 @@ export class AllVariablesMacro implements FormatVariable {
 
     for (const name of Object.keys(allVars)) {
       const variable = allVars[name];
+
+      if (variable instanceof MultiValueVariable && variable.hasAllValue() && !variable.state.allValue) {
+        params.push(format.formatter(ALL_VARIABLE_TEXT, [], variable))
```
torkelo, line 27: Don't think we should use the text value here but the value representation

torkelo (approved):

PR #924, opened 2024-09-28
useQueryRunner: Add more option props

torkelo:
Trying to use scenes-react for https://github.com/grafana/grafana/pull/92861 

Think that PR is a great first real use case to try scenes-react in

### DIFF
--- a/packages/scenes-react/src/hooks/useQueryRunner.ts
+++ b/packages/scenes-react/src/hooks/useQueryRunner.ts
@@ -8,6 +8,8 @@ export interface UseQueryOptions {
   queries: SceneDataQuery[];
   maxDataPoints?: number;
   datasource?: DataSourceRef;
+  liveStreaming?: boolean;
+  maxDataPointsFromWidth?: boolean;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


packages/scenes-react/src/components/VizGridLayout.tsx
```diff
@@ -0,0 +1,25 @@
...
+export interface Props {
+  minWidth?: number;
+  minHeight?: number;
+  children: React.ReactNode;
+}
+
+/**
+ * Simple css grid layout for visualizations
+ */
+export function VizGridLayout({ children, minWidth = 400, minHeight = 320 }: Props) {
```
oscarkilhed, line 13: Guessing this is supposed to replace the `VizGrid` component used in https://github.com/grafana/grafana/pull/93954 ?

oscarkilhed (approved):

packages/scenes-react/src/components/VizGridLayout.tsx
torkelo: Yes

torkelo: Just something simple to start with, annoying to have to create a custom layout container

PR #926, opened 2024-10-02
AdHocFilters: Better match old handling of queryparam formatting for ad hoc filters

kaydelaney:
In the old architecture, template variable adapters had a `getValueForUrl` method that could return a processed value suited for including in a URL if necessary. This was used in templateSrv:

https://github.com/grafana/grafana/blob/fb3b13b567ca074bcf9dca552e5ee4e6feff30ce/public/app/features/templating/template_srv.ts#L330-L335

To better match this behavior under scenes, I've moved some functions around and added an optional `getValueForUrl` method to the `SceneVariable` and `FormatVariable` interfaces. This method gets called from the `queryparam` formatter, if it exists.

Fixes: https://github.com/grafana/grafana/issues/94303

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -5,7 +5,7 @@ import { SceneObjectBase } from '../../core/SceneObjectBase';
 import { SceneVariable, SceneVariableState, SceneVariableValueChangedEvent, VariableValue } from '../types';
 import { ControlsLayout, SceneComponentProps } from '../../core/types';
 import { DataSourceRef } from '@grafana/schema';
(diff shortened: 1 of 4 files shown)


torkelo:
@kaydelaney can you describe / link the bug? Not sure I understand the changes, we have the URL sync handler interface, so not sure what we would need a new getValueForUrl function

Sergej-Vlasov:
@torkelo thread about the bug https://raintank-corp.slack.com/archives/C03KVDHTWAH/p1727773589691779

torkelo:
Ok, took me some time to understand the problem / difference. I think leveraging our existing URL sync handler interface inside the variable formatted looks like a smaller / less complex solution: https://github.com/grafana/scenes/pull/931/files

kaydelaney:
Closed for https://github.com/grafana/scenes/pull/931

PR #934, opened 2024-10-09
VizPanel: Fix issue where changing panel options wouldn't cause re-render

kaydelaney:
In scenes, changing certain panel options, for example the "Show threshold labels" option on Gauge panels, wouldn't re-render the panel. Looking at the GaugePanel component's render function, we have the following:

```typescript
  render() {
    const { height, width, data, renderCounter, options } = this.props;

    const { minVizHeight, minVizWidth } = this.calculateGaugeSize();

    return (
      <VizRepeater
        getValues={this.getValues}
        renderValue={this.renderValue}
        width={width}
        height={height}
        source={data}
        autoGrid={true}
        renderCounter={renderCounter}
        orientation={options.orientation}
        minVizHeight={minVizHeight}
        minVizWidth={minVizWidth}
      />
    );
  }
```

The options prop _does_ update, but `VizRepeater` is a pure component and since options isn't passed as a prop, the component doesn't re-render.
To fix this, I've added a `renderCounter` to `VizPanel` that gets incremented when the panel options change. Since `renderCounter` is passed to `VizRepeater`, a re-render is triggered.

<details>
  <summary>📦 Published PR as canary version: <code>5.20.0--canary.934.11258648927.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.20.0--canary.934.11258648927.0
  npm install @grafana/scenes@5.20.0--canary.934.11258648927.0
  # or 
  yarn add @grafana/scenes-react@5.20.0--canary.934.11258648927.0
  yarn add @grafana/scenes@5.20.0--canary.934.11258648927.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanel.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.tsx
@@ -83,6 +83,7 @@ export interface VizPanelState<TOptions = {}, TFieldConfig = {}> extends SceneOb
   _pluginLoadError?: string;
   /** Internal */
   _pluginInstanceState?: any;
+  _renderCounter?: number;
 }
 
 export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends SceneObjectBase<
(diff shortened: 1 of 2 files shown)


torkelo (approved):
hm... since the options prop is updated part of me feels like this is a bug in in the panels that use VizRepeater

PR #939, opened 2024-10-16
TimeRange: Fix timezone not being sync with url

axelavargas:
Related to #https://github.com/grafana/support-escalations/issues/12963

### Issue
Passing the timezone as a parameter in the URL is not working. 

After checking the code, I noticed we were returning early and not considering `timezone` URL parameter

### Testing change in core grafana

https://github.com/user-attachments/assets/f8ee25cd-42f9-4ea4-a75b-643ea977f847

### Todo
- [x] Allow `timezone` to be synced through rul
- [ ] Add unit test

### DIFF
--- a/packages/scenes/src/core/SceneTimeRange.tsx
+++ b/packages/scenes/src/core/SceneTimeRange.tsx
@@ -211,6 +211,7 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
 
     let from = parseUrlParam(values.from);
     let to = parseUrlParam(values.to);
+    let timeZone = values.timezone;
 
     if (values.time && values['time.window']) {
       const time = Array.isArray(values.time) ? values.time[0] : values.time;
@@ -225,7 +226,7 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
       }
     }
 
-    if (!from && !to) {
+    if (!from && !to && !timeZone) {
       return;
     }
 


packages/scenes/src/core/SceneTimeRange.tsx
```diff
@@ -225,7 +226,7 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
       }
     }
 
-    if (!from && !to) {
+    if (!from && !to && !timeZone) {
```
torkelo, line 229: maybe we can skip this condition and instead check further down if (Object.keys(update).length > 0) ?

axelavargas: Do you mean, something like this?
```
 
 // if (!from && !to && !timeZone) { ---removing this condition
 
.... old code ...

    update.value = evaluateTimeRange(
      update.from ?? this.state.from,
      update.to ?? this.state.to,
      update.timeZone ?? this.getTimeZone(),
      this.state.fiscalYearStartMonth,
      this.state.UNSAFE_nowDelay
    );
    
// NEW code
     if (Object.keys(update).length === 0) {
      return;
    }
    return this.setState(update);
  }
```

axelavargas: I push the changes :)

packages/scenes/src/core/SceneTimeRange.tsx
```diff
@@ -249,7 +245,10 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
       this.state.UNSAFE_nowDelay
     );
 
-    this.setState(update);
+    if (Object.keys(update).length === 0) {
+      return;
+    }
```
torkelo, line 250: eh, this not to low down? must it not be above where update.value is set to have any effect?

axelavargas: yes, you are right, did the changes

torkelo (approved):

PR #941, opened 2024-10-17
feat: sort adhoc filter options using ufuzzy

sd2k:
Prior to this change, we were leaving the order of the options in adhoc filters unchanged, which has led to some
[poor quality search results][example] across plugins using adhoc filters.

This change sorts the options in adhoc filters using ufuzzy, which should lead to better search results. This is the same approach used in Grafana core (see [1], [2], [3] for examples).

[example]: https://raintank-corp.slack.com/archives/C075BDBTX96/p1729076789303509
[1]: https://github.com/grafana/grafana/blob/ca1fd028a2a1aa0dd560d8220c9f04c243d77445/public/app/features/variables/pickers/OptionsPicker/reducer.ts#L264-L272
[2]: https://github.com/grafana/grafana/blob/ca1fd028a2a1aa0dd560d8220c9f04c243d77445/public/app/features/trails/services/search.ts#L13
[3]: https://github.com/grafana/grafana/blob/ca1fd028a2a1aa0dd560d8220c9f04c243d77445/public/app/features/explore/Logs/utils/uFuzzy.ts#L13

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/utils.ts
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/utils.ts
@@ -44,12 +44,17 @@ export function fuzzySearchOptions(options: Array<SelectableValue<string>>) {
         haystack.push(options[i].label || options[i].value!);
       }
     }
-    const idxs = ufuzzy.filter(haystack, search);
+    const [idxs, info, order] = ufuzzy.search(haystack, search);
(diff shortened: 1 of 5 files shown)


Sergej-Vlasov (approved):
LGTM! tested with ad hoc filters and group by variable

Sergej-Vlasov:
cc @leeoniya just in case

torkelo (approved):

PR #942, opened 2024-10-25
AdHocFilterCombobox: Improve backspace functionality to delete filter key, operator and values separately

Sergej-Vlasov:
Rework backspace functionality to delete filter key, operator, values individually instead of whole filter.

I am adding additional property `forceEdit` on the `AdHocFilterWithLabels` type that is purely used to trigger UI change via useEffect because that was the cleanest way I could think of to trigger edit in separate filter pills that has no reference to each other.

<details>
  <summary>📦 Published PR as canary version: <code>5.21.1--canary.942.11610448397.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.21.1--canary.942.11610448397.0
  npm install @grafana/scenes@5.21.1--canary.942.11610448397.0
  # or 
  yarn add @grafana/scenes-react@5.21.1--canary.942.11610448397.0
  yarn add @grafana/scenes@5.21.1--canary.942.11610448397.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -243,13 +243,13 @@ export class AdHocFiltersVariable
 
   public _handleComboboxBackspace(filter: AdHocFilterWithLabels) {
     if (this.state.filters.length) {
-      // default filter to forceEdit to last filter (when triggering from wip filter)
+      // default forceEdit last filter (when triggering from wip filter)
       let filterToForceIndex = this.state.filters.length - 1;
 
-      // adjust index if backspace triggered from non wip filter
+      // adjust filterToForceIndex index to -1 if backspace triggered from non wip filter
+      //  to avoid triggering forceEdit logic
       if (filter !== this.state._wip) {
-        const filterIndex = this.state.filters.findIndex((f) => f === filter);
-        filterToForceIndex = filterIndex - 1;
+        filterToForceIndex = -1;
       }
 
       // const filterToForceEditExists = !!this.state.filters[filterToForceIndex];
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -252,7 +252,6 @@ export class AdHocFiltersVariable
         filterToForceIndex = -1;
       }
 
-      // const filterToForceEditExists = !!this.state.filters[filterToForceIndex];
       this.setState({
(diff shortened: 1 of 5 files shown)


ashharrison90 (commented):
ok had a play around, nice job! 🙌

no code comments (yet 😂), just a couple of behaviour comments i spotted whilst clicking around:
- when you have multiple filters, backspacing the second key also deletes the first value
  - easiest way to repro is on the demo page. add `__name__ = ALERTS` and `action = add_client` chips
  - press backspace twice so the operator dropdown is showing
  - pressing backspace 1 more time
  - expected: second filter's key should be deleted the key dropdown visible
  - actual: it deletes the key and the first filter's value and shows the first filter's value dropdown.
- focus doesn't behave correctly when backspacing and readding a value.
  - add `__name__ = ALERTS` chip
  - backspace to remove the `ALERTS` value
  - readd it
  - expected: focus is on the input
  - actual: focus is on the chip
- this one is less important, but you can get in broken UI states by just holding enter to add everything:
  image
  - `ALERTS` isn't a valid key, `actions` isn't a valid operator, etc.
  - not a "proper" use case, but it points to a timing issue which is probably worse on slower machines

Sergej-Vlasov:
@ashharrison90 thank you for pointing out those!  Addressed all your points and ready for another look!

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -239,6 +241,41 @@ export class AdHocFiltersVariable
...
+      // default forceEdit last filter (when triggering from wip filter)
+      let filterToForceIndex = this.state.filters.length - 1;
+
+      // adjust filterToForceIndex index to -1 if backspace triggered from non wip filter
+      //  to avoid triggering forceEdit logic
+      if (filter !== this.state._wip) {
+        filterToForceIndex = -1;
+      }
+
+      // const filterToForceEditExists = !!this.state.filters[filterToForceIndex];
```
ashharrison90, line 255: we can remove this

ashharrison90 (approved):
behaviour is spot on! lgtm 🥇 

i think we're at a point where we can start to tidy things up in some follow up PRs? rough thoughts:
- start writing some unit tests to capture behaviour
  - things like what dropdown is shown and where focus is at any particular time and after specific keypresses
  - this will allow us to refactor with confidence we're not regressing anything
- refactor!
  - first thought is to pull out the `onKeyDown` and associated handlers into it's own file

wdyt? 🤔

Sergej-Vlasov:
@ashharrison90  Definitely agree on unit tests and tidying up everything!

PR #944, opened 2024-10-28
SceneObject: Improve cloning logic

torkelo:
We currently have a bug in scenes where you cannot "apply" changes in inspect drawer because there is an error thrown when you hit "Apply" (only happens when the dashboard is NOT in edit mode). The error happens because we try go into edit mode while the inspect drawer is open the error is `'Cannot clone a SceneObject with a SceneObjectRef in state` which we throw from the cloning. 

* Remove throwing this error (replaced with a console.warning)
* Now clone callers can opt not to clone some objects by passing them in as undefined in the "withState" argument.  (makes is so we can clone the state of dashboards with overlay: undefined)

https://github.com/grafana/grafana/issues/95482


oscarkilhed (approved):

mdvictor (approved):
LGTM

PR #945, opened 2024-10-28
Propagate variable changes through inactive objects

oscarkilhed:
There are some use cases when a query runner is active when it's parent is inactive, but the query runner will never get notified a variable it depends on has been updated.

This happens in grafana when the dashboard datasource is used to get data from a panel that is contained in a collapsed row.

This change makes the variable change propagate to children of deactivated objects to let the sceneObject affected decide if its going to react.


oscarkilhed:
@torkelo PR in grafana implementing your suggestion: https://github.com/grafana/grafana/pull/95507

PR #949, opened 2024-10-30
AdHocFiltersVariable: Pass scene reference to expressionBuilder

gtk-grafana:
What:
This PR allows scene developers to use application state to modify how ad hoc variable expressions are rendered. 
In order to avoid regression, a new state property `buildExpressionOnActivate` was added which will alter the behavior of AdHocFiltersVariable to wait until activation to build the filterExpression. This allows a reference to the AdHocFiltersVariable can be included in the `expressionBuilder` callback which is not available currently when the filterExpression is created during instantiation.

Why:
Currently the expressionBuilder method is unable to reference the application context, and use-cases where application state can influence how a particular ad hoc variable is interpolated is not supported.

How:
Specifically, for Explore Logs, we need to conditionally ignore filters that match the label in the active tab so users can include multiple values by clicking on buttons in the result panels for a single label without needing to modify the ad-hoc variable state, which would introduce undesirable implications in the UI.

An example implementation in the app-plugin could look like the following:

```
  const labelVariable = new AdHocFiltersVariable({
    name: VAR_LABELS,
    datasource: EXPLORATION_DS,
    layout: 'combobox',
    buildExpressionOnActivate: true,
    expressionBuilder: renderLogQLLabelFilters,
   // [...]
  });

function expressionBuilder(filters: AdHocVariableFilter[], scene: SceneObject) {
  // Get the key of another ad hoc variable
  const primaryLabel = getServiceSelectionPrimaryLabel(scene).state.filters[0]?.key
  
  // Exclude filters matching the first key in the other variable
  filters = filters.filter(f => f.key !== primaryLabel)
  let { positiveFilters, negative } = getLogQLLabelFilters(filters);
  const negativeFilters = negative.map((filter) => renderMetadata(filter)).join(', ');

  return trim(`${positiveFilters.join(', ')}, ${negativeFilters}`, ' ,');
}

```

<details>
  <summary>📦 Published PR as canary version: <code>5.21.2--canary.949.11612383247.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.21.2--canary.949.11612383247.0
  npm install @grafana/scenes@5.21.2--canary.949.11612383247.0
  # or 
  yarn add @grafana/scenes-react@5.21.2--canary.949.11612383247.0
  yarn add @grafana/scenes@5.21.2--canary.949.11612383247.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -176,7 +176,6 @@ export class AdHocFiltersVariable
 
 
   public constructor(state: Partial<AdHocFiltersVariableState>) {
-
     const initialFilterExpression = !state.buildExpressionOnActivate ? renderExpression(state.expressionBuilder, state.filters) : undefined
     super({
       type: 'adhoc',
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -99,7 +99,7 @@ export interface AdHocFiltersVariableState extends SceneVariableState {
   _wip?: AdHocFilterWithLabels;
 }
 
-export type AdHocVariableExpressionBuilderFn = (filters: AdHocFilterWithLabels[], sceneRef?: AdHocFiltersVariable) => string;
+export type AdHocVariableExpressionBuilderFn = (filters: AdHocFilterWithLabels[], sceneRef: AdHocFiltersVariable) => string;
 
 export type getTagKeysProvider = (
   variable: AdHocFiltersVariable,
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -81,6 +81,13 @@ export interface AdHocFiltersVariableState extends SceneVariableState {
   expressionBuilder?: AdHocVariableExpressionBuilderFn;
 
   /**
+   * When set, the filterExpression will not be created until after activation, and the expressionBuilder function
+   * will be passed a reference to this AdHocFiltersVariable scene object.
+   * This enables scene application context to influence how a particular filter expression is built.
+   */
+  buildExpressionOnActivate?: boolean
+
+  /**
    * Whether the filter supports new multi-value operators like =| and !=|
    */
   supportsMultiValueOperators?: boolean;
@@ -99,7 +106,7 @@ export interface AdHocFiltersVariableState extends SceneVariableState {
   _wip?: AdHocFilterWithLabels;
 }
 
-export type AdHocVariableExpressionBuilderFn = (filters: AdHocFilterWithLabels[], sceneRef: AdHocFiltersVariable) => string;
+export type AdHocVariableExpressionBuilderFn = (filters: AdHocFilterWithLabels[], sceneRef?: AdHocFiltersVariable) => string;
 
 export type getTagKeysProvider = (
   variable: AdHocFiltersVariable,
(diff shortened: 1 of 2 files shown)


packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -80,6 +80,13 @@ export interface AdHocFiltersVariableState extends SceneVariableState {
    */
   expressionBuilder?: AdHocVariableExpressionBuilderFn;
 
+  /**
+   * When set, the filterExpression will not be created until after activation, and the expressionBuilder function
+   * will be passed a reference to this AdHocFiltersVariable scene object.
+   * This enables scene application context to influence how a particular filter expression is built.
+   */
+  buildExpressionOnActivate?: boolean
```
torkelo, line 88: this I think misuse the expression builder , there expression builder is only meant to take in a series of filters and generate a string, it's stateless. 

If you want to dynamically build there expression based on scene state then update the filters in the ad-hoc filter

torkelo:
> Currently the expressionBuilder method is unable to reference the application context, and use-cases where application state can influence how a particular ad hoc variable is interpolated is not supported.

this does not make sense, and will break the state flow. As the ad-hoc filter will not update when the other scene state changes. 

If you want to link together some other scene state you need to update the ad-hoc filter based on that other scene state, this will make sure the state updates happen (IE some other variable or scene state update =>. update ad-hoc filter that depends on it => re-issue queries)

using state inside a the rendering of filter expression like this breaks the state flow/dependency (meaning this ad-hoc variable will not update when that depend state is updated)

torkelo:
> we need to conditionally ignore filters that match the label in the active tab so users can include multiple values by clicking on buttons in the result panels for a single label without needing to modify the ad-hoc variable state

Hm... not sure I understand, there is a ad-hoc filter with a label filter set but you you want to see all possible values like as if this filter was not there? why is it there if it not supposed to have an effect?

gtk-grafana:
> Hm... not sure I understand, there is a ad-hoc filter with a label filter set but you you want to see all possible values like as if this filter was not there? why is it there if it not supposed to have an effect?

Basically we want users to be able to add include filters in the viz:

https://github.com/user-attachments/assets/16969171-a758-41ff-a844-6317679c9985

And in the combobox:

https://github.com/user-attachments/assets/5ba5372b-3971-441f-b8e1-3333d2e0a98d

Otherwise adding any include filter that matches the key of the active tab will run the queries and hide other panels that the user might want to add.

But I think I understand why this approach won't work, and realizing it might be better/easier to add a new optional property to the `AdHocFilterWithLabels` so the application can keep certain filters in the combobox UI, but we can exclude them in the `expressionBuilder`

e.g.
```
export interface AdHocFilterWithLabels extends AdHocVariableFilter {
  keyLabel?: string;
  valueLabels?: string[];
  includeInQuery?: boolean // new prop that can be used in expressionBuilder, but doesn't impact UI layout
}
```

What do you think?

torkelo:
@gtk-grafana yea I was thinking something like an exclude or filter property.

But what purpose does filters have that are not used in a query?

gtk-grafana:
@torkelo Because they are used in different application contexts. Following the example of the video PoC above: in the "level" tab, I don't want including the "error" filter to exclude info, warn, etc, so I can include more level filters after adding the first one, but when I switch to the namespace tab, I do want those filters applied.

torkelo:
but if it's the same adhoc filter variable how can this be filter only be applied to one part of the scene? Feels like you need a locally scoped filter variable that excludes some filters

gtk-grafana:
I was able to get it working locally by updating the filter "meta" on dependent state changes with the proposal on https://github.com/grafana/scenes/pull/951.

Having more locally scoped variables was one of my initial attempts, but the variable I am attempting to convert is already rendered in the top scene and needs URL state to sync across all routes.

gtk-grafana:
I started running into issues with that approach because I'd need to subscribe to the variable and set the meta state on changes within the combobox which trigger duplicate queries. 

I'll try creating a "replica" of the "primary" top level variable that is not rendered at a lower scope which can subscribe to the primary and update the filters appropriately.

PR #952, opened 2024-11-01
Don't run migration when changing panel plugin

oscarkilhed:
When changing the panel plugin the pluginVersion will not be correct, this means the migration will always run. It shouldn't have to run when changing from one plugin type to another.


mdvictor (commented):
LGTM 💯

mdvictor (approved):

PR #955, opened 2024-11-05
AdHocFiltersCombobox - Pre-fill filter combobox value on edit

Sergej-Vlasov:
1. When editing value pre-fill input with the value for single value filters
2. Allow to select certain multi value to edit its content

<details>
  <summary>📦 Published PR as canary version: <code>5.23.1--canary.955.11702691002.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.23.1--canary.955.11702691002.0
  npm install @grafana/scenes@5.23.1--canary.955.11702691002.0
  # or 
  yarn add @grafana/scenes-react@5.23.1--canary.955.11702691002.0
  yarn add @grafana/scenes@5.23.1--canary.955.11702691002.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
@@ -17,6 +17,7 @@ export function AdHocFilterPill({ filter, model, readOnly, focusOnWipInputRef }:
   const [viewMode, setViewMode] = useState(true);
   const [shouldFocusOnPillWrapper, setShouldFocusOnPillWrapper] = useState(false);
   const pillWrapperRef = useRef<HTMLDivElement>(null);
+  const [populateInputOnEdit, setPopulateInputOnEdit] = useState(false);
 
   const keyLabel = filter.keyLabel ?? filter.key;
   // TODO remove when we're on the latest version of @grafana/data
@@ -52,6 +53,15 @@ export function AdHocFilterPill({ filter, model, readOnly, focusOnWipInputRef }:
     }
   }, [filter, model, viewMode]);
 
+  // reset populateInputOnEdit when pill goes into view mode
+  useEffect(() => {
+    if (populateInputOnEdit && viewMode) {
+      setPopulateInputOnEdit(false);
+    }
+    // excluding populateInputOnEdit dependency because we only care to reset it on viewMode change
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [viewMode]);
+
   if (viewMode) {
     const pillText = (
       <span className={styles.pillText}>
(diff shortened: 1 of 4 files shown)


ashharrison90 (commented):
nice! 👏 

some behaviour nits:
- when backspacing a filter pill, i don't think it should preserve the filter value(s). this only seems to happen for the value part. this behaviour should only happen if you shift-tab to the pill and press enter then.
- if i'm editing a filter and press shift-tab to go the previous part of the pill, it shouldn't preserve the input value (`ALERTS`) since that doesn't make sense for the previous part
image
image
- i feel like i should be able to shift-tab or click and edit the key in this scenario as i could with the operator, but i can't do it for key for some reason
image

shout if any of these don't make sense/need more explanation. once this goes in, i'm going to pick up the task to write some unit tests so we can capture all these behaviours

Sergej-Vlasov:
@ashharrison90 great catches! First and second points addressed! Third one - ability to edit key - I think we agreed to have keys non-editable and therefore there are no functionality around editing it. Let me know if you think we should enable editing of the key or not

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
```diff
@@ -52,6 +53,15 @@ export function AdHocFilterPill({ filter, model, readOnly, focusOnWipInputRef }:
...
 
+  // reset populateInputOnEdit when pill goes into view mode
+  useEffect(() => {
+    if (populateInputOnEdit && viewMode) {
+      setPopulateInputOnEdit(false);
+    }
+    // excluding populateInputOnEdit dependency because we only care to reset it on viewMode change
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [viewMode]);
+
```
ashharrison90, line 64: any time we ignore the rules of hooks we're creating technical debt for ourselves. there's a [new react compiler coming](https://react.dev/blog/2024/10/21/react-compiler-beta-release) and associated lint rules, and that will complain about any code that ignores the rules of hooks. so we should _really_ try to avoid it

think we can instead rewrite this `useEffect` for now to something like:
```tsx
  // reset populateInputOnEdit when pill goes into view mode
  useEffect(() => {
    if (viewMode) {
      setPopulateInputOnEdit((prevValue) => {
        return prevValue ? false : prevValue;
      });
    }
  }, [viewMode]);
```

ashharrison90 (approved):
idk, i kinda think we should be able to edit a key. feels weird that you can't but you can edit other parts 🤔 doesn't have to hold out this PR, but something to think about i guess

lgtm 👍

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
Sergej-Vlasov: @ashharrison90 I see your point about the key edit. Lets keep this to the other PR as this might have quite a lot of small changes so don't want to introduce it here. Refactored useEffect as you suggested 👍

PR #959, opened 2024-11-07
Variables: Saved default all value with non multi select fix

torkelo:
Fixes https://github.com/grafana/grafana/issues/96027 

The issue seems to be that both text & value are initialized to ALL_VARIABLE_VALUE, and in case the Select is not a multi value the correct label is not shown 

Think this could fix it 

<details>
  <summary>📦 Published PR as canary version: <code>5.23.2--canary.959.11726132276.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.23.2--canary.959.11726132276.0
  npm install @grafana/scenes@5.23.2--canary.959.11726132276.0
  # or 
  yarn add @grafana/scenes-react@5.23.2--canary.959.11726132276.0
  yarn add @grafana/scenes@5.23.2--canary.959.11726132276.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/variants/MultiValueVariable.test.ts
+++ b/packages/scenes/src/variables/variants/MultiValueVariable.test.ts
@@ -298,6 +298,24 @@ describe('MultiValueVariable', () => {
       expect(variable.state.value).toBe(ALL_VARIABLE_VALUE);
       expect(variable.state.text).toBe(ALL_VARIABLE_TEXT);
     });
+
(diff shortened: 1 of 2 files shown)


axelavargas (approved):
I tested locally. and it works as expected 🏅 , also the code looks good to me. I was checking why we could have that scenario of `text` being set to `$__all` in the old arch in dashboards and could not find anything. Maybe some provisioning mismatch

PR #961, opened 2024-11-08
AdHocFiltersCombobox: Collapse filter combobox

Sergej-Vlasov:
Filters UI collapse functionality

Implementation in grafana/grafana https://github.com/grafana/grafana/pull/98411

https://github.com/user-attachments/assets/dc1f28c5-71e6-4153-bfce-1cab75060cad

CONSIDERATIONS:
1. This is not entirely accurate when there are more variables on the same line (due to filter box growing) see in depth explanation -> https://github.com/grafana/scenes/issues/962

2. Perhaps worth adding extra setting `collapseThreshold` or have `collapseFilters` as `boolean | number` to allow more fine grain control on what pixel width to start collapse due to above

<details>
  <summary>📦 Published PR as canary version: <code>5.36.1--canary.961.12544951331.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.36.1--canary.961.12544951331.0
  npm install @grafana/scenes@5.36.1--canary.961.12544951331.0
  # or 
  yarn add @grafana/scenes-react@5.36.1--canary.961.12544951331.0
  yarn add @grafana/scenes@5.36.1--canary.961.12544951331.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersComboboxRenderer.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersComboboxRenderer.tsx
@@ -45,8 +45,10 @@ export const AdHocFiltersComboboxRenderer = memo(function AdHocFiltersComboboxRe
   );
 
   useLayoutEffect(() => {
-    // updateCollapseThreshold(!!model.state.collapseFilters ? true : false, filters.length);
-    updateCollapseThreshold(true, filters.length);
+    updateCollapseThreshold(!!model.state.collapseFilters ? true : false, filters.length);
+    // TODO: remove below before merging
+    // updateCollapseThreshold(true, filters.length); // for testing locally
+
     // needs to run only on first render
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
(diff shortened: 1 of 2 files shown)


torkelo:
I think something like this is really needed

Sergej-Vlasov:
@torkelo will clean up and finalise this over the next few days. This behaviour will be controlled via variable setting `collapseFilters`.

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -103,6 +103,11 @@ export interface AdHocFiltersVariableState extends SceneVariableState {
    * @internal state of the new filter being added
    */
   _wip?: AdHocFilterWithLabels;
+
+  /**
+   * Flag that will collapse combobox filters when they become too long
+   */
+  collapseFilters?: boolean;
```
ivanortegaalba, line 110: The first thing that comes to my mind is: Shouldn't this be something to control from the component rather than the variable declaration?

This is pure UI behavior. It is weird to create a SceneObject variable to represent the AdHoc filter and indicate that it is collapsible. Have you considered any other approach that allows you to configure this in the variables control, for example?

Sergej-Vlasov: I decided to go with this implementation because it does not involve any major design changes and offers persistence between reloads. Other approaches would struggle with persistence especially when this is opt in behaviour.

PR #965, opened 2024-11-11
Chore: Bump grafana dependencies

mdvictor:
This PR bumps scenes grafana dependencies to latest version v11.3


packages/scenes-react/package.json
```diff
@@ -37,16 +37,16 @@
...
   },
   "peerDependencies": {
-    "@grafana/data": "^11.0.0",
-    "@grafana/runtime": "^11.0.0",
-    "@grafana/schema": "^11.0.0",
-    "@grafana/ui": "^11.0.0",
+    "@grafana/data": "^11.3.0",
+    "@grafana/runtime": "^11.3.0",
+    "@grafana/schema": "^11.3.0",
+    "@grafana/ui": "^11.3.0",
```
jackw: nit: I've not looked at the code changes but unless scenes-react now requires ^11.3.0 due to api changes in those packages there shouldn't be a need to change this.

packages/scenes-react/package.json
```diff
@@ -37,16 +37,16 @@
     "url": "https://github.com/grafana/scenes/issues"
   },
   "dependencies": {
-    "@grafana/e2e-selectors": "^11.0.0",
+    "@grafana/e2e-selectors": "^11.3.0",
```
jackw, line 40: I think we should move this to a peerDep. Maybe we get [this PR](https://github.com/grafana/scenes/pull/940) reviewed and merged?

packages/scenes-react/package.json
mdvictor: Yeah, this makes sense, there shouldn't be a need to change this. Thank you!

packages/scenes-react/package.json
mdvictor: Sounds good! Seems like it already got approved!

jackw: Tis merged now. I apologise for any conflicts. 😬

torkelo (approved):

PR #966, opened 2024-11-12
SceneVariableSet: Notify scene objects that use time macros when time changes

torkelo:
Thought this was a clever solution (Making SceneVariableSet notify scene objects that use __from and __to macro when time range changes).

But it makes it so that SceneQueryRunner executes double queries when a query is using these macros (as SceneQueryRunner is already subscribing to time range).

Need a way to dedupe those updates.

Possible solutions
* dedupe the run queries somehow (when they are called with the same time range)
* Opt-in to being notified of time range "__from" and "__to" variables (or opt-out) in VariableDependencyConfig

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.test.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.test.ts
@@ -268,7 +268,7 @@ describe.each(['11.1.2', '11.1.1'])('SceneQueryRunner', (v) => {
       const queryRunner = new SceneQueryRunner({
         runQueriesMode: 'manual',
         queries: [{ refId: 'A' }],
-        $timeRange
+        $timeRange,
       });
 
       queryRunner.activate();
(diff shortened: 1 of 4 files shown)


kaydelaney:
Tested locally and it seems to work, and the code looks good. Maybe making it opt-in is safest for the moment?

torkelo:
@kaydelaney after thinking more about it I did not like this solution much, becomes a bit complex to have it be opt-in from SceneVariableSet (would need to have some kind of special handling for this variable in the propagation chain). 

Also feels a bit messy that when time range change we propagate variable change updates for for __from and __to (causing double updates when a scene object depends on both).

Opted for a simpler feature in VariableDependencyConfig that can handle the opt-in & and the time range subscription.

packages/scenes/src/querying/SceneQueryRunner.test.ts
```diff
@@ -277,21 +277,20 @@ describe.each(['11.1.2', '11.1.1'])('SceneQueryRunner', (v) => {
 
       expect(queryRunner.state.data).toBeUndefined();
 
-      queryRunner.runQueries()
+      queryRunner.runQueries();
```
torkelo, line 280: something is not right with our prettier config or pre-commit hooks?

kaydelaney (approved):

PR #967, opened 2024-11-14
Utils: Spread objects in cloneSceneObjectState

kaydelaney:
Fixes an issue where if a panel was duplicated, both the original panel and duplicated panel's `queries` state (in SceneQueryRunner) referred to the same object in memory, meaning when the duplicate query was changed, so was the original.

Thread: https://raintank-corp.slack.com/archives/C03KVDHTWAH/p1730897481961219

<details>
  <summary>📦 Published PR as canary version: <code>6.1.2--canary.967.13520771706.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.1.2--canary.967.13520771706.0
  npm install @grafana/scenes@6.1.2--canary.967.13520771706.0
  # or 
  yarn add @grafana/scenes-react@6.1.2--canary.967.13520771706.0
  yarn add @grafana/scenes@6.1.2--canary.967.13520771706.0
  ```
</details>


torkelo:
where do we mutate the queries object?

kaydelaney:
@torkelo It might be something strange some datasources do - could only reproduce with graphite/Mock datasource but not prometheus for example. I'll check.

kaydelaney:
Okay so looking into a bit more, at least in the case of Graphite, I think what's happening is it's taking the `target` that's passed to it and ultimately manipulating that when the query is modified via the graphite query editor.

https://github.com/grafana/grafana/blob/6abe99efd64b8867112d9d9c74971d96840ce32d/public/app/plugins/datasource/graphite/state/context.tsx#L91

Changing the above to `target: { ...query }` also seems to fix the issue, but other data sources also seem to make the same mistake, so it might be worthwhile "fixing" in scenes just the same.

packages/scenes/src/core/sceneGraph/utils.ts
```diff
@@ -45,6 +45,8 @@ export function cloneSceneObjectState<TState extends SceneObjectState>(
       for (const child of propValue) {
         if (child instanceof SceneObjectBase) {
           newArray.push(child.clone());
+        } else if (typeof child === 'object') {
+          newArray.push({ ...child });
```
dprokop, line 49: This is going to be problematic for properties that are `Array`. The individual array values will be spread as properties on the state.

dprokop (changes requested):

packages/scenes/src/core/sceneGraph/utils.ts
kaydelaney: Ah, good catch. Do you think something like this would be okay?
```suggestion
        } else if (typeof child === 'object' && !Array.isArray(child)) {
          newArray.push({ ...child });
```

leeoniya:
this bug was also reported for Google Cloud Monitoring datasource: https://github.com/grafana/grafana/issues/97223#issuecomment-2536294562

axelavargas (commented):
Hey @kaydelaney 👋🏾 , as other mentioned, can we fix this issue in the problematic data sources that are doing the mutation? There might be more side effects on implementing this approach

oscarkilhed:
Is there really more side effects to implementing this approach? If I were to clone a panel I wouldn't expect properties on it to be referring to the same property as the old panel, I think this has a great potential to cause more bugs in the future.

That said, the mutation in the graphite DS is also not great and should probably be fixed too.

packages/scenes/src/core/sceneGraph/utils.ts
```diff
@@ -45,6 +45,8 @@ export function cloneSceneObjectState<TState extends SceneObjectState>(
       for (const child of propValue) {
         if (child instanceof SceneObjectBase) {
           newArray.push(child.clone());
+        } else if (typeof child === 'object' && !Array.isArray(child)) {
+          newArray.push({ ...child });
```
mdvictor, line 49: An issue just popped up around transformations that is kind of related to this PR. Basically, duplicating a panel with some transformation and then modifying the transformation options within the second panel would also modify the transformation options in the initial panel.

This happens because of the spread operator won't cover objects that contain deeper nested arrays, which is the case for transformations where you'd have an object of the form: 
```
{
   id: "convertFieldType",
   options: {
       conversions: [{ ...some objects in array }]
   }
}
```

Modifying this line to use lodash `cloneDeep` instead of spread op fixes the issue with transformations, so maybe we can use that?

kaydelaney: Thanks for the info! I worry that using `cloneDeep` might introduce some performance issues considering how often scene objects are cloned, but I'd need to benchmark it to be sure. I'll look into it :)

mdvictor: Indeed, I share your worry, thought about it myself, not sure how to deal with it tho.

torkelo: @mdvictor did you locate where the transformations editors mutate the state?

kaydelaney:
Looked into the performance of cloning for three different solutions: Spreading object properties (which has its own limitations as mentioned by @mdvictor re: transformations), [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone), and [lodash's cloneDeep function](https://lodash.com/docs/#cloneDeep).

I evaluated the performance by taking a typical dashboard scene then cloning it ~1000 times, measuring the times taken to clone the scene, and calculating min, max, median, and mean times to clone for each algorithm.
I did this in Chrome, Firefox, and Safari, which you can see the results for below. Each metric has two values as I ran the test twice.

## Firefox ##

### Spread ###
Min: [0ms, 0ms]
Max: [8ms, 7ms]
Median: [0ms, 0ms]
Mean: [0.267ms, 0.188ms]

### structuredClone ###
Min: [0ms, 0ms],
Max: [9ms, 6ms],
Median: [0ms, 0ms]
Mean: [0.335ms, 0.397ms]

### cloneDeep ###
Min: [0ms, 0ms],
Max: [9ms, 7ms],
Median: [0ms, 0ms],
Mean: [0.362ms, 0.362ms]

## Chrome ##

### Spread ###
Min: [0ms, 0ms]
Max: [5.5ms, 6.1ms]
Median: [0.1ms, 0.1ms]
Mean: [0.175ms, 0.170ms]

### structuredClone ###
Min: [0.19ms, 0.19ms]
Max: [7.89ms, 5.4ms]
Median: [0.4ms, 0.3ms]
Mean: [0.44ms, 0.28ms]

### cloneDeep ###
Min: [0.1ms, 0.1ms]
Max: [6.1ms, 9.8ms]
Median: [0.2ms, 0.2ms]
Mean: [0.285ms, 0.37ms]

## Safari ##

### Spread ###
Min: [0ms, 0ms],
Max: [7ms, ~1ms],
Median: [0ms, 0ms],
Mean: [0.129ms, 0.122ms]

### structuredClone ###
Min: [0ms, 0ms]
Max: [3ms, 3ms]
Median: [0ms, 0ms]
Mean: [0.33ms, 0.3259ms]

### cloneDeep ###
Min: [0ms, 0ms]
Max: [2ms, 1ms]
Median: [0ms, 0ms]
Mean: [0.307ms, 0.316ms]
------

So, performance takes a small hit if we use `structuredClone` or `cloneDeep`, but I think it's still acceptable, and shouldn't negatively affect the UX in any perceivable way.

As for whether we should go with `structuredClone` or `cloneDeep`, `cloneDeep` seems to have slightly better performance for the median case, so I'm leaning towards that.

leeoniya:
> As for whether we should go with `structuredClone` or `cloneDeep`, `cloneDeep` seems to have slightly better performance for the median case, so I'm leaning towards that.

personally, i would opt for native methods if the perf difference is not large. less dependencies.

kaydelaney:
@leeoniya In general I'd agree, but lodash in particular is so ingrained in grafana (and its plugin ecosystem) that I'm not sure it's a big concern

leeoniya:
definitely not a big concern

axelavargas (approved):
@kaydelaney after extensive testing using this Scene version in a cloud Grafana instance (latest main 11.6.x) and local instance, I did not find anything else 🥳 

I checked:

- Editing panels that were duplicated works as expected ✅ 
- DataTrails - History steps works as expected ✅ 
- Editing Dashboard Json from editing ✅ 


https://github.com/user-attachments/assets/232ac086-4f63-43dc-859e-c937aedb56b5

PR #968, opened 2024-11-18
SceneObject: Render before activation

torkelo:
A big decision we made early on was to force activation of the scene object before it's component get's rendered. This was mainly to make sure parent SceneVariableSets where activated before child SceneQueryRunners come asking if their variables are "ready" or not. 

To solve this and ensure that all scene objects can run state initiation logic before children get rendered/activated, we made sure the in ComponentWrapper will not render the component until after activation. In retrospect I think this was a mistake. 

But this has some really big negative effects for React and rendering without flickering / intermediate empty states. 

An example of this problem was first scene in the SceneControlsSpacer object that renders a simple empty div but because it is rendered with a null value before activation the first dom hydration is without this element then on next render the element is added and a CSS reflow happens (and the controls on the right side of it are correctly pushed to the right), but because of this intermediate empty rendering the controls flicker / move as dashboards / scene are opened. 

We have a similar problem when opening dashboard settings from a dashboard with scrollbars. we first render an empty null view which causes the scrollbars to disappear and then after dashboard general settings is rendered the scrollbars reappear causing the whole page (at last the top nav right side icons) to jump around as the scrollbar is hidden and then shown again. 

Possible solutions

* Remove this logic from ComponentWrapper (this has some big sweeping effects, as we rely on it in a few places, mainly variables but also in scenes-react). But I think there are creative solutions/workarounds. The scary bit is how much scene apps could rely on this logic in subtle ways and removing it could cause all sorts of subtle or not-so-subtle crashes.  
* Opt-in (we could opt-in to this behavior for root scene objects, but really would need to be opt-in to all nodes that can have variables which is many so becomes a bit messy) 
* Opt-out (explored in this PR). 

**PR solution**

Add a way for SceneObjects to inform ComponentWrapper that it's fine to render it before activation 

<details>
  <summary>📦 Published PR as canary version: <code>5.27.0--canary.968.12050773878.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.27.0--canary.968.12050773878.0
  npm install @grafana/scenes@5.27.0--canary.968.12050773878.0
  # or 
  yarn add @grafana/scenes-react@5.27.0--canary.968.12050773878.0
  yarn add @grafana/scenes@5.27.0--canary.968.12050773878.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/core/types.ts
+++ b/packages/scenes/src/core/types.ts
@@ -61,6 +61,9 @@ export interface SceneObject<TState extends SceneObjectState = SceneObjectState>
   /** True when there is a React component mounted for this Object */
   readonly isActive: boolean;
 
+  /** Controls if activation blocks rendering */
+  readonly renderBeforeActivation: boolean;
+
   /** SceneObject parent */
   readonly parent?: SceneObject;
 
@@ -79,7 +82,7 @@ export interface SceneObject<TState extends SceneObjectState = SceneObjectState>
   /** Publish an event and optionally bubble it up the scene */
   publishEvent(event: BusEvent, bubble?: boolean): void;
 
-  /** Utility hook that wraps useObservable. Used by React components to subscribes to state changes */
+  /** Utility hook that wraps useObservableS. Used by React components to subscribes to state changes */
   useState(): TState;
 
   /** How to modify state */
--- a/packages/scenes/src/core/types.ts
+++ b/packages/scenes/src/core/types.ts
@@ -82,7 +82,7 @@ export interface SceneObject<TState extends SceneObjectState = SceneObjectState>
   /** Publish an event and optionally bubble it up the scene */
   publishEvent(event: BusEvent, bubble?: boolean): void;
 
-  /** Utility hook that wraps useObservableS. Used by React components to subscribes to state changes */
+  /** Utility hook that wraps useObservable. Used by React components to subscribes to state changes */
   useState(): TState;
 
   /** How to modify state */
--- a/packages/scenes/src/components/SceneControlsSpacer.tsx
+++ b/packages/scenes/src/components/SceneControlsSpacer.tsx
@@ -6,11 +6,8 @@ import { SceneComponentProps } from '../core/types';
 export class SceneControlsSpacer extends SceneObjectBase {
   public constructor() {
     super({});
-  }
 
-  public get Component() {
-    // Skipping wrapper component for this scene object so that it renders right away
-    return SceneControlsSpacer.Component;
+    this._renderBeforeActivation = true;
   }
 
   public static Component = (_props: SceneComponentProps<SceneControlsSpacer>) => {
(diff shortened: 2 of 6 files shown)


bfmatei:
> - Opt-in (we could opt-in to this behavior for root scene objects, but really would need to be opt-in to all nodes that can have variables which is many so becomes a bit messy)
> - Opt-out (explored in this PR).

I think this is a little bit confusing. From my pov the terminology of opt-in and opt-out should be swapped in this case as scene objects need to opt-in by having the `renderBeforeActivation` property defined as `true` in order for the rendering to happen before activation.

WDYT about having this gradually implemented: continue with the approach from this PR, test with core dashboards and other well established plugins that use Scenes and if we see that everything is OK with those we can decide if we go with opt-out or removing the logic from `ComponentWrapper`?

torkelo:
> I think this is a little bit confusing. From my pov the terminology of opt-in and opt-out should be swapped in this case as scene objects need to opt-in by having the renderBeforeActivation property defined as true in order for the rendering to happen before activation.

Not sure what you mean? Is it confusing that you opt-out by setting a property to true you mean (instead of false). Just that having a default/unset value be "true" feels wrong. 

We could name it "renderAfterActivation" and have it default to true. 

> WDYT about having this gradually implemented: continue with the approach from this PR, test with core dashboards and other well established plugins that use Scenes and if we see that everything is OK with those we can decide if we go with opt-out or removing the logic from ComponentWrapper?

Yea, something like that. Would be interesting to be able to change the default for all objects (And the opt-in where you want), ie to somehow control the default state of renderBeforeActivation (or renderAfterActivation)

bfmatei:
> Not sure what you mean? Is it confusing that you opt-out by setting a property to true you mean (instead of false). Just that having a default/unset value be "true" feels wrong.

Before this PR: no rendering before activation
In this PR: the render before activation happens by setting `renderBeforeActivation` to `true`. The default value is `false` which has the same effect as before the PR.

I consider this to be an `opt-in` approach if for a feature to work a flag has to be set to `true`.

In the description of the PR you mentioned this behavior as `Opt-out (explored in this PR)` which made me think that the new behavior applies by default to all scene objects.

torkelo:
@bfmatei your right, you opt-in to the new behavior.

torkelo:
@bfmatei took a look at this and can't come up with a good solution for variables, so I think this is the best we can do for now. 

For scenes-react this is not a big problem contexts (scene objects) are usually at the root and not unmounted/remounted for every page/route change

packages/scenes/src/core/types.ts
```diff
@@ -79,7 +82,7 @@ export interface SceneObject<TState extends SceneObjectState = SceneObjectState>
   /** Publish an event and optionally bubble it up the scene */
   publishEvent(event: BusEvent, bubble?: boolean): void;
 
-  /** Utility hook that wraps useObservable. Used by React components to subscribes to state changes */
+  /** Utility hook that wraps useObservableS. Used by React components to subscribes to state changes */
```
bfmatei, line 85: Just a nit:

```suggestion
  /** Utility hook that wraps useObservable. Used by React components to subscribes to state changes */
```

bfmatei (approved):
LGTM

bfmatei:
@torkelo I think it's fine tho'. Let's go with this approach!

PR #969, opened 2024-11-18
Variables: Fix search result ordering to use match quality

leeoniya:
ensures that when searching `namespace`, we get exact results first ~~(although case-insensitive)~~.

Ref: https://raintank-corp.slack.com/archives/C0630AYC4SY/p1731592718073559

the code is pretty much copy-pasted in 3 places, probably should be extracted?

sample list:

```
client_service_namespace
namespace
alert_namespace
container_namespace
Namespace
client_k8s_namespace_name
```

image

### DIFF
--- a/packages/scenes/package.json
+++ b/packages/scenes/package.json
@@ -38,7 +38,7 @@
   },
   "dependencies": {
     "@floating-ui/react": "^0.26.16",
-    "@leeoniya/ufuzzy": "^1.0.14",
+    "@leeoniya/ufuzzy": "^1.0.16",
     "@tanstack/react-virtual": "^3.9.0",
     "react-grid-layout": "^1.3.4",
     "react-use": "^17.5.0",
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/utils.ts
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/utils.ts
@@ -47,18 +47,16 @@ export function fuzzySearchOptions(options: Array<SelectableValue<string>>) {
     const filteredOptions: Array<SelectableValue<string>> = [];
 
     if (idxs) {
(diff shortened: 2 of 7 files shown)


dprokop (dismissed):
Looks good but a test case for this is missing. Agree this code could be extracted and deduplicated!

sd2k (commented):
LGTM too but agree that tests should be updated to include some case sensitive/insensitive exact matches. A refactor would also be nice!

leeoniya:
> tests should be updated to include some case sensitive/insensitive exact matches.

case-sensitive matching/sorting is out of scope here. we're just making sure that we sort properly by match quality.

i've added case-sensitivity as sorting/ranking metric in [uFuzzy v1.0.16](https://github.com/leeoniya/uFuzzy/releases/tag/1.0.16), so bumping that dep in another PR can bring those improvements.

i'll do the DRY pass and add some case-insensitive ranking tests here, though.

leeoniya:
okay, ended up just updating uFuzzy and added the ordering test. the DRY bits ended up being a bit nuanced since there are some differences in the option types and some other bits.

one thing to note here is that the haystack is now initialized eagerly rather than lazily (during first search). probably won't affect performance too much, but did not have quick way to check with real/giant datasets.

Sergej-Vlasov (approved):
Looks good! Tested with around 5k options and did not see any performance difference with the naked eye. 

On the side note, started getting this highlight only after update to newer version of uFuzzy:
Screenshot 2024-11-20 at 09 50 41

sd2k (approved):
LGTM! I can't reproduce the issue @Sergej-Vlasov is having with the import, it doesn't seem to have changed in ufuzzy since 1.0.14 either :thinking: could just be a transient thing?

Sergej-Vlasov:
It still builds and works as expected so could be a transient thing.

PR #972, opened 2024-11-21
SceneVariableSet: Store the whole error object instead of the error message

grafakus:
Currently, when an error happens during the `validateAndUpdate` observable / RxJS process, we [only store the error message](https://github.com/grafana/scenes/blob/main/packages/scenes/src/variables/sets/SceneVariableSet.ts#L250). 

After this PR, the whole error object will be stored, which can then be used to to inspect its [cause](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause), stack traces or any other custom property.

This issue relates to https://github.com/grafana/scenes/pull/371

### DIFF
--- a/packages/scenes/src/variables/sets/SceneVariableSet.test.tsx
+++ b/packages/scenes/src/variables/sets/SceneVariableSet.test.tsx
@@ -629,7 +629,7 @@ describe('SceneVariableList', () => {
 
       scene.activate();
 
-      expect(A.state.error).toBe('Danger!');
+      expect(A.state.error).toStrictEqual(new Error('Danger!'));
     });
 
     it('Should complete updating chained variables in case of error in all variables', () => {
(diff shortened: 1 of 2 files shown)


torkelo (commented):
Have you tested this with ControlsLabel ?

We should the error in the Variable label component, currently it expects a string

torkelo (commented):
would not logging the error be enough, do we need to store it?

grafakus:
> would not logging the error be enough, do we need to store it?

Hi @torkelo , sorry I've been busy on other fronts... Now, the error message is logged with the console. What if the app uses a custom logger that's interested in the stack trace or other properties that the error would contain? We need to store the whole error object to allow the custom logger to inspect it and send it wherever it wants (the console or Faro or...)

torkelo (commented):
See that error is very poorly typed on SceneVariableState

If we are to make this breaking change we need to change the error type to `Error`

PR #973, opened 2024-11-21
VizPanelBuilder: add setShowMenuOnHover to set PanelChrome showMenuOnHover prop

gtk-grafana:
Requires: https://github.com/grafana/grafana/pull/96868

What is this feature?
Adds scenes support for showMenuOnHover prop which will always show the vizPanelMenu, instead of only showing on hover.

Why do we need this feature?
See issue for more https://github.com/grafana/grafana/issues/96865.

TL;DR: Explore apps make use of right aligned actions in panel headers. If the menu icon is only visible on hover, users have the expectation that every panel has a menu on hover, or they don't realize a menu is there at all. Also having empty space next to the primary CTA in the panel header just looks weird.

Who is this feature for?
Mostly for users and developers of Explore apps.

Which issue(s) does this PR fix?:

Addresses the Scenes portion of: https://github.com/grafana/grafana/issues/96865

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanel.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.tsx
@@ -58,6 +58,10 @@ export interface VizPanelState<TOptions = {}, TFieldConfig = {}> extends SceneOb
    */
   hoverHeaderOffset?: number;
   /**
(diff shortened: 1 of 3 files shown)


torkelo (approved):
Feature/option makes sense to.

Issue #608, opened 2024-02-20
Migrate react-router from v5 to v6


leventebalogh:
### Why?

Grafana core allows plugins to start using the v6 version of react-router, and hopefully we will deprecate using v5 in plugins this year. `@grafana/scenes` is used by internal app plugins, which means that before migrating those we need to have a version of the scenes package that supports versions 6.x.x as well.

### How?
For the `scenes-app` follow our [migration guide](https://grafana.com/developers/plugin-tools/migration-guides/update-from-grafana-versions/migrate-9_x-to-10_x#update-to-react-router-v6) using create-plugin, and for `scenes` [rely on the official react-router v5->v6 migration guide](https://reactrouter.com/en/main/upgrading/v5).

**Some of the things that need too be changed:** _(list is not complete)_
- [ ] [SceneApp](https://github.com/grafana/scenes/blob/6dfa305e73e4751380b457a88f4a6352a6c4b201/packages/scenes/src/components/SceneApp/SceneApp.tsx#L2): Use `<Routes>` instead of `<Switch>`
- [ ] [SceneApp](https://github.com/grafana/scenes/blob/6dfa305e73e4751380b457a88f4a6352a6c4b201/packages/scenes/src/components/SceneApp/SceneApp.tsx#L2): Stop using the `render={}` prop in `Route` components, use the `element={}` prop instead and the `useParams()` hook
- [ ] [SceneAppPage](https://github.com/grafana/scenes/blob/6dfa305e73e4751380b457a88f4a6352a6c4b201/packages/scenes/src/components/SceneApp/SceneAppPage.tsx#L4): stop using the `render={}` prop and the `<Switch>` component
- [ ] [Routes](https://github.com/grafana/scenes/blob/6dfa305e73e4751380b457a88f4a6352a6c4b201/packages/scenes-app/src/components/Routes/Routes.tsx#L4): stop using the `<Switch>` component

PR #979, opened 2024-11-26
Update to `react-router@6`

leventebalogh:
**Fixes https://github.com/grafana/scenes/issues/608**

Updates the scenes packages to use react-router v6.
**This probably requires a major version bump, as it's going to be a breaking change for plugins that depend on scenes.**

### Why?
We are in the process of migrating Grafana core and also internal plugins to use react-router v6, however any plugin that is depending on scenes cannot easily do it (or at least [it's a complicated task](https://raintank-corp.slack.com/archives/C01C4K8DETW/p1730711123163989)). **Scenes still using react-router v5 is a blocker for plugins that would like to update.**

Huge thanks to @jackw for the help of discovering and removing _some_ of the circular dependences in the scenes package that were causing Jest to go mental.

# Release notes 

This release marks a big change where scene app based routing is using react-router v6 under the hood. This sadly requires several changes to plugins as react-router now requires all route paths to be relative. 

### Updating a plugin
[Example plugin update PR &rarr;](https://github.com/grafana/explore-logs/pull/1019/files)

1. **Bump `react-router-dom`** - `package.json`
```diff
-"react-router-dom": "^5.2.0",
+"react-router-dom": "^6.22.0",
```
2. **Remove `@types/react-router-dom`** - `package.json`
```diff
-"@types/react-router-dom": "^5.2.0",
```

3. **Stop externalising`react-router-dom`** - `.config/webpack/webpack.config.ts`
```diff
externals: [
      // Required for dynamic publicPath resolution
      { 'amd-module': 'module' },
      'lodash',
      'jquery',
      'moment',
      'slate',
      'emotion',
      '@emotion/react',
      '@emotion/css',
      'prismjs',
      'slate-plain-serializer',
      '@grafana/slate-react',
      'react',
      'react-dom',
      'react-redux',
      'redux',
      'rxjs',
      'react-router',
-    'react-router-dom',
      'd3',
      'angular',
      '@grafana/ui',
      '@grafana/runtime',
      '@grafana/data',
```

4. **Enable `react-router@v6`** - `.cprc.json`
Add the following line in the `.cprc.json`:
```diff
{
  "features": {
+   "useReactRouterV6": true
  }
}
```

4. **`<Route>`: use relative _wildcard_ routes**
```diff
-<Route path="/a/myorg-example-app/home ... />
+<Route path="home/*" ... />
```
6. **`<Route>`: use the `element` prop**
```diff
-<Route ... component={WithDrilldown} />
+<Route ... element={<WithDrilldown />} />
```
7. **`<Switch>`: replace with `Routes`**
```diff
-import { Switch } from "react-router-dom";
+import { Routes } from "react-router-dom";  

-<Switch> ... </Switch>
+<Routes> ... </Routes>
```

8. **Add a relative `routePath` prop to each `<SceneAppPage>`. It needs to include a wildcard if there are any drill-down or tab pages under it.**
```diff
-new SceneAppPage({
-  url: "/a/myorg-example-app/home",
-  ...
- });
+new SceneAppPage({
+  url: "/a/myorg-example-app/home",
+  routePath: "home/*",
+  ...
+})
```

9. **Make the `routePath` prop defined on drill-downs relative and include a wildcard.**

### Testing
- [x] Works with core Grafana ([pull request](https://github.com/grafana/grafana/pull/98426))
- [x] Scaffolding a new app plugin with scenes ([pull request](https://github.com/grafana/plugin-tools/pull/1422))
- [x] Grafana **_new scenes_** - Plugin **_old scenes (router v5)_** _(tested locally)_
- [x] Grafana **_new scenes_** - Plugin **_new scenes (router v6)_** _(tested locally)_
- [x] Grafana **_old scenes_** - Plugin **_new scenes (router v6)_** _(tested locally)_
- [x] Updating an existing app plugin ([explore-logs pull request](https://github.com/grafana/explore-logs/pull/1019)) 

<details>
  <summary>📦 Published PR as canary version: <code>6.0.0--canary.979.13200094417.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.0.0--canary.979.13200094417.0
  npm install @grafana/scenes@6.0.0--canary.979.13200094417.0
  # or 
  yarn add @grafana/scenes-react@6.0.0--canary.979.13200094417.0
  yarn add @grafana/scenes@6.0.0--canary.979.13200094417.0
  ```
</details>


torkelo:
Pushed an update to get SceneAppPage and most demos to work 

Scene apps needs a big update

* add routePath to all SceneAppPage definitions (with * wildcard if they have drilldowns) this needs to be relative not absolute 
* the url property on SceneAppPage needs to remain and be absolute as it's used to create breadcrumbs and tab links 
* routePath on all drilldowns needs to be relative and end in wildcard

packages/scenes-app/src/components/Routes/Routes.tsx
```diff
@@ -1,19 +1,17 @@
...
-      <Route path={prefixRoute(`${ROUTES.Demos}`)} component={DemoListPage} />
-      <Route path={prefixRoute(`${ROUTES.GrafanaMonitoring}`)} component={GrafanaMonitoringApp} />
-      <Route path={prefixRoute(`${ROUTES.ReactDemo}`)} component={ReactDemoPage} />
-      <Redirect to={prefixRoute(ROUTES.Demos)} />
-    </Switch>
+    <Routes>
+      <Route path={`${ROUTES.Demos}/*`} Component={DemoListPage} />
+      <Route path={`${ROUTES.GrafanaMonitoring}/*`} Component={GrafanaMonitoringApp} />
+      <Route path={`${ROUTES.ReactDemo}/*`} Component={ReactDemoPage} />
+      {/* <Redirect to={prefixRoute(ROUTES.Demos)} /> */}
```
torkelo, line 14: Need to add a v6 redirect here

leventebalogh:
I have spent sometime trying to green out all the test case, and although I think got _somewhere_, there are some weird things that might need someone with deeper knowledge of the scenes projects.

The main issue is with the tests under the scenes package: for some reason the inline `jest.mock()` calls don't seem to go through, [while mocking via a `__mocks__` directory is still working](https://github.com/grafana/scenes/pull/979/files#diff-00e970ba21eb6b9daa86839daa46eb35c95c819494bc28c46718d02120422e0fR16). Obviously it wouldn't make sense to transition all the tests to use a `__mocks__` folder solution, we should instead find the root cause. (There was one issue we discovered with @jackw in the jest config: we also needed to introduce a `"setupFiles"` and move parts of the `setupTestsAfterEnv.ts` in there, although that didn't fix this problem.)

oscarkilhed:
> I have spent sometime trying to green out all the test case, and although I think got _somewhere_, there are some weird things that might need someone with deeper knowledge of the scenes projects.
> 
> The main issue is with the tests under the scenes package: for some reason the inline `jest.mock()` calls don't seem to go through, [while mocking via a `__mocks__` directory is still working](https://github.com/grafana/scenes/pull/979/files#diff-00e970ba21eb6b9daa86839daa46eb35c95c819494bc28c46718d02120422e0fR16). Obviously it wouldn't make sense to transition all the tests to use a `__mocks__` folder solution, we should instead find the root cause. (There was one issue we discovered with @jackw in the jest config: we also needed to introduce a `"setupFiles"` and move parts of the `setupTestsAfterEnv.ts` in there, although that didn't fix this problem.)
> 
> @grafana/dashboards-squad

@leventebalogh I see the tests are passing, did you find the issue?

oscarkilhed (commented):
Thanks for working on this @leventebalogh

How does this work for apps built with scenes that run in Grafana? Will they continue to work as long as they are built with an older version of scenes, or will they break when Grafana and scenes runs on router@v6?

leventebalogh:
> @leventebalogh I see the tests are passing, did you find the issue?

Yes, thanks @oscarkilhed, we managed to green them out with @jackw 👍 


> How does this work for apps built with scenes that run in Grafana? Will they continue to work as long as they are built with an older version of scenes, or will they break when Grafana and scenes runs on router@v6?

I am currently testing out these changes with an example plugin, and as soon as I am done with that I'll update the PR description with correct instructions on what steps plugins need to take for updating. (We will probably also have a migration page for this)

At the time being Grafana still supports plugins both using react-router@v5 or react-router@v6. This means, that plugins built with an earlier scenes version will still be supported (at least until core is updated to use react-router@v6 fully - then they will break). As far as I know scenes is bundled (and not externalised), so it shouldn't be a problem that core is depending on a different version (for now). _I'll double check this though as well, and update the PR description!_ 

**This must be major version bump though**, as any plugins updating to this scenes version will also need to update some things in their source code.

leventebalogh:
@torkelo @dprokop could you please have a look if you think it's legit?

Sergej-Vlasov:
@leventebalogh do we need any additional steps for this to work in grafana/grafana? Did a [draft scenes bump PR](https://github.com/grafana/grafana/pull/98426) to launch a [drone pipeline](https://drone.grafana.net/grafana/grafana/214962) and quite a lot of tests are erroring out 🤔

leventebalogh:
Thanks @Sergej-Vlasov, nice catch! 🙏 

It looks like the issue was that core Grafana had a different version of `react-router-dom` than what was bundled with the `@grafana/scenes` package. To mitigate the issue I have moved the `react-router-dom` to be a peer and dev dependency of the scenes package. Hopefully the CI tests [on your PR](https://github.com/grafana/grafana/pull/98426) will get green, soon (I managed to green them out locally).

torkelo (approved):

Sergej-Vlasov (approved):
@leventebalogh great job! This seems to be working now! 

Since this is a major release lets postpone the merge a bit to allow any bugfixes/docs to land before hand. You can still start working on updating plugins in the mean time by using canary release of this PR. 

We have the release notes with how to handle this breaking change but I think it would also be beneficial to include this in https://grafana.com/developers/scenes/

packages/scenes-app/src/demos/variables.tsx
```diff
@@ -191,6 +194,7 @@ export function getVariablesDemo(defaults: SceneAppPageState) {
       new SceneAppPage({
         title: 'Many variable options',
         url: `${defaults.url}/many-values`,
+        routePath: 'many-values',
```
dprokop, line 197: I'm wondering if we shouldn't make `routePath` mandatory, since it can be either a fixed string or parametrizable one. And all over the place in this PR the routePath is being added to the app pages. Don't think I saw a single app page in this PR that would not require it. WDYT?

leventebalogh: Actually I think that's a good point, I couldn't think of any scenario where it wouldn't be necessary from now on (unless I am missing something). It could also help with migration I think for plugins (showing type errors for wherever it's missing). What do you think @torkelo?

leventebalogh:
@Sergej-Vlasov Thanks, sure 👍

> We have the release notes with how to handle this breaking change but I think it would also be beneficial to include this in https://grafana.com/developers/scenes/
Yeah, I think adding some migration docs would make sense as well, as this is a bit more "complicated" change, similar to what we have in https://grafana.com/developers/plugin-tools/migration-guides/update-from-grafana-versions/.

packages/scenes-app/src/demos/variables.tsx
torkelo: @dprokop yea, I agree. Should be mandatory

packages/scenes-app/src/components/Routes/Routes.tsx
leventebalogh: Resolved now.

packages/scenes-app/src/demos/variables.tsx
leventebalogh: Fixed in https://github.com/grafana/scenes/pull/979/commits/dc6a6e143e9980487136de7514672eafc061ccc2.

PR #981, opened 2024-11-28
Supply fixed now to datemath caluclations

oscarkilhed:
evaluateTimeRange uses dateMath.parse that is not deterministic. Sometimes the clock tics over to the next millisecond between calculating `to` and `from`. This makes the timerange `now-24 to now` be 86400001ms instead of 86400000ms (24 hours and one millisecond)

Requires https://github.com/grafana/grafana/pull/97151


Sergej-Vlasov (approved):

ivanortegaalba (approved):

oscarkilhed:
Hmm. Can't really merge this until there is a new version of grafana.data released, unfortunately this was not included in 11.4 released just a few days ago.

torkelo (commented):
Just import it with a tsignore , but you need a fall back to use the date math.parse if the scenes lib is used in a older version of Grafana that does not have this function

PR #983, opened 2024-11-29
GroupBy: Add <Select /> for single selection option

joannaWebDev:
What this PR does:
-  Introduced conditional rendering for <Select /> and <MultiSelect /> based on the isMultiSelect flag

**By default, isMultiSelect is true, which renders the <MultiSelect /> component (no breaking changes).**

-  When isMultiSelect is set to false, the <Select /> component is used for single selection.
-  Updated the groupByStatic Demo

**Demo**

https://github.com/user-attachments/assets/81ca7e56-bb99-46db-98f2-a333a6660995

### DIFF
--- a/packages/scenes/src/variables/groupby/GroupByVariable.tsx
+++ b/packages/scenes/src/variables/groupby/GroupByVariable.tsx
@@ -8,7 +8,7 @@ import { ValidateAndUpdateResult, VariableValueOption, VariableValueSingle } fro
 import { MultiValueVariable, MultiValueVariableState, VariableGetOptionsArgs } from '../variants/MultiValueVariable';
 import { from, lastValueFrom, map, mergeMap, Observable, of, take, tap } from 'rxjs';
 import { getDataSource } from '../../utils/getDataSource';
-import { InputActionMeta, MultiSelect } from '@grafana/ui';
+import { InputActionMeta, MultiSelect, Select } from '@grafana/ui';
 import { isArray } from 'lodash';
 import { dataFromResponse, getQueriesForVariables, handleOptionGroups, responseHasError } from '../utils';
 import { OptionWithCheckbox } from '../components/VariableValueSelect';
@@ -20,6 +20,8 @@ import { wrapInSafeSerializableSceneObject } from '../../utils/wrapInSafeSeriali
 export interface GroupByVariableState extends MultiValueVariableState {
   /** Defaults to "Group" */
   name: string;
+  /** Determines whether the component allows multiple selections (MultiSelect) or a single selection (Select). Defaults to true for MultiSelect. */
+  isMultiSelect?: boolean;
   /** The visible keys to group on */
   // TODO review this type and name (naming is hard)
   defaultOptions?: MetricFindValue[];
@@ -131,6 +133,7 @@ export class GroupByVariable extends MultiValueVariable<GroupByVariableState> {
   public constructor(initialState: Partial<GroupByVariableState>) {
     super({
       isMulti: true,
+      isMultiSelect: true,
       name: '',
       value: [],
       text: [],
@@ -204,11 +207,13 @@ export class GroupByVariable extends MultiValueVariable<GroupByVariableState> {
     return { value: [], text: [] };
   }
 }
+
 export function GroupByVariableRenderer({ model }: SceneComponentProps<MultiValueVariable>) {
   const {
     value,
     text,
     key,
+    isMultiSelect = true,
     maxVisibleValues,
     noValueOnClear,
     options,
@@ -262,7 +267,7 @@ export function GroupByVariableRenderer({ model }: SceneComponentProps<MultiValu
     [optionSearcher, inputValue]
   );
 
-  return (
+  return isMultiSelect ? (
     <MultiSelect<VariableValueSingle>
       data-testid={`GroupBySelect-${key}`}
       id={key}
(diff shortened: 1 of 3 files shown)


packages/scenes/src/variables/groupby/GroupByVariable.tsx
```diff
@@ -20,6 +20,10 @@ import { wrapInSafeSerializableSceneObject } from '../../utils/wrapInSafeSeriali
 export interface GroupByVariableState extends MultiValueVariableState {
   /** Defaults to "Group" */
   name: string;
+  /** Determines if user can select several options*/
+  isMulti: boolean;
+  /** Determines whether the component uses MultiSelect or Select. Defaults to true for MultiSelect. */
+  isMultiSelect?: boolean;
```
torkelo, line 26: why do we need two options here? one should be enough?

joannaWebDev: Looking into the SRP and ensuring separation of concerns, I've decoupled rendering from business logic:

- `isMulti` (rendering logic): Configures the behavior of the Select library to toggle between single-selection and multi-selection modes (i.e., can multiple options be selected?).
- `isMultiSelect` (business logic): Determines whether the component renders as a MultiSelect (checkbox-like) or Select (dropdown) component.

That being said, maybe renaming isMultiSelect to showMultiSelect would help reduce confusion and improve clarity. wdyt?

torkelo: @joannaWebDev 

> MultiSelect (checkbox-like) or Select (dropdown) component.

This does not make sense to, isMulti should always use a multi select and not a single select UI. What is the use case to have 

Can you explain why we would need a multi select variable with a single select dropdown? This variable should have the exact same behavior / options as the other multi select variables  (has a single isMulti option)

joannaWebDev: My inspiration was drawn from the [o11y-apps-platform](https://ops.grafana-ops.net/a/grafana-app-observability-app/services/service/o11y-apps-platform---apiserver?var-tempo=fds8vtxx3ao74b&var-prometheus=edprtf91hz01se&var-environmentValue=$__all&var-groupBy=cluster&var-groupBy=service_version), where the Select component is implemented as a dropdown but allows multiple selections when configured appropriately. This pattern seemed flexible and consistent with that approach, enabling a Select to function in both single and multi-selection scenarios based on the isMulti flag.

That said, I don’t have a strong preference here. I’m happy to align with a stricter approach where MultiSelect is exclusively used for multi-selections and Select only supports single selections. If this is the preferred direction, I can update the implementation accordingly.

packages/scenes/src/variables/groupby/GroupByVariable.tsx
```diff
@@ -131,6 +133,7 @@ export class GroupByVariable extends MultiValueVariable<GroupByVariableState> {
   public constructor(initialState: Partial<GroupByVariableState>) {
     super({
       isMulti: true,
+      isMultiSelect: true,
```
torkelo, line 136: we already have the isMulti, why the need for isMultiSelect??

What is the use case for a Multi group by with single select or a multi select with a single value variable (that would break/be buggy I suspect)

joannaWebDev: The reason for using isMultiSelect is that it controls which component is rendered — either a MultiSelect (with checkboxes for multiple selections) or a Select (single select dropdown), the latter being the component I am adding. It’s just a way to switch between the two UI components.

The isMulti prop, which I'm not using anymore this PR, is typically meant to determine whether multiple selections are allowed. But in this case, we’re only concerned with which component to use (multi-select with checkboxes or single select), so that’s why I’m just using isMultiSelect.

Sergej-Vlasov:
I believe it would be cleaner here to utilise `isMulti` prop because its basically the same as `isMultiSelect` in this case. `Select` uses `isMulti` to render `MultiSelect` in core:
```typescript
export function Select<T, Rest = {}>(props: SelectCommonProps<T> & Rest) {
  return <SelectBase {...props} />;
}

export function MultiSelect<T, Rest = {}>(props: MultiSelectCommonProps<T> & Rest) {
  // @ts-ignore
  return <SelectBase {...props} isMulti />;
}

```

so if the typescript is not complaining too much you can utilise `Select` component with dynamic `isMulti` prop and define different `onChange`, `onBlur` and other handlers/props for single/multi values. 

Another approach all together would be to implement new `Combobox` component but I am not sure if its fully flushed out yet

joannaWebDev:
> I believe it would be cleaner here to utilise `isMulti` prop because its basically the same as `isMultiSelect` in this case. `Select` uses `isMulti` to render `MultiSelect` in core:
> 
> ```ts
> export function Select<T, Rest = {}>(props: SelectCommonProps<T> & Rest) {
>   return <SelectBase {...props} />;
> }
> 
> export function MultiSelect<T, Rest = {}>(props: MultiSelectCommonProps<T> & Rest) {
>   // @ts-ignore
>   return <SelectBase {...props} isMulti />;
> }
> ```
> 
> so if the typescript is not complaining too much you can utilise `Select` component with dynamic `isMulti` prop and define different `onChange`, `onBlur` and other handlers/props for single/multi values.
> 
> Another approach all together would be to implement new `Combobox` component but I am not sure if its fully flushed out yet

Changed

packages/scenes/src/variables/groupby/GroupByVariable.tsx
torkelo: > where the Select component is implemented as a dropdown but allows multiple selections when configured appropriately.

Is that not what the isMulti does? 

> I’m happy to align with a stricter approach where MultiSelect is exclusively used for multi-selections and Select only supports single selections.

what is strict? I am confused :) are you proposing a multi select where you can only select a single value? what would be the point, and would that not be a very confusing user experience?

joannaWebDev: Makes sense. Changed the approach and to utilise `isMulti` to control the rendering of either MultiSelect or Select.  :-)

packages/scenes/src/variables/groupby/GroupByVariable.tsx
```diff
@@ -306,6 +308,51 @@ export function GroupByVariableRenderer({ model }: SceneComponentProps<MultiValu
         setIsOptionsOpen(false);
       }}
     />
+  ) : (
+    <Select
+      aria-label="Group by selector"
+      id={key}
+      placeholder="None"
```
Sergej-Vlasov, line 315: we should probably use the same `data-testid`, `placeholder` and `aria-label` props on both Select and MultiSelect components

Sergej-Vlasov (approved):

packages/scenes/src/variables/groupby/GroupByVariable.tsx
joannaWebDev: Added same data-testid, and placeholder, to Select. Added the placeholder to the MultiSelect as it was :-)

PR #984, opened 2024-12-02
Fix DataSourceVariable doesn't use `options` from initial state

svennergr:
Since `DataSourceVariable` is a `MultiSelectVariable` the initial state can contain `options` like this:

```ts
  const dataSourceList = getDataSourceSrv()
    .getList({
      filter: (datasource) => {
        if (
          [
            'mysql',
          ].includes(datasource.type)
        ) {
          return true;
        }
        return false;
      },
    })
    .map((ds) => {
      return { label: ds.name, value: ds.uid };
    });
  const dsVar = new DataSourceVariable({ label: 'Data Source', options: dataSourceList });
```

However, these `options` are not valued in its implementation of `getValueOptions`.

### DIFF
--- a/packages/scenes/src/variables/variants/DataSourceVariable.tsx
+++ b/packages/scenes/src/variables/variants/DataSourceVariable.tsx
@@ -45,10 +45,10 @@ export class DataSourceVariable extends MultiValueVariable<DataSourceVariableSta
   }
 
   public getValueOptions(args: VariableGetOptionsArgs): Observable<VariableValueOption[]> {
-    if (this.state.options) {
+    if (this.state.options?.length) {
       return of(this.state.options);
     }
-    
+
     if (!this.state.pluginId) {
       return of([]);
     }
--- a/packages/scenes/src/variables/variants/DataSourceVariable.tsx
+++ b/packages/scenes/src/variables/variants/DataSourceVariable.tsx
@@ -45,6 +45,10 @@ export class DataSourceVariable extends MultiValueVariable<DataSourceVariableSta
   }
 
   public getValueOptions(args: VariableGetOptionsArgs): Observable<VariableValueOption[]> {
+    if (this.state.options) {
+      return of(this.state.options);
+    }
+    
     if (!this.state.pluginId) {
       return of([]);
     }


cyriltovena (approved):
LGTM

packages/scenes/src/variables/variants/DataSourceVariable.tsx
```diff
@@ -45,6 +45,10 @@ export class DataSourceVariable extends MultiValueVariable<DataSourceVariableSta
   }
 
   public getValueOptions(args: VariableGetOptionsArgs): Observable<VariableValueOption[]> {
+    if (this.state.options?.length) {
+      return of(this.state.options);
+    }
+
```
torkelo, line 51: not sure about this, this will make it never update the options list (based on the pluginId), making it just a statically defined variable (might as well use CustomVariable) ?

svennergr: In my opinion we should then at least remove `options` from `initialState` in https://github.com/grafana/scenes/blob/4d4366eb57d841fbc55b5d9238fd2c4fe7598f64/packages/scenes/src/variables/variants/DataSourceVariable.tsx#L34

I needed a way to use multiple `pluginId` and thought `options` might be a good property, but it wasn't applied in the constructor. Instead I'm just setting the `DataSourceVariable` state later than the constructor now.

torkelo: > but it wasn't applied in the constructor. 

what do you mean?

svennergr: The example like this, where you pass `options` in the `DataSourceVariable` constructor does not work right now:

```ts
  const dataSourceList = getDataSourceSrv()
    .getList({
      filter: (datasource) => {
        if (
          [
            'mysql',
          ].includes(datasource.type)
        ) {
          return true;
        }
        return false;
      },
    })
    .map((ds) => {
      return { label: ds.name, value: ds.uid };
    });
  const dsVar = new DataSourceVariable({ label: 'Data Source', options: dataSourceList });
```

What I needed to do was to deliberately call `dsVar.setState({options: dataSourceList});` after I created the var.

torkelo: @svennergr confused, 

```ts
 super({
      type: 'datasource',
      value: '',
      text: '',
      options: [],
      name: '',
      regex: '',
      pluginId: '',
      ...initialState,
    });
```

looks like options passed in via initialState is set, of course it will be cleared by variable activates (or the set activates). But that is it's correct behavior, options is not to be set externally really it's something it looks up using the pluginId

svennergr: Yea, if it's the foreseen behavior it makes sense. From just looking at the types when using the variable as a "consumer"/developer, you see `options` and think it could be used.

What do you think about omitting that type in `initalState` like `constructor(initialState: Omit<Partial<DataSourceVariableState>, 'options'>);`

PR #985, opened 2024-12-02
Variables: Fixes issue when variable value was null

torkelo:
Fixes https://github.com/grafana/grafana/issues/96140 

A missed code in the old system was this: 
https://github.com/grafana/grafana/blob/main/public/app/features/variables/state/sharedReducer.ts#L157 

Where any non string text or value value was replaced with an empty string

### DIFF
--- a/packages/scenes/src/variables/variants/query/utils.test.ts
+++ b/packages/scenes/src/variables/variants/query/utils.test.ts
@@ -1,0 +1,9 @@
+import { VariableSort } from '@grafana/data';
+import { metricNamesToVariableValues } from './utils';
+
+describe('When null values are returned', () => {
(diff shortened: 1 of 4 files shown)


packages/scenes/src/variables/variants/query/utils.test.ts
```diff
@@ -0,0 +1,9 @@
+import { VariableSort } from '@grafana/data';
+import { metricNamesToVariableValues } from './utils';
+
+describe('When null values are returned', () => {
+  it('Should translated to emtpy strings', async () => {
```
dprokop, line 5:
```suggestion
  it('Should translated to empty strings', async () => {
```

dprokop (approved):

Issue #986, opened 2024-12-02
AdHocFiltersCombobox: Regression - selecting existing filter no longer visualizes results from getTagValuesProvider as of 5.23.1


gtk-grafana:
Regression introduced in https://github.com/grafana/scenes/pull/955 and released in https://github.com/grafana/scenes/releases/tag/v5.23.1.

As of 5.23.1, when clicking on an ad-hoc filter variable, the results of the getTagValues call are no longer presented to users
![Image](https://github.com/user-attachments/assets/cf380890-7191-4f43-808c-02581671beab)
Until the user manually deletes the existing value.

This is a detrimental change to the UX of the ad hoc combobox element as this significantly increases the burden of changing values, and there is no affordance to the user that other values will show up after deleting every char in the current value. 
Furthermore since this requires repeatedly tapping the delete key to see the other values, if you accidentally hit delete too many times the user is forced to re-enter the operator, or perhaps the entire filter if they hit delete 2 times more then necessary.  

It's very frustrating that this used to require 1 click/keypress to see if there are other values available, and 2 clicks/keypresses to change to another existing value, and now requires between 3 (if user knows about ctrl+delete) and 2 + filterValue.length for the user to clear out the existing value to see what other options are available to switch to.

Before https://github.com/grafana/scenes/pull/955:
![Image](https://github.com/user-attachments/assets/92282999-f540-44b6-8ef9-9646d7bff307)

After: 
![Image](https://github.com/user-attachments/assets/720193b7-c463-4590-a448-f5276f64e43a)

Showing "No options found" should **not** be the default behavior when clicking on an existing combobox filter, and needs to be configurable if there is good reason why clicking on an existing filter doesn't visualize results from the getTagValuesProvider in the UI until users delete the existing value.

PR #987, opened 2024-12-02
AdHocFiltersCombobox: Disable "pre-fill filter combobox value on edit" by default

gtk-grafana:
Address UX regression introduced by https://github.com/grafana/scenes/pull/955.

Since I don't understand what value https://github.com/grafana/scenes/pull/955 brings to variables without regex operators or other variables that return results from `getTagValuesProvider`, I'm proposing we disable this behavior by default, and developers can set the `populateInputOnEdit` on variables where this behavior is desired.

Fixes: https://github.com/grafana/scenes/issues/986

This PR solves the short-term problem of #955 being forced on all users of the combobox by default. 
As a follow-up it would be good to have some documentation added around what https://github.com/grafana/scenes/pull/955 is accomplishing, I assume that this is so users can more easily edit regex values, or other variables that don't return results from `getTagValuesProvider`. From there we might be able to come up with a better long-term solution (only drop into "edit" mode when there are no values returned by `getTagValuesProvider`, or only drop into edit mode when certain operators are selected, differentiate between edit and select a new value etc)


gtk-grafana:
Closing as https://github.com/grafana/scenes/pull/992 should fix https://github.com/grafana/scenes/issues/986

PR #990, opened 2024-12-04
Scopes: Add `ScopesBridge` object

bfmatei:
Continuing the discussion from here: https://github.com/grafana/grafana/pull/97176

So what this PR brings: allow SceneObjects to consume React Contexts.

<details>
  <summary>📦 Published PR as canary version: <code>6.3.1--canary.990.13785017927.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.3.1--canary.990.13785017927.0
  npm install @grafana/scenes@6.3.1--canary.990.13785017927.0
  # or 
  yarn add @grafana/scenes-react@6.3.1--canary.990.13785017927.0
  yarn add @grafana/scenes@6.3.1--canary.990.13785017927.0
  ```
</details>


torkelo (commented):
I think this is really cool, but until we have more examples / stronger need for this for many contexts I would recommend a more simple specific implementation for the ScopeContext specifically (maybe a base class that can handle some generic stuff). 

Like a ScopeContextConsumer scene object that you can add to your scene (we would add it to DashboardScene), maybe we add it as an optional state prop to EmbeddeddScene or SceneApp (DashboardSceneRenderer would have render the ScopeContextConsumer component to make the context => scene state active). 

Adding this to all scene objects, and to ComponentWrapper feels a bit too drastic until we know how much this will be used

torkelo:
But I do think something like ScopesContextConsumer or ScopeSceneBridge (what ever we call it), could be something we add to scenes lib (and it depends on the context in runtime). And then scene app plugins can use it

packages/scenes/src/core/SceneComponentWrapper.tsx
```diff
@@ -22,6 +43,13 @@ function SceneComponentWrapperWithoutMemo<T extends SceneObject>({ model, ...oth
   return <Component {...otherProps} model={model} />;
 }
 
+function SceneComponentWrapperWithoutMemo<T extends SceneObject>({ model, ...otherProps }: SceneComponentProps<T>) {
+  return [
+    <ContextsConsumer key="contexts" model={model} />,
+    <ComponentRenderer key="component" model={model} {...otherProps} />,
+  ];
+}
+
```
dprokop, line 52: Took me a bit to udnerstand this 🤔

dprokop (commented):
Agree with Torkel overall, this is a neat addition, but maybe would be worth talking it trough more widely if there's really a library need for this. It introduces quite a bit of complexity, and I'm not sure if we should solve this problem in classic scenes, but rather in react-scenes.

bfmatei:
@torkelo @dprokop 

I refactored this to only use React Context for Scopes (added a bridge between scopes and scenes).

A couple of notes:
- `SceneApp` seems the best place to place the bridge as it's top level, meaning that all nested scene objects can get it
- `SceneAppPage` have a `useScopes` flag that control wether a page should or should not show the scopes selector
- the bridge also handles URL sync as it's way better than controlling it via the service (I can change this tho')
- added usage for `SceneQueryRunner` as well as adhoc variables and groupby variables. No more double queries and no more data/filters enriching needed
- the context is defined here for testing purposes as it's easier than to link runtime to scopes and then scopes to grafana while doing development
- the `useScopes` hook has to spread the context as simply returning the context won't take effect for use-cases like
```ts
useEffect(() => {
  console.log(context);
}, [context]);
```
- the context has to be a service as it's way easier to offer interoperability between scopes dashboards, scopes selector and the scopes "facade"
- we need to do something with Prettier. Apparently there are a lot of files which are not formatted correctly 😢

packages/scenes/src/core/SceneScopesBridge.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  private get contextObservable(): Observable<ScopesContextValue | undefined> {
+    return this._contextSubject.asObservable();
+  }
+}
+
+function SceneScopesBridgeRenderer({ model }: SceneComponentProps<SceneScopesBridge>) {
+  const context = useScopes();
+
+  model.updateContext(context);
```
torkelo, line 133: I think this needs to be inside a useEffect to avoid any possibility of updating state while react is rendering

torkelo: then you can remove the setTimeout on lin 115

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -187,8 +194,20 @@ export class AdHocFiltersVariable
...
 
+  private _activationHandler = () => {
+    this._scopesBridge = sceneGraph.getScopesBridge(this);
+
+    if (this._scopesBridge) {
+      this._subs.add(
+        this._scopesBridge.subscribeToIsLoading((isLoading) => this.setState({ _isScopesLoading: isLoading }))
+      );
+    }
+  };
```
torkelo, line 209: I think we should be able to get rid of this. 

If a user managed to open the ad-hoc filter before scopes have loaded for the first time. If that becomes a problem we could just wait to render the ad-hoc filter (or the whole scene) until after scopes have loaded for the first time

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -140,6 +143,7 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
   private _onActivate() {
     if (this.isQueryModeAuto()) {
       const timeRange = sceneGraph.getTimeRange(this);
+      const scopesBridge = sceneGraph.getScopesBridge(this);
```
torkelo, line 146: Long term I think it would be cool to generalize this concept with a special variable type that all query runners and variables depend on by default (so will block initial query execution), and the variable can then modify the query. 

I tried to start a poc around it, a QueryEnhancerVariable but did not get very far. 

But I think this implementation is also ok, just a shame it adds more complexity to this already complex class :)

torkelo (commented):
I think this looking really good

I think the isScopesLoading code added to AdHocFilters and GroupByVarable feel a bit on the "too complex" / detail that it should not have to care so deeply about (ie subscribe to state etc). 

Wonder if we can skip that for now, and if it becomes a problem find a less invasive solution (like a loading spinner on the whole scene until scopes have initialized)

packages/scenes/src/querying/SceneQueryRunner.ts
bfmatei: I have something in mind regarding this. I'll start a PoC in the next weeks 🤔 .

packages/scenes/src/core/SceneScopesBridge.ts
bfmatei: Thanks! Fixed it now.

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
bfmatei: Removed for now

bfmatei (commented):
I'll just continue doing testing for this and see how it goes. I also need to fix the code in core and test that out as well.

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
torkelo: or alternatively subscribe to scopes from the rendering (instead via ad-hoc filter state)

packages/scenes/src/core/SceneScopesBridge.ts
```diff
@@ -0,0 +1,133 @@
...
+        ),
+        filter(([prevIsLoading, newIsLoading]) => prevIsLoading !== newIsLoading)
+      )
+      .subscribe(([_prevIsLoading, newIsLoading]) => {
+        cb(newIsLoading);
+      });
+  }
+
+  public enable() {
+    this.context?.enable();
```
dprokop, line 82: Either I'm blind or `this.context` is never set :D

packages/scenes/src/components/SceneApp/SceneAppPage.tsx
```diff
@@ -16,11 +18,30 @@ export class SceneAppPage extends SceneObjectBase<SceneAppPageState> implements
...
     super(state);
+
+    this.addActivationHandler(this._activationHandler);
   }
 
+  private _activationHandler = () => {
+    this._scopesBridge = sceneGraph.getScopesBridge(this);
+
+    if (this.state.useScopes) {
+      if (!this._scopesBridge) {
```
dprokop, line 33: Uber nit: we can try resolving the bridge only when useScopes is enabled

packages/scenes/src/core/SceneScopesBridge.ts
```diff
@@ -0,0 +1,133 @@
...
+            [prevContext?.state.value ?? [], newContext?.state.value ?? []] as [Scope[], Scope[]]
+        ),
+        filter(([prevScopes, newScopes]) => !isEqual(prevScopes, newScopes))
+      )
+      .subscribe(([prevScopes, newScopes]) => {
+        cb(newScopes, prevScopes);
+      });
+  }
+
+  public getIsLoading(): boolean {
```
dprokop, line 61: nit
```suggestion
  public isLoading(): boolean {
```

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -380,17 +406,19 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
...
     const timeRange = sceneGraph.getTimeRange(this);
-    if(this.isQueryModeAuto()){
+    const scopesBridge = sceneGraph.getScopesBridge(this);
+    if (this.isQueryModeAuto()) {
       this.subscribeToTimeRangeChanges(timeRange);
+      this.subscribeToScopesChanges(scopesBridge);
     }
 
-    this.runWithTimeRange(timeRange);
+    this.runWithTimeRangeAndScopes(timeRange, scopesBridge);
```
dprokop, line 421: Let's add tests for making sure the SQR works as expected w/ scopes.

dprokop (commented):
@bfmatei  - this looks good now! Excellent job on bringing this to scenes. I've left just a few nits to consider.

packages/scenes/src/core/SceneScopesBridge.ts
```diff
@@ -0,0 +1,125 @@
...
+export class SceneScopesBridge extends SceneObjectBase implements SceneObjectWithUrlSync {
+  static Component = SceneScopesBridgeRenderer;
+
+  protected _urlSync = new SceneObjectUrlSyncConfig(this, { keys: ['scopes'] });
+
+  protected _renderBeforeActivation = true;
+
+  private _contextSubject = new BehaviorSubject<ScopesContextValue | undefined>(undefined);
+
+  private _pendingScopes: string[] | null = null;
```
aocenas, line 21: Would be nice to add a comment, that this is used to backpropagate URL changes back to context value if context isn't loaded yet.

packages/scenes/src/core/SceneScopesBridge.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  public setReadOnly(readOnly: boolean) {
+    this.context?.setReadOnly(readOnly);
+  }
+
+  public updateContext(newContext: ScopesContextValue | undefined) {
+    if (this._pendingScopes && newContext) {
+      setTimeout(() => {
+        newContext?.changeScopes(this._pendingScopes!);
+        this._pendingScopes = null;
```
aocenas, line 93: Should we just `return` here and end the update? It seems like the `_pendingScopes` have priority (cause they are from the URL) so we need to propagate them into the context, so it does not seem important what is the current `newContext` value.

packages/scenes/src/core/SceneScopesBridge.ts
```diff
@@ -0,0 +1,125 @@
...
+      });
+    }
+
+    if (this.context !== newContext || this.context?.state !== newContext?.state) {
```
aocenas, line 97: I assume once we use memoization in the `useScopes` we don't really need to do this semi-deep checking here right?

packages/scenes/src/core/SceneScopesBridge.ts
```diff
@@ -0,0 +1,125 @@
...
+      const shouldUpdate = this.context?.state.value !== newContext?.state.value;
+
+      this._contextSubject.next(newContext);
+
+      if (shouldUpdate) {
+        this.forceRender();
```
aocenas, line 103: Unsure about the logic of this forceRendering here but maybe I am missing some scenes internal. We update context subject every time so anything that is subscribed to the context values should get updates from `subscribeToValue` and `subscribeToLoading` right? I assume the forceRerender triggers rerendering of the `SceneScopesBridgeRenderer` but that would just trigger the `useScopes` and `useEffect` but there should not be any updates (cause updates would rerender the component automatically anyway). It seems like this would trigger an infinite rerender loop if we didn't have the deep compare earlier here. So just wander why is this `forceRender` needed here?

packages/scenes/src/core/SceneScopesBridge.ts
```diff
@@ -0,0 +1,125 @@
...
+    this.context?.setEnabled(enabled);
+  }
+
+  public setReadOnly(readOnly: boolean) {
+    this.context?.setReadOnly(readOnly);
+  }
+
+  public updateContext(newContext: ScopesContextValue | undefined) {
+    if (this._pendingScopes && newContext) {
+      setTimeout(() => {
```
tskarhed, line 91: Why is this timeout needed?

packages/scenes/src/core/SceneScopesBridge.ts
bfmatei: Haven't tested out this case. At a first glance I would say we still need to save the new context to get access to other props (i.e. `enabled` or `loading`) even if scopes themselves were not loaded.

Something in the lines of:
- load page with scopes in URL
- trigger scopes loading in the context which results in setting the `loading` flag to `true`
- the bridge prevents the query runner to perform queries given that scopes are loading

packages/scenes/src/core/SceneScopesBridge.ts
bfmatei: This should be the case, yeah.

packages/scenes/src/core/SceneScopesBridge.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  public setEnabled(enabled: boolean) {
+    this.context?.setEnabled(enabled);
+  }
+
+  public setReadOnly(readOnly: boolean) {
+    this.context?.setReadOnly(readOnly);
+  }
+
+  public updateContext(newContext: ScopesContextValue | undefined) {
```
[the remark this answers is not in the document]
bfmatei: `updateScopes` was deliberately skipped until we have a need to expose it publicly. This should never be called by plugins. However, marking it as `private` would prevent us from accessing it in the model.

An alternative would be to expose a `SceneScopesBridgeLike` interface that only contains public methods and use that as a return value for `getScopesBridge` or for the state prop in `AppComponent`.

packages/scenes/src/core/SceneScopesBridge.ts
bfmatei: TBH, I really don't remember I this was needed. Might have been a temp fix back then.

mdvictor (commented):
This is amazing work! Small comment about removing the filter enriching call in adhocs

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -359,6 +367,7 @@ export class AdHocFiltersVariable
       filters: otherFilters,
       queries,
       timeRange,
+      scopes: this._scopesBridge?.getValue(),
       ...getEnrichedFiltersRequest(this),
```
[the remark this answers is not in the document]
mdvictor: Since the filters enrichers have all been removed [here](https://github.com/grafana/grafana/pull/97176) that call won't do anything so I guess we could leave it there for when/if we will ever enrich filters that way again?

mdvictor (approved):
LGTM!

packages/scenes/src/core/SceneScopesBridge.ts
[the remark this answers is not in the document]
tskarhed: We can't set this to private, as it is used when setting the context from `useScope`

packages/scenes/src/core/SceneScopesBridge.ts
```diff
@@ -0,0 +1,157 @@
...
+  public subscribeToValue(cb: (newScopes: Scope[], prevScopes: Scope[]) => void): Unsubscribable {
+    return this.contextObservable
+      .pipe(
+        filter((context) => !!context && !context.state.loading),
+        pairwise(),
+        map(
+          ([prevContext, newContext]) =>
+            [prevContext?.state.value ?? [], newContext?.state.value ?? []] as [Scope[], Scope[]]
+        ),
+        filter(([prevScopes, newScopes]) => !isEqual(prevScopes, newScopes))
```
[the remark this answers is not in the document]
tskarhed: Can we get around this by not filtering for undefined context above? I think having this safeguard can prevent unnecessary accidental updates. Maybe we can do a check if it is the first update here? We can do a `distinctUntilChanged` before mapping to the values. So it will end up emitting `[] and []`, but at least it will emit a value when going from an undefined context and the loading has finished..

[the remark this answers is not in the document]
aocenas: > when subscribeToValue triggers it compares prev and new scopes as being the same since it's the first value anyway.

I think you can do something like this though right:
```

current = SceneScopesBridge.context
SceneScopesBridge.subscribeToValue(() => do something with value)
```
I feel like both the `pairwise` and the `filter`s would mess up a bit the BehaviourSubject contract here. `pairwise` seem to need 2 events until it emits something, if we filter out undefined and loading context values this needs 3 changes to emit first time.

So I would agree that removing the first filter probably makes sense as you want to know if it was undefined before or goes to undefined value just has to be mapped right. Also as this suggests https://stackoverflow.com/questions/50059622/rxjs-observable-which-emits-both-previous-and-current-value-starting-from-first I would start it off with some empty value see this actually emits something from the start.

PR #993, opened 2024-12-05
SceneCSSGridLayout: Add support for dragging and dropping panels

kaydelaney:
There's a lot of code here but the gist of it is enables support for dragging and dropping panels in the responsive grid layout (when paired with the PR linked below)

Let me know if there are any questions or parts of the code that could use further explanation.

Related: https://github.com/grafana/grafana/pull/97495

### DIFF
--- a/packages/scenes/src/components/layout/CSSGrid/DragManager.tsx
+++ b/packages/scenes/src/components/layout/CSSGrid/DragManager.tsx
@@ -1,0 +1,250 @@
+import React from 'react';
+import { SceneObjectBase } from '../../../core/SceneObjectBase';
+import { SceneObject, SceneObjectState } from '../../../core/types';
+import { Rect, SceneCSSGridLayout } from './SceneCSSGridLayout';
+import { getPortalContainer } from '@grafana/ui';
+import { SceneCSSGridItem } from './SceneCSSGridItem';
+import { sceneGraph } from '../../../core/sceneGraph';
+import { createRoot, Root } from 'react-dom/client';
+
+export type DropZone = Rect & { layoutKey: string };
+
+interface DragManagerState extends SceneObjectState {
+  activeLayout?: SceneCSSGridLayout;
+  activeItem?: SceneObject;
+  dropZone?: DropZone;
+}
+
+export class DragManager extends SceneObjectBase<DragManagerState> {
+  private layouts: Record<string, SceneCSSGridLayout> = {};
+  public dropZones: DropZone[] = [];
+
+  private portalRoot!: Root;
+  private portal = React.createRef<HTMLDivElement>();
+  private preview = React.createRef<HTMLDivElement>();
+  private originLayout: SceneCSSGridLayout | undefined;
+
+  public constructor(state: DragManagerState) {
+    super(state);
+
+    this.onDrag = this.onDrag.bind(this);
+    this.onDragStart = this.onDragStart.bind(this);
+    this.onDragEnd = this.onDragEnd.bind(this);
+  }
+
+  public registerLayout(layout: SceneCSSGridLayout) {
+    this.layouts[layout.state.key!] = layout;
+  }
+
+  private oldBodyStyles = '';
+  public onDragStart(e: PointerEvent, layout: SceneCSSGridLayout, item: SceneObject) {
+    // find closest layout item
+    const layoutItem = sceneGraph.getAncestor(item, SceneCSSGridItem);
+    this.originLayout = layout;
+    const root = getPortalContainer().appendChild(document.createElement('div'));
+    this.portalRoot = createRoot(root);
+
+    document.addEventListener('pointermove', this.onDrag);
+    document.addEventListener('pointerup', this.onDragEnd);
+    this.oldBodyStyles = document.body.style.cssText;
+    document.body.style.cursor = 'move';
+    document.body.style.userSelect = 'none';
+
+    // request drop zones from registered layouts
+    const dropZones = [];
+    for (const l of Object.values(this.layouts)) {
+      dropZones.push(...l.getDropZones().map((v) => ({ ...v, layoutKey: l.state.key! })));
+    }
+    this.dropZones = dropZones;
+    const { closest, offset, scrollTop } = this.closestCell(this.dropZones, { x: e.clientX, y: e.clientY });
+
+    this.portalRoot.render(
+      <>
+        <div
+          style={{
+            position: 'fixed',
+            top: 0,
+            left: 0,
+            pointerEvents: 'none',
+            translate: `${offset.x}px ${offset.y}px`,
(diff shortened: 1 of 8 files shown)


torkelo:
Tried to test it with the css grid demo but just get an empty page 
Screenshot 2024-12-06 at 06 32 49

packages/scenes/src/components/layout/CSSGrid/SceneCSSGridItem.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export interface SceneCSSGridItemState extends SceneCSSGridItemPlacement, SceneObjectState {
+  body?: SceneObject;
+}
+
+export interface SceneCSSGridItemRenderProps<T> extends SceneComponentProps<T> {
+  parentState?: SceneCSSGridItemPlacement;
+}
+
+export class SceneCSSGridItem extends SceneObjectBase<SceneCSSGridItemState> {
```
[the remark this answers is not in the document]
kaydelaney: Oh this wasn't a POC, I just factored it out from SceneCSSGridLayout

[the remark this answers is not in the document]
torkelo: I understand, just know we are not using it dashboards. 

I think a challenge with this work is what logic needs live in scenes vs the core repo implementation (and the ResponsiveGridLayoutManager)

kaydelaney:
@torkelo The css grid demo should work now - made some fixes.

torkelo:
@kaydelaney works really well! 

I am not sure it helps or not but might be useful to go directly from this and explore how moving a panel between layout would work (as as an early exploratory poc/test, to understand the problem etc) .

torkelo:
and by "between layouts" i mean between two separate CSS grids in this case (moving between CSS grid and the default grid is a separate problem, one we might have to solve as well)

packages/scenes/src/components/layout/CSSGrid/SceneCSSGridLayout.tsx
```diff
@@ -57,126 +61,276 @@ export class SceneCSSGridLayout extends SceneObjectBase<SceneCSSGridLayoutState>
...
   }
 }
 
 function SceneCSSGridLayoutRenderer({ model }: SceneCSSGridItemRenderProps<SceneCSSGridLayout>) {
   const { children, isHidden, isLazy } = model.useState();
-  const style = useLayoutStyle(model.state);
+  const styles = useStyles2(getStyles, model.state);
+  const containerRef = useRef<HTMLDivElement>(null);
+  const oldDropZone = useRef<DropZone>();
+  const { activeItem, activeLayout, dropZone } = model.dragManager.useState();
```
mdvictor, line 147:
```suggestion
  const { activeItem, activeLayout, dropZone } = model.dragManager.useState() ?? {};
```

I think we need something like this for cases where the drag manager isn't set. For example if I go to the demos listing page [here](http://localhost:3000/a/grafana-scenes-app/demos) it crashes because it expects the drag manager to exist

kaydelaney: You're right! Sorry, I was so focused on the demo app that I'd forgotten to check it didn't break other things. Will fix this asap

packages/scenes/src/components/layout/CSSGrid/DragManager.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+export type DropZone = Rect & { layoutKey: string };
+
+interface DragManagerState extends SceneObjectState {
+  activeLayout?: SceneCSSGridLayout;
+  activeItem?: SceneObject;
+  dropZone?: DropZone;
+}
+
+export class DragManager extends SceneObjectBase<DragManagerState> {
```
torkelo, line 18: I think we should move as much as we can of the drag and drop stuff to core, (potentially the have a separate CSS grid layout here), might move it back here if there is need for this in other scene apps but think we can iterate on it faster if it's in core, and we already have a specific CSSGridItem there

kaydelaney:
Some additional context to hopefully help with reviewing:

My main goal with the design was to have as much of the DOM and event-level stuff in one place as possible, so we don’t have to reimplement the logic for a whole bunch of different layouts.
So to that end, there are two main components of the design: We have a `SceneLayout`, in this case `SceneCSSGridLayout`, and we have `DragManager`, a scene behavior that handles all the nitty-gritty dom manipulation and events.
The idea is that a layout registers itself with the `DragManager` and attaches the DragManager's `onDragStart` method to whatever bit of html they decide is the grid item's "handle".
When the `onDragStart` method is invoked, the `getDropZones` method of any registered layout is invoked, returning an array of rectangles representing valid places the dragged grid item could be placed.
Some math is then done to figure out which of these rectangles is closest to the current cursor position, which is then selected as the “placeholder” position.
Both the grid item placeholder and the grid item being dragged are put into a react portal while dragging is active.

Right now the code for `DragManager` is still specific to `SceneCSSGridLayout`, but after some details are ironed out and the requirements for other layouts reveals itself, I think we can make it more generically useful by changing the `SceneLayout` interface as needed and using that instead.

@mdvictor Let me know if any additional details would be helpful!

kaydelaney:
In preparation for moving this code over to the main grafana repo, I've created a much smaller version of this PR that just adds the plumbing needed for when things are moved. https://github.com/grafana/scenes/pull/1028

packages/scenes/src/components/layout/CSSGrid/SceneCSSGridLayout.tsx
```diff
@@ -57,126 +61,278 @@ export class SceneCSSGridLayout extends SceneObjectBase<SceneCSSGridLayoutState>
...
       ...state,
     });
+
+    this.onPointerDown = this.onPointerDown.bind(this);
+    this.addActivationHandler(() => {
+      const dragManager = findDragManager(this);
+      if (dragManager) {
+        this.dragManager = dragManager;
+        this.dragManager.registerLayout(this);
+      }
```
Sergej-Vlasov, line 71: getting typescript errors here. Might need a type cast in `findDragManager()` return
`Type 'SceneObject<SceneObjectState> | SceneStatelessBehavior<any>' is not assignable to type 'DragManager | undefined'.`

Sergej-Vlasov (commented):
Looks good, but getting the following errors and warnings in demo app when dragging:

`SceneObject already has a parent set that is different from the new parent.`
```
Uncaught TypeError: Cannot read properties of undefined (reading 'right')
at DragManager.onDrag (DragManager.js:110:62) ...
```

kaydelaney:
Grafana repo PR is up https://github.com/grafana/grafana/pull/99386

PR #995, opened 2024-12-06
JsonVariable: New type of object variable where each value user selects has many properties (POC)

torkelo:
The variable system has for a long time supported object property notation (Ex `${__org.name}`) 

Sadly we have not provided end users with the ability to define object variables. 

This PR explores a new type of variable (tentative name JsonVariable but it's a poor name).

Examples

```ts
new JsonVariable({
  name: 'env',
  value: 'test',
  provider: JsonVariableOptionProviders.fromString({
    json: `[
      { "id": 1, "name": "dev", "cluster": "us-dev-1", "status": "updating" }, 
      { "id": 2, "name": "prod", "cluster": "us-prod-2", "status": "ok" }
    ]`,
  }),
}),
new JsonVariable({
  name: 'testRun',
  label: 'Test run',
  provider: JsonVariableOptionProviders.fromObjectArray({
    options: [
      { runId: 'CAM-01', timeTaken: '10s', startTime: 1733492238318, endTime: 1733492338318 },
      { runId: 'SSL-02', timeTaken: '2s', startTime: 1733472238318, endTime: 1733482338318 }},
    ],
    valueProp: 'runId',
  }),
}),
```

Demo 
https://github.com/user-attachments/assets/f4c1c0b1-7510-4f23-b693-960fcfd8ec4b


packages/scenes-app/src/demos/jsonVariableDemo.tsx
```diff
@@ -0,0 +1,74 @@
...
+              name: 'testRun',
+              label: 'Test run',
+              value: 'test',
+              provider: JsonVariableOptionProviders.fromObjectArray({
+                options: [
+                  { runId: 'CAM-01', timeTaken: '10s', startTime: 1733492238318, endTime: 1733492338318 },
+                  { runId: 'SSL-02', timeTaken: '2s', startTime: 1733472238318, endTime: 1733482338318 },
+                  { runId: 'MRA-02', timeTaken: '13s', startTime: 1733462238318, endTime: 1733472338318 },
+                ],
+                valueProp: 'runId',
```
dprokop, line 43: so this will be the property used for default interpolation, when no path provided in the variable syntax? i.e. if i just used `${testRun}` the that would by default be equivalent to `${testRun.runId}`?

dprokop (commented):
I like this idea a lot. Not sure yet if this variable would make sense to be added to core, but starting with apps I think this is great.

PR #996, opened 2024-12-09
Variables: Interpolate datasource uid when used with datasource variable

Sergej-Vlasov:
Use case - use datasource variable as a datasource for ad hoc filter and in a dashboard query. 
so we have datasource variable `dsVar` and it equals to `prom-1` as example then:

1. ds variable in filter and panel (both `${dsVar}` )
2. ds variable in filter (`${dsVar}`) and named ds in panel (`prom-1`) that are the same
3. named variable in filter (`prom-1`) and ds variable in panel (`${dsVar}`) that are the same

Issues solved:
1. `findAndSubscribeToAdHocFilters` did not get interpolated datasource name.
2. ad hoc filter options would not filter down options based on query in dashboard therefore interpolated datasource uuids in `getQueriesForVariables` 

adding dashboard JSON that replicates issue:
[dsVar as filter datasource-1733769038417.json](https://github.com/user-attachments/files/18066201/dsVar.as.filter.datasource-1733769038417.json)

<details>
  <summary>📦 Published PR as canary version: <code>5.33.1--canary.996.12371059735.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.33.1--canary.996.12371059735.0
  npm install @grafana/scenes@5.33.1--canary.996.12371059735.0
  # or 
  yarn add @grafana/scenes-react@5.33.1--canary.996.12371059735.0
  yarn add @grafana/scenes@5.33.1--canary.996.12371059735.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.ts
@@ -445,7 +445,7 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
       const datasource = this.state.datasource ?? findFirstDatasource(queries);
       const ds = await getDataSource(datasource, this._scopedVars);
 
-      this.findAndSubscribeToAdHocFilters(datasource?.uid);
+      this.findAndSubscribeToAdHocFilters(ds.uid);
 
       const runRequest = getRunRequest();
       const { primary, secondaries, processors } = this.prepareRequests(timeRange, ds);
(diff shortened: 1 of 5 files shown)


packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -445,7 +445,7 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
       const datasource = this.state.datasource ?? findFirstDatasource(queries);
       const ds = await getDataSource(datasource, this._scopedVars);
 
-      this.findAndSubscribeToAdHocFilters(datasource?.uid);
+      this.findAndSubscribeToAdHocFilters(ds.uid);
```
torkelo, line 448: this could break existing logic, needs to be the same that was specified in SceneQueryRunner config (the string that has the variable expression), here you are just passing in the uid 

so if the ad-hoc filter has a variable expression as data source then this change will break that lookup

torkelo (approved):

packages/scenes/src/variables/utils.test.ts
```diff
@@ -176,6 +179,91 @@ describe('getQueriesForVariables', () => {
...
+  beforeAll(() => {
+    console.error = jest.fn();
+  });
+
+  afterAll(() => {
+    console.error = original;
+    jest.resetAllMocks();
+  });
+
+  it.only('should get queries for interpolated source object and query datasource uuids', () => {
```
ashharrison90, line 219: @Sergej-Vlasov stray `.only` here that needs removing. there is an eslint plugin (`eslint-plugin-jest`) and a corresponding rule (`jest/no-focused-tests`) that you should add as well to prevent this in the future 🙏

Sergej-Vlasov: good catch! Fix here https://github.com/grafana/scenes/pull/1008

Will follow up with the `eslint-plugin-jest` issue because it flags around 80 errors

PR #999, opened 2024-12-12
sceneUtils: expose setWindowGrafanaSceneContext

gtk-grafana:
So scene apps can set global timeSrv to enable copying time range

Potential usage:

```
  keybindings.addBinding({
    key: 't c',
    onTrigger: () => {
      const timeRange = sceneGraph.getTimeRange(scene)
      sceneUtils.setWindowGrafanaSceneContext(timeRange)
      appEvents.publish(new CopyTimeEvent());
    }
  })
```

### DIFF
--- a/packages/scenes/src/index.ts
+++ b/packages/scenes/src/index.ts
@@ -16,6 +16,7 @@ import {
   isIntervalVariable,
   isGroupByVariable,
 } from './variables/variants/guards';
+import { setWindowGrafanaSceneContext } from './utils/compatibility/setWindowGrafanaSceneContext';
 
 export * from './core/types';
 export * from './core/events';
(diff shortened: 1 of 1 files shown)


gtk-grafana:
I'm interested in finding solutions to automatically add keybindings for all scene apps: ideally every scene app would support copy/paste of time range automatically, or maybe by setting a new prop on `SceneTimeRangeState` in case apps have several time ranges.

torkelo:
setWindowGrafanaSceneContext is best called on-scene activation (already called automatically by EmbeddedScene) rather than in a command handler

gtk-grafana:
Thanks @torkelo, I had missed that this was already called in the EmbeddedScene, but it appears that doesn't always work when the time range is set on a child scene of the Embedded Scene. While we might be able to move our $timeRange up and get this working as expected, is this still something we'd want exposed to scene apps for cases where the time range that should get copied is not set when the Embedded scene is activated?

PR #1000, opened 2024-12-13
fix: wrap groupBy single-select value in an array

joannaWebDev:
When trying to use the Single select in db-o11y, realized we were sending incorrect data format. 

` "error": "json: cannot unmarshal string into Go struct field internalQueryModel.groupByKeys of type []string",
`

In this PR I'm wrapping the value in an array.

### DIFF
--- a/packages/scenes/src/variables/groupby/GroupByVariable.tsx
+++ b/packages/scenes/src/variables/groupby/GroupByVariable.tsx
@@ -345,7 +345,6 @@ export function GroupByVariableRenderer({ model }: SceneComponentProps<MultiValu
           model.changeValueTo([newValue.value]);
         }
       }}
-
       onOpenMenu={async () => {
         setIsFetchingOptions(true);
         await lastValueFrom(model.validateAndUpdate());
--- a/packages/scenes/src/variables/groupby/GroupByVariable.tsx
+++ b/packages/scenes/src/variables/groupby/GroupByVariable.tsx
@@ -342,9 +342,10 @@ export function GroupByVariableRenderer({ model }: SceneComponentProps<MultiValu
         }
         if (newValue?.value) {
           setUncommittedValue([newValue]);
-          model.changeValueTo(newValue.value, newValue.label);
+          model.changeValueTo([newValue.value]);
         }
       }}
+
       onOpenMenu={async () => {
         setIsFetchingOptions(true);
         await lastValueFrom(model.validateAndUpdate());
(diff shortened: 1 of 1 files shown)


packages/scenes/src/variables/groupby/GroupByVariable.tsx
```diff
@@ -342,7 +342,7 @@ export function GroupByVariableRenderer({ model }: SceneComponentProps<MultiValu
         }
         if (newValue?.value) {
           setUncommittedValue([newValue]);
-          model.changeValueTo(newValue.value, newValue.label);
+          model.changeValueTo([newValue.value]);
```
ashharrison90, line 345: think this should be `model.changeValueTo([newValue.value], [newValue.label]);` 🤔

ashharrison90 (commented):
question

packages/scenes/src/variables/groupby/GroupByVariable.tsx
joannaWebDev: oops, dunno why i removed it. fixed

packages/scenes/src/variables/groupby/GroupByVariable.tsx
```diff
@@ -340,9 +340,9 @@ export function GroupByVariableRenderer({ model }: SceneComponentProps<MultiValu
           }
           return;
         }
-        if (newValue?.value) {
+        if (newValue?.value && newValue?.label) {
```
ashharrison90, line 343: don't think you want this extra condition. label is optional

ashharrison90: maybe something like:
```tsx
        if (newValue?.value) {
          setUncommittedValue([newValue]);
          model.changeValueTo([newValue.value], newValue.label ? [newValue.label] : undefined);
        }
```
would work then

joannaWebDev: done

packages/scenes/src/variables/groupby/GroupByVariable.tsx
```diff
@@ -342,7 +342,10 @@ export function GroupByVariableRenderer({ model }: SceneComponentProps<MultiValu
         }
         if (newValue?.value) {
           setUncommittedValue([newValue]);
-          model.changeValueTo(newValue.value, newValue.label);
+          if (newValue?.value) {
+            setUncommittedValue([newValue]);
+            model.changeValueTo([newValue.value], newValue.label ? [newValue.label] : undefined);
+          }
```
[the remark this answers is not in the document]
joannaWebDev: i am so sorry

ashharrison90 (approved):
lgtm 👍

packages/scenes/src/variables/groupby/GroupByVariable.tsx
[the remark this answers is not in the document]
ashharrison90: no problem 🙂

PR #1004, opened 2024-12-16
AdHocFiltersVariable: provide updateFilters method to allow updating filters without emitting SceneVariableValueChangedEvent

gtk-grafana:
Use case: We have 2 ad-hoc variables at different scopes in the application, an "editing" variable at a lower scope, and a "submitted" variable at a higher scope. Both variables are interpolated by certain query providers. Our un-interpolated expression would look like this: `${AD_HOC_SUBMITTED} ${AD_HOC_EDITING}`. The submitted variable is rendered on the top scene, and on every route, but the "editing" variable only exists for a single route. Both variables are rendered in different areas of the UI.

When users are done configuring the lower scope variable, it is submitted, and moved to the higher scope variable. The value in the editing variable is deleted.

However since the `SceneVariableValueChangedEvent` is emitted whenever an individual ad-hoc variable's expression has changed, this triggers duplicate queries in our application, even though the final interpolated expression doesn't change.

Usage: 
```
// Move filter up
submittedVariable.updateFilters([ {key: 'key', operator: '=', value: 'val'} ], {skipPublish: true});

// Delete lower variable filters
editingVariable.updateFilters([], {
  skipPublish: true
});
```

Why this change:
Otherwise applications will either need to live with running duplicate queries, extend the base `AdHocFiltersVariable` class, or convert every query provider that consumes these variables to manual execution.

TL;DR:
This PR will allow the caller of updateFilters to override the default behavior and skip sending the `SceneVariableValueChangedEvent`, to allow setting the ad-hoc filters state without triggering queries.

# Release notes
New AdHocFiltersVariable method `updateFilters` to allow updating filters state. Allows skipping emit of `SceneVariableValueChangedEvent` to prevent filter changes from notifying dependent scene objects.

<details>
  <summary>📦 Published PR as canary version: <code>5.36.0--canary.1004.12431956361.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.36.0--canary.1004.12431956361.0
  npm install @grafana/scenes@5.36.0--canary.1004.12431956361.0
  # or 
  yarn add @grafana/scenes-react@5.36.0--canary.1004.12431956361.0
  yarn add @grafana/scenes@5.36.0--canary.1004.12431956361.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -213,7 +213,7 @@ export class AdHocFiltersVariable
     filters: AdHocFilterWithLabels[],
     options?: {
       skipPublish?: boolean;
-      forcePublish?: boolean
+      forcePublish?: boolean;
     }
   ): void {
     let filterExpressionChanged = false;
@@ -229,7 +229,7 @@ export class AdHocFiltersVariable
       filterExpression,
     });
 
-    if (filterExpressionChanged && options?.skipPublish !== true || options?.forcePublish) {
+    if ((filterExpressionChanged && options?.skipPublish !== true) || options?.forcePublish) {
       this.publishEvent(new SceneVariableValueChangedEvent(this), true);
     }
   }
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -190,18 +190,7 @@ export class AdHocFiltersVariable
   }
 
   public setState(update: Partial<AdHocFiltersVariableState>): void {
-    let filterExpressionChanged = false;
-
-    if (update.filters && update.filters !== this.state.filters && !update.filterExpression) {
-      update.filterExpression = renderExpression(this.state.expressionBuilder, update.filters);
-      filterExpressionChanged = update.filterExpression !== this.state.filterExpression;
-    }
-
-    super.setState(update);
-
-    if (filterExpressionChanged) {
-      this.publishEvent(new SceneVariableValueChangedEvent(this), true);
-    }
+    this.updateFilters(update)
   }
 
   /**
@@ -210,23 +199,19 @@ export class AdHocFiltersVariable
    * allowing consumers to update the filters without triggering dependent data providers.
    */
   public updateFilters(
-    filters: AdHocFilterWithLabels[],
+    update: Partial<AdHocFiltersVariableState>,
     options?: {
       skipPublish?: boolean;
     }
   ): void {
     let filterExpressionChanged = false;
-    let filterExpression = undefined;
 
-    if (filters && filters !== this.state.filters) {
-      filterExpression = renderExpression(this.state.expressionBuilder, filters);
-      filterExpressionChanged = filterExpression !== this.state.filterExpression;
+    if (update.filters && update.filters !== this.state.filters) {
+      update.filterExpression = renderExpression(this.state.expressionBuilder, update.filters);
+      filterExpressionChanged = update.filterExpression !== this.state.filterExpression;
     }
 
-    super.setState({
-      filters,
-      filterExpression,
-    });
+    super.setState(update);
 
     if (filterExpressionChanged && options?.skipPublish !== true) {
       this.publishEvent(new SceneVariableValueChangedEvent(this), true);
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -202,18 +202,19 @@ export class AdHocFiltersVariable
     update: Partial<AdHocFiltersVariableState>,
     options?: {
       skipPublish?: boolean;
+      forcePublish?: boolean;
     }
   ): void {
     let filterExpressionChanged = false;
 
-    if (update.filters && update.filters !== this.state.filters) {
+    if (update.filters && update.filters !== this.state.filters && !update.filterExpression) {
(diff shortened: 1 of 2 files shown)


packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -189,7 +189,7 @@ export class AdHocFiltersVariable
     }
   }
 
-  public setState(update: Partial<AdHocFiltersVariableState>): void {
+  public setState(update: Partial<AdHocFiltersVariableState>, options?: { skipPublish?: boolean }): void {
```
dprokop, line 192: this is quite badly named public interface :( I think for what you are trying to achieve (and probably most filters consumers) we should simply have `updateFilters` that would allow filters and filterExpression modification, with the options to skip publishing the event. Feels bad that the `setState`, a standard scene object method, is used for that and it's interface becomes different from the std one.  

I'd suggest refactoring AdhocFilters interface to reflect the above and add release notes to the new version of scenes that this is changing. That would be in line with i.e. what we have for `MultiValueVariable.changeValueTo` that is en explicit method for changing the value and publishing the relevant event.

torkelo: yea, a a new updateFilters function with a signature like 

```ts
updateFilters(filters: AdHocFilterWithLabels[], options?: { skipPublish?: boolean })
```

looks better

gtk-grafana: That makes a lot of sense, thank you both for the feedback! These changes have been implemented and pushed up.

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -204,6 +204,35 @@ export class AdHocFiltersVariable
     }
   }
 
+  /**
+   * Updates the variable's `filters` and `filterExpression` state.
+   * If `skipPublish` option is true, this will not emit the `SceneVariableValueChangedEvent`,
+   * allowing consumers to update the filters without triggering dependent data providers.
+   */
+  public updateFilters(
```
Sergej-Vlasov, line 212: Wondering if we should adjust updateFilters to:

```typescript
 public updateFilters(
    update: Partial<AdHocFiltersVariableState>,
    options?: {
      skipPublish?: boolean;
    }
  )....
```
and then reuse that in the setState so we don't get out of sync on functions and have a single source of truth for updating
```typescript
  public setState(update: Partial<AdHocFiltersVariableState>): void {
    this.updateFilters(update);
  }
```

gtk-grafana: Good catch, this change was implemented in https://github.com/grafana/scenes/pull/1004/commits/cdffe34b5409a96ede19be4635b1e1ec1d2630cc

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -190,16 +190,30 @@ export class AdHocFiltersVariable
...
+    update: Partial<AdHocFiltersVariableState>,
+    options?: {
+      skipPublish?: boolean;
+    }
+  ): void {
     let filterExpressionChanged = false;
 
-    if (update.filters && update.filters !== this.state.filters && !update.filterExpression) {
+    if (update.filters && update.filters !== this.state.filters) {
```
Sergej-Vlasov, line 209: @gtk-grafana looks like accidental delete of ` && !update.filterExpression`

gtk-grafana: Yeah, it was not so accidental in updateFilters, and now I realize my mistake in DRYing setState and updateFilters. With the updated merge the expression will have already been updated when the event emit is skipped, and then when we actually want to emit the event, it won't send because the filter expression is unchanged.

I had to add an additional option to force the publish of the event, which should restore prior functionality to setState.

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
dprokop: I disagree with this. `updateFilters` should update filters, not entire state of the object. What i think is needed here is to remove the `setState` overload all together. If someone want's to update filters, then we have an API for that. If someone wants to update the state, then they can rely on the base class's `setState` (without side effect of the event being published). This is a breaking change, hence i suggested adding release notes :)

gtk-grafana: Thanks @dprokop, the setState overload has been reverted, but I kept the option to force the publish in the updateFilters.

dprokop (approved):
LGTM. I think the `setState` method overload in filters should be _completely removed_, it currently duplicates the `updateFilters`, but without the options. 

Since we have the `updateFilters` function, it's this one that should be one for updating that particular state. The `setState` is inherited by default from `SceneObjectBase` and it should simply update the state with no effects. 

I approve this work to not block it during xmas, but I think we can do the removal in this PR given that we are notifying about this change in the release notes.

gtk-grafana:
@dprokop if we want to update and use this PR to introduce that breaking change after the holidays that's fine with me, I'm personally not really blocked as I extended the base `AdHocFiltersVariable` in Explore Logs to add this new updateFilters method until we can upgrade to a version of scenes that supports this functionality.

To make sure I'm understanding your request, the `setState` method on `AdHocFiltersVariable` should be completely removed, and everything should be handled in `updateFilters`? Then `updateFilters` can call `this.setState` instead of calling the super.setState on `SceneObjectBase`?

gtk-grafana:
I lied, I do need this to get unblocked, we can follow this up in another PR.

PR #1010, opened 2024-12-19
fix: groupBy manual applyMode

joannaWebDev:
This PR fixes a bug in **the GroupByVariable where the applyMode property was not being respected**. Specifically, the addActivationHandler logic ran regardless of the applyMode value, causing unintended behavior.

To address this, we introduced a conditional check for `this.state.applyMode === 'auto'` before invoking addActivationHandler. This ensures that activation logic only applies when the applyMode is set to auto.

Changes have been tested locally using `npm link --force ../scenes/packages/scenes`.

**Demo**

https://github.com/user-attachments/assets/7c13836a-416d-4213-83d2-df3332f46655

### DIFF
--- a/packages/scenes/src/variables/groupby/GroupByVariable.tsx
+++ b/packages/scenes/src/variables/groupby/GroupByVariable.tsx
@@ -144,11 +144,13 @@ export class GroupByVariable extends MultiValueVariable<GroupByVariableState> {
       noValueOnClear: true,
     });
 
-    this.addActivationHandler(() => {
-      allActiveGroupByVariables.add(this);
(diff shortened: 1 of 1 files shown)


joannaWebDev:
> Hey @joannaWebDev 😊, great work with this one 🏅 , can you add a unit test for this scenario, https://github.com/grafana/scenes/blob/main/packages/scenes/src/variables/groupby/GroupByVariable.test.tsx

Sure thing. Done

axelavargas (approved):
🥇 Great work, @joannaWebDev !

PR #1011, opened 2024-12-19
Template Variables: Use Combobox behind toggle

tskarhed:
It turns out that Combobox actually isn't exported until 11.5. I would really appreciate some feedback on the way of handling feature toggling here. This PR will likely be WIP until mid-January.

Try running with grafana/ui at `11.5.0-214403`.

### DIFF
--- a/packages/scenes/src/variables/components/VariableValueSelect.tsx
+++ b/packages/scenes/src/variables/components/VariableValueSelect.tsx
@@ -9,6 +9,7 @@ import {
   getSelectStyles,
   useStyles2,
   useTheme2,
+  Combobox,
 } from '@grafana/ui';
 
 import { SceneComponentProps } from '../../core/types';
@@ -47,12 +48,37 @@ export function toSelectableValue<T>(value: T, label?: string): SelectableValue<
 }
 
 export function VariableValueSelect({ model }: SceneComponentProps<MultiValueVariable>) {
-  const { value, text, key, options, includeAll, isReadOnly, allowCustomValue = true } = model.useState();
+  const {
+    value,
+    text,
+    key,
+    options,
+    includeAll,
+    isReadOnly,
+    allowCustomValue = true,
+    renderWithCombobox,
+  } = model.useState();
   const [inputValue, setInputValue] = useState('');
   const [hasCustomValue, setHasCustomValue] = useState(false);
   const selectValue = toSelectableValue(value, String(text));
 
   const optionSearcher = useMemo(() => getOptionSearcher(options, includeAll), [options, includeAll]);
+
+  if (renderWithCombobox) {
+    const comboboxOptions = options.map((o) => ({ value: o.value.toString(), label: o.label }));
+    return (
+      <Combobox
+        value={value.toString()}
+        width="auto"
+        minWidth={20}
+        options={comboboxOptions}
+        createCustomValue={allowCustomValue}
+        onChange={(newValue) => {
+          model.changeValueTo(newValue.value!, newValue.label!);
+        }}
+      />
+    );
+  }
 
   const onInputChange = (value: string, { action }: InputActionMeta) => {
     if (action === 'input-change') {


packages/scenes/src/variables/components/VariableValueSelect.tsx
```diff
@@ -47,13 +48,38 @@ export function toSelectableValue<T>(value: T, label?: string): SelectableValue<
...
+        minWidth={10}
+        options={comboboxOptions}
+        createCustomValue={allowCustomValue}
+        onChange={(newValue) => {
+          model.changeValueTo(newValue.value, newValue.label);
+        }}
+      />
+    );
+  }
+
```
Sergej-Vlasov, line 82: @tskarhed  I think instead of adding this here it would be cleaner to add this in `renderSelectForVariable` function

packages/scenes/src/variables/components/VariableValueSelect.tsx
```diff
@@ -256,7 +258,36 @@ const getOptionStyles = (theme: GrafanaTheme2) => ({
...
+  let comboboxOptions = options.map((o) => ({ value: o.value.toString(), label: o.label }));
+
+  if (includeAll) {
+    comboboxOptions.unshift({ value: ALL_VARIABLE_VALUE, label: ALL_VARIABLE_TEXT });
+  }
+
+  return (
+    <Combobox
+      id={'var-' + key}
```
tskarhed, line 272: Not sure where this `var-` is added for the Select

packages/scenes/src/variables/components/VariableValueSelect.tsx
```diff
@@ -256,7 +258,36 @@ const getOptionStyles = (theme: GrafanaTheme2) => ({
   }),
 });
 
+function VariableValueCombobox({ model }: SceneComponentProps<MultiValueVariable>) {
+  const { value, key, options, includeAll, isReadOnly, allowCustomValue = true } = model.useState();
+
+  let comboboxOptions = options.map((o) => ({ value: o.value.toString(), label: o.label }));
+
+  if (includeAll) {
+    comboboxOptions.unshift({ value: ALL_VARIABLE_VALUE, label: ALL_VARIABLE_TEXT });
```
tskarhed, line 267: Are custom All values handled somewhere else?

packages/scenes/src/variables/components/VariableValueSelect.tsx
```diff
@@ -256,7 +258,36 @@ const getOptionStyles = (theme: GrafanaTheme2) => ({
   }),
 });
 
+function VariableValueCombobox({ model }: SceneComponentProps<MultiValueVariable>) {
+  const { value, key, options, includeAll, isReadOnly, allowCustomValue = true } = model.useState();
+
+  let comboboxOptions = options.map((o) => ({ value: o.value.toString(), label: o.label }));
```
tskarhed, line 264: Do options always have a label in this context?

tskarhed: Memoize this

packages/scenes/src/variables/components/VariableValueSelect.tsx
```diff
@@ -256,7 +260,61 @@ const getOptionStyles = (theme: GrafanaTheme2) => ({
...
+
+      console.log('search change ', newInputValue);
+
+      // Same functionality as in _updateOptionsBasedOnSearchFilter, although deboucning is managed in Combobox
+      return (await lastValueFrom(model.getValueOptions({ searchFilter: newInputValue }))).map((o) => ({
+        value: o.value.toString(),
+        label: o.label,
+      }));
+    };
+  }, [model]);
```
tskarhed, line 291: This always seem to return all options. Unclear why.

packages/scenes/src/variables/components/VariableValueSelect.tsx
```diff
@@ -256,7 +275,61 @@ const getOptionStyles = (theme: GrafanaTheme2) => ({
...
+
+        model.changeValueTo(newValue.value, newValue.label);
+      }}
+      loading={loading}
+    />
+  );
+}
+
 export function renderSelectForVariable(model: MultiValueVariable) {
+  if (model.state.renderWithCombobox && !model.state.isMulti) {
```
torkelo, line 329: instead of a feature toggle in state, which is very hard to remove without a breaking change, can we use a regular feature toggle instead?

tskarhed: That's what I was hoping was possible. How do we use feature toggles in Scenes?

packages/scenes/src/variables/components/VariableValueSelect.tsx
```diff
@@ -256,7 +275,59 @@ const getOptionStyles = (theme: GrafanaTheme2) => ({
...
+    return async (newInputValue: string) => {
+      model.onSearchChange!(newInputValue);
+
+      // Same functionality as in _updateOptionsBasedOnSearchFilter, although deboucning is managed in Combobox
+      return (await lastValueFrom(model.getValueOptions({ searchFilter: newInputValue }))).map((o) => ({
+        value: o.value.toString(),
+        label: o.label,
+      }));
+    };
+  }, [model]);
```
tskarhed, line 303: I don't think this is working properly, and I lack the insight into how value updates are made in Scenes to make an informed decision

packages/scenes/src/variables/variants/query/QueryVariable.tsx
```diff
@@ -140,23 +141,32 @@ export class QueryVariable extends MultiValueVariable<QueryVariableState> {
...
-  private _updateOptionsBasedOnSearchFilter = debounce(async (searchFilter: string) => {
+  // Used for Combobox, as debouncing is done inside the component
+  private _getNewOptions = async (searchFilter: string) => {
     const result = await lastValueFrom(this.getValueOptions({ searchFilter }));
     this.setState({ options: result, loading: false });
-  }, 400);
+    return result;
+  };
+
+  private _updateOptionsBasedOnSearchFilter = debounce(this._getNewOptions, 400, { leading: true });
```
tskarhed, line 162: `leading` is needed to not return undefined. If this is desired behavior is up for discussion.

packages/scenes/src/variables/components/VariableValueSelect.tsx
```diff
@@ -256,7 +277,69 @@ const getOptionStyles = (theme: GrafanaTheme2) => ({
...
+  const onInputChange = useMemo(() => {
+    if (!model.onSearchChange) {
+      return;
+    }
+
+    // If we don't use searchFilter in QueryVariable, don't bother to make a query. TODO: Proper typing
+    // @ts-ignore
+    if ('query' in model.state && !containsSearchFilter(model.state.query)) {
+      return;
+    }
```
tskarhed, line 300: Some feedback to do some proper typing on this would be nice. `query` only exists on a subclass.

packages/scenes/src/variables/components/VariableValueSelect.tsx
```diff
@@ -256,7 +277,69 @@ const getOptionStyles = (theme: GrafanaTheme2) => ({
...
+
+    return async (newInputValue: string) => {
+      const res = model.onSearchChange!(newInputValue).then((result) => {
+        return result.map((o) => ({
+          value: o.value.toString(),
+          label: o.value.toString() === ALL_VARIABLE_VALUE ? ALL_VARIABLE_TEXT : o.label,
```
tskarhed, line 306: Unclear how `All` should be treated in this case.

packages/scenes/src/variables/components/VariableValueSelect.tsx
```diff
@@ -188,6 +205,10 @@ export function VariableValueSelectMulti({ model }: SceneComponentProps<MultiVal
         if (action.action === 'clear' && noValueOnClear) {
           model.changeValueTo([]);
         }
+
+        const allSelected = newValue.find((v) => v.value === '$__all');
```
tskarhed, line 209: Consider using a `Set`

packages/scenes/src/variables/components/VariableValueSelect.tsx
torkelo: You can simply call the function and use the same logic we have above (for the old select )

packages/scenes/src/variables/components/VariableValueSelect.tsx
```diff
@@ -256,7 +277,69 @@ const getOptionStyles = (theme: GrafanaTheme2) => ({
...
+      return;
+    }
+
+    // If we don't use searchFilter in QueryVariable, don't bother to make a query. TODO: Proper typing
+    // @ts-ignore
+    if ('query' in model.state && !containsSearchFilter(model.state.query)) {
+      return;
+    }
+
+    return async (newInputValue: string) => {
```
torkelo, line 302: Think you can remove this is you simply called the function, like the current impl

packages/scenes/src/variables/variants/query/QueryVariable.tsx
```diff
@@ -140,23 +141,33 @@ export class QueryVariable extends MultiValueVariable<QueryVariableState> {
...
+      return Promise.resolve([]); // Not called for Combobox, Select will treat this as void
     }
 
-    this._updateOptionsBasedOnSearchFilter(searchFilter);
+    // @ts-ignore
+    if (config?.featureToggles?.templateVariablesUsesCombobox) {
+      return this._getNewOptions(searchFilter);
+    }
+
+    // Undefined for first call
```
torkelo, line 152: What is undefined?

torkelo (commented):
It seems hang / crash a browser when there are many options (many values) returned by the variable query: 

In this demo example: 
http://localhost:3000/a/grafana-scenes-app/demos/variables/many-values?var-interval=1d&from=now-6h&to=now&timezone=browser&var-server=AA&var-pod=AAA&var-handler=pod&var-keyValue=1&var-queryVar=A&var-namespace=broken&__feature.templateVariablesUsesCombobox&var-manyOptions=CEeAKU5LETNWXReRPQY7IddhyHDK5eYQi2kdATodj6m4AYPHJ8

Tested this a bit now with a prometheus variable query that uses __searchFilter, 

With this query:
```ts
  new QueryVariable({
                  name: 'namespace',
                  query: {
                    qryType: 1,
                    query: 'label_values({__name__=~"${__searchFilter}.*"},__name__)',
                    refId: 'PrometheusVariableQueryEditor-VariableQuery',
                  },
                  datasource: { uid: 'se-demo' },
                }),
```

The call to onSearchChange and getValueOptions causes double queries (at least on first open). 

Dan we really need to move the loading state from the label to input as well causing the input to jump around (not related to this PR, just an observation)

tskarhed:
> It seems hang / crash a browser when there are many options (many values) returned by the variable query:

There is a possibility this is fixed with the latest changes to Combobox. This seems to only be an issue when first opened... 
Something is creating event listeners, as they are close to 100k, looking into it.

tskarhed:
> It seems hang / crash a browser when there are many options (many values) returned by the variable query:
> 
> In this demo example: http://localhost:3000/a/grafana-scenes-app/demos/variables/many-values?var-interval=1d&from=now-6h&to=now&timezone=browser&var-server=AA&var-pod=AAA&var-handler=pod&var-keyValue=1&var-queryVar=A&var-namespace=broken&__feature.templateVariablesUsesCombobox&var-manyOptions=CEeAKU5LETNWXReRPQY7IddhyHDK5eYQi2kdATodj6m4AYPHJ8

A ton of event listeners are being created. I have no idea why. The basic use case is handled without any issues directly in Combobox.

image

torkelo:
@tskarhed yea, it's strange, the combobox looks to recursive be rendering itself or something. 

thought it has something to do with passing options via the onInputChange but commented and it still hangs the same way

tskarhed:
I still don't know exactly why this issue appears here and nowhere else. However, it seems like the initial render left the maxHeight undefined, hence the list virtualization never kicked in. I fixed this by setting arbitrary initial dimensions (which are overridden soon after measuring things).

https://github.com/grafana/grafana/pull/100188

tskarhed:
> With this query:
> 
> ```ts
>   new QueryVariable({
>                   name: 'namespace',
>                   query: {
>                     qryType: 1,
>                     query: 'label_values({__name__=~"${__searchFilter}.*"},__name__)',
>                     refId: 'PrometheusVariableQueryEditor-VariableQuery',
>                   },
>                   datasource: { uid: 'se-demo' },
>                 }),
> ```
> 
> The call to onSearchChange and getValueOptions causes double queries (at least on first open).

Posting this recording for reference.

https://github.com/user-attachments/assets/b3ae35a1-c2e0-4e7e-8b2e-cd1f47573965

It doesn't seem to happen when having an empty value. I'm looking into it.

tskarhed:
@torkelo A call in onOpen in Combobox isn't debounced. A PR is open for it now https://github.com/grafana/grafana/pull/100219

Other than that, is there anything blocking this PR, or do you feel okay approving it so we can start testing it in our instances? I assume the debounce PR will be in order soon enough.

packages/scenes/src/variables/components/VariableValueSelect.tsx
```diff
@@ -256,7 +278,83 @@ const getOptionStyles = (theme: GrafanaTheme2) => ({
   }),
 });
 
+function VariableValueCombobox({ model }: SceneComponentProps<MultiValueVariable>) {
+  const { value, key, isReadOnly, options, includeAll, allowCustomValue = true, loading } = model.useState();
+
+  const isAsync = useMemo(
+    () =>
+      //@ts-ignore
+      model.onSearchChange && 'query' in model.state && containsSearchFilter(model.state.query),
```
torkelo, line 287: We just need to fix this now, and the empty promise return of onSearchChange.

packages/scenes/src/variables/variants/query/QueryVariable.tsx
tskarhed: THe return value. This is because of the debounce. It can be fixed by setting `{ leading: true }`

packages/scenes/src/variables/components/VariableValueSelect.tsx
tskarhed: Although it is a bit misnamed, as `QueryVariable` always makes at least one async request.

PR #1012, opened 2024-12-20
AdHocFilters: add hidden option 

mikkancso:
We'd like to add a search input field to a page which already has AdHocFilters with a restricted set of keywords. We have the following requirements:
  - when a search term is supplied, then all the queries on the page should be filtered as if there was a filter added for the a specific, fixed label. More precisely it should be equivalent to adding `specific_label=~.*searchTerm.*`
  - but we don't want this new filter to be shown by the AdHocFilter control (we currently use VariableValueControl for that) or the URL

This PR enables us to achieve this by introducing the `hidden` property on AdHocFilters.

Related discussion: https://raintank-corp.slack.com/archives/C04FRQ2S0AV/p1732712478401569


dprokop (commented):
Let's add tests for that and should be ok to go!

dprokop (approved):

PR #1015, opened 2024-12-24
QueryVariable: Use correct option property for variable options sorting

Sergej-Vlasov:
sorting was carried out agains `option.text` instead of `option.label`

<details>
  <summary>📦 Published PR as canary version: <code>5.36.2--canary.1015.12584248290.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.36.2--canary.1015.12584248290.0
  npm install @grafana/scenes@5.36.2--canary.1015.12584248290.0
  # or 
  yarn add @grafana/scenes-react@5.36.2--canary.1015.12584248290.0
  yarn add @grafana/scenes@5.36.2--canary.1015.12584248290.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/variants/query/utils.ts
+++ b/packages/scenes/src/variables/variants/query/utils.ts
@@ -130,16 +130,10 @@ function sortByNumeric(opt: VariableValueOption) {
   }
 }
 
+const collator = new Intl.Collator(undefined, { sensitivity: 'accent', numeric: true });
+
 function sortByNaturalSort(options: VariableValueOption[]) {
   return options.sort((a, b) => {
-    if (!a.label) {
-      return -1;
-    }
-
-    if (!b.label) {
-      return 1;
-    }
-
-    return a.label.localeCompare(b.label, undefined, { numeric: true });
+    return collator.compare(a.label, b.label);
   });
 }
--- a/packages/scenes/src/variables/variants/query/utils.ts
+++ b/packages/scenes/src/variables/variants/query/utils.ts
@@ -133,7 +133,7 @@ function sortByNumeric(opt: VariableValueOption) {
 const collator = new Intl.Collator(undefined, { sensitivity: 'accent', numeric: true });
 
 function sortByNaturalSort(options: VariableValueOption[]) {
-  return options.sort((a, b) => {
+  return options.slice().sort((a, b) => {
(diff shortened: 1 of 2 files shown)


torkelo (approved):

packages/scenes/src/variables/variants/query/utils.ts
```diff
@@ -146,3 +117,29 @@ export const sortVariableValues = (options: any[], sortOrder: VariableSort) => {
...
+  return options.sort((a, b) => {
+    if (!a.label) {
+      return -1;
+    }
+
+    if (!b.label) {
+      return 1;
+    }
+
+    return a.label.localeCompare(b.label, undefined, { numeric: true });
```
leeoniya, line 143: localeCompare is a super slow api.

see https://github.com/grafana/grafana/pull/96862/files#diff-666e28cd8f4263c8acbeec76db74f0d091d0d7a4c1221e906f9c29325c195a61R547

leeoniya:
also, for numeric sort you can just subtract the values from each other and skip the if/conditions.

leeoniya:
i think a lot of the sorting funcs can be removed and/or simplified, like `sortByNumeric` can likely just rely on the same `numeric: true` collator. but not in this PR :)

packages/scenes/src/variables/variants/query/utils.ts
```diff
@@ -146,3 +117,23 @@ export const sortVariableValues = (options: any[], sortOrder: VariableSort) => {
...
+    return -1;
+  } else {
+    return parseInt(matches[1], 10);
+  }
+}
+
+const collator = new Intl.Collator(undefined, { sensitivity: 'accent', numeric: true });
+
+function sortByNaturalSort(options: VariableValueOption[]) {
+  return options.sort((a, b) => {
```
leeoniya, line 136: just double checking that you want to mutate original array here. instead of `options.slice().sort(...)`.

leeoniya (approved):

packages/scenes/src/variables/variants/query/utils.ts
Sergej-Vlasov: good catch. No harm here in modifying but other sorts use `sortBy` from lodash which returns new copy, so implementing the same for natural sort as well

PR #1023, opened 2025-01-08
improve demo app navigation and include  code highlighter

yduartep:
This PR includes:
- Improvement on the navigation when switching between the 3 different demo apps. Currently It's necessary to open the left nav in order to go to another demo app:

Screenshot 2025-01-08 at 19 25 22

Screenshot 2025-01-08 at 19 25 36

- Add code highlighter feature to watch the code associated to each demo, search by title and description and copy the code to clipboard.

Screenshot 2025-01-08 at 19 29 35

Screenshot 2025-01-10 at 16 20 06


packages/scenes-app/package.json
```diff
@@ -68,6 +72,7 @@
     "@grafana/schema": "^11.3.0",
     "@grafana/ui": "^11.3.0",
     "@types/lodash": "latest",
+    "prismjs": "^1.29.0",
```
dprokop, line 75: Please don't add this dependency. There's a `CodeEditor` component available from grafana/ui that you can use, sth like this, instead of introducing `HighlightCode` component:


```tsx
<Modal className={styles.modalContent} title={'Demo source code'} isOpen={isModalOpen} onDismiss={onModalClose}>
        <div className={styles.bodyContent}>
          <p>{text}</p>
          <div className={styles.wrapper}>
            <CodeEditor
              language="typescript"
              value={fileContent?.trim() ?? ''}
              readOnly={true}
              containerStyles={styles.codeEditor}
            />
          </div>
        </div>
        <Modal.ButtonRow>
          <Button variant="primary" fill="solid" onClick={onModalClose}>
            OK
          </Button>
        </Modal.ButtonRow>
      </Modal>
```


where corresponding styles are:
```tsx
 wrapper: css({
      display: 'flex',
      height: '100%',
      flexDirection: 'column',
      gap: theme.spacing(2),
    }),
    codeEditor: css({
      flexGrow: 1,
    }),
```

dprokop (changes requested):
Looks great, but let's get rid of unnecessary prismjs dependency.

packages/scenes-app/package.json
yduartep: Now I am using the CodeEditor component and I added a copy to clipboard button. I had to include a new function to update the height of the modal on each resize event so the code editor auto adjust the heigh. Let me know what do you think

packages/scenes-app/src/components/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -0,0 +1,35 @@
...
+  clipboardText: string;
+
+  onClipboardCopy(): void;
+
+  onClipboardError?(): void;
+
+  variant?: ButtonVariant;
+}
+
+export const CopyToClipboard = (props: Props) => {
```
dprokop, line 15: Another one that's available in grafana/ui : `ClipboardButton` :)

dprokop (changes requested):

dprokop (approved):

Issue #1027, opened 2025-01-20
CustomVariable with no options defined clears the current value when url value is same as saved value


torkelo:
Replicated here: 

https://play.grafana.org/d/aeakefqp551j4e/replicating-variable-url-sync-issue?orgId=1&from=now-6h&to=now&timezone=ut&var-test=customValue

https://play.grafana.org/d/aeakefqp551j4e/replicating-variable-url-sync-issue?orgId=1&from=now-6h&to=now&timezone=ut&var-test=customValue

because the saved value is the same as the URL value no call to updateFromUrl is called so the skipNextValidation is not set to true, this is the flag that allows URL values to override and set variable values that are not valid. 

https://github.com/grafana/scenes/blob/main/packages/scenes/src/variables/variants/MultiValueVariable.ts#L438

Not sure what the correct fix is here

torkelo:
https://github.com/grafana/support-escalations/issues/14254

PR #1033, opened 2025-01-23
Fix issue with custom variables losing url value when there are no options

mdvictor:
Attempts to fix https://github.com/grafana/scenes/issues/1027, not sure if it's the way to go or if it doesn't break other stuff that I might have missed. 

It seems that it only affects CustomVariables so we could isolate it there and turn on the `skipNextValidation` flag if there are no options and we have a value in the url.

### DIFF
--- a/packages/scenes/src/variables/variants/CustomVariable.tsx
+++ b/packages/scenes/src/variables/variants/CustomVariable.tsx
@@ -44,7 +44,7 @@ export class CustomVariable extends MultiValueVariable<CustomVariableState> {
       }
     });
 
-    if (!options.length && this.hasKeyInUrl()) {
+    if (!options.length && this.hasValueInUrl()) {
       this.skipNextValidation = true;
     }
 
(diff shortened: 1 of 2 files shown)


packages/scenes/src/variables/variants/CustomVariable.tsx
```diff
@@ -44,10 +44,21 @@ export class CustomVariable extends MultiValueVariable<CustomVariableState> {
...
   public static Component = ({ model }: SceneComponentProps<MultiValueVariable>) => {
     return renderSelectForVariable(model);
   };
+
+  private hasValueInUrl() {
+    const urlValues = this.urlSync?.getUrlState();
+    const urlKey = this.urlSync?.getKeys()?.find((key) => key === `var-${this.state.name}`);
+
+    return !!urlValues?.[urlKey ?? ''];
+  }
```
torkelo, line 63: not sure this does anything? it just checks if the a key specific in getKeys exists in getUrlState, wont this always be true for a variable that has a value? 

You might as well check:

```
if (!options.length && this.state.value) {
``` 

because custom variable is special (in the sense it's always designed to have the possible values hard coded), in this case only the starting value is hard coded, so it should be fine to always set skipNextValidation = true for a custom variable (I think)

mdvictor: oh you're right, i was pretty sure the state.value was empty but it's not. I'll test it out with `skipNextValidation` always set to true

torkelo (approved):

PR #1034, opened 2025-01-23
AdHocFilters: allow setting filter meta in getTagKeysProvider

gtk-grafana:
### Problem:
Assume the API endpoint that is called by `getTagKeysProvider` returns metadata needed for an ad-hoc filter value, that is not the key or key label. Currently, when defining `getTagKeysProvider` in an adHocVariable, we are only able to update the `key` or `keyLabel` props of the _wip ad-hoc filter. This means that we are forced to call the same API again in the `getTagValuesProvider` that we called in `getTagKeysProvider` to re-fetch the metadata associated with the selected key, or we must save the results of the API call made in `getTagKeysProvider` outside of the ad-hoc variable state.

### Specific use case: 
In Explore Logs, we call the `detected_fields` api endpoint to get keys for Fields variables. This returns the desired parser to use when building Loki queries, but isn't part of the actual value that is queried.

For example, when adding a `caller` field, the API tells us that the field only contains values that should be parsed by the logfmt parser. The api response for this field is:
```
{
    "label": "caller",
    "type": "string",
    "cardinality": 7,
    "parsers": [
        "logfmt"
    ]
}
```

When a value is selected for this field, we build a query like this:
```
{service_name=~"nginx-json-mixed"} | logfmt | caller=`broadcast.go:48`
```

Additionally, in the future we want to differentiate fields by "type", so we can provide a different set of operators for "string" typed fields vs "duration", "float", "int", "bytes", etc. We should be able to use the meta set in this PR to overwrite the `_getOperators` after selecting a key to keep the user from seeing operators which are not valid for that field type.

### This PR:
Attempts to solve the problem by allowing custom `meta` on the SelectableValue, which is saved to the adHoc `AdHocFilterWithLabels` filter state after the user selects a key provided by `getTagKeysProvider`.

### Limitations:
Currently passing meta from the `getTagValuesProvider` to the ad-hoc filter state is not supported by this PR, however it would be trivial to pass this in the `generateFilterUpdatePayload` method.

### Related PRs:
Allowing users to set valueLabels or encode meta with custom values: https://github.com/grafana/scenes/pull/1038

<details>
  <summary>📦 Published PR as canary version: <code>5.39.0--canary.1034.12990010339.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@5.39.0--canary.1034.12990010339.0
  npm install @grafana/scenes@5.39.0--canary.1034.12990010339.0
  # or 
  yarn add @grafana/scenes-react@5.39.0--canary.1034.12990010339.0
  yarn add @grafana/scenes@5.39.0--canary.1034.12990010339.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -17,7 +17,7 @@ import { getEnrichedFiltersRequest } from '../getEnrichedFiltersRequest';
 import { AdHocFiltersComboboxRenderer } from './AdHocFiltersCombobox/AdHocFiltersComboboxRenderer';
 import { wrapInSafeSerializableSceneObject } from '../../utils/wrapInSafeSerializableSceneObject';
 
-export interface AdHocFilterWithLabels<M = any> extends AdHocVariableFilter {
+export interface AdHocFilterWithLabels<M extends Record<string, any> = {}> extends AdHocVariableFilter {
   keyLabel?: string;
   valueLabels?: string[];
   // this is used to externally trigger edit mode in combobox filter UI
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -17,13 +17,16 @@ import { getEnrichedFiltersRequest } from '../getEnrichedFiltersRequest';
 import { AdHocFiltersComboboxRenderer } from './AdHocFiltersCombobox/AdHocFiltersComboboxRenderer';
 import { wrapInSafeSerializableSceneObject } from '../../utils/wrapInSafeSerializableSceneObject';
 
-export interface AdHocFilterWithLabels extends AdHocVariableFilter {
+export interface AdHocFilterWithLabels<M = any> extends AdHocVariableFilter {
   keyLabel?: string;
   valueLabels?: string[];
   // this is used to externally trigger edit mode in combobox filter UI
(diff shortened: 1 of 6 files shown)


packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -17,13 +17,14 @@ import { getEnrichedFiltersRequest } from '../getEnrichedFiltersRequest';
 import { AdHocFiltersComboboxRenderer } from './AdHocFiltersCombobox/AdHocFiltersComboboxRenderer';
 import { wrapInSafeSerializableSceneObject } from '../../utils/wrapInSafeSerializableSceneObject';
 
-export interface AdHocFilterWithLabels extends AdHocVariableFilter {
+export interface AdHocFilterWithLabels<M = any> extends AdHocVariableFilter {
```
torkelo, line 20:
```suggestion
export interface AdHocFilterWithLabels<M extends Record<string, any> = {}> extends AdHocVariableFilter {
```

It might be helpful to expect an object here

torkelo (approved):

dprokop (commented):
LGTM

> Additionally, in the future we want to differentiate fields by "type", so we can provide a different set of operators for "string" typed fields vs "duration", "float", "int", "bytes", etc. We should be able to use the meta set in this PR to overwrite the _getOperators after selecting a key to keep the user from seeing operators which are not valid for that field type.

@gtk-grafana - let's talk about this when you start looking at it. I'd love us to provide that possibility in core DataSourceAPI so that data sources can provide type information as well.

PR #1035, opened 2025-01-24
MultiValueVariable: Add `getDefaultSingleState` method

domasx2:
:wave:  have a need to select default variable option that isn't necessarily the first one. Refactored a little bit to expose `getDefaultSingleState` method for picking the default option so it can be easily overridden, in the same way as `getDefaultMultiState`.

### DIFF
--- a/packages/scenes/src/variables/variants/MultiValueVariable.ts
+++ b/packages/scenes/src/variables/variants/MultiValueVariable.ts
@@ -247,7 +247,7 @@ export abstract class MultiValueVariable<TState extends MultiValueVariableState 
     }
   }
 
-  public getDefaultSingleState(options: VariableValueOption[]) {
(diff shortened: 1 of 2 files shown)


torkelo (approved):

packages/scenes/src/variables/variants/MultiValueVariable.ts
```diff
@@ -252,6 +247,16 @@ export abstract class MultiValueVariable<TState extends MultiValueVariableState
     }
   }
 
+  public getDefaultSingleState(options: VariableValueOption[]) {
```
torkelo, line 250: maybe change this to protected? to make it clear it's for variables that extend this?

domasx2: fair, done

PR #1038, opened 2025-01-24
AdHocVariables: Allow setting valueLabel and customizing value when adding custom values

gtk-grafana:


packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -427,6 +436,10 @@ export class AdHocFiltersVariable
       description,
     }));
   }
+
+  public _getOnAddCustomValue() {
```
torkelo, line 440: public but has a private "_" prefix? , don't think we need this function

dprokop: agree

dprokop (approved):

PR #1040, opened 2025-01-29
Dependencies: Bump Grafana packages to v11.5

tskarhed:
Needed for Combobox rollout.

- Fix `hoverHeader` and `collapsible` type incompatibility in PanelChrome
- Update VizPanelBuilder snapshot (no idea if this is correct and expected)
- Update SceneTimeRange tests (Moment added an additional `_i` field which caused the tests to fail)
- Mock ReactInlineSVG
- Mock canvas
- Made numericOption optional to be compatible with the `{}` type


packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
```diff
@@ -177,29 +177,29 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
...
               dragHooks.onDragStart?.(e, model);
             }}
+            {...(collapsible
+              ? {
+                  collapsible: Boolean(collapsible),
+                  collapsed,
+                  onToggleCollapse: model.onToggleCollapse,
+                  showMenuAlways,
+                }
+              : { hoverHeader, hoverHeaderOffset })}
```
torkelo, line 202: some information on this change?

tskarhed: This was giving a type error, since these two options are mutually exclusive (See types for PanelChrome).

packages/scenes-app/src/demos/runtimePanelPlugin.tsx
```diff
@@ -45,7 +45,7 @@ interface CustomVizOptions {
 }
 
 interface CustomVizFieldOptions {
-  numericOption: number;
+  numericOption?: number;
```
tskarhed, line 48: These changes are due to the incompatibility with the `{}` type, which previously was `any`. Let me know if there is some other fundamental change needed to solve this.

packages/scenes/src/core/PanelBuilders/__snapshots__/VizPanelBuilder.test.ts.snap
```diff
@@ -369,7 +369,7 @@ exports[`VizPanelBuilder options and field config standard panels defaults provi
 {
   "defaults": {
     "custom": {
-      "label": "auto",
+      "fillOpacity": 50,
```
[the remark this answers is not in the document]
mdvictor: yeah this should be good, seems like the xychart schema changed. thank you for updating this

mdvictor (approved):
LGTM! Tested in both demo app and some core dashboards, works well!

PR #1042, opened 2025-01-31
SceneGraph: Expose getClosest method

kaydelaney:
Adds `getClosest` function to exported `sceneGraph` object. It's a useful function that I've found myself needing in the main repo from time to time, so I thought it would be worth exposing.


packages/scenes/src/core/sceneGraph/index.ts
```diff
@@ -2,34 +2,36 @@ import { lookupVariable } from '../../variables/lookupVariable';
...
+  getDataLayers,
   getLayout,
   getVariables,
-  getDataLayers,
   hasVariableDependencyInLoadingState,
   interpolate,
-  getAncestor,
-  findDescendents,
 } from './sceneGraph';
+import { getClosest } from './utils';
```
bfmatei, line 18: Shall we also move `getClosest` from `utils` to `sceneGraph` file?

ivanortegaalba (approved):
Let's apply Bogdan's comment and LGTM! 💯

PR #1050, opened 2025-02-06
Themes: Fixes variable labels to support border radius

torkelo:
The new Grafanacon themes have higher borderRadius which make the variable controls look a bit bad 

Before: 
Screenshot 2025-02-06 at 15 19 14

After: 
Screenshot 2025-02-06 at 15 18 41

### DIFF
--- a/packages/scenes/src/variables/components/VariableValueSelectors.tsx
+++ b/packages/scenes/src/variables/components/VariableValueSelectors.tsx
@@ -88,8 +88,7 @@ function VariableLabel({ variable, layout, hideLabel }: VariableSelectProps) {
 
 const containerStyle = css({
   display: 'flex',
-  // No border for second element (input)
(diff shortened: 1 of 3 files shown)


packages/scenes/src/variables/components/VariableValueSelectors.tsx
```diff
@@ -86,5 +86,13 @@ function VariableLabel({ variable, layout, hideLabel }: VariableSelectProps) {
   );
 }
 
-const containerStyle = css({ display: 'flex' });
+const containerStyle = css({
+  display: 'flex',
+  // No border for second element (inputs) as label and input border is shared
+  '> :nth-child(2)': css({
```
ashharrison90, line 92: this feels a little brittle, but i can't really see a better way to do it 🤷

ashharrison90 (approved):
lgtm 👍

PR #1054, opened 2025-02-13
Fix issue were operators do not appear properly in adhoc vars

mdvictor:
Fixes an issue in dashboards where adhocs do not show the regex operators even though they should appear since the variable settings UI page shows `Allow custom values` true, by default. This only happens when the flag is set by default and not explicitly in the model.

To maintain backwards comp, I added the `allowCustomValues` flag and set it to true by default (as custom values worked by default before this flag), but not in the model, so that it wouldn't change the dashboard model.

If I were to have added the flag in the constructor with a default value when undefined this issue wouldn't have happened, but then all dashboards would have their models changed and a new line for allowing custom values would be serialized in adhoc vars in already saved dashboards.


joannaWebDev (approved):
lgtm

PR #1055, opened 2025-02-13
Perform history step on MultiValueVariable

adrapereira:
Small PR to "copy" the approach used in `SceneObjectUrlSyncConfig` to create history steps when the value is updated. I added the `performBrowserHistoryAction` function to the `SceneObjectUrlSyncHandler` interface and implemented it in `MultiValueUrlSyncHandler`. If this approach sounds good I can also do it for other classes implementing `SceneObjectUrlSyncHandler` like in AdHocFiltersVariable, GroupByVariable, etc.

This is needed because in [traces-drilldown](https://github.com/grafana/traces-drilldown) we make extensive use of `CustomVariable` and are trying to create history steps whenever a variable updates. 

There should be no breaking changes.


torkelo:
@adrapereira can you take a look at https://github.com/grafana/scenes/pull/882/ ?  It adds browser history to variables (that extend MultiValueVariable), and only creates history steps when user initiates the change (which I think is important)

PR #1056, opened 2025-02-14
Dashboards: Fix interaction with multi variable value pills

mdvictor:
Fixes an issue in dashboards where the multi-var value pills could not be interacted with.  Seems to be an issue stemming from `react-select` itself.

Related fix in grafana-ui [here](https://github.com/grafana/grafana/pull/100728)


mdvictor:
Closing this as it is no longer needed. Will be fixed in core by [this PR](https://github.com/grafana/grafana/pull/100950)

PR #1057, opened 2025-02-18
Fallback page: fix infinite loop

domasx2:
:wave: 

Upgrading App O11y to scenes v6, found it's possible to get into infinite loop when a `SceneAppPage` specifies it's fallback page as itself (see test).

The use case is that we want `SceneApp` to render initialization page regardless of route to do some initial loading of metadata. Did not find a better way to do `any route` than to have a page fallback to itself, let me know if I'm missing something :bow: 

The cause of infinite loop was fallback `Route` using `Component` insted of `element`, which causes fallback page to re-mount on every render. And remounting was causing it to update `initializedScene` state prop of parent, causing further re-renders.

### DIFF
--- a/packages/scenes/src/components/SceneApp/SceneApp.test.tsx
+++ b/packages/scenes/src/components/SceneApp/SceneApp.test.tsx
@@ -416,6 +416,30 @@ describe('SceneApp', () => {
         expect(await screen.findByTestId('custom-fallback-content')).toBeInTheDocument();
       });
     });
+
+    describe('Custom fallback page that loops on itself', () => {
+      const page: SceneAppPage = new SceneAppPage({
+        title: 'Test',
+        url: '/test',
+        routePath: 'test/*',
+        getScene: () => {
+          return new EmbeddedScene({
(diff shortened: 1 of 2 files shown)


leventebalogh (approved):
Nice catch, the change looks legit to me 👍 

I am unfortunately not sure from the top of my head if there's a better to achieve what you would like to. I am not sure that I understood correctly, but what you would like to do is to load some metadata on an initial rendering? If so, would it be an option to not use scenes itself for this but a custom route in your app plugin? So something before / after the SceneApp in your plugin?
```ts
<Routes>
  <Route path="*" element={<YourComponent />}>
</Routes>
```

domasx2:
>  so, would it be an option to not use scenes itself for this but a custom route in your app plugin?

If I was rewriting from scratch, I'd likely do that and avoid `ScenesApp` altogether :thinking:  but now looking for easy upgrade path for existing code.  Initially there were reasons to try to do it "in-scenes" so to speak, most importantly to use the `ScenesQueryRunner` for the metadata loading

PR #1060, opened 2025-02-24
Show AdHoc `baseFilters` with a source as readonly in the UI

mdvictor:
`baseFilters` are usually hidden away within the adhoc variable but affect the returned keys and values.

With this PR we allow showing `baseFilters` in the UI if they are specifically sourced (e.g.: from a scope). They are currently read-only and work as a visual aid to translating selected scopes into actual filters.

Related to https://github.com/grafana/grafana/pull/101217

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -25,9 +25,17 @@ export interface AdHocFilterWithLabels<M extends Record<string, any> = {}> exten
   // hide the filter from AdHocFiltersVariableRenderer and the URL
   hidden?: boolean;
   meta?: M;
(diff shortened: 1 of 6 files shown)


packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -25,10 +25,18 @@ export interface AdHocFilterWithLabels<M extends Record<string, any> = {}> exten
   // hide the filter from AdHocFiltersVariableRenderer and the URL
   hidden?: boolean;
   meta?: M;
+  // filter source, it can be either scopes, dashboards or undefined,
+  // which means it won't appear in the UI
+  source?: FilterSource;
```
dprokop, line 30: uber nit: we use `origin` term in data frames to identify origin field/data frame. Think we can align the terminiology.

dprokop (approved):

PR #1062, opened 2025-02-26
AdHocVariable: Edit injected filters

mdvictor:
allows adhocs to show and edit base filters if they have a source. 

Holds base filters into url sync to allow for passing them over from one dashboard to another, alongside the source and original value of a injected filter.

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
@@ -27,14 +27,14 @@ export function AdHocFilterPill({ filter, model, readOnly, focusOnWipInputRef }:
   const handleChangeViewMode = useCallback(
     (event?: React.MouseEvent, shouldFocusOnPillWrapperOverride?: boolean) => {
       event?.stopPropagation();
-      if (readOnly || filter.source) {
+      if (readOnly) {
         return;
       }
 
       setShouldFocusOnPillWrapper(shouldFocusOnPillWrapperOverride ?? !viewMode);
       setViewMode(!viewMode);
     },
-    [readOnly, viewMode, filter.source]
+    [readOnly, viewMode]
   );
 
   useEffect(() => {
@@ -115,12 +115,25 @@ export function AdHocFilterPill({ filter, model, readOnly, focusOnWipInputRef }:
           />
         ) : null}
 
-        {filter.source && (
+        {filter.source && !filter.originalValue && (
           <IconButton
             name="info-circle"
             size="md"
             className={styles.pillIcon}
             tooltip={`This is a ${filter.source} injected filter`}
+          />
+        )}
+
+        {filter.source && filter.originalValue && (
+          <IconButton
+            onClick={(e) => {
+              e.stopPropagation();
+              model.restoreBaseFilter(filter);
+            }}
+            name="history"
+            size="md"
+            className={styles.pillIcon}
+            tooltip={`Restore filter to its original value`}
           />
         )}
       </div>
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -27,7 +27,8 @@ export interface AdHocFilterWithLabels<M extends Record<string, any> = {}> exten
   meta?: M;
   // filter source, it can be either scopes, dashboards or undefined,
   // which means it won't appear in the UI
-  source?: FilterSource;
+  source?: FilterSource | string;
+  originalValue?: string[];
 }
 
 export type AdHocControlsLayout = ControlsLayout | 'combobox';
@@ -216,8 +217,15 @@ export class AdHocFiltersVariable
   public setState(update: Partial<AdHocFiltersVariableState>): void {
     let filterExpressionChanged = false;
 
-    if (update.filters && update.filters !== this.state.filters && !update.filterExpression) {
-      update.filterExpression = renderExpression(this.state.expressionBuilder, update.filters);
+    if (
+      ((update.filters && update.filters !== this.state.filters) ||
+        (update.baseFilters && update.baseFilters !== this.state.baseFilters)) &&
+      !update.filterExpression
+    ) {
+      update.filterExpression = renderExpression(this.state.expressionBuilder, [
+        ...(update.baseFilters?.filter((filter) => filter.source) ?? []),
+        ...(update.filters ?? []),
+      ]);
       filterExpressionChanged = update.filterExpression !== this.state.filterExpression;
     }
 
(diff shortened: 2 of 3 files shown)


packages/scenes/src/variables/adhoc/AdHocFiltersVariable.test.tsx
```diff
@@ -745,6 +750,274 @@ describe.each(['11.1.2', '11.1.1'])('AdHocFiltersVariable', (v) => {
...
+        // no origin, so this does not get synced
+        { key: 'baseKey3', keyLabel: 'baseKey3', operator: '!=', value: 'baseValue3', valueLabels: ['baseValue3'] },
+      ],
+    });
+
+    act(() => {
+      filtersVar._updateFilter(filtersVar.state.filters[0], { key: 'newKey', keyLabel: 'newKey' });
+    });
+
+    // injected filters stored in the following format: normal|adhoc|values\original|values\filterOrigin
```
dprokop, line 782: I think the format is fine, question is if you sync them always? I think we should probably sync them only when they have changed, otherwise one will navigate to a dashboard in the same scope.

packages/scenes/src/variables/adhoc/AdHocFiltersVariableUrlSyncHandler.ts
```diff
@@ -10,37 +11,71 @@ import { escapeUrlPipeDelimiters, toUrlCommaDelimitedString, unescapeUrlDelimite
 export class AdHocFiltersVariableUrlSyncHandler implements SceneObjectUrlSyncHandler {
   public constructor(private _variable: AdHocFiltersVariable) {}
 
+  private getInjectedKey(): string {
+    return `var-injected-${this._variable.state.name}`;
+  }
```
dprokop, line 16: Why do we need to do this? I think we could sync the injected alongside the regular filters (keys that are injected can't be really overwritten by the user) and figure out which filters were scope-injected based on keys matching and origin. WDYT?

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.test.tsx
mdvictor: I think you are right. I kept thinking about this and initially thought that both normal filters and baseFilters with an origin should be treated the same, thus having these baseFilters in the URL, same as your average filter, but I agree that there is no point in having both unmodified scope injected filter and the scope itself in there

packages/scenes/src/variables/adhoc/AdHocFiltersVariableUrlSyncHandler.ts
mdvictor: Indeed, no need for another key other than for verbosity, we can identify injected filters based on origin. Removing the extra key

packages/scenes/src/variables/adhoc/AdHocFiltersVariableUrlSyncHandler.ts
```diff
@@ -20,16 +26,42 @@ export class AdHocFiltersVariableUrlSyncHandler implements SceneObjectUrlSyncHan
...
+          .map((filter) =>
+            toArray(filter)
+              .map(escapeInjectedFilterUrlDelimiters)
+              .join('|')
+              .concat(
+                `#${filter.originalValue?.map(escapeInjectedFilterUrlDelimiters).join('|') ?? ''}#${filter.origin}`
+              )
+          )
+      );
+    }
```
Sergej-Vlasov, line 60: This part could use some comments

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
```diff
@@ -115,14 +115,27 @@ export function AdHocFilterPill({ filter, model, readOnly, focusOnWipInputRef }:
...
+            onClick={(e) => {
+              e.stopPropagation();
+              model.restoreOriginalFilter(filter);
+            }}
+            name="history"
+            size="md"
+            className={styles.pillIcon}
+            tooltip={`Restore filter to its original value`}
+          />
+        )}
```
Sergej-Vlasov, line 138: nit: could be combined into:
```typescript
{filter.origin && filter.originalValue ? (
      <IconButton
            onClick={(e) => {
              e.stopPropagation();
              model.restoreOriginalFilter(filter);
            }}
            name="history"
            size="md"
            className={styles.pillIcon}
            tooltip={`Restore filter to its original value`}
          />
       ) : (
       <IconButton
            name="info-circle"
            size="md"
            className={styles.pillIcon}
            tooltip={`This is a ${filter.origin} injected filter`}
          />
       )}

```

Sergej-Vlasov (commented):
Looking great!
Some safeguards are needed for editing scope injected filters:

1. prevent editing `operator` using `shift+tab` keyboard shortcut
2. prevent editing `operator` using backspace (when deleting last value)
3. deleting `operator` triggers focus shift to wip (add new) filter and scopes filter stays in broken edit mode 
4. clicking on scope inject filter to edit and then clicking on `operator` pill leaves scope injected filter in edit mode and focuses on wip filter
5. backspace from wip filter does not start editing scope injected filters

https://github.com/user-attachments/assets/3f96edeb-ac72-4ff0-80df-760464a14667


https://github.com/user-attachments/assets/d649366b-fdfa-4ea7-b3fb-05b65a6cd1ca

Other than those, looks good ✅

packages/scenes/src/variables/adhoc/AdHocFiltersVariableUrlSyncHandler.ts
mdvictor: I added a comment explaining the url sync format for injected filters

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -37,6 +38,7 @@ export interface AdHocFilterWithLabels<M extends Record<string, any> = {}> exten
   // filter origin, it can be either scopes, dashboards or undefined,
   // which means it won't appear in the UI
   origin?: FilterOrigin;
+  originalValue?: string[];
```
tskarhed, line 41: I find it hard to understand what this is and why it should display the history icon when it is defined. My initial thought was that it is defined when not edited, not the other way around.

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -214,7 +216,12 @@ export class AdHocFiltersVariable
       filters: [],
       datasource: null,
       applyMode: 'auto',
-      filterExpression: state.filterExpression ?? renderExpression(state.expressionBuilder, state.filters),
+      filterExpression:
+        state.filterExpression ??
+        renderExpression(state.expressionBuilder, [
+          ...(state.baseFilters?.filter((filter) => filter.origin) ?? []),
+          ...(state.filters ?? []),
```
tskarhed, line 223: This is being used a three different places? Would it make sense to extract this into a helper that describes what it does?

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -235,28 +242,66 @@ export class AdHocFiltersVariable
...
+    // we overwrite everything with the newly set scopes
+    if (!overwrite) {
+      const editedScopeFilters = scopeInjectedFilters.filter((filter) => filter.originalValue?.length);
+      const editedScopeFilterKeys = editedScopeFilters.map((filter) => filter.key);
+      const scopeFilterKeys = scopeFilters.map((filter) => filter.key);
+
+      // if the scope filters contain the key of an edited scope one, we replace
+      // with the edited one, otherwise if an edited filter is not found in
+      // the scope filters (this can happend when changing scopes alltogether)
+      // we drop the edited one and use the new filters from the new scopes
```
tskarhed, line 278: Isn't this similar to what's being done in `_updateFilter`? I am not familiar with the order of things, how come it is needed to be done in both places?

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
mdvictor: I see what you mean, this could use better naming, maybe something like `savedOriginalValue`? The point of the property is indeed to hold the initial filter value before it is being edited

PR #1063, opened 2025-02-26
Profiling: explicitly set profiling names

oscarkilhed:
It turns out when the profiling measurements are reported the name is missing. I suspect this is because of some optimization when building the prod artifacts is removing `constructor.name`.

We can instead set the profiling names explicitly to make sure they are submitted.


dprokop (approved):
🥳

PR #1064, opened 2025-02-27
sceneInterpolator: Ensure unescaped backslashes are escaped

kaydelaney:


packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -236,7 +237,11 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
 
     // If all transformations are config object we can interpolate them all at once
     if (onlyObjects) {
-      return JSON.parse(sceneGraph.interpolate(this, JSON.stringify(transformations), data.request?.scopedVars));
+      return deepIterate(transformations, (v) => {
+        if (typeof v === 'string') {
+          return sceneGraph.interpolate(this, v, data.request?.scopedVars)
+        }
```
leeoniya, line 243: should return v otherwise, right?

leeoniya: oh, never mind, it's handled in the caller.

mdvictor:
I think this fixes the escaping issue, but doesn't fully fix https://github.com/grafana/grafana/issues/101037. The dashboard is still broken because of that panel although the error is now different

leeoniya:
please look at https://github.com/grafana/grafana/pull/101438#discussion_r1974356450 and make sure this scenes variant works with the dashboard attached in that PR description.

leeoniya:
think we need to fix this one as well (repro in link):

https://github.com/grafana/grafana/issues/103048#issuecomment-2764651282

Issue #1065, opened 2025-02-27
AdHocFilters: Allow custom value to be first


gtk-grafana:
Users have requested the ability for the custom value to show up at the top of the list, instead of at the bottom. Variables with many values may not even display the custom option to the user, which is frustrating when attempting to write regex operations which are not returned by getTagValuesProvider.

<img width="419" alt="Image" src="https://github.com/user-attachments/assets/5d988a60-034e-421c-bf2e-8780e04f81f0" />

Related: https://github.com/grafana/logs-drilldown/issues/1116

Ideally this could be set depending on the operator used, so regex operations would push custom value to the top, and non-regex operations would push custom value to the bottom, but being able to set the position for a variable would still be an improvement. 

This could be worked around by setting the variable `allowCustomValue` state in the getTagKeysProvider and getTagValuesProvider

PR #1066, opened 2025-02-27
AdHocVariables: Allow custom value to be first

gtk-grafana:
Allows pushing the custom variable to the top of the list in ad hoc variables.

Notes for reviewer:
The type for allowCustomVariable: `allowCustomValue?: boolean | 'first';` is admittedly not very ergonomic, but I wasn't sure if we wanted to add an additional state property for this.

Fixes: https://github.com/grafana/scenes/issues/1065.

<details>
  <summary>📦 Published PR as canary version: <code>6.2.0--canary.1066.13575913588.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.2.0--canary.1066.13575913588.0
  npm install @grafana/scenes@6.2.0--canary.1066.13575913588.0
  # or 
  yarn add @grafana/scenes-react@6.2.0--canary.1066.13575913588.0
  yarn add @grafana/scenes@6.2.0--canary.1066.13575913588.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -108,7 +108,7 @@ export interface AdHocFiltersVariableState extends SceneVariableState {
   /**
    * Flag that decides whether custom values can be added to the filter
    */
-  allowCustomValue?: boolean;
+  allowCustomValue?: boolean | 'first';
 
   /**
    * @internal state of the new filter being added
--- a/packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFilterRenderer.tsx
@@ -115,10 +115,10 @@ export function AdHocFilterRenderer({ filter, model }: Props) {
     <Select
       virtualized
       allowCustomValue={model.state.allowCustomValue !== false ?? true}
+      createOptionPosition={model.state.allowCustomValue === 'first' ? 'first' : 'last'}
       isValidNewOption={(inputValue) => inputValue.trim().length > 0}
       allowCreateWhileLoading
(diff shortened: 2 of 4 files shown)


torkelo:
@gtk-grafana 

This seems like a bit of a bug, Why is the dropdown showing a bunch of values that do not match what I have typed (in terms of simple search) 

Screenshot 2025-09-18 at 17 31 11

but still unsure why we need an option for this, feels like this should be the default (at least for regex mode)

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -108,7 +108,7 @@ export interface AdHocFiltersVariableState extends SceneVariableState {
   /**
    * Flag that decides whether custom values can be added to the filter
    */
-  allowCustomValue?: boolean;
+  allowCustomValue?: boolean | 'first';
```
leeoniya, line 111: this is pretty weird. i'd break out another prop for this, e.g. `customValuePosition: string`

gtk-grafana: Agreed, if the approach in this PoC is otherwise acceptable I'll get that updated.

leeoniya:
> but if we are using uFuzzy the custom value should match 100% and should be a the top

if the operator is regexp, then it doesnt make sense to use uFuzzy since you're already building a regexp (which is what uFuzzy does internally)

gtk-grafana:
> how is this possible? it will know the operator is =~ or !=~

I'm not sure where the line is between remaining datasource agnostic and keeping the API simple, but checking for `=~` or `!~` would work for PromQL/LogQL and give us a quick fix without needing to change existing state props.

I can try to knock out a solution for that now if folks don't mind introducing some datasource specific checks in the combobox code?

gtk-grafana:
Closing in favor of https://github.com/grafana/scenes/pull/1250

Issue #1068, opened 2025-03-04
Dashboard variables still broken for DataSourceVariableSupport when used with DataSourceWithBackend plugin


rnterbush:
This issue is related to https://github.com/grafana/grafana/issues/98101 and the associated fix https://github.com/grafana/scenes/pull/1006

The net outcome is that dashboard variable queries throw an error and fail to load any values.

The PR above adds the necessary handling for variables of type VariableSupportType.Datasource in [createQueryVariableRunner.ts], but the call to runRequest passes a reference to the datasource "query" function as the third argument. If a valid function is passed, it will be called by the logic in runRequest with no invoked object. This breaks the DataSourceWithBackend-specific implementation of query(), which assumes "this" to be defined.

PR #1069, opened 2025-03-05
Variables: Fix for data source variable support

torkelo:
Fixes https://github.com/grafana/scenes/issues/1068

Aligns it with the old implementation: https://github.com/grafana/grafana/blob/main/public/app/features/variables/query/queryRunners.ts#L180


mdvictor (approved):
LGTM! Managed to hack up a testing env and seems like it's working!

PR #1072, opened 2025-03-06
Inject scopes in adhoc

mdvictor:
This PR moves scope to adhoc conversion code from [grafana](https://github.com/grafana/grafana/pull/101217) after the scopes lifting work and removal of [facades](https://github.com/grafana/grafana/pull/97176/files#diff-ff0ca5646e987fbf2d56d6c59e7fc2edbe2deba4ae94413b64a35511d39ff3a5).

TODO:
- [ ] write some tests for the base filters logic
- [ ] fix scenario where reloading a dashboard with some scopes selected won't update the adhocs.

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -1,5 +1,12 @@
 import React from 'react';
-import { AdHocVariableFilter, GetTagResponse, GrafanaTheme2, MetricFindValue, SelectableValue } from '@grafana/data';
+import {
+  AdHocVariableFilter,
+  GetTagResponse,
+  GrafanaTheme2,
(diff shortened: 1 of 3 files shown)


packages/scenes/src/core/SceneScopesBridge.ts
```diff
@@ -51,8 +51,7 @@ export class SceneScopesBridge extends SceneObjectBase implements SceneObjectWit
         map(
           ([prevContext, newContext]) =>
             [prevContext?.state.value ?? [], newContext?.state.value ?? []] as [Scope[], Scope[]]
-        ),
-        filter(([prevScopes, newScopes]) => !isEqual(prevScopes, newScopes))
```
mdvictor, line 55: With this filter removed we always the the latest value in scenarios in which we wouldn't otherwise, e.g.: when reloading a dashboard with scopes already set

PR #1073, opened 2025-03-11
AdhocVariable: Inject scopes in baseFilters

mdvictor:
This PR moves scope to adhoc conversion code from https://github.com/grafana/grafana/pull/101217 after the scopes lifting work and removal of [facades](https://github.com/grafana/grafana/pull/97176/files#diff-ff0ca5646e987fbf2d56d6c59e7fc2edbe2deba4ae94413b64a35511d39ff3a5).

<details>
  <summary>📦 Published PR as canary version: <code>6.3.2--canary.1073.13788678829.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.3.2--canary.1073.13788678829.0
  npm install @grafana/scenes@6.3.2--canary.1073.13788678829.0
  # or 
  yarn add @grafana/scenes-react@6.3.2--canary.1073.13788678829.0
  yarn add @grafana/scenes@6.3.2--canary.1073.13788678829.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -1,5 +1,12 @@
 import React from 'react';
-import { AdHocVariableFilter, GetTagResponse, GrafanaTheme2, MetricFindValue, SelectableValue } from '@grafana/data';
+import {
+  AdHocVariableFilter,
+  GetTagResponse,
+  GrafanaTheme2,
+  MetricFindValue,
+  Scope,
+  SelectableValue,
+} from '@grafana/data';
 import { SceneObjectBase } from '../../core/SceneObjectBase';
 import { SceneVariable, SceneVariableState, SceneVariableValueChangedEvent, VariableValue } from '../types';
 import { ControlsLayout, SceneComponentProps } from '../../core/types';
@@ -17,6 +24,7 @@ import { getEnrichedFiltersRequest } from '../getEnrichedFiltersRequest';
 import { AdHocFiltersComboboxRenderer } from './AdHocFiltersCombobox/AdHocFiltersComboboxRenderer';
 import { wrapInSafeSerializableSceneObject } from '../../utils/wrapInSafeSerializableSceneObject';
 import { SceneScopesBridge } from '../../core/SceneScopesBridge';
+import { getAdHocFiltersFromScopes } from './getAdHocFiltersFromScopes';
 
 export interface AdHocFilterWithLabels<M extends Record<string, any> = {}> extends AdHocVariableFilter {
   keyLabel?: string;
@@ -219,6 +227,29 @@ export class AdHocFiltersVariable
 
   private _activationHandler = () => {
     this._scopesBridge = sceneGraph.getScopesBridge(this);
+
+    const scopes = this._scopesBridge?.getValue();
+
+    if (scopes) {
+      this._updateScopesFilters(scopes);
+    }
+
+    const sub = this._scopesBridge?.subscribeToValue((n, _) => {
+      this._updateScopesFilters(n);
+    });
+
+    return () => {
+      sub?.unsubscribe();
+    };
+  };
+
+  private _updateScopesFilters = (scopes: Scope[]) => {
+    // maintain other baseFilters in the array, only update scopes ones
+    const newFilters = [
+      ...(this.state.baseFilters?.filter((f) => f.origin && f.origin !== FilterOrigin.Scopes) ?? []),
+      ...getAdHocFiltersFromScopes(scopes),
+    ];
+    this.setState({ baseFilters: newFilters });
   };
 
   public setState(update: Partial<AdHocFiltersVariableState>): void {
--- a/packages/scenes/src/variables/adhoc/getAdHocFiltersFromScopes.ts
+++ b/packages/scenes/src/variables/adhoc/getAdHocFiltersFromScopes.ts
@@ -1,0 +1,98 @@
+import { Scope, ScopeFilterOperator, ScopeSpecFilter, scopeFilterOperatorMap } from '@grafana/data';
+import { AdHocFilterWithLabels, FilterOrigin } from './AdHocFiltersVariable';
+
+export type EqualityOrMultiOperator = Extract<ScopeFilterOperator, 'equals' | 'not-equals' | 'one-of' | 'not-one-of'>;
+
+export const reverseScopeFilterOperatorMap: Record<ScopeFilterOperator, string> = Object.fromEntries(
+  Object.entries(scopeFilterOperatorMap).map(([symbol, operator]) => [operator, symbol])
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+) as Record<ScopeFilterOperator, string>;
+
+export function isEqualityOrMultiOperator(value: string): value is EqualityOrMultiOperator {
+  const operators = new Set(['equals', 'not-equals', 'one-of', 'not-one-of']);
+  return operators.has(value);
+}
+
+export function getAdHocFiltersFromScopes(scopes: Scope[]): AdHocFilterWithLabels[] {
+  const formattedFilters: Map<string, AdHocFilterWithLabels> = new Map();
+  // duplicated filters that could not be processed in any way are just appended to the list
+  const duplicatedFilters: AdHocFilterWithLabels[] = [];
+  const allFilters = scopes.flatMap((scope) => scope.spec.filters);
+
+  for (const filter of allFilters) {
+    processFilter(formattedFilters, duplicatedFilters, filter);
+  }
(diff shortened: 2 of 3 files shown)


packages/scenes/src/variables/adhoc/getAdHocFiltersFromScopes.ts
```diff
@@ -0,0 +1,98 @@
...
+  Object.entries(scopeFilterOperatorMap).map(([symbol, operator]) => [operator, symbol])
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+) as Record<ScopeFilterOperator, string>;
+
+export function isEqualityOrMultiOperator(value: string): value is EqualityOrMultiOperator {
+  const operators = new Set(['equals', 'not-equals', 'one-of', 'not-one-of']);
+  return operators.has(value);
+}
+
+export function getAdHocFiltersFromScopes(scopes: Scope[]): AdHocFilterWithLabels[] {
```
mdvictor, line 16: These functions were all moved from grafana without any changes and have been previously approved. The only difference is that I've also extracted the above guard and types from `grafana/data` since builds fail because we need a grafana release for the package. I'll deal with these imports when grafana v11.6 gets released.

aocenas: not sure I understand the implications of the Grafana release. I assume if those types were not yet in the grafana/data package specified in scenes, the imports would fail not the type assertion. Or are the types just different then what they should be in the end?

packages/scenes/src/variables/adhoc/getAdHocFiltersFromScopes.ts
```diff
@@ -0,0 +1,98 @@
...
+  const formattedFilters: Map<string, AdHocFilterWithLabels> = new Map();
+  // duplicated filters that could not be processed in any way are just appended to the list
+  const duplicatedFilters: AdHocFilterWithLabels[] = [];
+  const allFilters = scopes.flatMap((scope) => scope.spec.filters);
+
+  for (const filter of allFilters) {
+    processFilter(formattedFilters, duplicatedFilters, filter);
```
aocenas, line 23: I know this is just copied from previous code, but I find it very strange that we have to supply the empty map and array, to be populated, and get the return value that way, instead of the function just having a return value. Is there some specific usecase this is required?

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -219,6 +227,29 @@ export class AdHocFiltersVariable
...
+    }
+
+    const sub = this._scopesBridge?.subscribeToValue((n, _) => {
+      this._updateScopesFilters(n);
+    });
+
+    return () => {
+      sub?.unsubscribe();
+    };
+  };
```
tskarhed, line 244: Will you not get the initial value when subscribing? Is the `getValue` call necessary?

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -219,6 +227,29 @@ export class AdHocFiltersVariable
...
+
+  private _updateScopesFilters = (scopes: Scope[]) => {
+    // maintain other baseFilters in the array, only update scopes ones
+    const newFilters = [
+      ...(this.state.baseFilters?.filter((f) => f.origin && f.origin !== FilterOrigin.Scopes) ?? []),
```
tskarhed, line 249: This filters out baseFilters that were previously scope injected, right?

packages/scenes/src/variables/adhoc/getAdHocFiltersFromScopes.ts
aocenas, line 16: Would be nice to have a bit of comment here. I assume there is some way ad hoc filters support more operators or something so some scope filters can be merged into single ad-hoc filter and that is what we are trying to process here but it's not so easy to get from the code alone.

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -219,6 +227,29 @@ export class AdHocFiltersVariable
...
+    const sub = this._scopesBridge?.subscribeToValue((n, _) => {
+      this._updateScopesFilters(n);
+    });
+
+    return () => {
+      sub?.unsubscribe();
+    };
+  };
+
+  private _updateScopesFilters = (scopes: Scope[]) => {
```
aocenas, line 246: Is there a convention in scenes for having a `_` prefix for methods?

aocenas (approved):
some smaller comments but overall looks good

packages/scenes/src/variables/adhoc/getAdHocFiltersFromScopes.ts
mdvictor: The types are the same. I went with this cuz it was a bit easier. Although now that I think about it I might be able to go around without duplicating. Will give it another test and see.

IIRC the problem is that these types were added to grafana/data and that is not released yet. So I can't just bump the scenes deps to use the new version of grafana/data since that gets released together with grafana, right? So when grafana 11.6 launches, so will grafana/data bump to 11.6, which will let me then update scenes to use 11.6 and thus those new imports. 

I could add a ts-ignore to the imports and it would work locally, but causes issues in the ci build. That's why I just went with this easier approach, but if there is a better way I'll use it

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
tskarhed: I think it is supposed to be extra on top of having `private`

packages/scenes/src/variables/adhoc/getAdHocFiltersFromScopes.ts
aocenas: Ah ok I think I just misunderstood what exactly is missing from the grafana/data. Yeah I think duplicating makes sense. I did a ts-ignore on import but then the code would work if the import was `undefined` here I am not sure that would be easy to deal with anyway.

packages/scenes/src/variables/adhoc/getAdHocFiltersFromScopes.ts
mdvictor: My reasoning behind this was to just have one map that deals with everything. The alternative would be to return the map (would still need to be a map to check for colisions), copy it in the parent map and deal with collisions, rinse and repeat.

packages/scenes/src/variables/adhoc/getAdHocFiltersFromScopes.ts
mdvictor: I agree, adding it now. Yes, the main reasoning is around operator issues, because multiple scopes can contain filters with the same keys, but differ in op and value, and the processing part just verifies whether all those scope filters with the same key can or cannot be merged into one single adhoc filter.

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
mdvictor: yeah, I've seen that `_` convention used around for extra brevity on a method being private. Have no strong opinions on it tho

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
mdvictor: Both are needed for different scenarios. the sub will update adhocs everytime the scopes are changed, while the getValue  part will always push current scopes into the adhoc when changing dashboards, for example, while scopes remain the same

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
mdvictor: Yes it does, while maintaining whatever other baseFilters there might be. I see I've further modified this in the other [PR](https://github.com/grafana/scenes/pull/1062/files#diff-56fcc6003a9ae4350d1c2dc6241b534e02a997ed25933cec1a303dced5436fd5R279)

But yes, that is what it does

tskarhed (approved):

packages/scenes/src/variables/adhoc/getAdHocFiltersFromScopes.ts
mdvictor: Added a comment to the method [here](https://github.com/grafana/scenes/pull/1062)

packages/scenes/src/variables/adhoc/getAdHocFiltersFromScopes.ts
aocenas: Ah ok got it, didn't realise that the map and array serves to keep state in between the loops.

PR #1074, opened 2025-03-17
fix: explore button to preserve datasource selection in mixed datasource panels

joannaWebDev:
In a Time Series panel using VizPanel from Grafana Scenes, I encountered an issue when using a mixed datasource setup with both Prometheus and Loki:

- The panel itself correctly displays data from both datasources.
- However, **clicking the Explore button redirects the user to Explore mode, but only Loki queries appear, while the Prometheus queries are missing**.

This happens because `getExploreURL` was not preserving the original datasource assignments from the panel when constructing the Explore URL. Instead, it was defaulting to the last interpolated query’s datasource, causing some queries to be omitted in the Explore view.

**Solution**
- If multiple unique datasource UIDs exist, the `getExploreURL` function now marks the Explore request as` "-- Mixed --"` instead of defaulting to only one datasource. This allows all queries, including those from different datasources, to be passed correctly to Explore
Updated tests. 

**The Explore UI before/after**

before

after

Fixes issue https://github.com/grafana/grafana-dbo11y-app/issues/738

### DIFF
--- a/packages/scenes/src/utils/explore.ts
+++ b/packages/scenes/src/utils/explore.ts
@@ -19,7 +19,6 @@ export async function getExploreURL(
   }
 
   const { from, to } = timeRange;
-
   const filters = data.request?.filters;
 
   const scopedVars: ScopedVars = {
@@ -35,28 +34,27 @@ export async function getExploreURL(
     )
   )
     .filter((promise): promise is PromiseFulfilledResult<DataQuery> => promise.status === 'fulfilled')
-    .map((q) => q.value);
+    .map((q) => q.value)
+    .map((q) => transform?.(q) ?? q);
 
-  let queries: DataQuery[];
+  const queries: DataQuery[] = interpolatedQueries?.flat() ?? [];
 
-  const hasMixedDatasources = new Set(interpolatedQueries.map((q) => q.datasource?.uid)).size > 1;
+  // Check if we have mixed datasources (more than one unique datasource)
+  const hasMixedDatasources = new Set(queries.map((q) => q.datasource?.uid)).size > 1;
 
-  // Apply transform based on whether we're dealing with multiple queries or a single query
-  if (transform) {
-    const result = transform(
-      hasMixedDatasources || interpolatedQueries.length > 1 ? interpolatedQueries : interpolatedQueries[0]
-    );
-    queries = Array.isArray(result) ? result : [result];
-  } else {
-    queries = interpolatedQueries;
-  }
+  // For mixed datasources, mark the datasource as "-- Mixed --"
+  let datasource = hasMixedDatasources
+    ? '-- Mixed --'
+    : queries.find((query) => !!query.datasource?.uid)?.datasource?.uid;
 
-  // For mixed datasource panels, use "-- Mixed --" as the datasource
-  let datasource;
+  // If there are multiple queries with different data sources, ensure all queries are included.
+  // Maintain all queries and just update the datasource for mixed case
   if (hasMixedDatasources) {
-    datasource = '-- Mixed --';
-  } else {
-    datasource = queries.find((query) => !!query.datasource?.uid)?.datasource?.uid;
+    queries.forEach((query) => {
+      if (!query.datasource?.uid) {
+        query.datasource = { uid: '-- Mixed --' }; // For mixed datasource, mark as "-- Mixed --"
+      }
+    });
   }
 
   if (queries?.length && datasource && from && to) {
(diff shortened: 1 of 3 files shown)


packages/scenes/src/utils/explore.ts
```diff
@@ -35,12 +35,29 @@ export async function getExploreURL(
...
+    .map((q) => q.value);
 
-  const queries: DataQuery[] = interpolatedQueries ?? [];
+  let queries: DataQuery[];
 
-  const datasource = queries.find((query) => !!query.datasource?.uid)?.datasource?.uid;
+  const hasMixedDatasources = new Set(interpolatedQueries.map((q) => q.datasource?.uid)).size > 1;
+
+  // Apply transform based on whether we're dealing with multiple queries or a single query
+  if (transform) {
```
domasx2, line 45: Hey, could you elaborate on why the change to transform is needed?  It changes transform from being called once per query to being called once with all queries as the param. I don't see how this is related to mixed data sources issue :thinking: 

Also it's a breaking change, anyone who uses it will need to update their implmentation. Maybe worth adding a `transformAll`  or something similar instead

packages/scenes/src/utils/explore.ts
```diff
@@ -11,15 +11,14 @@ export async function getExploreURL(
   data: PanelData,
   model: SceneObject,
   timeRange: RawTimeRange,
-  transform?: (query: DataQuery) => DataQuery
+  transform?: (query: DataQuery | DataQuery[]) => DataQuery | DataQuery[]
```
domasx2, line 14: since the  way transform works is not changed anymore, I guess parameter/return types do not need to be changed either. It no longer receives or produces array of queries

packages/scenes/src/utils/explore.ts
```diff
@@ -38,9 +37,25 @@ export async function getExploreURL(
...
+  let datasource = hasMixedDatasources
+    ? '-- Mixed --'
+    : queries.find((query) => !!query.datasource?.uid)?.datasource?.uid;
+
+  // If there are multiple queries with different data sources, ensure all queries are included.
+  // Maintain all queries and just update the datasource for mixed case
+  if (hasMixedDatasources) {
+    queries.forEach((query) => {
+      if (!query.datasource?.uid) {
+        query.datasource = { uid: '-- Mixed --' }; // For mixed datasource, mark as "-- Mixed --"
```
domasx2, line 55: hm, is there a use case for _individual_ queries to have "mixed" data source? I thought the overall data source is mixed, but each individual query should specify a concrete one

packages/scenes/src/utils/explore.ts
joannaWebDev: Very true. Reverted

packages/scenes/src/utils/explore.ts
joannaWebDev: I don't think there's an use case for an individual query having `-- Mixed --` as its datasource. However the entire set of queries in the panel can be marked as `-- Mixed --` when they span multiple datasources.

See example [here](https://github.com/grafana/grafana-dbo11y-app/blob/2e796fa9e76bcae6c25293cf551c14dc5a54d647/src/pages/QueryDetails/querySample/querySamplesPanel/querySamplesPanel.ts#L37-L36)



```
function getData(job: string, digest: string, schema: string, instance: string) {
  return new SceneQueryRunner({
    datasource: {
      uid: '-- Mixed --',
      type: 'datasource',
    },
    queries: [
      {
        refId: 'Query Samples Panel',
        datasource: {
          uid: VAR_LOKI_DATASOURCE,
          type: 'loki',
        },
        expr: `first_over_time({job="${JOB_NAME}", container="agent-poc"} | logfmt | op="query_sample_summary" | digest="${digest}" | unwrap duration_seconds(elapsed_time)[$__auto])`,
        instant: true,
        queryType: 'range',
      },
      {
        refId: 'P99',
        datasource: PROMETHEUS_DS,
        expr: quantile(job, digest, schema, instance, '99'),
      },
    ],
  });
}

```

So each query should specify a concrete datasource (e.g., Prometheus, Loki). The "mixed" concept only applies at the panel level when you're passing multiple queries from multiple datasources to Grafana Explore. Does that make sense?

torkelo: but line 55 sets query datasource to mixed, which is strange it only applies to a set of queries (ie on SceneQueryRunner level). Not on individual query level

torkelo: I don't think you need to modify any query level datasource, they should all remain as is. think just updating so that the datasource variable set on line 46 is set to mixed when hasMixedDataSources is true

joannaWebDev: Oh, I see my mistake. Fixed

torkelo (approved):

PR #1076, opened 2025-03-19
Local dev: link to all @grafana modules in core

ashharrison90:
- updates the linking script to link to all `@grafana` core modules
- this removes the need for [webpack aliasing in core](https://github.com/grafana/grafana/blob/main/scripts/webpack/webpack.dev.js#L55-L58) when linking scenes locally to core
- adds a cleanup function to cleanup the links once the script shuts down
- adds systemjs types to devdeps to resolve the systemjs types used in data and runtime correctly otherwise this repo falls back to declarations in testing-library and gets grumpy 😠 
- see [related PR in core](https://github.com/grafana/grafana/pull/102449)

### DIFF
--- a/packages/scenes/package.json
+++ b/packages/scenes/package.json
@@ -75,6 +75,7 @@
     "@types/react-dom": "18.2.24",
     "@types/react-grid-layout": "1.3.2",
     "@types/react-virtualized-auto-sizer": "1.0.1",
+    "@types/systemjs": "^6.15.1",
     "@types/uuid": "8.3.4",
     "@typescript-eslint/eslint-plugin": "^4.33.0",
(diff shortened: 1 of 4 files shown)


jackw (approved):
Changes look good to me and appears to work locally for me even with the TS errors coming from grafana core dev builds. Pretty sure the issue is [this types pointing to source code](https://github.com/grafana/scenes/blob/main/packages/scenes/package.json#L9) which the ForkTsChecker in core Grafana then picks up. We either change that to point to the d.ts files which would also improve type checking performance everywhere scenes is used OR we update scenes to use the new jsx transform / make it aligned with the tsconfig settings in core grafana.

Would be best to get an additional review from someone that knows the library better before merging.

PR #1092, opened 2025-04-09
VizPanel: Add shouldMigrate() callback to migration logic

leeoniya:
depends on https://github.com/grafana/grafana/pull/103710

related: https://github.com/grafana/grafana/pull/108998

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanel.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.tsx
@@ -233,13 +233,10 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
 
     _UNSAFE_customMigrationHandler?.(panel, plugin);
 
-    if (
-      plugin.onPanelMigration &&
-      (currentVersion !== pluginVersion ||
-        //@ts-expect-error (TODO: remove after upgrading with https://github.com/grafana/grafana/pull/108998)
-        plugin.shouldMigrate?.(panel)) &&
-      !isAfterPluginChange
-    ) {
+    //@ts-expect-error (TODO: remove after upgrading with https://github.com/grafana/grafana/pull/108998)
+    const needsMigration = currentVersion !== pluginVersion || plugin.shouldMigrate?.(panel);
+
+    if (plugin.onPanelMigration && needsMigration && !isAfterPluginChange) {
       // These migration handlers also mutate panel.fieldConfig to migrate fieldConfig
       panel.options = await plugin.onPanelMigration(panel);
     }


packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -230,7 +230,11 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
 
     _UNSAFE_customMigrationHandler?.(panel, plugin);
 
-    if (plugin.onPanelMigration && currentVersion !== pluginVersion && !isAfterPluginChange) {
+    if (
+      plugin.onPanelMigration &&
+      (currentVersion !== pluginVersion || plugin.shouldMigrate?.(panel)) &&
```
torkelo, line 235: I assume you will need a ts ignore error here as shouldMigrate does not exist yet (until we upgrade packages).
could you refactor this to a separate module function named `shouldMigrate(plugin, panel, currentVersion, pluginVersion). 

So we can get the `if (plugin.shouldMigrate?.(panel) on a single line` so we can ts-ignore it specifically (And add a todo to remove the ts-ignore on next upgrade)

adela-almasan (approved):

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -233,7 +233,13 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
 
     _UNSAFE_customMigrationHandler?.(panel, plugin);
 
-    if (plugin.onPanelMigration && currentVersion !== pluginVersion && !isAfterPluginChange) {
+    if (
+      plugin.onPanelMigration &&
+      (currentVersion !== pluginVersion ||
+        //@ts-expect-error (TODO: remove after upgrading with https://github.com/grafana/grafana/pull/108998)
+        plugin.shouldMigrate?.(panel)) &&
```
torkelo: I would find this if statement more readable/understandable if it was something like 

```ts
 //@ts-expect-error (TODO: remove after upgrading with https://github.com/grafana/grafana/pull/108998)
const needsMigration = currentVersion !== pluginVersion || plugin.shouldMigrate?.(panel))

if (plugin.onPanelMigration && needsMigration && !isAfterPluginChange)
```

torkelo (approved):

packages/scenes/src/components/VizPanel/VizPanel.tsx
leeoniya: updated

PR #1093, opened 2025-04-09
fix: (VizPanelRenderer) - move showMenuAlways prop to the top level

L2D2Grafana:
# Description
Move showMenuAlways VizPanelRenderer.tsx prop out of collapse conditional. In Logs Drilldown we have a situation where we'd like to `showMenuAlways` but do not use collapsible. [upgrade to Scenes6](https://github.com/grafana/logs-drilldown/pull/1019/files#diff-0c9eb85b8eb906fbac0d474a7c68685514172ee72d0597664fa8618f27826087)
I'm not totally familiar with the change to making `showMenuAlways` conditional on the `collapsible` property [PR](https://github.com/grafana/scenes/pull/1040/files#diff-b4e762e26aeb297a8f36c1ada3213ecd5a9bcd972799a753018dccf96fc75a30R195
).🤔 @tskarhed let me know what you think.

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanel.test.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.test.tsx
@@ -874,7 +874,7 @@ describe('VizPanel', () => {
     it('should toggle collapsed state', () => {
       panel.onToggleCollapse(true);
       expect(panel.state.collapsed).toBe(true);
-      
+
       panel.onToggleCollapse(false);
       expect(panel.state.collapsed).toBe(false);
     });
@@ -934,7 +934,7 @@ describe('VizPanel', () => {
     it('should toggle menu visibility', () => {
       panel.setState({ showMenuAlways: false });
       expect(panel.state.showMenuAlways).toBe(false);
-      
+
       panel.setState({ showMenuAlways: true });
       expect(panel.state.showMenuAlways).toBe(true);
     });
--- a/packages/scenes/src/components/VizPanel/VizPanel.test.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.test.tsx
@@ -849,6 +849,96 @@ describe('VizPanel', () => {
       });
     });
   });
+
+  describe('Collapsible panel functionality', () => {
+    let panel: VizPanel<OptionsPlugin1, FieldConfigPlugin1>;
+
+    beforeEach(async () => {
+      panel = new VizPanel<OptionsPlugin1, FieldConfigPlugin1>({
+        title: 'Collapsible Panel',
+        pluginId: 'custom-plugin-id',
+        collapsible: true,
+        collapsed: false,
+        $data: getDataNodeWithTestData(),
+      });
+
+      pluginToLoad = getTestPlugin1();
+      panel.activate();
+    });
+
+    it('should initialize with correct collapsible state', () => {
+      expect(panel.state.collapsible).toBe(true);
+      expect(panel.state.collapsed).toBe(false);
+    });
+
+    it('should toggle collapsed state', () => {
+      panel.onToggleCollapse(true);
+      expect(panel.state.collapsed).toBe(true);
+      
+      panel.onToggleCollapse(false);
+      expect(panel.state.collapsed).toBe(false);
+    });
+  });
+
+  describe('Hover header functionality', () => {
+    let panel: VizPanel<OptionsPlugin1, FieldConfigPlugin1>;
+
+    beforeEach(async () => {
+      panel = new VizPanel<OptionsPlugin1, FieldConfigPlugin1>({
+        title: 'Hover Panel',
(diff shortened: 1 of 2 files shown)


packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
```diff
@@ -192,14 +192,14 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
...
-                  collapsed,
-                  onToggleCollapse: model.onToggleCollapse,
-                  showMenuAlways,
-                }
-              : { hoverHeader, hoverHeaderOffset })}
+            {...{
+              ...(collapsible !== undefined && { collapsible: Boolean(collapsible) }),
+              ...(collapsed !== undefined && { collapsed }),
+              ...(model.onToggleCollapse && { onToggleCollapse: model.onToggleCollapse }),
+              ...(showMenuAlways !== undefined && { showMenuAlways }),
```
gtk-grafana, line 199: I think the `showMenuAlways` can be moved outside of the collapsible conditional, I had added the type incorrectly (but it was [fixed recently](https://github.com/grafana/grafana/pull/103553)) to the collapsible interface, but it should be a valid prop whenever the hoverHeader is not set. Might require a `@ts-expect-error` though until the version of `@grafana/grafana-ui` is added as a dep

packages/scenes/src/components/VizPanel/VizPanel.test.tsx
```diff
@@ -849,6 +849,96 @@ describe('VizPanel', () => {
...
+      expect(panel.state.collapsed).toBe(false);
+    });
+
+    it('should toggle collapsed state', () => {
+      panel.onToggleCollapse(true);
+      expect(panel.state.collapsed).toBe(true);
+
+      panel.onToggleCollapse(false);
+      expect(panel.state.collapsed).toBe(false);
+    });
```
gtk-grafana, line 880: let's assert in here that we can set the `showMenuAlways` when collapsible is true

packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
```diff
@@ -192,12 +192,12 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
             onDragStart={(e: React.PointerEvent) => {
               dragHooks.onDragStart?.(e, model);
             }}
+            showMenuAlways={showMenuAlways}
```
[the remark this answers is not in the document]
gtk-grafana: let's make sure to add a comment explaining why, and link to the PR with the fix: https://github.com/grafana/grafana/pull/103553

packages/scenes/src/components/VizPanel/VizPanel.test.tsx
gtk-grafana: and we should also assert that we can set the `showMenuAlways` when the width is set and also when undefined.

L2D2Grafana: Updated the collapsible test with `showMenuAlways`. Not sure I understand, how is that dependent on width?

gtk-grafana: Just another way to assert that the showMenuAlways prop is valid for all panel chrome subtypes except for `HoverHeader`. Look at the [PanelChromeProp types definition](https://github.com/grafana/grafana/blob/main/packages/grafana-ui/src/components/PanelChrome/PanelChrome.tsx#L28).

It is probably overkill but would be nice to demonstrate that the ability to set the panel menu as always visible should be an option if the `PanelChromeProps` type is `Collapsible`, `FixedDimensions`, or `AutoSize` but not `HoverHeader`.

This way if someone tries to remove that ts-expect-error in the future before the type fix is pulled in the unit tests will provide an extra layer of insulation against a breaking change and hint to people that the types are wrong.

torkelo (commented):
I don't understand all the tests, a don't see any corresponding logic ? 

looks like they just test the constructor passes state props to state?

gtk-grafana:
> I don't understand all the tests, a don't see any corresponding logic ?
> 
> looks like they just test the constructor passes state props to state?

These tests are just trying to ensure that we don't break the desired functionality to match the types that were incorrect until this [recent fix in g12](https://github.com/grafana/grafana/pull/103553), like what happened here: https://github.com/grafana/scenes/pull/1040/files#diff-b4e762e26aeb297a8f36c1ada3213ecd5a9bcd972799a753018dccf96fc75a30R195.

L2D2Grafana:
> I don't understand all the tests, a don't see any corresponding logic ?
> 
> looks like they just test the constructor passes state props to state?

Updated it to check for the dom element as well.

gtk-grafana (approved):
Looks good, thanks for taking a look at this @L2D2Grafana!

PR #1095, opened 2025-04-10
AdHocFilters: Dashboard level filters

mdvictor:
This PR adds support for dashboard level filters, which are filters set in the schema, under the adhoc variables `baseFilter` property. This new type of baseFilter will need a specific `origin: 'dashboards'` property to indicate it is a dashboard level filter.

Dashboard level filters will stick on your dashboard, they can be cleared (become a regex match-all filter) but not removed, their value can be edited and afterwards restored. Going to another dashboard with this filter will maintain it if there is another dashboard level filter with the same key or remove it, unless the filter is edited. If it is edited it will maintain it across dashboards as a normal filter.

This PR also simplifies the code a bit by not carrying original values (the values that an edited filter restores to), but instead keeping originals in runtime and setting a flag on edited filters as `restorable`. Modified `baseFilters` will be maintained in the URL so they can be moved between dashboards.

https://github.com/user-attachments/assets/3c524f32-fbfa-40f3-b7d0-c7734308719b

<details>
  <summary>📦 Published PR as canary version: <code>6.10.0--canary.1095.14635793130.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.10.0--canary.1095.14635793130.0
  npm install @grafana/scenes@6.10.0--canary.1095.14635793130.0
  # or 
  yarn add @grafana/scenes-react@6.10.0--canary.1095.14635793130.0
  yarn add @grafana/scenes@6.10.0--canary.1095.14635793130.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -211,6 +211,9 @@ export class AdHocFiltersVariable
   private _scopedVars = { __sceneObject: wrapInSafeSerializableSceneObject(this) };
   private _dataSourceSrv = getDataSourceSrv();
   private _scopesBridge: SceneScopesBridge | undefined;
+  // holds the originalValues of all baseFilters in a map. The values
+  // are set on construct and used to restore a baseFilter with an origin
+  // to its original value if edited at some point
   private _originalValues: Map<string, { value: string[]; operator: string }> = new Map();
 
   protected _urlSync = new AdHocFiltersVariableUrlSyncHandler(this);
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -288,14 +288,9 @@ export class AdHocFiltersVariable
 
   private _updateScopesFilters(scopes: Scope[]) {
     if (!scopes.length) {
-      const remainingBaseFilters = this.state.baseFilters?.filter((filter) => filter.origin !== FilterOrigin.Scopes);
-
-      if (remainingBaseFilters?.length) {
-        this.setState({
-          baseFilters: remainingBaseFilters,
-        });
-      }
-
+      this.setState({
+        baseFilters: this.state.baseFilters?.filter((filter) => filter.origin !== FilterOrigin.Scopes),
+      });
       return;
     }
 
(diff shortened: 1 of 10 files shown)


packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -233,6 +235,16 @@ export class AdHocFiltersVariable
       patchGetAdhocFilters(this);
     }
 
+    // set dashboard lvl original values in constructor, since we have them from schema
+    this.state.baseFilters?.forEach((baseFilter) => {
+      if (baseFilter.origin === FilterOrigin.Dashboards) {
```
dprokop, line 240: I see the origin ain't yet introduced in the schema?

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -210,6 +211,7 @@ export class AdHocFiltersVariable
   private _scopedVars = { __sceneObject: wrapInSafeSerializableSceneObject(this) };
   private _dataSourceSrv = getDataSourceSrv();
   private _scopesBridge: SceneScopesBridge | undefined;
+  private _originalValues: Map<string, { value: string[]; operator: string }> = new Map();
```
dprokop, line 214: Can you please add a comment in the code what this represents, for a future reference?

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
mdvictor: @dprokop WDYM? origin shows up in the dashboard schema. Is there something else I'm missing?

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -378,20 +423,18 @@ export class AdHocFiltersVariable
...
-        update.originalValue = undefined;
+      const isRestorableOverride = update.hasOwnProperty('restorable');
+      if (
+        !isRestorableOverride &&
+        ((updateValues && !isEqual(updateValues, originalValues?.value)) ||
+          (update.operator && update.operator !== originalValues?.operator))
+      ) {
+        update.restorable = true;
+      } else if (updateValues && isEqual(updateValues, originalValues?.value)) {
+        update.restorable = false;
```
dprokop, line 437: please add comment for this piece of code, it's a bit hard to understand

PR #1096, opened 2025-04-11
Variables: Switch between multi and single select if state changes

torkelo:
* In order to switch betwen single and multi select while visible (mounted) we need to subscribe to state on a higher level than we did
* Renamed renderSelectForVariable MultiOrSingleValueSelect and made it a proper react component, it now subscribe to state. Removed state subscription from the lower level components, they get the state instead. 
* This is for the live editing of variable options in the edit pane

### DIFF
--- a/packages/scenes/src/variables/components/VariableValueSelect.tsx
+++ b/packages/scenes/src/variables/components/VariableValueSelect.tsx
@@ -11,8 +11,7 @@ import {
   useTheme2,
 } from '@grafana/ui';
 
-import { SceneComponentProps } from '../../core/types';
-import { MultiValueVariable } from '../variants/MultiValueVariable';
+import { MultiValueVariable, MultiValueVariableState } from '../variants/MultiValueVariable';
 import { VariableValue, VariableValueSingle } from '../types';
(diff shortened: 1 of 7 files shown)


packages/scenes/src/variables/components/VariableValueSelect.tsx
```diff
@@ -256,10 +261,12 @@ const getOptionStyles = (theme: GrafanaTheme2) => ({
   }),
 });
 
-export function renderSelectForVariable(model: MultiValueVariable) {
-  if (model.state.isMulti) {
-    return <VariableValueSelectMulti model={model} />;
+export function MultiOrSingleValueSelect({ model }: { model: MultiValueVariable }) {
+  const state = model.useState();
```
dprokop, line 265: Ah, I see, so subscribing here instead of downstream makes this component re-adjust based on the isMulti = fals | true. Makes sense!

dprokop (approved):

packages/scenes/src/variables/components/VariableValueSelect.tsx
torkelo: @dprokop yes, exactly

PR #1098, opened 2025-04-11
feat: (sceneQueryRunner) - allow manual timerange override, pass props to useQueryRunner

L2D2Grafana:
POC💡  [original thread](https://raintank-corp.slack.com/archives/C04FRQ2S0AV/p1744213676586359)
- Allow timeRange prop to override 
- Expose props to scenes-react

### DIFF
--- a/packages/scenes-react/src/hooks/useQueryRunner.ts
+++ b/packages/scenes-react/src/hooks/useQueryRunner.ts
@@ -1,10 +1,9 @@
 import { useEffect } from 'react';
-import { SceneDataQuery, SceneQueryRunner } from '@grafana/scenes';
+import { SceneDataQuery, SceneQueryRunner, SceneTimeRangeLike } from '@grafana/scenes';
 import { DataSourceRef } from '@grafana/schema';
 import { isEqual } from 'lodash';
 import { CacheKey } from '../caching/SceneObjectCache';
 import { useSceneObject } from './useSceneObject';
-
 export interface UseQueryOptions {
   queries: SceneDataQuery[];
   maxDataPoints?: number;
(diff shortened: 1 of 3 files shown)


packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -75,6 +75,8 @@ export interface QueryRunnerState extends SceneObjectState {
   dataLayerFilter?: DataLayerFilter;
   // Private runtime state
   _hasFetchedData?: boolean;
+  // Optional timeRange to use instead of sceneGraph timeRange
+  timeRange?: SceneTimeRangeLike;
```
torkelo, line 79: not sure this makes sense, all scene objects can already have local time range via $timeRange

packages/scenes-react/src/hooks/useQueryRunner.ts
```diff
@@ -1,17 +1,19 @@
...
 export interface UseQueryOptions {
   queries: SceneDataQuery[];
   maxDataPoints?: number;
   datasource?: DataSourceRef;
   cacheKey?: CacheKey;
   liveStreaming?: boolean;
   maxDataPointsFromWidth?: boolean;
+  // Optional timeRange to use instead of sceneGraph timeRange
+  runQueriesMode?: 'auto' | 'manual';
+  timeRange?: SceneTimeRangeLike;
```
torkelo, line 16: 1) the comment is on the wrong prop 
2) Feels like passing a SceneTimeRangeLike is wrong here vs a plain TimeRange prop. Also would need to add diffing so when prop changes query is executed (when runQueriesMode is auto)  
3) VizPanel uses sceneGraph.getTimeRange as well to change time range, for zoom in on graphs so that will not work with this approach.

PR #1101, opened 2025-04-14
SceneTimeRange: Fix calculatePercentOfInterval

edvard-falkskar:
Fixes incorrect percent of interval calculation.

`intervalMs / percent` works for the default ` { percent: 10 }` `refreshOnActivate` value but not any other value.

Found the problem by setting it to 1% which should have invalidated my 5m time range in 2.4s but it didn't.


javiruiz01 (approved):
Works!

PR #1109, opened 2025-04-23
Validate timezone

scottlepp:
when a timezone is not valid, default to `browser`

<details>
  <summary>📦 Published PR as canary version: <code>6.10.1--canary.1109.14643606089.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.10.1--canary.1109.14643606089.0
  npm install @grafana/scenes@6.10.1--canary.1109.14643606089.0
  # or 
  yarn add @grafana/scenes-react@6.10.1--canary.1109.14643606089.0
  yarn add @grafana/scenes@6.10.1--canary.1109.14643606089.0
  ```
</details>


packages/scenes/src/core/constants.ts
```diff
@@ -0,0 +1 @@
+export const BROWSER = 'browser';
```
torkelo, line 1: think you can use defaultTimeZone from `@grafana/schema` instead

packages/scenes/src/core/SceneTimeRange.tsx
```diff
@@ -122,13 +122,13 @@ export class SceneTimeRange extends SceneObjectBase<SceneTimeRangeState> impleme
...
     // Return local time zone if provided
-    if (this.state.timeZone) {
+    if (this.state.timeZone && getValidTimeZone(this.state.timeZone)) {
       return this.state.timeZone;
     }
 
     // Resolve higher level time zone source
     const timeZoneSource = this.getTimeZoneSource();
-    if (timeZoneSource !== this) {
+    if (timeZoneSource !== this && getValidTimeZone(timeZoneSource.state.timeZone)) {
```
torkelo, line 131: don't think we need these, validating in ctor and onTimeZoneChange should be enough

torkelo (approved):

packages/scenes/src/core/SceneTimeRange.tsx
scottlepp: ctor doesn't modify state.  the new tests will fail

PR #1110, opened 2025-04-24
Use log util.  Direct console.warn fails tests in grafana.

scottlepp:
console.warn causes build failures in grafana.  use the log utility instead

```
 PublicDashboardPageProxy › when scene feature enabled › should render PublicDashboardScenePage if publicDashboardsScene is enabled

    Expected test not to call console.warn().
```


harisrozajac (approved):

scottlepp:
> what unit test uses an invalid timezone?
@torkelo you're right.  needed to also handle empty string

PR #1113, opened 2025-04-25
SceneTimePicker: Add defaultQuickRanges to override the default time range options

joshhunt:
There's a PR open to make the default quick time ranges configurable in the config.ini https://github.com/grafana/grafana/pull/102254

This PR adds a new option to SceneTimePicker as a way to pass in these time ranges without having them serialised back into the dashboard JSON.

<details>
  <summary>📦 Published PR as canary version: <code>6.10.3--canary.1113.14667240921.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.10.3--canary.1113.14667240921.0
  npm install @grafana/scenes@6.10.3--canary.1113.14667240921.0
  # or 
  yarn add @grafana/scenes-react@6.10.3--canary.1113.14667240921.0
  yarn add @grafana/scenes@6.10.3--canary.1113.14667240921.0
  ```
</details>


packages/scenes/src/components/SceneTimePicker.tsx
```diff
@@ -93,7 +94,7 @@ function SceneTimePickerRenderer({ model }: SceneComponentProps<SceneTimePicker>
       history={timeRangeHistory}
       //TODO: remove once grafana/grafana updated to 11.6
       //@ts-expect-error
-      quickRanges={quickRanges}
+      quickRanges={defaultQuickRanges || quickRanges}
```
Sergej-Vlasov, line 97: @joshhunt I believe this should be
```typescript
quickRanges={quickRanges || defaultQuickRanges}
```
quickRanges - coming from dashboard JSON and should override defaults

joshhunt:
In order to get tests to run, I bumped grafana-ui in the scenes-app workspace. Not sure if this is the right approach or not. I presume this is well trodden territory, but it kinda accidental that the tests actually run despite not depending on `@grafana/ui`.

Sergej-Vlasov (approved):
@joshhunt Looks good! I tidied up a bit as there was a huge change to yarn.lock. Ideally we leave version updates to a separate PR to avoid any complications so we will handle version update separately!

packages/scenes/src/components/SceneTimePicker.test.tsx
```diff
@@ -116,7 +116,7 @@ describe('SceneTimePicker', () => {
     expect(screen.getByText('2024-09-04 00:00:00 to 2024-09-14 23:59:59')).toBeInTheDocument();
   });
 
-  it.only('renders default quick ranges from server config', async () => {
```
joshhunt: oops thanks, sorry!

CollinFingar (approved):

joshhunt:
@Sergej-Vlasov The problem is that this change uses the new `quickRanges` prop on `<TimeRangePicker />`, which is not in the grafana-ui version in this repo, so the test will fail without it. The existing timePicker functionality in the scenes library is not able to be tested without this.

It's kind of funky in this repo how the scenes package doesn't have an explicit/dev dependency on grafana-ui, an it just relies on the package from scenes-app to be hoisted. The test work accidentally!

Sergej-Vlasov:
> @Sergej-Vlasov The problem is that this change uses the new `quickRanges` prop on `<TimeRangePicker />`, which is not in the grafana-ui version in this repo, so the test will fail without it. The existing timePicker functionality in the scenes library is not able to be tested without this.
> 
> It's kind of funky in this repo how the scenes package doesn't have an explicit/dev dependency on grafana-ui, an it just relies on the package from scenes-app to be hoisted. The test work accidentally!

@joshhunt bumped to 11.6 and addressed ts errors. Should be good to go!

PR #1114, opened 2025-04-28
Tighten security for github actions in repo

mdvictor:


scottlepp (approved):
I think by harden you are saying tighten up security

PR #1115, opened 2025-04-30
AdHocFiltersCombobox: Read-only filters improvements

adrapereira:
- Prevents keyboard from changing read-only filters
- Updates filter tooltip to show when key+operator+value exceeds the preset limit
- Shows key+operator+value in tooltip instead of only the value
- Adds lock icon to read-only filters
- Updates style of read-only filters

image

<details>
  <summary>📦 Published PR as canary version: <code>6.10.3--canary.1115.14859615413.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.10.3--canary.1115.14859615413.0
  npm install @grafana/scenes@6.10.3--canary.1115.14859615413.0
  # or 
  yarn add @grafana/scenes-react@6.10.3--canary.1115.14859615413.0
  yarn add @grafana/scenes@6.10.3--canary.1115.14859615413.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
@@ -83,7 +83,7 @@ export function AdHocFilterPill({ filter, model, readOnly, focusOnWipInputRef }:
             handleChangeViewMode();
           }
         }}
-        role="button"
+        role={readOnly ? undefined : 'button'}
         aria-label={`Edit filter with key ${keyLabel}`}
         tabIndex={0}
         ref={pillWrapperRef}


ifrost (approved):
LGTM! 🎉 



A few nit picks:
- We could force the cursor be to default instead of pointer on the lock as it's not possible to click on it
- I was wondering if we could re-use existing styles `readOnlyCombinedFilter` but not sure what it would break 🤔 
- Might be also worth updating the demo to showcase readonly/restorable filters

I put suggestions in https://github.com/grafana/scenes/pull/1117/files, feel free to pick any (or none)

mdvictor:
Hello! What is the usage of `filter.origin` in this scenario? Seems like this should affect only read only filters?

mdvictor:
> Using `filter.origin` will add the origin to the tooltip of the lock icon. In the example, "Asserts" is the origin.
> 
> image

I see, thank you, makes sense!

PR #1118, opened 2025-05-05
AdHocVariable: Add more suggestive injected filters tooltips

mdvictor:
Adds better tooltips for the new injected filters.

<details>
  <summary>📦 Published PR as canary version: <code>6.10.4--canary.1118.14886789462.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.10.4--canary.1118.14886789462.0
  npm install @grafana/scenes@6.10.4--canary.1118.14886789462.0
  # or 
  yarn add @grafana/scenes-react@6.10.4--canary.1118.14886789462.0
  yarn add @grafana/scenes@6.10.4--canary.1118.14886789462.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
@@ -3,7 +3,7 @@ import { GrafanaTheme2 } from '@grafana/data';
 import { useStyles2, IconButton, Tooltip, Icon } from '@grafana/ui';
 import React, { useState, useRef, useCallback, useEffect } from 'react';
 import { AdHocCombobox } from './AdHocFiltersCombobox';
-import { AdHocFilterWithLabels, AdHocFiltersVariable, isMatchAllFilter } from '../AdHocFiltersVariable';
+import { AdHocFilterWithLabels, AdHocFiltersVariable, FilterOrigin, isMatchAllFilter } from '../AdHocFiltersVariable';
 
 const LABEL_MAX_VISIBLE_LENGTH = 20;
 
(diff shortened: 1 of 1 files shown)


Sergej-Vlasov (approved):

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
```diff
@@ -61,6 +61,16 @@ export function AdHocFilterPill({ filter, model, readOnly, focusOnWipInputRef }:
   }, [viewMode]);
 
   if (viewMode) {
+    let injectedInfoText = '';
```
scottlepp, line 64: if you wrap this with a function you won't need to declare the variables using let
something like 
```
function getTooltips(filter) {
    if (filter.origin === 'dashboard') {
      return {injectedInfoText: 'Applied by default in this dashboard. If edited, it carries over to other dashboards.',
      injectedRestoreText: 'Restore the value defined by this dashboard.'}
    }
   // is `scope` the only other type?
     return { injectedInfoText: 'Applied automatically from your selected scope.',
      injectedRestoreText: 'Restore the value set by your selected scope.'}
    }
}
```

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
```diff
@@ -61,6 +61,16 @@ export function AdHocFilterPill({ filter, model, readOnly, focusOnWipInputRef }:
...
+    let injectedRestoreText = '';
+    if (filter.origin === 'dashboard') {
+      injectedInfoText = 'Applied by default in this dashboard. If edited, it carries over to other dashboards.';
+      injectedRestoreText = 'Restore the value defined by this dashboard.';
+    } else if (filter.origin === 'scope') {
```
scottlepp, line 69: is `scope` and `dashboard` the only values?  

If so, you don't need the `else if`.  just default the values to the scope values.

If not, don't you need some default tooltips?

mdvictor: Hmm that's a good point. For now those 2 are the only values, more might be added in the future. I went for an else if just for verbosity but alas, this is set in the schema and one might be able to set filter.origin to be an arbitrary value, so we should have some default message. I'll address this, thank you @scottlepp !

PR #1120, opened 2025-05-09
[WIP] Macros: Add Scopes macro with queryparam formatter

tskarhed:
Work in progress. Add scopes macro, mainly for adding it to link URL's.

Related to https://github.com/grafana/hyperion-planning/issues/208

### DIFF
--- a/packages/scenes/src/variables/macros/index.ts
+++ b/packages/scenes/src/variables/macros/index.ts
@@ -5,6 +5,7 @@ import { AllVariablesMacro } from './AllVariablesMacro';
 import { DataMacro, FieldMacro, SeriesMacro, ValueMacro } from './dataMacros';
 import { UrlMacro } from './urlMacros';
 import { OrgMacro, UserMacro } from './contextMacros';
+import { ScopesMacro } from './scopesMacros';
 
 export const macrosIndex = new Map<string, MacroVariableConstructor>([
   [DataLinkBuiltInVars.includeVars, AllVariablesMacro],
@@ -21,6 +22,7 @@ export const macrosIndex = new Map<string, MacroVariableConstructor>([
   ['__org', OrgMacro],
   ['__interval', IntervalMacro],
   ['__interval_ms', IntervalMacro],
+  ['__scopes', ScopesMacro],
 ]);
 
 /**
--- a/packages/scenes/src/variables/macros/scopesMacros.ts
+++ b/packages/scenes/src/variables/macros/scopesMacros.ts
@@ -1,0 +1,87 @@
+import { Unsubscribable } from 'rxjs';
+import { sceneGraph } from '../../core/sceneGraph';
+import { SceneObject } from '../../core/types';
+import { SceneScopesBridge } from '../../core/SceneScopesBridge';
+import { RefreshEvent } from '@grafana/runtime';
+import { CustomVariableValue, VariableCustomFormatterFn } from '../types';
+import { FormatVariable, formatRegistry } from '../interpolation/formatRegistry';
+import { VariableFormatID } from '@grafana/schema';
+
+export class ScopesMacro implements FormatVariable {
+  public state: { name: string; type: string };
+  private _scopesBridge?: SceneScopesBridge;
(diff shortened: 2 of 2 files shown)


packages/scenes/src/variables/macros/scopesMacros.ts
```diff
@@ -0,0 +1,87 @@
...
+    this.state = { name: name, type: 'scopes_macro' };
+    this._scopesBridge = sceneGraph.getScopesBridge(this._sceneObject);
+
+    // Subscribe to scope changes
+    if (this._scopesBridge) {
+      this._subscription = this._scopesBridge.subscribeToValue((newScopes, prevScopes) => {
+        // Force a re-render when scopes change
+
+        //this._sceneObject.forceRender();
+        this._sceneObject.publishEvent(new RefreshEvent());
```
tskarhed, line 25: I can't seem to get the rendered values update when I change scopes. Both of these options cause some bugs.

packages/scenes/src/variables/macros/scopesMacros.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    // For all other formats, use the default formatter from the registry
+    const formatter = formatRegistry.getIfExists(formatNameOrFn as string);
+    if (formatter) {
+      return formatter.formatter(this._scopes, [], {
+        state: { name: 'scopes', type: 'scopes_macro' },
+        getValue: () => this._scopes,
+      });
+    }
```
tskarhed, line 82: Not a fan of this. Would appreciate any suggestion of falling back to the standard formatting

packages/scenes/src/variables/macros/scopesMacros.ts
torkelo: I dont think you need this part, if scopes where part of the data request you can get the scopes from the dataContext

torkelo: see dataMacro for how to access it

torkelo: hm.. wait, does not look like request is part of the dataContex only the data frames

torkelo: anyway, this constructor is only called when the macro is evaluated so you cannot handle any subscriptions via the macro

packages/scenes/src/variables/macros/index.ts
```diff
@@ -21,6 +22,7 @@ export const macrosIndex = new Map<string, MacroVariableConstructor>([
   ['__org', OrgMacro],
   ['__interval', IntervalMacro],
   ['__interval_ms', IntervalMacro],
+  ['__scopes', ScopesMacro],
```
aocenas, line 25: @torkelo do macros need to be defined separately for scenes? I would probably expect any scenes string interpolation to also go through template_srv (or having an interface where such service can be injected) and use the same macros/values as other parts of Grafana.

torkelo: @aocenas no they are part of the scenes variable system. template_srv will call the scenes variable interpolator if it detects a scene context (via scopedVars or a global context)

aocenas: does that mean we have to make sure these things are synchronized? Or is this meant to be different in some way? Just want to make sure what this means for any work that needs to add some interpolation macro.

torkelo: @aocenas not sure what synchronization you mean. macros only need to be registered here and will work from any scene app interpolation, or a datasource that calls templateSrv.replace

aocenas: So is this the "source of truth" for what macros exist around in grafana? Will these also work outside of scenes context? The `a datasource that calls templateSrv.replace` would suggest this is the case just want to make sure.

torkelo: @aocenas yes, since templateSrv will call the scene interpolator

PR #1121, opened 2025-05-14
Scopes: Replace ScopesBridge with ScopesVariable 

torkelo:
Problem

* We have use case to access scopes as a variable (in data links and dashboard links) 
* We have many places that need to manually look up scopes bridge and subscribe to it, check it's loading state, block query execution when loading, issue new queries when updated etc. This we all get for free with the variable system. 

Solution

* [x] Replace SceneScopesBridge with ScopesVariable 
* [x] ScopesVariable can fill the same purpose with a lot less special lookup/subscription code 
* [x] New unit tests (could not find any unit tests for ScenesScopesBridge) 
* [x] ScopesVariable can enable/disable scopes just like scopes bridge could (or if you leave enable unset leave it up to the app to control it on higher level using the ScopesContext) 
* [x] sceneGraph.getScopesBridge => sceneGraph.getScopes (now just returns the scopes value, no need to return an object that can be subscribed to, as that is now handled by the variable system) 
* [x] Add dependsOnScopes to VariableDependencyConfig (simple option that just adds the `__scopes` variable to the dependencies) 
* [x] Adds new scopesDemo so we can more easily test scopes when developing and for scene app devs to see how to use it. 
* [x] Update SceneQueryRunner tests
* [x] Update AdhocFilterVariable tests
* [x] Write release notice 
* [x] Manual (in app) testing (Been unable to get any scopes to select in the UI, could not get of the scripts to insert scopes to work) 
* [x] New tests for updates to SceneVariableSet 

OSS PR: https://github.com/grafana/grafana/pull/105399

# Release notes

SceneScopesBridge is now replaced by ScopesVariable, which now needs to be added to a SceneVariableSet attached to every SceneAppPage or EmbeddedScene (whatever suits your scenario) where you want to use scopes. This variable is not visible in the UI. You can control whether scopes are enabled or disabled on a specific page/scene using the ScopesVariable or control this on a higher level using ScopesContext. 

sceneGraph.getScopesBridge is replaced by sceneGraph.getScopes, which just returns the scopes (by looking up the ScopesVariable and getting it from it's value). 

<details>
  <summary>📦 Published PR as canary version: <code>6.12.0--canary.1121.15070832166.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.12.0--canary.1121.15070832166.0
  npm install @grafana/scenes@6.12.0--canary.1121.15070832166.0
  # or 
  yarn add @grafana/scenes-react@6.12.0--canary.1121.15070832166.0
  yarn add @grafana/scenes@6.12.0--canary.1121.15070832166.0
  ```
</details>


packages/scenes-app/src/demos/scopesDemo.tsx
```diff
@@ -0,0 +1,39 @@
...
+  VizPanel,
+} from '@grafana/scenes';
+import { getEmbeddedSceneDefaults, getPromQueryInstant } from './utils';
+import { EmbeddedSceneWithContext } from '@grafana/scenes-react';
+
+export function getScopesDemo(defaults: SceneAppPageState) {
+  return new SceneAppPage({
+    ...defaults,
+    $variables: new SceneVariableSet({
+      variables: [new ScopesVariable({ enable: true })],
```
tskarhed, line 17: Is this the only thing you have to do to enable scopes in an app? Where is this being set in order to get the support in dashboards?

torkelo: @tskarhed yes, this is the only thing needed. In dashboards, we would do the same, add this variable when feature toggle is enabled.

packages/scenes-app/src/demos/scopesDemo.tsx
```diff
@@ -0,0 +1,40 @@
...
+  VizPanel,
+} from '@grafana/scenes';
+import { getEmbeddedSceneDefaults, getPromQueryInstant } from './utils';
+import { EmbeddedSceneWithContext } from '@grafana/scenes-react';
+
+export function getScopesDemo(defaults: SceneAppPageState) {
+  return new SceneAppPage({
+    ...defaults,
+    $variables: new SceneVariableSet({
+      variables: [new ScopesVariable({ enable: true }), new AdHocFiltersVariable({})],
```
tskarhed, line 18:
```suggestion
      variables: [new ScopesVariable({ enable: true }), new AdHocFiltersVariable({ layout: 'combobox' })],
```
This is needed to see scope injected filters

dprokop (approved):
@torkelo  - fantastic refactor. All crystal clear with a single question from me that I'm not certain about.

I've spent some time testing this, and have not noticed any regressions. Scopes are correctly passed to queries and ad hoc calls. Sync between context and variable seems to be working OK.

@bfmatei @mdvictor  - please have a look at this and review. It would be great if we could land this soon as this is quite an impactful refactor.

packages/scenes/src/variables/sets/SceneVariableSet.ts
```diff
@@ -294,7 +291,11 @@ export class SceneVariableSet extends SceneObjectBase<SceneVariableSetState> imp
             otherVariable.onCancel();
           }
 
-          this._variablesToUpdate.add(otherVariable);
+          if (otherVariable.validateAndUpdate) {
+            this._variablesToUpdate.add(otherVariable);
+          }
+
+          otherVariable.variableDependency.variableUpdateCompleted(variableThatChanged, true);
```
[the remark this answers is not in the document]
torkelo: @dprokop I can add a comment here (still need to write tests for the updates to SceneVariableSet). 

The change is because before SceneVariableSet did not propagate variable dependency changes to other variables inside it (this was only done via adding it to update queue, which calls validateAndUpdate on the variable). So far, that has been enough as variables that set _variableDependency all implement validateAndUpdate, but the ScopesVariable does notneed to do that. it just wants to set _variableDependency and handle the variable dependency change itself.  

With this change we make sure variables get's notified of other variable dependency changes like any object

bfmatei:
Just had a look over as well. Super awesome work!

bfmatei (approved):

aocenas:
Looks good to me but can't say much about it in the context of scenes architecture, definitely looks simpler now, though.

mdvictor (approved):
LGTM! Tested adhocs scope injected filters and works well. there is one quirk when setting the scope and editing the injected filter and moving to another dashboard it doesn't maintain the edited value but changes to the scope value, but this happens only the first time you set the scope, editing again and changing dashboards maintains the correct value.

IIRC it comes from how the scopes are being emitted and not related to this PR, seems like the same happens with scopesBridge. I'll have a look into this when refactoring this next week.

bfmatei (approved):

PR #1124, opened 2025-05-15
VizPanel: Cloning should ignore pluginInstanceState

torkelo:
Because plugins (panels) can put anything into pluginInstanceState (like recursive structures), it's not safe to deep clone (nor safe to ref copy), so leaving it out of the clone is best. 

Fixes perf issue with canvas panel (and hanging issue)

* Move clone to separate file
* Move tests closer to code 
* Add new test for VizPanel clone 

https://github.com/grafana/support-escalations/issues/16251
https://github.com/grafana/support-escalations/issues/16210

<details>
  <summary>📦 Published PR as canary version: <code>6.11.0--canary.1124.15040845199.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.11.0--canary.1124.15040845199.0
  npm install @grafana/scenes@6.11.0--canary.1124.15040845199.0
  # or 
  yarn add @grafana/scenes-react@6.11.0--canary.1124.15040845199.0
  yarn add @grafana/scenes@6.11.0--canary.1124.15040845199.0
  ```
</details>


kaydelaney (approved):

kaydelaney:
Nice! I was thinking, since we introduced `cloneDeep` into the scene object cloning procedure to avoid an issue with duplication (see https://raintank-corp.slack.com/archives/C03KVDHTWAH/p1730897481961219), maybe we could change the code to only use `cloneDeep` when we're duplicating, but not for repeats. To me, it sort of actually makes sense for repeat panels' to share most of the same underlying data, since they're just "repeats" and not fully independent panels, like a duplicated one might be. But maybe the performance of cloning scene objects is good enough after this PR that it doesn't matter? What do you think?

torkelo:
@kaydelaney yea, maybe. I dont really like that we need to make those deepClones at all, the non-scene object states should be immutable, wish we could freeze objects (to trigger errros when ever plugins try to mutate state that should be immutable) and revert https://github.com/grafana/scenes/pull/967

PR #1125, opened 2025-05-15
Scopes: Add queryparam formatting

tskarhed:
Adds `queryparam` formatting to the ScopesVariable. Defaults other formatting to `, ` delimiter between scopes.

### DIFF
--- a/packages/scenes/src/variables/variants/ScopesVariable.test.tsx
+++ b/packages/scenes/src/variables/variants/ScopesVariable.test.tsx
@@ -5,14 +5,13 @@ import { act, render, screen } from '@testing-library/react';
 import { SceneVariableSet } from '../sets/SceneVariableSet';
 import { EmbeddedScene } from '../../components/EmbeddedScene';
 import { VariableValueSelectors } from '../components/VariableValueSelectors';
-import { ScopesVariable, ScopesVariableFormatter } from './ScopesVariable';
+import { ScopesVariable } from './ScopesVariable';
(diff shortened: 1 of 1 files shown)


packages/scenes/src/variables/variants/ScopesVariable.test.tsx
```diff
@@ -47,6 +47,26 @@ describe('ScopesVariable', () => {
...
+          metadata: { name: 'scope2' },
+          spec: { title: 'scope2', type: 'type2', description: 'description2', category: 'scope', filters: [] },
+        },
+      ],
+    });
+
+    const value = variable.getValue() as ScopesVariableFormatter;
+
+    expect(value.formatter('queryparam')).toEqual('scope=scope1&scope=scope2');
+    expect(value.formatter()).toEqual('scope1, scope2');
```
torkelo, line 68: not sure why it needs to return this formatted? why not just return all the scopes (as string array)? should result in the exact same behavior.

torkelo: hm.., I guess that will result in `__scopes=scope1&__scopes=scope2`, so maybe custom formater is needed

torkelo (approved):

packages/scenes/src/variables/variants/ScopesVariable.test.tsx
tskarhed: The other option would be to check for `__scopes` in the queryparam default implementation and handle it there, but I think that's a bad pattern.

PR #1137, opened 2025-05-28
chore: unify license change in #327

gtk-grafana:
We only updated the main package.json in #327, but npm is still saying AGPL-3. Lets unify these licenses so there's no ambiguity that scenes is licensed under Apache-2.0.


cbos:
Good to see this PR, to make it clear.

Based on the original change, I asked a question on the PR: https://github.com/grafana/scenes/pull/327#issuecomment-2753479296 

To me it sounds logical to correct this.

bossinc (approved):

PR #1143, opened 2025-06-08
Variables: URL parameter namespaces

gtk-grafana:
Proof of concept allowing variables to specify namespaces of url parameters.

Exposes `getVariableUrlName` method for consuming applications that need to know computed url parameter name for variables.

WIP - PoC - Worked on in a jetlagged state on busses and airports. Needs careful audit before opening review.

### DIFF
--- a/packages/scenes-app/src/monitoring-app/GrafanaMonitoringApp.tsx
+++ b/packages/scenes-app/src/monitoring-app/GrafanaMonitoringApp.tsx
@@ -1,7 +1,7 @@
 // Libraries
 import React, { useMemo } from 'react';
 
-import { SceneApp, SceneAppPage, SceneRouteMatch, SceneAppPageLike, DEFAULT_VARIABLE_NAMESPACE } from '@grafana/scenes';
+import { SceneApp, SceneAppPage, SceneRouteMatch, SceneAppPageLike, VARIABLE_NAMESPACE } from '@grafana/scenes';
 
 import {
   getOverviewScene,
@@ -39,14 +39,14 @@ export function getMainPageScene() {
         url: prefixRoute('grafana-monitoring'),
         routePath: '',
         getScene: getOverviewScene,
-        preserveUrlKeys: ['from', 'to', `${DEFAULT_VARIABLE_NAMESPACE}-instance`],
+        preserveUrlKeys: ['from', 'to', `${VARIABLE_NAMESPACE}-instance`],
       }),
       new SceneAppPage({
         title: 'HTTP handlers',
         url: prefixRoute('grafana-monitoring/handlers'),
         routePath: 'handlers/*',
         getScene: getHttpHandlerListScene,
-        preserveUrlKeys: ['from', 'to', `${DEFAULT_VARIABLE_NAMESPACE}-instance`],
+        preserveUrlKeys: ['from', 'to', `${VARIABLE_NAMESPACE}-instance`],
         drilldowns: [
           {
             routePath: prefixRoute('grafana-monitoring/handlers/:handler'),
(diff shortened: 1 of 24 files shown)


matyax (commented):
LGTM!

grafakus (commented):
@gtk-grafana Thank you for the PoC and trying to find a way to solve this issue!

I have only one suggestion:

Instead of having to pass a namespace to _all_ the variables whenever we instantiate them, could we just pass it once, maybe to `UrlSyncContextProvider`?

The namespace is likely to be the same within the app... Also, conceptually, it's an URL-related issue, so it seems reasonable to solve it closer to where URLs are handheld. 

Lastly, if the app uses `SceneObjectUrlSyncConfig`, it would also be handheld by Scenes without having to explicitly specify the namespace in several places in the app code.

What do you think?

gtk-grafana:
> could we just pass it once?

@grafakus Great question, and I have to admit that sounds lovely from an implementation standpoint, however we have the requirement that there are some variables that we do want to conflict, e.g. time range changes (`from` and `to` url params) are expected to conflict so when changes are made in one embedded app, that state change can be shared with the embedder or other embedded sibling plugins.

It is for this reason we've given developers the ability to specify namespaces for every variable, the same way developers already have that level of granularity in url bindings defined in the _urlSync keys.

Take a look at the [design doc](https://docs.google.com/document/d/1BHvligFtTpJT7B5Pm-_QYG_WYeZGcdXtCrVQNmCTIY8/edit?tab=t.0) for more background

matyax:
+1 to 

```ts
<UrlSyncContextProvider namespace={plugin.id} excludeFromNamespace={["from", "to"]} ...>
  <trail.Component model={trail} />
</UrlSyncContextProvider>
```

gtk-grafana:
So I'm not sure that passing anything into `UrlSyncContextProvider` is going to work without some serious changes to scenes architecture, but the url syncing in scenes is not simple so I could very well be missing something. Here is my current understanding in case someone else wants to take a stab at this proposed alternate implementation.

### Investigation

The `SceneObjectUrlSyncHandler` is a react context, and it is not aware of its children as scene objects, so there doesn't appear to be an apparent solution for how we could make the relationship between the scene objects and the urlSyncManager, which is required so we could get the namespace from the context and into the scene object internals so they know what their key is. In my PoC I added the urlSyncManager to the scene state so it could be queried, but noticed that while my urls were updated, and things appeared to be rendering, setting state throughout the application wasn't working, and values weren't getting pulled from the url...

The UrlSyncManager subscribes to scene state changes, and data flows unidirectionally from scenes into the react context which handles updates to the locationService singleton.
It appears that this is the case because the keys are needed to get the initial value before activation of the scene objects (i.e. `getUrlState` is called before activation), otherwise variables would get rendered and any subscriptions to their state would get an incorrect initial value, so storing the urlSyncManager in the scene tree won't work as a way of getting the namespace into the keys. 

Since we need to know the keys of variables before they are activated, and since scene objects are added to the scene tree during activation, we don't have any way to get the namespace from the UrlSyncManager in a scene object at the time we call `getUrlState` in an orphaned scene object that is not yet attached to the tree. 
We also can't call a hook to fetch the data from the react context because the react object doesn't exist before activation.

I believe we would need to store the namespace outside of scenes and react, like storing the namespace globally somewhere (e.g. a singleton), but since we can't assume a single instance of scenes in an embedded context, we'd still need another way to make the relationship between each specific scene tree and the namespaces stored in the global variable, which would need to be through data exposed in a react context or in scene state, which we've already ruled out won't work as the scene objects that would need that information need it before they are attached to the scene tree, and before they are rendered.

### Summary
@grafakus @matyax that would have been lovely, but it looks like every variable that is namespaced will need to know what its namespace is at time of instantiation, and not during scene activation or during render, as we need to pull values from the url before a scene object is activated. 

For reference, here is the initial work required to implement [this PR](https://github.com/grafana/logs-drilldown/pull/1328/files) in Logs Drilldown. It's not a huge lift adding a namespace to the constructor of each variable, especially in comparison with the changes required in core scenes to support a simpler API, which would likely be a breaking change (although I did find a few bugs with the link extensions I need to look into).

Here is the (failed) [PoC branch](https://github.com/grafana/scenes/tree/gtk-grafana/variable-namespace-poc-url-sync-manager) if anyone wants to play around with it, but I'm going to set this down and continue on the path outlined in this PR.

matyax:
Understood. Thanks for the investigation and the detailed explanation. If the current architecture doesn't allow it, the current implementation is a good alternative that also has a small footprint on the plugins.

grafakus:
@gtk-grafana Great analysis!

Based on the info you've shared, I've managed to make something work in [this branch](https://github.com/grafana/scenes/compare/main...grafakus/poc-url-search-params-namespace)

It's currently limited to passing a `urlSyncOptions.namespace` when instantiating a `SceneApp`. Let me know what you think.

gtk-grafana:
@grafakus Nice! Yeah that does work quite well, although I don't think we can use the scene graph to lookup the config as we need to know the namespaces before scene object activation, but since the current embedding solution already requires that `UrlSyncContextProvider` is wrapping the exposed component I'm not sure it matters if the embedding application is using a SceneApp since the `UrlSyncContextProvider` is required, and will pass the options into the `UrlSyncManager`. 

Great work, if you can open that as a PR and add some test coverage we might have a more palatable solution for embedding apps then what's proposed here.

matyax:
Thank you @grafakus !

PR #1145, opened 2025-06-11
VariableDependencyConfig: Simplify bits and pieces

kaydelaney:
Was looking through `VariableDependencyConfig.ts` while debugging an issue and thought some things could use a little tweak for easier readability. This is partly personal preference so I'm happy to go with the crowd if others don't find it more readable. 😛


scottlepp (approved):
Looks a lot cleaner.  I see there is an existing test for this.  Not sure if the test cases cover all these scenarios.  Might want coverage tools to confirm.

PR #1146, opened 2025-06-11
fix groupBy edge case

mdvictor:
Builds on top of https://github.com/grafana/scenes/pull/1142

Fixes another edge case that happens on first load of a dashboard without a default groupBy or any value set on it, then moving to a dashboard with a default value, but instead of showing the default value it shows whatever value was saved on the groupBy with the possibility to restore. This is not desired and should show the default value since there are no other values in the URL from the transition.

All these scenarios will benefit a lot from some E2Es that I will be working on as soon as possible to cover both groupBys and adHocs.

### DIFF
--- a/packages/scenes/src/variables/groupby/GroupByVariableUrlSyncHandler.ts
+++ b/packages/scenes/src/variables/groupby/GroupByVariableUrlSyncHandler.ts
@@ -56,11 +56,7 @@ export class GroupByVariableUrlSyncHandler implements SceneObjectUrlSyncHandler 
 
       if (restorableValue === 'false') {
         if (this._sceneObject.state.defaultValue) {
-          this._sceneObject.changeValueTo(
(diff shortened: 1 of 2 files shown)


packages/scenes/src/variables/groupby/GroupByVariableUrlSyncHandler.ts
```diff
@@ -78,6 +78,12 @@ export class GroupByVariableUrlSyncHandler implements SceneObjectUrlSyncHandler
       }
 
       this._sceneObject.changeValueTo(values, texts);
+    } else if (this._sceneObject.state.defaultValue) {
```
ivanortegaalba: Any test we can write for this case we're solving? To avoid regressions mostly :)

mdvictor: yes, added tests for this

PR #1148, opened 2025-06-13
LazyLoading: Do not issue queries for panels outside view port

torkelo:
Alternative to https://github.com/grafana/scenes/pull/1147 

* Adds LazyLoaderInViewContext with simple boolean value (if something is in view)
* New function isInViewChanged on DataProvider interface so that viz panel can forward it 

One big requested feature is a ability to scroll to and highlight a panel (in a dashboard or scene app), but because the lazy loading is on the layout level the panels out of view are not rendered so there is no panel level key/ref we can easily scroll. I am thinking we should move the lazy loader to be on the panel level (that way we do not need it on layout level)

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
@@ -1,4 +1,4 @@
-import React, { RefCallback, useCallback, useMemo } from 'react';
+import React, { RefCallback, useCallback, useEffect, useMemo } from 'react';
 import { useMeasure } from 'react-use';
 
 // @ts-ignore
@@ -14,6 +14,8 @@ import { VizPanel } from './VizPanel';
 import { css, cx } from '@emotion/css';
 import { debounce } from 'lodash';
 import { VizPanelSeriesLimit } from './VizPanelSeriesLimit';
+import { useLazyLoaderIsInView } from '../layout/LazyLoader';
+import { SceneQueryRunner } from '../../querying/SceneQueryRunner';
 
 export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
   const {
@@ -61,9 +63,16 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
   const timeZone = sceneTimeRange.getTimeZone();
   const timeRange = model.getTimeRange(dataWithFieldConfig);
 
+  // Switch to manual query execution if the panel it outside viewport
+  const isInView = useLazyLoaderIsInView();
+  useEffect(() => {
+    if (dataObject.isInViewChanged) {
+      dataObject.isInViewChanged(isInView);
+    }
+  }, [isInView, dataObject]);
+
   // Interpolate title
   const titleInterpolated = model.interpolate(title, undefined, 'text');
-
   const alertStateStyles = useStyles2(getAlertStateStyles);
 
   if (!plugin) {
--- a/packages/scenes/src/querying/SceneQueryRunner.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.ts
@@ -420,6 +420,13 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
       return;
     }
 
+    if (this.isQueryModeAuto() && !this._isInView) {
+      this._queryNotExecutedWhenOutOfView = true;
+      return;
+    }
+
+    this._queryNotExecutedWhenOutOfView = false;
+
     // If data layers subscription doesn't exist, create one
     if (!this._dataLayersSub) {
       this._handleDataLayers();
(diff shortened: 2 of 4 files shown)


torkelo:
@bfmatei 

> once a panel is loaded, it doesn't get unloaded hence the queries of panels outside of viewport are still issued. Is this the intended behavior?

no, the entire point of the PR is to avoid queries when out of view, this works when I am testing this PR

torkelo:
@bfmatei 

on http://localhost:3000/a/grafana-scenes-app/demos/lazy-load?from=now-6h&to=now&timezone=browser , if I scroll down and wait for all panels to load, then scroll up and hit refresh (or zoom in while at the bottom), only the visible panels re-issue queries, and if wait and then scroll the panels that where out of view start loading (issuing queries )

bfmatei (approved):
Rebased the branch. Also tested again in core dashboards. The queries are not issued for panels outside of viewport (taking the margin into consideration as well).

packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
```diff
@@ -63,9 +64,16 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
   const timeZone = sceneTimeRange.getTimeZone();
   const timeRange = model.getTimeRange(dataWithFieldConfig);
 
+  // Switch to manual query execution if the panel it outside viewport
```
oscarkilhed:
```suggestion
  // Switch to manual query execution if the panel is outside viewport
```

packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
```diff
@@ -63,9 +64,16 @@ export function VizPanelRenderer({ model }: SceneComponentProps<VizPanel>) {
...
+  const isInView = useLazyLoaderIsInView();
+  useEffect(() => {
+    if (dataObject.isInViewChanged) {
```
mdvictor, line 70: I'm not sure, but would this not potentially call the `isInViewChanged` in `SceneDataTransformer` for panels with transformations? And since it is not implemented in the SceneDataTransformer, the queryRunner will never get around to updating isInView?

I'll test this theory soon in a dashboard, I've yet to try out these changed.

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -661,6 +668,17 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
   private isQueryModeAuto(): boolean {
     return (this.state.runQueriesMode ?? 'auto') === 'auto';
   }
+
+  private _isInView = true;
+  private _queryNotExecutedWhenOutOfView = false;
```
mdvictor: Nit: maybe we can have these properties grouped with the others at the top so we keep everything together?

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -661,6 +668,17 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
...
+
+  public isInViewChanged(isInView: boolean): void {
+    this._isInView = isInView;
```
oscarkilhed, line 677:
```suggestion
    writeSceneLog('SceneQueryRunner', `isInViewChanged: ${isInView}`, this.state.key);
    this._isInView = isInView;
```

packages/scenes/src/querying/SceneQueryRunner.ts
bfmatei: Fixed

packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
bfmatei: Nice catch! Fixed

bfmatei (dismissed):
Testing again. There are some issues with referenced queries:
1. load dashboard. Panel with referenced query is in view and the panel from which we reference query is not in view.
   - same query ran twice: once for the visible panel and (I guess) once for the hidden panel ❌ 
2. scroll down to panel from which we reference query
   - no additional query is issued - this is fine since the query is already issued at step 1 ✅ 
3. scroll up to first panel and back down
   - no additional queries are ran - this is fine since there's no need to issue new queries ✅ 
4. hit the refresh button while scrolled down to the panel from which we reference the query
   - queries for the visible panels are issued - this is fine since this is the intended behavior ✅ 
5. scroll back to the panel that references a query
   - no queries are issued - this is fine since since the query is already issued at step 4 ✅ 
6. while having in view the panel that references a query, hit refresh
   - no queries are issued - this is not fine - my guess is that the query runner that handles it has "isInView" property set to false because it's outside of the view panel ❌ 

Looking for a fix now.

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -116,6 +116,9 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
   private _scopedVars = { __sceneObject: wrapInSafeSerializableSceneObject(this) };
   private _layerAnnotations?: DataFrame[];
   private _resultAnnotations?: DataFrame[];
+  private _isInView = true;
+  private _bypassIsInView = false;
```
bfmatei, line 120: @torkelo @mdvictor - I added this to allow the dashboard datasource to override the isInView behavior temporary. WDYT?

torkelo: @bfmatei looks good.

PR #1150, opened 2025-06-13
Update eslint to v9

joshhunt:
Updates Eslint in the repo to v9, migrates the configs to the new flag config format, and fixes some new lint rules from updated packages

There was previously an incompatibility with the `@typescript-eslint/explicit-member-accessibility` rule where it wouldn't lint class properties. I've disabled this rule on class properties in this PR for ya'll to turn it on and fix the existing errors in your own time

Need https://github.com/grafana/eslint-config-grafana/pull/48 for this.


mdvictor (approved):
lgtm!

PR #1151, opened 2025-06-13
Internationalisation: Add markup for translations

joshhunt:
This PR adds support for internationalisation in Scenes, primarily to ensure all of the Grafana Dashboarding experience is translated.

 - adds `@grafana/i18n` package - this is developer's interface to the i18n framework
 - adds i18next tooling for phrase extraction 
 - adds lint rule to ensure that new phrases are translated
 - adds translation markup + extracted phrases

To test in Grafana

 - Copy `en-US/grafana-scenes.json` to `fr-FR/grafana-scenes.json` and AI translate it or something
 - `yarn build`, `mkdir -p npm-artifacts && yarn lerna exec --no-private -- yarn pack --out "../../npm-artifacts/%s-%v.tgz"`
 - Check out [`jh/scenes-i18n`](https://github.com/grafana/grafana/compare/jh/scenes-i18n) branch
   - Edit package.json to point scenes to the `@grafana-scenes-6.19.0.tgz` you created
   - `rm -rf node_modules/@grafana* node_modules/.cache`
   - `yarn install`
 - Run Grafana and change language (`?lang=fr-FR`).
 - Load a dashboard. The aria label on the row toggle should be in french.

Fixes https://github.com/grafana/grafana/issues/106705

Later we'll have some PRs for the Github Actions to sync the translations with Crowdin.

---

<details>
  <summary>📦 Published PR as canary version: <code>6.20.3--canary.1151.15712722900.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.20.3--canary.1151.15712722900.0
  npm install @grafana/scenes@6.20.3--canary.1151.15712722900.0
  # or 
  yarn add @grafana/scenes-react@6.20.3--canary.1151.15712722900.0
  yarn add @grafana/scenes@6.20.3--canary.1151.15712722900.0
  ```
</details>


crowdin.yml
```diff
@@ -0,0 +1,12 @@
+# The following are pulled from env variables
+project_id_env: CROWDIN_PROJECT_ID
+api_token_env: CROWDIN_PERSONAL_TOKEN
```
mdvictor, line 3: Where are these vars pulled from?

mdvictor:
This looks really great atm, thanks for adding i18n in scenes!

crowdin.yml
joshhunt: These will come from a Github Action that @ashharrison90 is preparing :)

https://github.com/grafana/grafana-github-actions/blob/main/.github/workflows/crowdin-download.yml
https://github.com/grafana/grafana-github-actions/blob/main/.github/workflows/crowdin-upload.yml

packages/scenes/package.json
```diff
@@ -6,7 +6,7 @@
   "license": "AGPL-3.0-only",
   "homepage": "https://grafana.com/developers/scenes",
   "main": "dist/index.js",
-  "types": "src/index.ts",
+  "types": "dist/index.d.ts",
```
mdvictor: Why do we need this change? Seems like it's causing the `./scripts/demo.sh` script to throw errors, although the demo app compiles properly. Not sure why this happens tho

```
Property '#private' is missing in type 'import("/Users/mdvictor/Workspace/scenes/packages/scenes/src/core/SceneObjectRef").SceneObjectRef<import("/Users/mdvictor/Workspace/scenes/packages/scenes/src/core/types").SceneObject<import("/Users/mdvictor/Workspace/scenes/packages/scenes/src/core/types").SceneObjectState>>' but required in type 'import("/Users/mdvictor/Workspace/scenes/packages/scenes/dist/index").SceneObjectRef<import("/Users/mdvictor/Workspace/scenes/packages/scenes/dist/index").SceneObject<import("/Users/mdvictor/Workspace/scenes/packages/scenes/dist/index").SceneObjectState>>'.
scenes-app:dev:     104 |   private addScene() {
scenes-app:dev:     105 |     const scene = buildNewSubScene();
scenes-app:dev:   > 106 |     this.publishEvent(new NewSceneObjectAddedEvent(scene), true);
scenes-app:dev:         |                                                    ^^^^^
scenes-app:dev:     107 |     this.setState({ scene });
scenes-app:dev:     108 |   }
scenes-app:dev:     109 |
scenes-app:dev:
scenes-app:dev: Found 2 errors in 127 ms.
```

joshhunt: I'm not sure! I don't recall doing that on purpose, but that would explain the build errors I started getting 'randomly'!

packages/scenes/eslint.config.js
```diff
@@ -26,4 +27,21 @@ module.exports = [
...
+    name: 'grafana/i18n-overrides',
+    plugins: {
+      '@grafana/i18n': grafanaI18nPlugin,
+    },
+    files: ['src/**/*.{ts,tsx,js,jsx}'],
+    ignores: ['src/**/*.{test,spec,story}.{ts,tsx}'],
+    rules: {
+      '@grafana/i18n/no-untranslated-strings': [
+        'error',
+        { calleesToIgnore: ['^css$', 'use[A-Z].*'], forceFix: ['src'] },
```
ashharrison90: re: `forceFix` - i feel like it's good as a one time thing, but leaving it permanently is a bit too much magic? wdyt? 🤔

packages/scenes/src/components/layout/grid/SceneGridRow.tsx
```diff
@@ -89,9 +90,14 @@ export function SceneGridRowRenderer({ model }: SceneComponentProps<SceneGridRow
     <div className={cx(styles.row, isCollapsed && styles.rowCollapsed)}>
       <div className={styles.rowTitleAndActionsGroup}>
         <button
+          data-josh="hello"
```
ashharrison90: let's remove `data-josh` 😄

packages/scenes/src/locales/fr-FR/grafana-scenes.json
ashharrison90, line 1: we should probably remove this before merging 🤔

packages/scenes/src/variables/variants/TestVariable.tsx
```diff
@@ -42,7 +43,7 @@ export class TestVariable extends MultiValueVariable<TestVariableState> {
       type: 'custom',
       name: 'Test',
       value: 'Value',
-      text: 'Text',
+      text: t('variables.test-variable.text.text', 'Text'),
```
ashharrison90: given this is a test variable that is "only designed for unit tests and potentially e2e tests.", we could ignore here. but suppose it shouldn't do any harm to translate either 🤷

package.json
```diff
@@ -40,7 +40,8 @@
     "lint": "yarn turbo lint --filter=!website",
     "typecheck": "yarn turbo typecheck --filter=!website",
     "prettier:check": "prettier --check --list-different=false --log-level=warn \"**/*.{ts,tsx,scss,md,mdx,json,js}\"",
-    "prettier:write": "prettier --list-different \"**/*.{js,ts,tsx,scss,md,mdx,json}\" --write"
+    "prettier:write": "prettier --list-different \"**/*.{js,ts,tsx,scss,md,mdx,json}\" --write",
+    "i18n:extract": "yarn workspace @grafana/scenes run i18n:extract"
```
ashharrison90: i think everywhere else we've used `i18n-extract` instead of `i18n:extract`

don't think it really matters, but may matter if we want a reusable workflow to check translation extraction has been committed? 🤔

packages/scenes/package.json
```diff
@@ -31,15 +31,18 @@
     "test": "jest",
     "lint": "eslint --cache .",
     "prepack": "node ./scripts/prepack.js",
-    "postpack": "node ./scripts/postpack.js"
+    "postpack": "node ./scripts/postpack.js",
+    "i18n:extract": "i18next --config src/locales/i18next-parser.config.cjs"
```
ashharrison90: same here re: `i18n:extract` vs `i18n-extract`

packages/scenes/src/locales/loadTranslationResources.ts
```diff
@@ -0,0 +1,15 @@
...
+
+const loadTranslationResources: ResourceLoader = async (locale: string) => {
+  switch (locale) {
+    case 'fr-FR':
+      return (await import('./fr-FR/grafana-scenes.json')).default;
+    case 'en-US': // Don't load the en-US bundle, it's the default in the source
+    default:
+      return Promise.resolve({});
+  }
+};
```
ashharrison90: i guess this will need to change to be something like https://github.com/grafana/grafana/blob/main/packages/grafana-prometheus/src/loadResources.ts?

also think i've just spotted something... the `loadResources` example above won't work with pseudo in the `LANGUAGES` array will it? 😬 or at the very least it will attempt to make an unnecessary request

packages/scenes/src/locales/loadTranslationResources.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const initTranslations = () => {
+  return initPluginTranslations('grafana-scenes', [loadTranslationResources]);
+};
```
ashharrison90: pretty minor, but we could remove this and only expose the resource loader and leave it up to the consumer to handle this e.g. grafana core

that way we don't have to worry about namespace collisions across disparate repos. we'd only have to worry about it in grafana core wherever the `initPluginTranslations` call is 🤔 although tbh i don't know what other cases there are, so maybe it's not a big deal

ashharrison90 (commented):
nits mostly, great job 👍

packages/scenes/src/variables/variants/TestVariable.tsx
joshhunt: I think its fine to leave in. It's just one word 🤷 

It better tests i18n :)

package.json
joshhunt: i was trying to match the scenes convention, but you're right, this should be unified for shared workflow

packages/scenes/src/locales/loadTranslationResources.ts
joshhunt: I'll double check, but I don't believe rollup supports dynamic 'glob imports' out of the box, and I was hesitant to add a plugin to handle it. Maybe one for @mdvictor to see if the team has a preference.

Downside of this approach is the switch needs to be updated manually for each language.

packages/scenes/src/locales/loadTranslationResources.ts
joshhunt: This is a good idea... except because scenes library depends on its own version of `@grafana/i18n`, it has its own instance that needs to be initialised. We can't call it from grafna core because that won't create tFunc within scenes.

joshhunt: Or actually. I've now noticed that scenes depends on `@grafana/ui` as a peer-dependency, so it's not actually install as a direct dep of this package, and it would share the instance from core.

Lets dig more into this...

packages/scenes/src/locales/loadTranslationResources.ts
ashharrison90: i don't follow - those aren't glob imports, it's just setting up an object containing all the individual language imports 🤔

joshhunt: ``import(`./locales/${lang.code}/grafana-prometheus.json`)`` is a dynamic import that the bundler special-cases in order to actually bundle arbitary files based on a variable

ashharrison90: hmmmmm... doesn't `grafana-prometheus` use rollup? 🤔

mdvictor: I reckon it would be nicer to have it dynamically imported if possible. Still not sure if it is? if prometheus uses rollup too, I reckon that means it should be possible to have something like in `loadResources`?

mdvictor (approved):
LGTM! All that seems to remain are the 2 open conversations @ashharrison90 started

packages/scenes/src/locales/loadTranslationResources.ts
joshhunt: @mdvictor done!

packages/scenes/package.json
ashharrison90: is this what's causing the type errors in CI as well? 🤔

ashharrison90 (approved):
looks great - just the question about the package.json types changes 👍

packages/scenes/package.json
joshhunt: argh! this is back! I think `yarn build` that does this?

packages/scenes/src/locales/loadTranslationResources.ts
joshhunt: Changed to peer dep to match how this will work in other packages.

joshhunt:
@mdvictor Finally, this should be all good now!

Since you last reviewed
 - use the dynamic variable/glob import to load the translations
 - expose the same `loadResources` function that other packages use
 - prefixed all keys with 'grafana-scenes' to prevent key collisions with grafana itself

mdvictor (approved):
@joshhunt thank you for the update on changes!

Code looks good and everything builds as expected so LGTM!

gtk-grafana:
This is a breaking change for Grafana <12.1 as the `@grafana/i18n` package was released in 12.1.
We need to revert and fallback, or revert and re-release this as a major/breaking release

ashharrison90:
> This is a breaking change for Grafana <12.1 as the `@grafana/i18n` package was released in 12.1. We need to revert and fallback, or revert and re-release this as a major/breaking release

@gtk-grafana why? we specifically thought about this and came to the conclusion it was not breaking - did we miss something? 🤔

gtk-grafana:
@ashharrison90 
Yeah it breaks plugins when updating to scenes >= 6.21.0. Building scene dependent plugins returns a bunch of errors like:

```
ERROR in ../node_modules/@grafana/scenes/dist/esm/components/NestedScene.js 1:0-34
Module not found: Error: Can't resolve '@grafana/i18n' in '/Users/galen/projects/grafana/explore-logs/node_modules/@grafana/scenes/dist/esm/components'
 @ ../node_modules/@grafana/scenes/dist/esm/index.js 54:0-58 54:0-58
 @ ./services/datasource.ts 85:0-64 100:43-60 571:4-40
 @ ./module.tsx 42:45-76
 
 ERROR in ../node_modules/@grafana/scenes/dist/esm/variables/variants/TestVariable.js 1:0-34
Module not found: Error: Can't resolve '@grafana/i18n' in '/Users/galen/projects/grafana/explore-logs/node_modules/@grafana/scenes/dist/esm/variables/variants'
 @ ../node_modules/@grafana/scenes/dist/esm/index.js 38:0-68 38:0-68
 @ ./services/datasource.ts 85:0-64 100:43-60 571:4-40
 @ ./module.tsx 42:45-76
```

But the [workaround above](https://github.com/grafana/scenes/pull/1151#issuecomment-3005223611) allows plugins to build again.

joshhunt:
Just to make sure everyone's on the same page:
 - The `@grafana/i18n` package is not 'provided by grafana' in the same way that other packages like `@grafana/ui` are, so there should be no change to minimum Grafana version for your plugin
 - `@grafana/i18n` is defined as a peer dependency (like `@grafana/ui` etc), so must be installed in the consuming package
 - `@grafana/i18n` is published to NPM, already with the `latest` tag (ahead of 12.1), and is intended to be bundled into your code 
 - There shouldn't be any runtime breaking changes to the scenes library. `@grafana/i18n` is designed to gracefully degrade down to English-only when used in older versions of Grafana.

Ultimately, to use newer versions of Scenes consumers (plugins) will need to add `@grafana/i18n` as a dependency to their package. The only breaking change question is around whether we should have bumped the major version or not, which we did not consider when releasing these changes.

PR #1152, opened 2025-06-16
AdHocFiltersVariable: Non-applicable filters 

mdvictor:
WIP

This PR adds the possibility to view filters within the dashboard which are not applicable to any query in the dashboard.

https://github.com/user-attachments/assets/2a6220c3-6ab1-4612-a13a-052cf72b0f1c

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -37,6 +37,8 @@ export interface AdHocFilterWithLabels<M extends Record<string, any> = {}> exten
   readOnly?: boolean;
   // whether this specific filter is restorable to some value from _originalValues
   restorable?: boolean;
+  // sets this filter as non-applicable
+  nonApplicable?: boolean;
 }
 
 export type AdHocControlsLayout = ControlsLayout | 'combobox';
@@ -247,7 +249,9 @@ export class AdHocFiltersVariable
   }
 
   private _activationHandler = () => {
+    console.log('LOL');
     this._updateScopesFilters();
+    this._verifyNonApplicableFilters();
 
     return () => {
       // we clear both scopes and dashboard level filters before leaving a dashboard to maintain accuracy for both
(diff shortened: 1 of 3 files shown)


packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -534,6 +541,58 @@ export class AdHocFiltersVariable
...
+        const isNonApplicable = response.find((filter) => filter.value === key && filter.nonApplicable);
+
+        if (isNonApplicable) {
+          const filter = this._originalValues.get(key);
+          if (filter) {
+            filter.nonApplicable = true;
+          }
+        }
+      }
+    });
```
dprokop: @mdvictor  - could you leave a comment why is this needed?

mdvictor: On further thought this isn't needed. My logic was that one could move from one dashboard to another with a dashboard or scope level filter that is set to match-all. Then, on the second dashboard, one could reset the matchall to whatever value it had and it would be valid, although the key might not be applicable. 

But that doesn't make sense because if the dashboard level filter is moved to the other dashboard, then the key is set on that dashboard, so it's obviously valid. Same for scope filter: the scope would be carried across due to the scopes dashboard selector, which again only shows valid dashboards to that scope.

I will remove this

mdvictor: I've updated the code around this too, there is another scenario I forgot about: when turning a non applicable filter to a match-all, we need to keep a reference that it is non applicable in the originalValues map, so if/when we restore that match-all it still keeps it as non applicable. Latest commit deals with that

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -534,6 +546,55 @@ export class AdHocFiltersVariable
     }
   }
 
+  public async _verifyNonApplicableFilters() {
+    const filters = [...this.state.filters, ...(this.state.originFilters ?? [])];
```
dprokop, line 550: Consider this a nit, but  it would improve readability of the code a bit: I think the `originFilters` is easily to be confused with `_originalValues`. How about renaming it to `filtersWithOrigin`?

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -534,6 +546,55 @@ export class AdHocFiltersVariable
...
+
+    const ds = await this._dataSourceSrv.get(this.state.datasource, this._scopedVars);
+    // @ts-ignore (TODO)
```
dprokop: I think we should have `@ts-expect-error` here. Then, when we update to grafana-data version that has this API, ts will mark this directive as no longer needed. With `@ts-ignore` the ignore can live forever unnoticed.

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -534,6 +546,55 @@ export class AdHocFiltersVariable
...
+    if (!ds || !ds.getApplicableFilters) {
+      return [];
+    }
+
+    const timeRange = sceneGraph.getTimeRange(this).state.value;
+    const queries = this.state.useQueriesAsFilterForOptions ? getQueriesForVariables(this) : undefined;
+
+    // @ts-ignore (TODO)
```
dprokop: Same here, and in the lines below.

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -534,6 +546,55 @@ export class AdHocFiltersVariable
...
+        if (!f.matchAllFilter) {
+          this._updateFilter(f, { nonApplicable: true });
+        }
+
+        const originalValue = this._originalValues.get(f.key);
+        if (originalValue) {
+          originalValue.nonApplicable = true;
+        }
+      }
+    });
```
dprokop: Why can't we do a batch update of the filters, but instead you update filters one by one?

mdvictor: You're right, there's no point in updating individually, I've changed to a batch update. I also realized `renderExpression` was not building the correct expression taking nonApplicables into account, that's been changed as well in the latest commit

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
mdvictor: I agree, it is quite hard to read between the two, although I won't push for this name change in this PR, but will create a separate one as it touches a number of files and would pollute this one.

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
dprokop: thank you!

dprokop (approved):
@mdvictor thank you for considering my feedback. I've made another review - no more comments from my side.

PR #1155, opened 2025-06-17
SceneQueryRunner: Detect local variable change

torkelo:
This is related to row repeats: https://github.com/grafana/grafana/pull/104312 

The issue with the current way scene query runner records variable values on deactive, is that it can then record a value that was either just set or removed. So it will record a value that has not actually been used in a query. 

By changing to record values on every query we can be sure that when the scene query runner re-activates that it will compare against the variables that it used when it last issued a query. 

This scenario is mainly relevant for row / tab repeats where the first row get's a local variable value or it get's removed, this should trigger a new query on the first row but because scene query runner records the value on deactivate it records the new value and when it- reactivates it thinks nothing has changed 

<details>
  <summary>📦 Published PR as canary version: <code>6.22.0--canary.1155.15756683630.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.22.0--canary.1155.15756683630.0
  npm install @grafana/scenes@6.22.0--canary.1155.15756683630.0
  # or 
  yarn add @grafana/scenes-react@6.22.0--canary.1155.15756683630.0
  yarn add @grafana/scenes@6.22.0--canary.1155.15756683630.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/types.ts
+++ b/packages/scenes/src/variables/types.ts
@@ -11,14 +11,7 @@ export interface SceneVariableState extends SceneObjectState {
   label?: string;
   hide?: VariableHide;
   skipUrlSync?: boolean;
-  /**
-   * True when possible values are being fetched
-   */
   loading?: boolean;
-  /**
-   * True when current value is not yet validated/updated and a dependency is in loading state
-   **/
-  dependencyLoading?: boolean;
   error?: any | null;
   description?: string | null;
 }
--- a/packages/scenes/src/variables/types.ts
+++ b/packages/scenes/src/variables/types.ts
@@ -11,7 +11,14 @@ export interface SceneVariableState extends SceneObjectState {
   label?: string;
   hide?: VariableHide;
   skipUrlSync?: boolean;
+  /**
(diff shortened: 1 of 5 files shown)


packages/scenes/src/variables/types.ts
```diff
@@ -11,7 +11,14 @@ export interface SceneVariableState extends SceneObjectState {
...
   hide?: VariableHide;
   skipUrlSync?: boolean;
+  /**
+   * True when possible values are being fetched
+   */
   loading?: boolean;
+  /**
+   * True when current value is not yet validated/updated and a dependency is in loading state
+   **/
+  dependencyLoading?: boolean;
```
bfmatei: This doesn't seem to be used anywhere.

bfmatei (approved):
One small nit. Other than that, this looks good

packages/scenes/src/variables/types.ts
torkelo: @bfmatei ah, good catch, did not end up using this (tried to solve this various ways)

PR #1157, opened 2025-06-18
SceneTimeRangeCompare: Add option to hide checkbox

drew08t:
Related to: https://github.com/grafana/grafana/pull/104672

For some use cases of `SceneTimeRangeCompare`, requiring the checkbox to toggle the comparison selection is redundant. For example, if time comparison is enabled in panel options (like in the PR above), it becomes an additional step to enable time comparison selection, especially considering that `NO_PERIOD_VALUE` is always included in the select options and can be used to turn comparison off.

In this PR, we add optional `hideCheckbox` option to `SceneTimeRangeCompare`.

Before (now the default behavior, hideCheckbox undefined):
Jun-17-2025 17-47-48

After (hideCheckbox = true):
Jun-17-2025 17-48-17

### DIFF
--- a/packages/scenes/jest.config.ts
+++ b/packages/scenes/jest.config.ts
@@ -3,6 +3,7 @@ const esModules = ['ol', 'd3', 'd3-color', 'd3-interpolate', 'delaunator', 'inte
 module.exports = {
   moduleNameMapper: {
     '\\.css$': '<rootDir>/utils/test/__mocks__/style.ts',
+    '\\.svg$': '<rootDir>/utils/test/__mocks__/style.ts',
     'react-inlinesvg': '<rootDir>/utils/test/__mocks__/react-inlinesvg.tsx',
   },
   testEnvironment: 'jsdom',
--- a/packages/scenes/src/components/SceneTimeRangeCompare.test.tsx
+++ b/packages/scenes/src/components/SceneTimeRangeCompare.test.tsx
@@ -26,7 +26,7 @@ describe('SceneTimeRangeCompare', () => {
         expect(result).toMatchInlineSnapshot(`
               [
                 {
-                  "label": "No comparison",
+                  "label": "None",
                   "value": "__noPeriod",
                 },
                 {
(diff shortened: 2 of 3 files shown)


packages/scenes/jest.config.ts
```diff
@@ -3,6 +3,7 @@ const esModules = ['ol', 'd3', 'd3-color', 'd3-interpolate', 'delaunator', 'inte
 module.exports = {
   moduleNameMapper: {
     '\\.css$': '<rootDir>/utils/test/__mocks__/style.ts',
+    '\\.svg$': '<rootDir>/utils/test/__mocks__/style.ts',
```
drew08t, line 6: I was having trouble getting tests to run without adding this. Not sure if it's needed, might have just been my environment.

dprokop (changes requested):
This not only hides the checkbox, but also the "label". IMO without the label the UI is simply a dropdown that's hard to figure out what it should be used for.

Nothing agains hiding the checkbox - i think it makes sense for the Grafana core use case, but we need to make sure time comparison UI is understandable and discoverable to the user of Grafana dashboards.

natellium:
@dprokop What about changing the text of "No Comparison" to "No time comparison"?

drew08t:
> > What about changing the text of "No Comparison" to "No time comparison"?
> 
> @natellium it's fine with me (knowing how it works). However, to me it's a good UX practice to have a control _with_ a label, rather than a functional button being the label. Alternatively, maybe the control for this shouldn't be a `ToolbarButton` but rather a `Select` so that we could use a placeholder as a label?
> 
> Also, since UI changes here, please have a look at this comment: [grafana/grafana#104672 (review)](https://github.com/grafana/grafana/pull/104672#pullrequestreview-2938553256)

I pushed a commit that uses the select built in label. Also renames "No comparison" to "None".
Jun-18-2025 14-51-23

dprokop (approved):
Think this looks good now, thanks for consideing my feedback.

PR #1159, opened 2025-06-18
Annotations: Add scopes variable dependency

tskarhed:
Closes https://github.com/grafana/hyperion-planning/issues/246


packages/scenes/src/querying/layers/SceneDataLayerBase.ts
```diff
@@ -56,6 +56,7 @@ export abstract class SceneDataLayerBase<T extends SceneDataLayerProviderState>
 
   protected _variableDependency: VariableDependencyConfig<T> = new VariableDependencyConfig(this, {
```
tskarhed, line 57: I tried overriding this in `AnnotationsDataLayer`, with little success. Please let me know if there is any side effect to add the scope dependency here instead.

torkelo (approved):

PR #1162, opened 2025-06-19
i18n: fixes peer dependencies

hugohaggmark:
Scenes is currently targetting `@grafana/i18n` `>=12.1.0` for peerDependencies but any package manager that automatically installs peer deps will now fail on install as 12.1.0 doesn't exist in the registry. This PR sets it to `*` so installs complete whilst we wait on the 12.1.0 release.


jackw (approved):
LGTM! 🚀

PR #1164, opened 2025-06-24
Add join templating format for joining arrays with a custom separator

joshhunt:
Part of https://github.com/grafana/grafana/issues/97570

This adds support to variable interpolation for a custom `join` formatter to join arrays of values with an explicit separator, to be used like `${myVar:join: | }`. This functions identically to `Array.prototype.join`.

 - If the variable value is a string, or only 1 item in the array, it should just return the input
 - If the argument is not specified, it should join the values with `,`

This join operator is a bit too awkward to use to solve the original feature request, so we're going to add a new `customURL` format as well, but just a generic join operator is useful for all as well!

There was also a discussion about supporting quoted arguments, with a more robust variable-format parser. We've decided to not do this for now ([branch still exists here](https://github.com/grafana/scenes/tree/jh/variable-format-parser)) because of the additional complexity. It can be added separately if the need arises.

### DIFF
--- a/packages/scenes/src/variables/interpolation/formatRegistry.test.ts
+++ b/packages/scenes/src/variables/interpolation/formatRegistry.test.ts
@@ -5,11 +5,8 @@ import { TestVariable } from '../variants/TestVariable';
 import { formatRegistry } from './formatRegistry';
 import { VariableFormatID } from '@grafana/schema';
 
-// 'join' is a new supported format not yet in the depended @grafana/schema version
-type FormatId = VariableFormatID | 'join';
-
 function formatValue<T extends VariableValue>(
-  formatId: FormatId,
+  formatId: VariableFormatID,
   value: T,
   text?: string,
   args: string[] = []
@@ -77,9 +74,13 @@ describe('formatRegistry', () => {
     expect(formatValue(VariableFormatID.Date, 1594671549254, 'text', ['YYYY-MM', 'ss'])).toBe('2020-07:09');
     expect(formatValue(VariableFormatID.Date, 1594671549254, 'text', ['YYYY', 'MM', 'DD'])).toBe('2020:07:13');
 
+    // @ts-expect-error join not in depended @grafana/schema yet
     expect(formatValue('join', 'hello', 'text', undefined)).toBe('hello'); // handles non-arrays
+    // @ts-expect-error
     expect(formatValue('join', ['hello'], 'text', undefined)).toBe('hello'); // handles arrays of 1 length
+    // @ts-expect-error
     expect(formatValue('join', ['hello', 'world'], 'text', undefined)).toBe('hello,world'); // has a default separator
+    // @ts-expect-error
     expect(formatValue('join', ['hello', 'world'], 'text', [' | '])).toBe('hello | world'); // has a custom separator
 
     expect(formatValue(VariableFormatID.UriEncode, '/any-path/any-second-path?query=foo()bar BAZ')).toBe(


packages/scenes/src/variables/interpolation/sceneInterpolator.test.ts
```diff
@@ -116,6 +116,85 @@ describe('sceneInterpolator', () => {
     expect(sceneInterpolator(scene, '${test:queryparam}')).toBe('var-test=hello');
   });
 
+  describe('join format', () => {
```
joshhunt: I'll probably remove these tests, but I wanted a place to test the full variable syntax

packages/scenes/src/variables/interpolation/formatRegistry.test.ts
```diff
@@ -74,6 +74,12 @@ describe('formatRegistry', () => {
     expect(formatValue(VariableFormatID.Date, 1594671549254, 'text', ['YYYY-MM', 'ss'])).toBe('2020-07:09');
     expect(formatValue(VariableFormatID.Date, 1594671549254, 'text', ['YYYY', 'MM', 'DD'])).toBe('2020:07:13');
 
+    // TODO: use enum when it's available
+    expect(formatValue('join', 'hello', 'text', undefined)).toBe('hello'); // handles non-arrays
```
dprokop, line 78: ts-expect-error i suppose?

packages/scenes/src/variables/interpolation/sceneInterpolator.test.ts
```diff
@@ -116,6 +116,85 @@ describe('sceneInterpolator', () => {
...
+          variables: [
+            new ConstantVariable({
+              name: 'test',
+              value: ['hello', 'world'],
+            }),
+          ],
+        }),
+      });
+
+      // TODO - idk about this one (it currently fails). Should the argument support quotes? I imagine this
```
dprokop: I'd say either we explicitly require quotes for join symbol, or we do not support them at all (meaning they become the separator).

packages/scenes/src/variables/interpolation/formatRegistry.test.ts
joshhunt: I'll get this out https://github.com/grafana/grafana/pull/107129 and a canary package published, and see if that can be used here? Otherwise, yeah I guess.

packages/scenes/src/variables/interpolation/sceneInterpolator.test.ts
joshhunt: Yeah, I agree with that. 'Problem' is that I don't know whether to support it at all. Looking at how variables are captured from the syntax:

https://github.com/grafana/scenes/blob/40d7bf0772d77022d392a583dbf7738ad801f45f/packages/scenes/src/variables/interpolation/sceneInterpolator.ts#L120-L126

It's pretty simple - just splitting on `:` for seperate arguments. This means it's not possible to have a join separator that itself contains a `:`. Ordinarily, I guess quoting the value is how you could achieve this, but I'm not sure how to add support for quoted values and how thorough it should be - should `${var:function:"first:arg":second-arg}`be supported?

I'm not sure what to do about all of this. Is this something worth supporting? Or just leave it as is and accept the awkwardness of `${var:join: | }` and inability to have separators with `:` in them?

torkelo: The variable system is in great need of a proper lexer/parser (instead of the regex we use) to support more complex syntax and escaping. 

if we add the option to add quotes around values later I guess this syntax would break `${var:join:"a"}`

(ie any value that had quotes), but that feels like a very uncommon/unusual separator :)

joshhunt: @torkelo What do you suggest I do for this one?

torkelo: this this scenario for now (quotes / escaping).

torkelo (commented):
I think this has one big fatal flaw (to solve the customer scenario), in that values need to be URL escaped to support all possible values. 

So either we need to be able to chain formats (something I have thought about but really needs a new function syntax), or we add a joinUrl format that does the same as join but URL encodes values

packages/scenes/src/variables/interpolation/sceneInterpolator.test.ts
```diff
@@ -116,6 +116,85 @@ describe('sceneInterpolator', () => {
...
+          variables: [
+            new ConstantVariable({
+              name: 'test',
+              value: ['hello', 'world'],
+            }),
+          ],
+        }),
+      });
+
+      expect(sceneInterpolator(scene, 'p-tst=v-${test:join:&p-tst=v-}')).toBe('p-tst=v-hello&p-tst=v-world');
```
torkelo: feels very complex/messy to have to specify the first pattern outside the expression and then again in the join param (but with `&` prefix). 

for this use case maybe something like this could be enough (And simpler) `${test:customUrl:p-tst:v-}` where first arg is the URL key, the second is a value prefix.

packages/scenes/src/variables/interpolation/formatRegistry.test.ts
```diff
@@ -5,8 +5,11 @@ import { TestVariable } from '../variants/TestVariable';
 import { formatRegistry } from './formatRegistry';
 import { VariableFormatID } from '@grafana/schema';
 
+// 'join' is a new supported format not yet in the depended @grafana/schema version
+type FormatId = VariableFormatID | 'join';
```
dprokop: @joshhunt WDYT about skipping this type and instead use`ts-expect-error` annotations so that we can rely on TS hinting us we need to update the code when we update grafana/schema version?

joshhunt: ahh yeah, that makes sense!

PR #1165, opened 2025-06-25
sceneGraph: findObject should return first match

torkelo:
Noticed that findObject would always return the last match (if multiple objects match predicate), this fixes it by adding support for early exit from forEachChild (by returning false, same way lodash forEach works )

<details>
  <summary>📦 Published PR as canary version: <code>6.23.0--canary.1165.15877190588.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.23.0--canary.1165.15877190588.0
  npm install @grafana/scenes@6.23.0--canary.1165.15877190588.0
  # or 
  yarn add @grafana/scenes-react@6.23.0--canary.1165.15877190588.0
  yarn add @grafana/scenes@6.23.0--canary.1165.15877190588.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/core/SceneObjectBase.tsx
+++ b/packages/scenes/src/core/SceneObjectBase.tsx
@@ -357,6 +357,7 @@ export abstract class SceneObjectBase<TState extends SceneObjectState = SceneObj
   /**
    * Loop through state and call callback for each direct child scene object.
    * Checks 1 level deep properties and arrays. So a scene object hidden in a nested plain object will not be detected.
+   * Return false to exit loop early.
    */
   public forEachChild(callback: (child: SceneObjectBase) => void) {
     forEachChild(this.state, callback);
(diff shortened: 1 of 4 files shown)


packages/scenes/src/core/SceneObjectBase.tsx
```diff
@@ -421,18 +422,31 @@ export function useSceneObjectState<TState extends SceneObjectState>(
   return model.state;
 }
 
-function forEachChild<T extends object>(state: T, callback: (child: SceneObjectBase) => void) {
+function forEachChild<T extends object>(state: T, callback: (child: SceneObjectBase) => void | false) {
```
scottlepp, line 425: not sure why we return false.  the caller will check this?  a unit test for this scenario would help

bfmatei (approved):
I'm a little bit concerned about introducing this to return early since it's a common pattern in JS/TS to do something like:
```ts
[].forEach(() => {
  if (Math.random() < 0.5) {
    return myFn();
  }

  myFn2();
}
```

If `myFn` return `false`, this behavior would introduce a silent bug.

However I guess this wouldn't be a problem tho' since it's pretty much protected by TS' type system where defining `myFn` as `myFn(): boolean` would make the compiler complain.

packages/scenes/src/core/SceneObjectBase.tsx
bfmatei: Basically returning `false` would be the equivalent of a `break;` in a traditional for loop.

`forEachChild` is used internally by `findObject`

Issue #1167, opened 2025-06-25
AdHocFiltersCombobox: tab navigation in Grafana Drawer crashes browser


gtk-grafana:
Tabbing to a combobox within a Drawer causes an infinite loop of setting and removing focus.

This can be reproduced by simply dropping in a combobox in a `@grafana/ui` Drawer component and tabbing to the input: https://github.com/gtk-grafana/grafana-drawerscenesrepro-app.

PR #1166, opened 2025-06-25
AdHocFiltersCombobox: Update z-index

ashharrison90:
- updates the `z-index` of the `AdHocFiltersCombobox`
- this ensures it can correctly render within `Drawer`/`Modal` components which currently have [lower z-indices](https://github.com/grafana/grafana/blob/main/packages/grafana-data/src/themes/zIndex.ts#L7)

<details>
  <summary>📦 Published PR as canary version: <code>6.24.1--canary.1166.15878540264.0</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install @grafana/scenes-react@6.24.1--canary.1166.15878540264.0
  npm install @grafana/scenes@6.24.1--canary.1166.15878540264.0
  # or 
  yarn add @grafana/scenes-react@6.24.1--canary.1166.15878540264.0
  yarn add @grafana/scenes@6.24.1--canary.1166.15878540264.0
  ```
</details>

PR #1168, opened 2025-06-25
fix: infinite focus loop

gtk-grafana:
* Updates the `FloatingFocusManager` `modal` prop to true, which seems to fix the infinite focus loop.
* Without changes from https://github.com/grafana/scenes/pull/1166, the combobox dropdown will not be visible.

Fixes: https://github.com/grafana/scenes/issues/1167

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersCombobox.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersCombobox.tsx
@@ -616,7 +616,7 @@ export const AdHocCombobox = forwardRef(function AdHocCombobox(
       {optionsLoading ? <Spinner className={styles.loadingIndicator} inline={true} /> : null}
       <FloatingPortal>
         {open && (
-          <FloatingFocusManager context={context} initialFocus={-1} visuallyHiddenDismiss modal={false}>
+          <FloatingFocusManager context={context} initialFocus={-1} visuallyHiddenDismiss modal={true}>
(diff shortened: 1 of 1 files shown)


packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersCombobox.tsx
```diff
@@ -616,7 +616,7 @@ export const AdHocCombobox = forwardRef(function AdHocCombobox(
       {optionsLoading ? <Spinner className={styles.loadingIndicator} inline={true} /> : null}
       <FloatingPortal>
         {open && (
-          <FloatingFocusManager context={context} initialFocus={-1} visuallyHiddenDismiss modal={false}>
+          <FloatingFocusManager context={context} initialFocus={-1} visuallyHiddenDismiss modal={true}>
```
gtk-grafana, line 619: @Sergej-Vlasov do you remember if there was a reason this prop was set to false?

Sergej-Vlasov: @gtk-grafana sorry for missing this PR! I do not remember a particular reason for that prop, so might be a relic of early implementation attempts (because that prop is kinda negated by `<FloatingPortal>`.

Don't think that your changes should break anything but doing some testing to confirm all is good

Sergej-Vlasov (approved):
LGTM! 👍 

Did not notice any regressions during filters ui testing

PR #1173, opened 2025-07-01
VizPanel: Apply transformations returned by the migration handler

bfmatei:
Some panels  (i.e. geomap) mutate the transformations on the panel as part of the migration process.

Unfortunately, these transformations are not available until the panel plugin is loaded hence they can't be appended automatically when a dashboard is loaded and the data provider is created.

Alternatively to this, we could apply the transformations at render time similarly to what we do with series limit (React hook) or field config (memoized data via state object).

<details>
  <summary>📦 Published PR as canary version: <code>6.26.2--canary.1173.16026799419.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.26.2--canary.1173.16026799419.0
  npm install @grafana/scenes@6.26.2--canary.1173.16026799419.0
  # or 
  yarn add @grafana/scenes-react@6.26.2--canary.1173.16026799419.0
  yarn add @grafana/scenes@6.26.2--canary.1173.16026799419.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanel.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.tsx
@@ -246,6 +246,7 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
       if ($data instanceof SceneDataTransformer) {
         $data.setState({ transformations: panel.transformations });
       } else if ($data instanceof SceneQueryRunner) {
+        $data.clearParent();
         $data = new SceneDataTransformer({
           transformations: panel.transformations,
           $data,
--- a/packages/scenes/src/components/VizPanel/VizPanel.test.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.test.tsx
@@ -14,17 +14,22 @@ import {
   PanelData,
   PanelProps,
   toUtc,
+  DataTransformerConfig,
 } from '@grafana/data';
 import * as grafanaData from '@grafana/data';
 import { getPanelPlugin } from '../../../utils/test/__mocks__/pluginMocks';
 
-import { VizPanel } from './VizPanel';
+import { VizPanel, VizPanelState } from './VizPanel';
 import { SceneDataNode } from '../../core/SceneDataNode';
 import { SeriesVisibilityChangeMode } from '@grafana/ui';
 import { SceneTimeRange } from '../../core/SceneTimeRange';
 import { act, render, screen } from '@testing-library/react';
 import { RefreshEvent } from '@grafana/runtime';
 import { VizPanelMenu } from './VizPanelMenu';
+import { SceneDataTransformer } from '../../querying/SceneDataTransformer';
+import { EmptyDataNode } from '../../variables/interpolation/defaults';
+import { mockTransformationsRegistry } from '../../utils/mockTransformationsRegistry';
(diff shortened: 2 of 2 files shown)


packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -236,6 +238,21 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
...
+    if (panel.transformations && $data) {
+      if ($data instanceof SceneDataTransformer) {
+        $data.setState({ transformations: panel.transformations });
+      } else if ($data instanceof SceneQueryRunner) {
+        $data = new SceneDataTransformer({
+          transformations: panel.transformations,
+          $data,
+        });
+      }
+    }
```
dprokop: Think we can only apply this when there's is a `onPanelMigration` and it run?

bfmatei: `panel.transformations` is populated only if `onPanelMigration` ran as `panel` is not the original panel model but something we build manually a little bit above this code.

Will add it tho' for the sake of visibility.

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -236,6 +238,22 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
...
     }
 
+    // Some panels mutate the transformations on the panel as part of migrations.
+    // Unfortunately, these mutations are not available until the panel plugin is loaded.
+    // At this time, the data provider is already set, so this is the easiest way to fix it.
+    let $data = this.state.$data;
+    if (panel.transformations && $data) {
+      if ($data instanceof SceneDataTransformer) {
+        $data.setState({ transformations: panel.transformations });
+      } else if ($data instanceof SceneQueryRunner) {
```
torkelo, line 248: think we only need to handle his case, we always wrap the scene query runner in a SceneDataTransformer in core dashboards.


We do not give panel migrations access to transformations:
https://github.com/grafana/scenes/blob/a67572bce8f0e7bacebfcbd6b5bba764f73b4287/packages/scenes/src/components/VizPanel/VizPanel.tsx#L340

So panel plugins do mutate this could be destructive (ie loose existing transformations)

torkelo, line 248: think we only need to handle his case, we always wrap the scene query runner in a SceneDataTransformer in core dashboards.


We do not give panel migrations access to transformations:
https://github.com/grafana/scenes/blob/a67572bce8f0e7bacebfcbd6b5bba764f73b4287/packages/scenes/src/components/VizPanel/VizPanel.tsx#L340

So panel plugins do mutate this could be destructive (ie loose existing transformations)

bfmatei: I thought about it as well.

However there are a couple of issues:
- `SceneQueryRunner`s are always wrapped in a `SceneDataTransformer` in dashboards
-  backwards compatibility: if a dashboard was saved in a non-scenes architecture, with transformations applied on top, we would miss the transformations brought by panel migrations :-/

From what I've looked in the Grafana/plugins codebase, all our plugins are returning the original transformations as well.

PR #1177, opened 2025-07-02
QueryVariable: Add ability to provide custom static options

domasx2:
One day hackathon effort :) 

Adds ability to add addition "static" options to query variable options, defined in the same way as custom variable options.
They are prepended by default, but can be optionally appened or sorted in per defined sort order.

Use case is to be able to add options like `{ label: "not set", value: "" }` or `{ label: "any production cluster", value: ".*-prod-.*"}`

This is needed for https://github.com/grafana/grafana/issues/102229, with Grafana PR to follow.

### DIFF
--- a/packages/scenes/src/variables/variants/query/QueryVariable.test.tsx
+++ b/packages/scenes/src/variables/variants/query/QueryVariable.test.tsx
@@ -30,6 +30,7 @@ import { act, render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { config, setRunRequest } from '@grafana/runtime';
 import { SafeSerializableSceneObject } from '../../../utils/SafeSerializableSceneObject';
+import { VariableSort } from '@grafana/schema';
 
 const runRequestMock = jest.fn().mockReturnValue(
   of<PanelData>({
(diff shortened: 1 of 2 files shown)


packages/scenes/src/variables/variants/query/QueryVariable.test.tsx
```diff
@@ -425,4 +426,107 @@ describe.each(['11.1.2', '11.1.1'])('QueryVariable', (v) => {
...
     });
   });
+
+  describe('When static options are provided', () => {
+    it('Should prepend static options to the query results when no static options order is provided', async () => {
+      const variable = new QueryVariable({
+        name: 'test',
+        datasource: { uid: 'fake-std', type: 'fake-std' },
+        query: 'query',
+        staticOptions: 'A,B,C',
```
torkelo: rather than using csv string here I think providing an array of VariableValueOption would be better. The csv string format was kind a legacy mistake in CustomVariable (we should have opted for an object model vs parsed csv string as the main state).

domasx2: makes sense!

domasx2: done!

domasx2:
Related Grafana PR: https://github.com/grafana/grafana/pull/107514

torkelo (approved):

PR #1181, opened 2025-07-08
AdHocFilters: Show reason for non-applicable filters

mdvictor:
Screenshot 2025-07-23 at 16 31 10

Builds on top of https://github.com/grafana/scenes/pull/1152 and refactors things a bit, mainly calls a different DS method from `getApplicableFilters` to `getFiltersApplicability`

Non-applicable filters can now also show a reason why the filter pill is invalid. It will also re-evaluate applicability on filter removal.

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.test.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.test.ts
@@ -682,6 +682,8 @@ describe.each(['11.1.2', '11.1.1'])('SceneQueryRunner', (v) => {
       scene.setState({ $variables: new SceneVariableSet({ variables: [filtersVar] }) });
       deactivationHandlers.push(filtersVar.activate());
 
+      await new Promise((r) => setTimeout(r, 1));
+
       filtersVar.setState({
         filters: [
           { key: 'A', operator: '=', value: 'B' },
--- a/packages/scenes/src/querying/__snapshots__/SceneQueryRunner.test.ts.snap
+++ b/packages/scenes/src/querying/__snapshots__/SceneQueryRunner.test.ts.snap
@@ -57,7 +57,7 @@ exports[`SceneQueryRunner when running query should build DataQueryRequest objec
     "from": "now-6h",
     "to": "now",
   },
-  "requestId": "SQR185",
+  "requestId": "SQR186",
   "scopes": undefined,
   "startTime": 1689063488000,
   "targets": [


packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
```diff
@@ -172,7 +172,7 @@ export function AdHocFilterPill({ filter, model, readOnly, focusOnWipInputRef }:
           </Tooltip>
         )}
 
-        {filter.origin && !filter.restorable && !filter.readOnly && (
+        {filter.origin && !filter.restorable && !filter.readOnly && !filter.nonApplicable && (
```
dprokop: Can we extract this condition evaluation to either a function or a variable with a comment that describes what this condition is actually doing?

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -546,58 +570,68 @@ export class AdHocFiltersVariable
...
-        f.nonApplicable = true;
+      if (filter) {
+        f.nonApplicable = !filter.applicable;
+        f.nonApplicableReason = filter.reason;
       }
     });
 
     update.originFilters?.forEach((f) => {
-      const isApplicable = response.includes(f.key);
+      const filter = responseMap.get(`${f.key}-${f.origin}`);
```
dprokop, line 606: Does this code depend on https://github.com/grafana/scenes/pull/1185/?

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -546,58 +570,68 @@ export class AdHocFiltersVariable
...
+    setStateOnCheckFail?: boolean
+  ) {
     const ds = await this._dataSourceSrv.get(this.state.datasource, this._scopedVars);
     // @ts-expect-error (temporary till we update grafana/data)
-    if (!ds || !ds.getApplicableFilters) {
-      return [];
+    if (!ds || !ds.getFiltersApplicability()) {
+      if (setStateOnCheckFail) {
+        this.setState(update);
+      }
```
dprokop: I don't think I understand the `setStateOnCheckFail` flag. Isn't this an equivalent of `!ds || !ds.getFiltersApplicability` really? IIUC what you want to do here is to allow immediate `setState` in case ds does not support the `getFiltersApplicability`.

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -546,58 +570,68 @@ export class AdHocFiltersVariable
...
-
+  public async _setStateWithFiltersApplicabilityCheck(
+    update: Partial<AdHocFiltersVariableState>,
+    setStateOnCheckFail?: boolean
+  ) {
     const ds = await this._dataSourceSrv.get(this.state.datasource, this._scopedVars);
     // @ts-expect-error (temporary till we update grafana/data)
-    if (!ds || !ds.getApplicableFilters) {
-      return [];
+    if (!ds || !ds.getFiltersApplicability()) {
```
dprokop: Why are you calling `getFiltersApplicability` method here? This should be just a check whether or not the data source has this API implemented. The actual execution is happening after that check.

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
mdvictor: no, this is totally independent although the key is similar

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
mdvictor: yess, this is a typo 😞  thank you for catching this!

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
mdvictor: Yeah this is a bit confusing, I refactored it a bit, hoping it makes more sense. Basically that extra flag is for not setting state in the activation handler. So in most cases I replace a `setState` with this new call, but in the activation handler I am not replacing any `setState`: if there is no filters applicability call then I simply want to return because there is for sure no state update. That is what this flag does

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.test.tsx
```diff
@@ -837,6 +837,8 @@ describe.each(['11.1.2', '11.1.1'])('AdHocFiltersVariable', (v) => {
 
     scopesVariable.update();
 
+    await new Promise((r) => setTimeout(r, 1));
```
mdvictor, line 840: need to add this to all scopes updating tests since now it goes through `_setStateWithFiltersApplicabilityCheck` which `awaits` for the ds query

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -325,7 +336,10 @@ export class AdHocFiltersVariable
     ];
 
     // maintain other originFilters in the array, only update scopes ones
-    this.setState({ originFilters: [...finalFilters, ...remainingFilters] });
+    this._setStateWithFiltersApplicabilityCheck({
+      filters: this.state.filters,
```
mdvictor, line 337: I reckon we do not need to clone this as scenes deals nicely with it and keeps things cleaner imo in the adhocs `setState` call when checking update state changes.

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -546,58 +567,68 @@ export class AdHocFiltersVariable
     }
   }
 
-  public async _verifyNonApplicableFilters() {
-    const filters = [...this.state.filters, ...(this.state.originFilters ?? [])];
-
+  public async _setStateWithFiltersApplicabilityCheck(
+    update: Partial<AdHocFiltersVariableState>,
+    returnWithoutStateUpdate?: boolean
```
dprokop: @mdvictor please add a verbose explanation of this flage.

samjewell:
This sounds like a nice improvement 👏 

Small request from me - might you be able to put a screenshot on the PR description, so we can easily see what this will look like? 🙏

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
mdvictor: removed flag alltogether, the code management overhead is quite large for this small of a performance optimization imo

PR #1184, opened 2025-07-09
 Add customqueryparam templating format for joining arrays with a custom separator 

joshhunt:
Part of https://github.com/grafana/grafana/issues/97570

Adds a new templating formatter for customising query parameter strings. It supports two arguments, which are both optional:
 - arg 1: parameter name, defaults to exactly the variable name (without the `var-` prefix)
 - arg 2: value prefix

| Value                 | Template                                   | Result                                  |
| --------------------- | ------------------------------------------ | --------------------------------------- |
| `['api']`             | `?${servers:customqueryparam}`             | `?servers=api`                          |
| `['api']`             | `?${servers:customqueryparam:p-server}`    | `?p-server=api`                         |
| `['api']`             | `?${servers:customqueryparam:p-server:x-}` | `?p-server=x-api`                       |
| `['api', 'database']` | `?${servers:customqueryparam:p-server:x-}` | `?p-server=x-api&p-server=x-database`   |
| `['api', 'mysql db']` | `?${servers:customqueryparam:p-server:x-}` | `?p-server=x-api&p-server=x-mysql%20db` |

Reviewer questions:
 - Instead of a new format, could we just add these arguments to the already existing `queryparam` format?
   - It wouldn't have an option to default to the plain variable name, you would need to specify that explicitly
 - If we keep the seperate format, thoughts on the name? I went for `customqueryparam` to match the existing `queryparam`, but it is quite long!

### DIFF
--- a/packages/scenes/src/variables/interpolation/formatRegistry.test.ts
+++ b/packages/scenes/src/variables/interpolation/formatRegistry.test.ts
@@ -110,9 +110,16 @@ describe('formatRegistry', () => {
     expect(formatValue('customqueryparam', 'api', 'text', ['p-server'])).toBe('p-server=api'); // custom name, optional value
     // @ts-expect-error
     expect(formatValue('customqueryparam', 'api', 'text', ['p-server', 'v-'])).toBe('p-server=v-api'); // value prefix
-    expect(formatValue('customqueryparam', 'mysql&postgres', 'text', ['p-server', 'v-'])).toBe(
-      'p-server=v-mysql%26postgres'
-    ); // value prefix
+
+    // customqueryparam - url encoding
+    // @ts-expect-error
+    expect(formatValue('customqueryparam', ['mysql&postgres', 'databases!'], 'text', ['p-server', 'v-'])).toBe(
+      'p-server=v-mysql%26postgres&p-server=v-databases%21'
+    ); // variable value encoded
+    expect(
+      // @ts-expect-error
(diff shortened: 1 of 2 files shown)


packages/scenes/src/variables/interpolation/formatRegistry.test.ts
```diff
@@ -5,8 +5,11 @@ import { TestVariable } from '../variants/TestVariable';
 import { formatRegistry } from './formatRegistry';
 import { VariableFormatID } from '@grafana/schema';
 
+// 'join' is a new supported format not yet in the depended @grafana/schema version
+type FormatId = VariableFormatID | 'join' | 'customqueryparam';
```
dprokop: you know, same comment you already addressed in `join` pr :)

packages/scenes/src/variables/interpolation/formatRegistry.ts
```diff
@@ -328,6 +340,22 @@ export const formatRegistry = new Registry<FormatRegistryItem>(() => {
         return formatQueryParameter(variable.state.name, value);
       },
     },
+    {
+      id: 'customqueryparam',
```
dprokop, line 344: Let's make sure https://grafana.com/docs/grafana/latest/dashboards/variables/variable-syntax/#advanced-variable-format-options is updated with these new params.

joshhunt: I'll add docs in the PR that updates scenes in Grafana

PR #1186, opened 2025-07-09
Time zone selection: When selecting Default in time zone picker, interpolate timezone from user profile setting

harisrozajac:
This PR fixes a bug where user couldn't select default timezone. See the issue below for additional info and steps to reproduce. 

Fixes: https://github.com/grafana/grafana/issues/107734

This notion of default time zone is a bit confusing because [here](https://github.com/grafana/grafana/blob/e14c17ef69ca86bf5397a6c0d0d2459316f1cf0a/packages/grafana-data/src/datetime/timezones.ts#L8-#L12) we have default as empty string - and in the old arch when we have this empty string we resolve it to default time zone from user profile using `config.bootData.user.timezone` [here](https://github.com/grafana/grafana/blob/20c700dd52373c7b4e6f9560d0c2e35985308cfb/public/app/features/profile/state/reducers.ts#L113-L115).

But in https://github.com/grafana/scenes/pull/1109, we've introduced defaulting to browser and started relying on [defaultTimeZone](https://github.com/grafana/grafana/blob/0fdcae4e264bd9bba70024d7505ee46eacd0f255/packages/grafana-schema/src/common/common.gen.ts) from grafana-schema, which is set to browser.

### DIFF
--- a/packages/scenes/src/core/SceneTimeRange.test.tsx
+++ b/packages/scenes/src/core/SceneTimeRange.test.tsx
@@ -1,4 +1,4 @@
-import { toUtc, dateMath } from '@grafana/data';
+import { toUtc, dateMath, InternalTimeZones } from '@grafana/data';
 import { SceneFlexItem, SceneFlexLayout } from '../components/layout/SceneFlexLayout';
 import { PanelBuilders } from './PanelBuilders';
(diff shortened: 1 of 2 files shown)


Proximyst (approved):
Looks great! Take my review with a big grain of salt :)

PR #1189, opened 2025-07-23
Accessibility: Add `role="menuitem"` to `VizPanelMenu` items

ashharrison90:
- adds `role="menuitem"` to all `VizPanelMenu` items
- that's all the items in this menu:
  image
- @ckbedwell i've read through https://github.com/grafana/grafana/pull/86886, and it sounds like this is fine/desired for this particular implementation of menu/menuitem. but i'd appreciate your 👀 just to make sure 🙏 

Fixes https://github.com/grafana/grafana/issues/106223

<details>
  <summary>📦 Published PR as canary version: <code>6.27.3--canary.1189.16466706909.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.27.3--canary.1189.16466706909.0
  npm install @grafana/scenes@6.27.3--canary.1189.16466706909.0
  # or 
  yarn add @grafana/scenes-react@6.27.3--canary.1189.16466706909.0
  yarn add @grafana/scenes@6.27.3--canary.1189.16466706909.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanelMenu.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanelMenu.tsx
@@ -52,6 +52,7 @@ function VizPanelMenuRenderer({ model }: SceneComponentProps<VizPanelMenu>) {
           return (
             <Menu.Item
               key={item.text}
+              role="menuitem"
               label={item.text}
               icon={item.iconClassName}
               childItems={item.subMenu ? renderItems(item.subMenu) : undefined}


packages/scenes/src/components/VizPanel/VizPanelMenu.tsx
```diff
@@ -52,6 +52,7 @@ function VizPanelMenuRenderer({ model }: SceneComponentProps<VizPanelMenu>) {
           return (
             <Menu.Item
               key={item.text}
+              role="menuitem"
```
joshhunt, line 55: I'm sure we've had this discussion before, but is there a reason we don't add this to the Menu.Item component itself in `@grafana/ui`?

ashharrison90: see the old linked PR in the description. the way I understand it: there's 2 subtly different ways of using this component:
- this way, where it makes sense that everything is set as `menuitem`s
- the help menu example, where having links actually described as links is more useful

rather than set it in grafana/ui and have it apply everywhere, I think the behaviour was kept the same but an optional override was exposed. we're now using this optional override 👍

I don't know which usage is more prevalent or whether it makes sense to switch the default in grafana/ui. safest option is probably leave it as-is?

joshhunt: Sorry, I should have paid more attention to you already answering this :(

PR #1190, opened 2025-07-23
[WIP] Chore: Enable automatic jsx

jackw:
**What is this PR?**

This PR refactors the scenes library to use the automatic jsx transform.

It also updates the rollup config to only build the ts declarations for prod as during dev the package.json `types` property points to `src/index.ts`.

Draft right now whilst I test some things out.

**Why?**

This should solve the 100 odd TS errors that occur when deving scenes against core Grafana.


torkelo:
@jackw thanks! anything we can help with?

PR #1192, opened 2025-07-23
GroupBy: Non-applicable values

mdvictor:


packages/scenes/src/variables/groupby/GroupByVariable.test.tsx
```diff
@@ -515,6 +515,199 @@ describe.each(['11.1.2', '11.1.1'])('GroupByVariable', (v) => {
...
+        // @ts-expect-error (temporary till we update grafana/data)
+        getFiltersApplicability: getFiltersApplicabilitySpy,
+      });
+
+      await act(async () => {
+        await variable._verifyApplicability();
+      });
+
+      expect(getFiltersApplicabilitySpy).toHaveBeenCalled();
+      expect(variable.state.keysApplicability).toBeNull();
```
dprokop: Nit - can we undefined this as `undefined`? looking at the type definition of `keysApplicability` `null` should not be allowed at all:

```ts
keysApplicability?: FiltersApplicability[]
```

packages/scenes/src/variables/groupby/GroupByVariable.tsx
```diff
@@ -179,6 +200,53 @@ export class GroupByVariable extends MultiValueVariable<GroupByVariableState> {
...
+    if (!ds.getFiltersApplicability) {
+      return;
+    }
+
+    const queries = getQueriesForVariables(this);
+    const timeRange = sceneGraph.getTimeRange(this).state.value;
+    const value = this.state.value;
+
+    // @ts-expect-error (temporary till we update grafana/data)
+    const response = await ds.getFiltersApplicability({
```
dprokop, line 235: Just a sanity check, shouldn't we pass filters here too? and same about group by in the filters applicability check?

packages/scenes/src/variables/groupby/GroupByValueContainer.tsx
```diff
@@ -0,0 +1,46 @@
...
+  nonApplicablePill: css({
+    background: theme.colors.action.selected,
+    color: theme.colors.text.disabled,
+    border: 0,
+    '&:hover': {
+      background: theme.colors.action.selected,
+    },
+    textDecoration: 'line-through',
+  }),
+});
```
dprokop: Any chance we could unify non-applicable pill rendering between group by and filters? Now we have two separate rendering paths for similar style.

dprokop (approved):
Looks good, but I've left some things to consider before merging this.

packages/scenes/src/variables/groupby/GroupByVariable.tsx
mdvictor: Should they both be passed? I don't know. It makes sense indeed if filters and groupBys depend on one another. I assumed they didn't.

packages/scenes/src/variables/groupby/GroupByValueContainer.tsx
mdvictor: Makes sense, just added this modification

packages/scenes/src/variables/groupby/GroupByVariable.test.tsx
mdvictor: Indeed, good catch!

PR #1196, opened 2025-07-25
Add option to replace variable macros

oscarkilhed:
In special circumstances we need to be able to replace built in macros.  This PR adds a replace parameter to the registerVariableMacro that allows us to do that. The replaced macro will be restored if the macro is unregistered.

<details>
  <summary>📦 Published PR as canary version: <code>6.28.4--canary.1196.16600049151.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.28.4--canary.1196.16600049151.0
  npm install @grafana/scenes@6.28.4--canary.1196.16600049151.0
  # or 
  yarn add @grafana/scenes-react@6.28.4--canary.1196.16600049151.0
  yarn add @grafana/scenes@6.28.4--canary.1196.16600049151.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/macros/index.ts
+++ b/packages/scenes/src/variables/macros/index.ts
@@ -30,7 +30,7 @@ export const macrosIndex = new Map<string, MacroVariableConstructor>([
  * @returns a function that unregisters the macro
  */
 export function registerVariableMacro(name: string, macro: MacroVariableConstructor, replace = false): () => void {
-  if (!replace &&macrosIndex.get(name)) {
+  if (!replace && macrosIndex.get(name)) {
     throw new Error(`Macro already registered ${name}`);
   }
 


packages/scenes/src/variables/macros/index.ts
```diff
@@ -26,16 +26,26 @@ export const macrosIndex = new Map<string, MacroVariableConstructor>([
 /**
  * Allows you to register a variable expression macro that can then be used in strings with syntax ${<macro_name>.<fieldPath>}
  * Call this on app activation and unregister the macro on deactivation.
+ * @param replace WARNING! If true, this will replace the existing macro with the same name. This is global and will affect all apps. Unregistering the macro will restore the original macro.
  * @returns a function that unregisters the macro
  */
-export function registerVariableMacro(name: string, macro: MacroVariableConstructor): () => void {
-  if (macrosIndex.get(name)) {
+export function registerVariableMacro(name: string, macro: MacroVariableConstructor, replace = false): () => void {
+  if (macrosIndex.get(name) && !replace) {
```
bfmatei: This can be simplified a little bit and hurry up the execution. We only have a `macrosIndex.get` instead of two and a `macrosIndex.has` call.

```ts
export function registerVariableMacro(name: string, macro: MacroVariableConstructor, replace = false): () => void {
  const existingMacro = macrosIndex.get(name);

  if (!replace) {
    throw new Error(`Macro already registered ${name}`);
  }

  macrosIndex.set(name, macro);

  return () => {
    if (replace && existingMacro) {
      macrosIndex.set(name, existingMacro);
    } else {
      macrosIndex.delete(name);
    }
  }
```

packages/scenes/src/variables/macros/index.ts
```diff
@@ -26,16 +26,26 @@ export const macrosIndex = new Map<string, MacroVariableConstructor>([
...
+  if (macrosIndex.has(name) && !replace) {
+    replacedMacro = macrosIndex.get(name);
+  }
+
   macrosIndex.set(name, macro);
 
   return () => {
-    macrosIndex.delete(name);
+    if (replacedMacro) {
+      macrosIndex.set(name, replacedMacro);
```
bfmatei: I don't know if setting the old macro back is a good idea as the old one might not get unregistered.

bfmatei (commented):
Some comments below.

packages/scenes/src/variables/macros/index.ts
oscarkilhed: Yeah, you're right. Maybe we should just treat this as the exception it is, and not let you unregister a macro that was replaced.

packages/scenes/src/variables/macros/index.ts
oscarkilhed: Updated the code to be similar to this given the changes to not allow unregistering replaced macros.

bfmatei (approved):

PR #1197, opened 2025-07-29
LazyLoader: Improve the lazy loading mechanism

bfmatei:
The lazy loader was migrated as-was from Grafana core and it was a little bit buggy and also it contained some unnecessary features.

This PR brings some changes to it:
- Remove function-type children as we don't need it. There is no need to pass a function that propagates the `isInView` property
- Load the content only when it is in view. With this change we aim at:
   - improving the runtime performance given that we don't need to keep the
   - removing the number of queries that we execute

There is a follow up PR that I'll open in Grafana core where this can be seen in action.


kaydelaney (approved):

Sergej-Vlasov (approved):
LGTM!

PR #1204, opened 2025-08-05
Forward variable field path to formatters

oscarkilhed:
In order for variable.getValue text to properly work we need to pass in the field path. An example of this is the __dashboard macro in Grafana that has properties on it like __dashboard.title __dashboard.uuid.


packages/scenes/src/variables/interpolation/formatRegistry.ts
```diff
@@ -314,9 +314,9 @@ export const formatRegistry = new Registry<FormatRegistryItem>(() => {
       id: VariableFormatID.Text,
       name: 'Text',
       description: 'Format variables in their text representation. Example in multi-variable scenario A + B + C.',
-      formatter: (value, _args, variable) => {
+      formatter: (value, _args, variable, fieldPath) => {
         if (variable.getValueText) {
-          return variable.getValueText();
+          return variable.getValueText(fieldPath);
```
torkelo, line 319: is there a bug report for someone using text formatter with fieldPath? 

this feels a but messy / wrong fix, the cleaner solutionn feels like it woudl be to pass in the value and text representation to the formatter, but feels a bit bad from a performance standpoint to always call getValue and getValueText

PR #1205, opened 2025-08-05
SceneRenderProfiler: Handle tab inactivity 

dprokop:
# Add tab inactivity detection to SceneRenderProfiler

## Problem

When users switch browser tabs during performance profiling, `requestAnimationFrame` callbacks get paused and can resume later when tab is activated back. This results in massive frame times (i.e. hours instead of milliseconds) being recorded and meaningless profiling data.

## Solution

**Primary Protection: Page Visibility API**
- Immediately cancels profiling when tab becomes inactive
- Uses `document.visibilitychange` event for instant response

**Fallback Protection: Frame Length Detection**
- Detects tab inactivity via frame duration > 1000ms
- Serves as backup if visibility events are missed

```typescript
// Primary: Page Visibility API
document.addEventListener('visibilitychange', () => {
  if (document.hidden && this.#profileInProgress) {
    this.cancelProfile();
  }
});

// Fallback: Frame length detection
if (frameLength > TAB_INACTIVE_THRESHOLD) {
  this.cancelProfile();
  return;
}
```

## Additional Changes

- Fixed race condition in `cancelProfile()` that caused runtime errors
- Added `cleanup()` method to prevent memory leaks
- Added defensive guards against accessing null `profileInProgress`

<details>
  <summary>📦 Published PR as canary version: <code>6.29.2--canary.1205.16748456616.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.29.2--canary.1205.16748456616.0
  npm install @grafana/scenes@6.29.2--canary.1205.16748456616.0
  # or 
  yarn add @grafana/scenes-react@6.29.2--canary.1205.16748456616.0
  yarn add @grafana/scenes@6.29.2--canary.1205.16748456616.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/behaviors/SceneRenderProfiler.ts
+++ b/packages/scenes/src/behaviors/SceneRenderProfiler.ts
@@ -3,6 +3,7 @@ import { SceneQueryControllerLike } from './types';
 
 const POST_STORM_WINDOW = 2000; // Time after last query to observe slow frames
 const SPAN_THRESHOLD = 30; // Frames longer than this will be considered slow
+const TAB_INACTIVE_THRESHOLD = 1000; // Tab inactive threshold in ms
 
 export class SceneRenderProfiler {
   #profileInProgress: {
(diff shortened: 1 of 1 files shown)


ivanortegaalba (approved):
This approach looks great to me! One thought that came to mind was: why not consider using a web worker? That way, they can keep working even when tabs are inactive. However, I then realized that `requestAnimationFrame` doesn't run in background tabs, so you won’t get real render frames while the tab is hidden.

Let's move forward with this! My only concern is that we might miss very slow dashboards, especially since people often navigate to other tabs while loading. Ideally, we could keep the network measurement waiting until the tab becomes active again, but that might add too much complexity for the value it provides. So, no worth it IMO :)

PR #1207, opened 2025-08-06
Transformations: Support series <-> annotations conversions

leeoniya:
we'd like to be able to write transformations that convert between series frames and anno frames. e.g. to use sql or loki datasources to produce Exemplars that can be rendered by TimeSeries, Heatmap, etc.

related: https://github.com/grafana/grafana/pull/109220/files

### DIFF
--- a/packages/scenes/src/querying/SceneDataTransformer.ts
+++ b/packages/scenes/src/querying/SceneDataTransformer.ts
@@ -1,4 +1,11 @@
-import { DataTopic, DataTransformerConfig, LoadingState, PanelData, transformDataFrame } from '@grafana/data';
+import {
+  DataFrame,
+  DataTopic,
+  DataTransformerConfig,
+  LoadingState,
+  PanelData,
+  transformDataFrame,
+} from '@grafana/data';
 import { toDataQueryError } from '@grafana/runtime';
 import { catchError, forkJoin, map, of, ReplaySubject, Unsubscribable } from 'rxjs';
 import { sceneGraph } from '../core/sceneGraph';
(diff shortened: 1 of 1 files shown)


dprokop (commented):
@leeoniya i've pushed some tests that I think may be surfacing problems with this line: https://github.com/grafana/scenes/pull/1207/files#diff-672934780f78622c0c2f2c605ff8848860c289c3e67c924bd1664ceef8d84ae0R223

I'm not 100% sure what the intent here is, but I suppose preserving the original series/annotations after the transformation is not something that you want?

packages/scenes/src/querying/SceneDataTransformer.test.ts
```diff
@@ -760,6 +760,341 @@ describe('SceneDataTransformer', () => {
...
+
+      // console.log(JSON.stringify({ series: data?.series, annotations: data?.annotations }))
+
+      expect({ series: data?.series, annotations: data?.annotations }).toEqual({
+        series: [],
+        annotations: [
+          {
+            fields: [
+              { name: '0', config: {}, values: [210, 410, 610], type: 'number' },
+              { name: '1', config: {}, values: [12, 14, 16], type: 'number' },
```
leeoniya, line 1098: @dprokop thanks for the tests, this one is failing but couldn't figure out why the series -> anno frames are not getting the `+10` treatment. would appreciate some help here. i'm not seeing how it would be something introduced by this PR :thinking:

packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -194,17 +195,25 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
...
+            for (const frame of frames) {
+              if (frame.meta?.dataTopic === DataTopic.Annotations) {
+                annotations.push(frame);
+              } else {
+                series.push(frame);
+              }
+            }
+          });
+
+          return { ...data, series, annotations };
```
leeoniya, line 216: @dprokop this is now copy/paste from core: https://github.com/grafana/grafana/pull/109220

Sergej-Vlasov:
@leeoniya  I think the issue lies in [here](https://github.com/grafana/scenes/blob/51add1279212be420e33df1136c52dbc3a6c8882/packages/scenes/src/querying/SceneDataTransformer.ts#L191-L195)

we do not apply annotation transformation to annotations that are transformed from series, vice versa will apply. We need to figure out how to apply transformations consecutively within one stream. I imagine we need another wrapper to account for change between series/annotations

leeoniya:
> we do not apply annotation transformation to annotations that are transformed from series

thanks for checking. i think we can skip this test for now and fix the situation later. it's an edge case that can be worked around by running the transforms on series first and doing the conversion to annos at the end.

PR #1209, opened 2025-08-07
SceneRenderProfiler: Only capture network requests within measurement window

dprokop:
# Fix network timing capture to only include requests within measurement window

## Problem
By observing data that we currently capture I've noticed the `captureNetwork` function incorrectly captured network requests that started within the profiling time window but ended after `endTs`. This led to inaccurate network timing measurements in performance profiles. This is caused because not all network requests are considered influential for the rendering time, for example see the screenshot below:

image

The last request captured is fired by faro collector and influences the `networkDuration` captured, while it should not. This results in `networkDuration > duration`.

## Solution
Updated the filtering logic in `captureNetwork` to ensure only network requests that both **start AND end** within the specified time window are captured.

### Before:
```typescript
// Only checked start time
entry.startTime >= startTs && entry.startTime <= endTs
```

### After:
```typescript
// Checks both start AND end times
entry.startTime >= startTs && 
entry.startTime <= endTs && 
entry.responseEnd >= startTs && 
entry.responseEnd <= endTs
```

## Changes
- Fixed timing filter logic in `captureNetwork` function
- Added test coverage
- Added Performance API mocking to Jest test environment setup
- Exported `captureNetwork` function for testability

## Test Coverage
Added tests for:
- Valid requests within time window
- Requests starting within but ending after window (excluded)
- Requests starting before but ending within window (excluded)
- Requests completely outside window (excluded)
- Multiple overlapping requests
- Boundary conditions
- Zero-duration requests
- Empty scenarios

### DIFF
--- a/packages/scenes/src/behaviors/SceneRenderProfiler.test.ts
+++ b/packages/scenes/src/behaviors/SceneRenderProfiler.test.ts
@@ -1,4 +1,4 @@
-import { calculateNetworkTime, processRecordedSpans } from './SceneRenderProfiler';
+import { calculateNetworkTime, processRecordedSpans, captureNetwork } from './SceneRenderProfiler';
 
 describe('calculateNetworkTime', () => {
   it('should return the duration of a single request', () => {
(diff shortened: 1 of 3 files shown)


ivanortegaalba (approved):
Good catch! 💯

mdvictor (approved):
lgtm!

PR #1210, opened 2025-08-07
Add new operators to AdHocFiltersVariable

joey-grafana:
Adds new operators to the AdHocFiltersVariable.

This is the result of an internal Slack chat where it was discussed that having `>=` and `<=` would be useful for example when searching for HTTP status codes.

Screenshot 2025-08-07 at 08 36 49


ifrost:
btw. Status code is always a number or it could be a string? 🤔

PR #1227, opened 2025-08-26
SceneGridRow: Adjust row panel count to account for repeats

Sergej-Vlasov:
adjusting row panel count to account for the new `repeatedPanels` property

<details>
  <summary>📦 Published PR as canary version: <code>6.31.1--canary.1227.17262127065.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.31.1--canary.1227.17262127065.0
  npm install @grafana/scenes@6.31.1--canary.1227.17262127065.0
  # or 
  yarn add @grafana/scenes-react@6.31.1--canary.1227.17262127065.0
  yarn add @grafana/scenes@6.31.1--canary.1227.17262127065.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/layout/grid/types.ts
+++ b/packages/scenes/src/components/layout/grid/types.ts
@@ -13,7 +13,6 @@ export interface SceneGridItemPlacement {
 export interface SceneGridItemStateLike extends SceneGridItemPlacement, SceneObjectState {
   isResizable?: boolean;
   isDraggable?: boolean;
-  repeatedPanels?: VizPanel[];
 }
 
 export interface SceneGridItemLike extends SceneObject<SceneGridItemStateLike> {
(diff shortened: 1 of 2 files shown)


packages/scenes/src/components/layout/grid/types.ts
```diff
@@ -13,6 +13,7 @@ export interface SceneGridItemPlacement {
 export interface SceneGridItemStateLike extends SceneGridItemPlacement, SceneObjectState {
   isResizable?: boolean;
   isDraggable?: boolean;
+  repeatedPanels?: VizPanel[];
```
oscarkilhed: Rather than leaking this implementation into scenes, can't we include a `getChildrenCount` on the `SceneGridItemLike` or `SceneObject` interface.

In Grafana `DashboardGridItem` already has this function (that we can rename). https://github.com/grafana/grafana/blob/main/public/app/features/dashboard-scene/scene/layout-default/DashboardGridItem.tsx#L99

oscarkilhed: The change in scenes would then be something like this:

https://github.com/grafana/scenes/commit/f84c5426cd60a6aa1aa399563d5f4da087420398

And then we implement the `getChildCount` on `DashboardGridItem` in grafana.

Sergej-Vlasov: Your approach looks cleaner will adjust!

PR #1231, opened 2025-09-01
SceneGraph: Filter out dashboard controls from `getVariables()`

leventebalogh:
**Related PR:** https://github.com/grafana/grafana/pull/109225

### What changed?
As discussed in https://github.com/grafana/grafana/pull/109225#discussion_r2306953250 we wanted to move the functionality of filtering out variables that are meant to be rendered under a dashboard-controls dropdown menu to scenes. This PR also adds another function for getting the "dashboard control" variables. 

<details>
  <summary>📦 Published PR as canary version: <code>6.32.1--canary.1231.17376394106.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.32.1--canary.1231.17376394106.0
  npm install @grafana/scenes@6.32.1--canary.1231.17376394106.0
  # or 
  yarn add @grafana/scenes-react@6.32.1--canary.1231.17376394106.0
  yarn add @grafana/scenes@6.32.1--canary.1231.17376394106.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/core/sceneGraph/index.ts
+++ b/packages/scenes/src/core/sceneGraph/index.ts
@@ -15,10 +15,12 @@ import {
   getAncestor,
   findDescendents,
   getScopes,
+  getDashboardControls,
 } from './sceneGraph';
 
 export const sceneGraph = {
   getVariables,
+  getDashboardControls,
   getData,
   getTimeRange,
   getLayout,
--- a/packages/scenes/src/core/sceneGraph/sceneGraph.ts
+++ b/packages/scenes/src/core/sceneGraph/sceneGraph.ts
@@ -2,7 +2,7 @@ import { Scope, ScopedVars } from '@grafana/data';
 import { EmptyDataNode, EmptyVariableSet } from '../../variables/interpolation/defaults';
 
 import { sceneInterpolator } from '../../variables/interpolation/sceneInterpolator';
-import { VariableCustomFormatterFn, SceneVariables } from '../../variables/types';
+import { VariableCustomFormatterFn, SceneVariables, SceneVariable } from '../../variables/types';
 
 import { isDataLayer, SceneDataLayerProvider, SceneDataProvider, SceneLayout, SceneObject } from '../types';
 import { lookupVariable } from '../../variables/lookupVariable';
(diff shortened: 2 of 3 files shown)


packages/scenes/src/core/sceneGraph/sceneGraph.ts
```diff
@@ -17,9 +17,25 @@ import { SCOPES_VARIABLE_NAME } from '../../variables/constants';
...
+  return filterVariables(variables, (v) => v.state.showInControlsMenu !== true);
+}
+
+/**
+ * Get the closest node with controls
+ */
+export function getDashboardControls(sceneObject: SceneObject): SceneVariables {
+  const variables = getClosest(sceneObject, (s) => s.state.$variables) ?? EmptyVariableSet;
+
+  return filterVariables(variables, (v) => v.state.showInControlsMenu === true);
```
torkelo, line 31: Could we maybe extend the show/hide label/variable option we already have to cover this new display option rather than a new Boolean option?

packages/scenes/src/core/sceneGraph/index.ts
```diff
@@ -15,10 +15,12 @@ import {
   getAncestor,
   findDescendents,
   getScopes,
+  getDashboardControls,
 } from './sceneGraph';
 
 export const sceneGraph = {
   getVariables,
+  getDashboardControls,
```
torkelo, line 23: Do we really need this? 

Feels like this is just a new display mode on variables, think all who call getVariables should get all

packages/scenes/src/core/sceneGraph/sceneGraph.ts
leventebalogh: Yea, that's a possible way to tackle this problem indeed, and might be better than what we have in place now. Let's discuss it offline (Slack) and I will come back here with a summary.

packages/scenes/src/core/sceneGraph/index.ts
leventebalogh: If were to follow the way it is currently (having a separate `showInControlsMenu` prop), and [if we would move the filtering to scenes](https://github.com/grafana/grafana/pull/109225#discussion_r2306953250), then we would need it to have access to these dashboard controls.

If we decide that `getVariables()` just returns all the variables - which is totally fine with me - then it's not necessary 👍

leventebalogh:
Closing this as we discussed another way forward offline.

PR #1232, opened 2025-09-01
Pass both filters and groupByKeys in annotation requests

mdvictor:
We have a need to pass filters and groupBy keys in the annotation requests as well, similar to how we pass them for queries.

This PR allows annotation requests to also contain filters and groupBy keys that update on var state changes, similar to how they work for scopes.

To achieve this I've refactored and reused code from the SceneQueryRunner, creating a new DrilldownDependenciesManager

<details>
  <summary>📦 Published PR as canary version: <code>6.33.0--canary.1232.17398095824.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.33.0--canary.1232.17398095824.0
  npm install @grafana/scenes@6.33.0--canary.1232.17398095824.0
  # or 
  yarn add @grafana/scenes-react@6.33.0--canary.1232.17398095824.0
  yarn add @grafana/scenes@6.33.0--canary.1232.17398095824.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.ts
@@ -128,7 +128,7 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
     dependsOnScopes: true,
   });
 
-  private _drilldownDependenciesManager?: DrilldownDependenciesManager<QueryRunnerState> =
+  private _drilldownDependenciesManager: DrilldownDependenciesManager<QueryRunnerState> =
     new DrilldownDependenciesManager(this._variableDependency);
 
   public constructor(initialState: QueryRunnerState) {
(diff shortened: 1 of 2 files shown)


packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -132,6 +128,9 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
     dependsOnScopes: true,
   });
 
+  private _drilldownDependenciesManager?: DrilldownDependenciesManager<QueryRunnerState> =
```
torkelo: why the optional ? typing, it is always defined / created here. (and could then remove the ? when used)

mdvictor: Yes that is pointless there, I've removed it there and in related code, thank you!

PR #1236, opened 2025-09-04
Variables: exploration to support object-based values with multiple properties

grafakus:
Follow-up of https://github.com/grafana/scenes/pull/1228

### DIFF
--- a/packages/scenes/src/variables/types.ts
+++ b/packages/scenes/src/variables/types.ts
@@ -76,6 +76,7 @@ export interface VariableValueOption {
   label: string;
   value: VariableValueSingle;
   group?: string;
+  properties?: Record<string, any>;
 }
 
 export interface SceneVariableSetState extends SceneObjectState {
--- a/packages/scenes/src/variables/variants/CustomVariable.test.ts
+++ b/packages/scenes/src/variables/variants/CustomVariable.test.ts
@@ -280,4 +280,45 @@ label-3 : value-3,`,
       expect(B.state.options[2].value).toBe('value1');
     });
   });
+
+  describe('multi prop / object support', () => {
+    it('Can have object values (JSON array of objects)', async () => {
+      const variable = new CustomVariable({
+        name: 'test',
+        isMulti: false,
+        isJson: true,
+        query: `
+[
+  { "label": "Test", "value": "test", "properties": { "id": "test", "display": "Test", "location": "US" } },
+  { "label": "Prod", "value": "prod", "properties": { "id": "prod", "display": "Prod", "location": "EU" } }
+]
+        `,
+        value: 'prod',
+        text: 'Prod',
+        options: [],
+      });
+
+      await lastValueFrom(variable.validateAndUpdate());
+
+      expect(variable.getValue()).toEqual('prod');
+      expect(variable.getValue('location')).toEqual('EU');
+    });
+
+    it('Can have object values (JSON array of strings)', async () => {
+      const variable = new CustomVariable({
+        name: 'test',
+        isMulti: false,
+        isJson: true,
+        query: `["test", "prod"]`,
+        value: 'prod',
+        text: 'prod',
+        options: [],
+      });
+
+      await lastValueFrom(variable.validateAndUpdate());
+
+      expect(variable.state.value).toBe('prod');
+      expect(variable.state.text).toBe('prod');
+    });
+  });
 });
--- a/packages/scenes/src/variables/variants/CustomVariable.tsx
+++ b/packages/scenes/src/variables/variants/CustomVariable.tsx
@@ -11,6 +11,7 @@ import React from 'react';
 
 export interface CustomVariableState extends MultiValueVariableState {
   query: string;
+  isJson: boolean;
 }
 
 export class CustomVariable extends MultiValueVariable<CustomVariableState> {
@@ -26,24 +27,48 @@ export class CustomVariable extends MultiValueVariable<CustomVariableState> {
       text: '',
       options: [],
       name: '',
+      isJson: false,
       ...initialState,
     });
   }
 
   public getValueOptions(args: VariableGetOptionsArgs): Observable<VariableValueOption[]> {
     const interpolated = sceneGraph.interpolate(this, this.state.query);
-    const match = interpolated.match(/(?:\\,|[^,])+/g) ?? [];
 
-    const options = match.map((text) => {
-      text = text.replace(/\\,/g, ',');
-      const textMatch = /^\s*(.+)\s:\s(.+)$/g.exec(text) ?? [];
-      if (textMatch.length === 3) {
-        const [, key, value] = textMatch;
-        return { label: key.trim(), value: value.trim() };
-      } else {
-        return { label: text.trim(), value: text.trim() };
+    let options: VariableValueOption[] = [];
+
+    if (this.state.isJson) {
+      try {
+        const parsedOptions = JSON.parse(interpolated);
+
+        if (!Array.isArray(parsedOptions)) {
+          throw new Error('Query must be a JSON array');
+        }
+
+        if (typeof parsedOptions[0] === 'string') {
+          options = parsedOptions.map((value) => ({ label: value.trim(), value: value.trim() }));
+        } else if (typeof parsedOptions[0] === 'object' && parsedOptions[0] !== null) {
+          options = parsedOptions;
+        } else {
+          throw new Error('Query must be a JSON array of strings or objects');
+        }
+      } catch (error) {
+        throw error;
       }
-    });
+    } else {
+      const match = interpolated.match(/(?:\\,|[^,])+/g) ?? [];
+
+      options = match.map((text) => {
+        text = text.replace(/\\,/g, ',');
+        const textMatch = /^\s*(.+)\s:\s(.+)$/g.exec(text) ?? [];
+        if (textMatch.length === 3) {
+          const [, key, value] = textMatch;
+          return { label: key.trim(), value: value.trim() };
+        } else {
+          return { label: text.trim(), value: text.trim() };
+        }
+      });
+    }
 
     if (!options.length) {
       this.skipNextValidation = true;
--- a/packages/scenes/src/variables/variants/MultiValueVariable.test.ts
+++ b/packages/scenes/src/variables/variants/MultiValueVariable.test.ts
@@ -758,7 +758,7 @@ describe('MultiValueVariable', () => {
         skipUrlSync: true,
         optionsToReturn: [
           { label: 'Test', value: 'test', properties: { id: 'test', display: 'Test', location: 'US' } },
-          { label: 'Prod', value: 'pod', properties: { id: 'prod', display: 'Prod', location: 'EU' } },
+          { label: 'Prod', value: 'prod', properties: { id: 'prod', display: 'Prod', location: 'EU' } },
         ],
       });
 
--- a/packages/scenes/src/variables/variants/query/QueryVariable.test.tsx
+++ b/packages/scenes/src/variables/variants/query/QueryVariable.test.tsx
@@ -189,6 +189,51 @@ describe.each(['11.1.2', '11.1.1'])('QueryVariable', (v) => {
       expect(variable.state.loading).toEqual(true);
     });
 
+    describe('multi prop / object support', () => {
+      it('Can have object values', async () => {
+        setCreateQueryVariableRunnerFactory(
+          () =>
+            new FakeQueryRunner(
+              fakeDsMock,
+              jest.fn().mockReturnValue(
+                of<PanelData>({
+                  state: LoadingState.Done,
+                  series: [
+                    toDataFrame({
+                      fields: [
+                        { name: 'text', type: FieldType.string, values: ['Test', 'Prod'] },
+                        { name: 'value', type: FieldType.string, values: ['test', 'prod'] },
+                        {
+                          name: 'properties',
+                          type: FieldType.other,
+                          values: [
+                            { id: 'test', display: 'Test', location: 'US' },
+                            { id: 'prod', display: 'Prod', location: 'EU' },
+                          ],
+                        },
+                      ],
+                    }),
+                  ],
+                  timeRange: getDefaultTimeRange(),
+                })
+              )
+            )
+        );
+
+        const variable = new QueryVariable({
+          name: 'test',
+          datasource: { uid: 'fake-std', type: 'fake-std' },
+          query: 'query',
+          isMultiDimensions: true,
+        });
+
+        await lastValueFrom(variable.validateAndUpdate());
+
+        expect(variable.getValue('location')).toEqual('US');
+        expect(variable.getValue()).toEqual('test');
+      });
+    });
+
     it('Should pass variable scene object when resolving data source and via request scoped vars', async () => {
       const variable = new QueryVariable({
         name: 'test',
--- a/packages/scenes/src/variables/variants/query/QueryVariable.tsx
+++ b/packages/scenes/src/variables/variants/query/QueryVariable.tsx
@@ -34,6 +34,7 @@ export interface QueryVariableState extends MultiValueVariableState {
   type: 'query';
   datasource: DataSourceRef | null;
   query: string | SceneDataQuery;
+  isMultiDimensions: boolean;
   regex: string;
   refresh: VariableRefresh;
   sort: VariableSort;
@@ -62,6 +63,7 @@ export class QueryVariable extends MultiValueVariable<QueryVariableState> {
       datasource: null,
       regex: '',
       query: '',
+      isMultiDimensions: false,
       refresh: VariableRefresh.onDashboardLoad,
       sort: VariableSort.disabled,
       ...initialState,
--- a/packages/scenes/src/variables/variants/query/toMetricFindValues.test.ts
+++ b/packages/scenes/src/variables/variants/query/toMetricFindValues.test.ts
@@ -24,6 +24,20 @@ describe('toMetricFindValues', () => {
       { name: 'expandable', type: FieldType.boolean, values: [true, false, true] },
     ],
   });
+  const frameWithPropertiesField = toDataFrame({
+    fields: [
+      { name: 'label', type: FieldType.string, values: ['A', 'B', 'C'] },
+      {
+        name: 'properties',
+        type: FieldType.other,
+        values: [
+          { value: 'A', displayvalue: 'Alpha' },
+          { value: 'B', displayvalue: 'Beta' },
+          { value: 'C', displayvalue: 'Gamma' },
+        ],
+      },
+    ],
+  });
 
   // it.each wouldn't work here as we need the done callback
   [
@@ -70,6 +84,14 @@ describe('toMetricFindValues', () => {
         { text: 'A', value: 'A', expandable: true },
         { text: 'B', value: 'B', expandable: false },
         { text: 'C', value: 'C', expandable: true },
+      ],
+    },
+    {
+      series: [frameWithPropertiesField],
+      expected: [
+        { text: 'A', value: 'A', properties: { displayvalue: 'Alpha', value: 'A' } },
+        { text: 'B', value: 'B', properties: { displayvalue: 'Beta', value: 'B' } },
+        { text: 'C', value: 'C', properties: { displayvalue: 'Gamma', value: 'C' } },
       ],
     },
   ].map((scenario) => {
--- a/packages/scenes/src/variables/variants/query/toMetricFindValues.ts
+++ b/packages/scenes/src/variables/variants/query/toMetricFindValues.ts
@@ -8,7 +8,11 @@ import {
 } from '@grafana/data';
 import { map, OperatorFunction } from 'rxjs';
 
-export function toMetricFindValues(): OperatorFunction<PanelData, MetricFindValue[]> {
+interface MetricFindValueWithProperties extends MetricFindValue {
+  properties?: Record<string, any>;
+}
+
+export function toMetricFindValues(): OperatorFunction<PanelData, MetricFindValueWithProperties[]> {
   return (source) =>
     source.pipe(
       map((panelData) => {
@@ -26,12 +30,13 @@ export function toMetricFindValues(): OperatorFunction<PanelData, MetricFindValu
         }
 
         const processedDataFrames = getProcessedDataFrames(frames);
-        const metrics: MetricFindValue[] = [];
+        const metrics: MetricFindValueWithProperties[] = [];
 
         let valueIndex = -1;
         let textIndex = -1;
         let stringIndex = -1;
         let expandableIndex = -1;
+        let propertiesIndex = -1;
 
         for (const frame of processedDataFrames) {
           for (let index = 0; index < frame.fields.length; index++) {
@@ -40,6 +45,10 @@ export function toMetricFindValues(): OperatorFunction<PanelData, MetricFindValu
 
             if (field.type === FieldType.string && stringIndex === -1) {
               stringIndex = index;
+            }
+
+            if (field.type === FieldType.other && propertiesIndex === -1) {
+              propertiesIndex = index;
             }
 
             if (fieldName === 'text' && field.type === FieldType.string && textIndex === -1) {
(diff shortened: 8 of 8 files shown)


packages/scenes/src/variables/variants/query/QueryVariable.test.tsx
```diff
@@ -189,6 +189,51 @@ describe.each(['11.1.2', '11.1.1'])('QueryVariable', (v) => {
...
+        setCreateQueryVariableRunnerFactory(
+          () =>
+            new FakeQueryRunner(
+              fakeDsMock,
+              jest.fn().mockReturnValue(
+                of<PanelData>({
+                  state: LoadingState.Done,
+                  series: [
+                    toDataFrame({
+                      fields: [
```
grafakus: Not sure about the structure, since there's some redundancy.

packages/scenes/src/variables/variants/CustomVariable.test.ts
```diff
@@ -280,4 +280,45 @@ label-3 : value-3,`,
...
+    it('Can have object values (JSON array of objects)', async () => {
+      const variable = new CustomVariable({
+        name: 'test',
+        isMulti: false,
+        isJson: true,
+        query: `
+[
+  { "label": "Test", "value": "test", "properties": { "id": "test", "display": "Test", "location": "US" } },
+  { "label": "Prod", "value": "prod", "properties": { "id": "prod", "display": "Prod", "location": "EU" } }
+]
```
torkelo, line 365: dont think this form should be required, better to just support plain array of objects, and then options for what property is the value & label (label being optional)

packages/scenes/src/variables/variants/query/QueryVariable.test.tsx
torkelo: should not require any structure, we just need options for value and id field name

grafakus: I've made the changes in https://github.com/grafana/scenes/pull/1236/commits/da04a3b86529e146d6852209dbaa9fe263f92fd4 :

- Adding support for `valueProp` and `textProp` in `MultiValueVariable`
- Making sure that `valueProperties` is always updated when new options are received
- Then updating `CustomVariable` and `QueryVariable` to use `valueProp` and `textProp`

I haven't looked yet into `MultiValueVariable.changeValueTo` but the code start to feel a bit messy IMO...

While writing the [design doc](https://docs.google.com/document/d/1zY20vaJqItK_78LTraAVQGjFHqrU3syFQLiALoStxFM/edit?tab=t.0#heading=h.5sybau7waq2q) (very much in draft), I've been thinking that two ideas could help simplifying variables management:
1. Options providers (introduced in your [1st poc](https://github.com/grafana/scenes/pull/995)), to unify Custom and Query variables and provide a nice extension point for external consumers of the library to register their own custom providers without having to support them in Scenes
2. Not 100% clear in my mind: why not creating an abstraction for the current value instead of having to deal with `value`, `text` and `valueProperties` all around?

packages/scenes/src/variables/types.ts
```diff
@@ -72,10 +72,12 @@ export interface CustomVariableValue {
 }
 
 export interface ValidateAndUpdateResult {}
+export interface VariableValueOptionProperties extends Record<string, any> {}
 export interface VariableValueOption {
   label: string;
   value: VariableValueSingle;
   group?: string;
+  properties?: VariableValueOptionProperties | VariableValueOptionProperties[];
```
torkelo: dont think we need to support an array here

packages/scenes/src/variables/variants/MultiValueVariable.ts
```diff
@@ -171,13 +179,23 @@ export abstract class MultiValueVariable<TState extends MultiValueVariableState
         }
       }
 
+      stateUpdate.valueProperties = (stateUpdate.value as VariableValueSingle[]).map(
+        (v) => options.find((o) => o.value === v)?.properties
+      );
+
```
torkelo: not sure I understand this

packages/scenes/src/variables/variants/CustomVariable.test.ts
```diff
@@ -280,4 +280,46 @@ label-3 : value-3,`,
...
+        query: `
+[
+  { "id": "test", "display": "Test", "location": "US" },
+  { "id": "prod", "display": "Prod", "location": "EU" }
+]
+        `,
+        valueProp: 'id',
+        textProp: 'display',
+        value: 'prod',
+        text: 'Prod',
```
torkelo, line 368: think we should probably introduce different providers here like in

https://github.com/grafana/scenes/pull/995 

so that we can have a CSV provider, object provider, json provider, then the options can be on each provider and we do not need "isJson" on the top level etc.

packages/scenes/src/variables/variants/query/toMetricFindValues.test.ts
```diff
@@ -24,6 +24,19 @@ describe('toMetricFindValues', () => {
...
+        name: 'properties',
+        type: FieldType.other,
+        values: [
+          { value: 'A', displayValue: 'Alpha' },
+          { value: 'B', displayValue: 'Beta' },
+          { value: 'C', displayValue: 'Gamma' },
+        ],
+      },
+    ],
+  });
```
torkelo: properties is not a field, it's all fields. 

A data frame retured by a query variable should be able to have many fields (1 used for value, 1 optionally used for label, the others are the "extra property" fields.

torkelo: think of SQL query like 

`SELECT id, displayName, country, city` 

would result in a DataFrame with fields id, displayName, country and city. id, displayName, country and city should all end up in the "properties" object in the MetricFindValueWithProperties  type

packages/scenes/src/variables/types.ts
grafakus: You're right
```suggestion
```

packages/scenes/src/variables/variants/MultiValueVariable.ts
grafakus: To keep `valueProperties` in sync with `value` so `getValue` can use it directly. Myabe it's easier if `getValue` uses the current options instead

packages/scenes/src/variables/variants/query/toMetricFindValues.test.ts
grafakus: Got it, it makes sense (I haven't worked so much with data frames so far)

packages/scenes/src/variables/variants/MultiValueVariable.ts
torkelo: I just dont understand what this is trying to solve, and it should crash for scenarios when stateUpdate.value is not an array?

grafakus: 🤦🏾‍♂️ Ok, I see now. Fixed in https://github.com/grafana/scenes/pull/1236/commits/7a39e48cd00fd2d3370088ac05c8962e92050a88

packages/scenes/src/variables/variants/CustomVariable.test.ts
grafakus: First step (CSV and JSON): https://github.com/grafana/scenes/pull/1236/commits/b83b4655b4cc6b95e9727271fa8f3e2138ce7465

packages/scenes/src/variables/variants/query/toMetricFindValues.test.ts
grafakus: Implemented in https://github.com/grafana/scenes/pull/1236/commits/20f655bc790801573e5c25596821211ef5c4ba00

packages/scenes/src/variables/variants/CustomVariable.tsx
```diff
@@ -1,16 +1,17 @@
...
 
-import { MultiValueVariable, MultiValueVariableState, VariableGetOptionsArgs } from './MultiValueVariable';
-import { sceneGraph } from '../../core/sceneGraph';
 import React from 'react';
+import { buildOptionsProvider, OptionsProviderSettings } from './CustomOptionsProviders';
+import { MultiValueVariable, MultiValueVariableState, VariableGetOptionsArgs } from './MultiValueVariable';
 
 export interface CustomVariableState extends MultiValueVariableState {
   query: string;
+  optionsProvider: OptionsProviderSettings;
```
torkelo: not sure I like this, think it would be better if this was a simple interface.

 something like

```ts
export interface VariableValueProvider {
  getValues(arg: VariableGetOptionsArgs): Observable<MetricFindValueWithOptionalProperties>
}
```

Hm.. Interested in trying to do this in a new SelectVariable (and leave CustomVariable and QueryVariable as is for now). just to get a sense of how that would work/look. 

The query string state property of CustomVariable should not really exist, it's an option that should be part of provider object/class (as it's type and value depends on the provider).

packages/scenes/src/variables/variants/MultiValueVariable.ts
```diff
@@ -177,24 +179,60 @@ export abstract class MultiValueVariable<TState extends MultiValueVariableState
...
+      }
+
+      // No early return here as want to continue to look a value match
+      if (o.label === currentText) {
+        textMatch = o;
+      }
+    }
+
+    return textMatch;
+  }
```
torkelo: dont undestand why this new function is needed vs the old, options has a value & label prop, we can use them.  

And options are returned by the value provider, the is nothing in MultiValueVariable that needs to know abvout textProp and valueProp

packages/scenes/src/variables/variants/MultiValueVariable.ts
```diff
@@ -217,33 +255,46 @@ export abstract class MultiValueVariable<TState extends MultiValueVariableState
...
+      if (allValue) {
+        return new CustomAllValue(allValue, this);
       }
-
-      value = this.state.options.map((x) => x.value);
+      if (optionsProvider?.valueProp) {
+        return options.map((o) =>
+          this.getValueFromValueProperties(o.properties, fieldPath || optionsProvider.valueProp!)
+        );
+      }
```
torkelo: dont think this logic is needed, if we have a fieldPath we use it otherwise we use o.value

packages/scenes/src/variables/variants/CustomVariable.tsx
torkelo: did a quick test here: https://github.com/grafana/scenes/commit/1449343886fd1a9af4669ff4d04980f003e0b3a5 

hm.. the problem with that approach is that it becomes a bit of a pain to update the state of the value providers (have to create a new one), would have to make the value providers scene objects to solve that, which feels a bit overkill / hoping to avoid that :(

grafakus: We can inject an instance of the provider, that would be neat. My thought here is to try keeping the changes minimal from a consumer perspective and not to break the existing contract (like passing `query`):

```ts
new CustomVariable({ query: '[]', optionsProvider: { type: 'json' } });
```

vs

```ts
new CustomVariable({ optionsProvider: { getOptions(arg) { ... } } });
```

> Interested in trying to do this in a new SelectVariable (and leave CustomVariable and QueryVariable as is for now). just to get a sense of how that would work/look.

👍🏾👍🏾 It would give us a good idea of how [proposal 3 of the design doc](https://docs.google.com/document/d/1zY20vaJqItK_78LTraAVQGjFHqrU3syFQLiALoStxFM/edit?tab=t.0#heading=h.5sybau7waq2q) would look like.

packages/scenes/src/variables/variants/MultiValueVariable.ts
grafakus: It's not needed indeed

packages/scenes/src/variables/variants/MultiValueVariable.ts
grafakus: indeed

packages/scenes/src/variables/variants/query/QueryVariable.tsx
```diff
@@ -65,6 +51,7 @@ export class QueryVariable extends MultiValueVariable<QueryVariableState> {
       refresh: VariableRefresh.onDashboardLoad,
       sort: VariableSort.disabled,
       ...initialState,
+      optionsProvider: { ...initialState.optionsProvider, type: 'query' },
```
torkelo: This feels unnecessary here, as it's inside the QueryVariable, we know the provider

packages/scenes/src/variables/variants/query/QueryVariable.tsx
```diff
@@ -75,86 +62,27 @@ export class QueryVariable extends MultiValueVariable<QueryVariableState> {
...
-              options = options.filter((option) => !customOptions.find((custom) => custom.value === option.value));
-              if (this.state.staticOptionsOrder === 'after') {
-                options.push(...customOptions);
-              } else if (this.state.staticOptionsOrder === 'sorted') {
-                options = sortVariableValues(options.concat(customOptions), this.state.sort);
-              } else {
-                options.unshift(...customOptions);
-              }
+    return new Observable((subscriber) => {
+      buildOptionsProvider(this as unknown as MultiValueVariable)
```
torkelo: this looks less then ideal

packages/scenes/src/variables/variants/query/toMetricFindValues.ts
```diff
@@ -25,68 +33,59 @@ export function toMetricFindValues(): OperatorFunction<PanelData, MetricFindValu
           return [];
         }
 
-        const processedDataFrames = getProcessedDataFrames(frames);
```
torkelo, line 28: Both elastic and postgres use the metricFindQuery so already return MetricFindValue, so this function will return already on line 29 (unless we update those data sources to handle variables differently)

packages/scenes/src/variables/variants/CustomVariable.tsx
torkelo: If we are to make a change to CustomVariable and still use the `query` property for the csv/json content, I think we should try to make the smallest / simplest thing, just adding a new option. 

```ts
valuesFormat: 'csv'  | 'json' 
```  
  
I think we can wait with textProp and valueProp for now (trying to simplify, remove all options that are not 100% needed). we can use same heurstic as toMetricFindValues  (property named value and text, if not use first string for value, rest goes to properties)

packages/scenes/src/variables/variants/MultiValueVariable.test.ts
```diff
@@ -749,22 +749,75 @@ describe('MultiValueVariable', () => {
...
+          delayMs: 0,
+          skipUrlSync: true,
+          value: ['prod', 'stag'],
+          text: 'Prod + Staging',
+          optionsToReturn: [
+            { label: 'Test', value: 'test', properties: { id: 'test', display: 'Test', location: 'US' } },
+            { label: 'Stag', value: 'stag', properties: { id: 'stag', display: 'Stag', location: 'SG' } },
+            { label: 'Prod', value: 'prod', properties: { id: 'prod', display: 'Prod', location: 'EU' } },
+          ],
+          optionsProvider: { type: 'json', valueProp: 'id', textProp: 'display' },
```
torkelo: this is not needed, the optionsToReturn have everything

packages/scenes/src/variables/variants/query/QueryVariable.tsx
grafakus: It was added it to be able to pass custom `valueProp` and `textProp`

packages/scenes/src/variables/variants/query/QueryVariable.tsx
grafakus: On the pros side, it's quite simple to rely on the factory function ; on the cons side, a new instance has to be built every time.

To mitigate the cons, could we rely on memoization?

packages/scenes/src/variables/variants/CustomVariable.tsx
grafakus: About the heuristic to find values and labels, wouldn't it be better not to assume we will be able to rely on them (or at that it'll be easy to rely on them) for all the datasources that will start getting support (Elastic and Postgres) and for all the scenarios from a user perspective?

packages/scenes/src/variables/variants/query/toMetricFindValues.ts
grafakus: I think it was mentioned when we reached the consensus on the design doc. Besides pinging the owners of the Elastic and Postgres datasources, is the change in this PR enough?

packages/scenes/src/variables/variants/CustomOptionsProviders.ts
```diff
@@ -0,0 +1,261 @@
...
+import { getDataSource } from '../../utils/getDataSource';
+import { wrapInSafeSerializableSceneObject } from '../../utils/wrapInSafeSerializableSceneObject';
+import { createQueryVariableRunner } from './query/createQueryVariableRunner';
+import { toMetricFindValues } from './query/toMetricFindValues';
+import { metricNamesToVariableValues, sortVariableValues } from './query/utils';
+import { VariableGetOptionsArgs } from './MultiValueVariable';
+
+type BuilderFunction = (variable: CustomVariable | QueryVariable) => CustomOptionsProvider;
+
+const OPTIONS_PROVIDERS_REGISTRY = new Map<string, BuilderFunction>([
```
dprokop: Do we anticipate more formats introduction? I don't see a justification for having this complexity at this stage. Eventually when we decide to unify custom/query variable under List variable, this will become unnecessary.

packages/scenes/src/variables/variants/CustomVariable.tsx
dprokop: +1 on @torkelo comment of having minimal interface change here, `type: 'csv' | 'json'`, or `valuesFormat: 'csv' | 'json'`. With custom variable it's really `valuesFormat?:  'json'` as CSV is the default already.

grafakus:
> In my opinion this PR introduces a lot of unjustified infrastructure(providers, runtime registration, factories) for really solving the problem of supporting json values. The provider approach would be nice if we were to scale the format options, but I cannot see introducing any new formats any time soon - _maybe_ yaml would make sense, but I doubt truly.

Hi @dprokop @torkelo , resuming work after a long PTO... Thank you for all the comments. 
With fresh eyes, it feels indeed that a simpler approach would be better, I'll make the changes.

packages/scenes/src/variables/variants/query/toMetricFindValues.ts
grafakus: Well actually maybe a better question is, should we remove the check L29 / should each data source handle the frames with extra properties?

packages/scenes/src/locales/cs-CZ/grafana-scenes.json
```diff
@@ -158,4 +158,4 @@
       }
     }
   }
-}
```
grafakus, line 161: Not sure how this happens after merging changes from. main 🤷🏾‍♂️

packages/scenes/src/variables/variants/query/toMetricFindValues.test.ts
```diff
@@ -3,100 +3,246 @@ import { of } from 'rxjs';
...
+    const frameWithPropertiesAndExpandableField = toDataFrame({
+      fields: [
+        ...frameWithPropertiesField.fields,
+        { name: 'expandable', type: FieldType.boolean, values: [true, false, true] },
+      ],
+    });
+
+    [
+      {
+        series: [frameWithPropertiesField],
```
dprokop, line 144: Nit: i'd rather like to have series defined here inline, otherwise it's tricky to follow what the actual input is. And the test while cleverly structured, it's hard to read.

packages/scenes/src/variables/variants/query/toMetricFindValues.ts
dprokop, line 1: Think there are some opportunities to simplify this file after recent changes, see the attached diff:

```ts
import {
  DataFrame,
  FieldType,
  getFieldDisplayName,
  getProcessedDataFrames,
  isDataFrame,
  MetricFindValue,
  PanelData,
} from '@grafana/data';
import { map, OperatorFunction } from 'rxjs';

interface MetricFindValueWithOptionalProperties extends MetricFindValue {
  properties?: Record<string, any>;
}

export function toMetricFindValues(
  valueProp?: string,
  textProp?: string
): OperatorFunction<PanelData, MetricFindValueWithOptionalProperties[]> {
  return (source) =>
    source.pipe(
      map((panelData) => {
        const frames = panelData.series;
        if (!frames?.length) {
          return [];
        }

        if (areMetricFindValues(frames)) {
          return frames;
        }

        if (frames[0].fields.length === 0) {
          return [];
        }

        const indices = findFieldsIndices(frames);
        validateIndices(indices, valueProp, textProp);

        const metrics: MetricFindValueWithOptionalProperties[] = [];

        for (const frame of frames) {
          for (let index = 0; index < frame.length; index++) {
            const fieldValue = (fieldIndex: number) =>
              fieldIndex !== -1 ? frame.fields[fieldIndex].values.get(index) : undefined;

            const value = fieldValue(indices.value);
            const text = fieldValue(indices.text);
            const expandable = fieldValue(indices.expandable);

            const properties = indices.properties.length
              ? indices.properties.reduce((acc, p) => {
                  acc[p.name] = frame.fields[p.index].values.get(index);
                  return acc;
                }, {} as Record<string, string>)
              : undefined;

            metrics.push({
              value: getValueOrFallback(value, text, valueProp, textProp, properties),
              text: getValueOrFallback(text, value, textProp, valueProp, properties),
              ...(properties && { properties }),
              ...(expandable !== undefined && { expandable }),
            });
          }
        }

        return metrics;
      })
    );
}

function validateIndices(indices: Indices, valueProp?: string, textProp?: string): void {
  const hasNoValueOrText = indices.value === -1 && indices.text === -1;

  if (hasNoValueOrText && !indices.properties.length) {
    throw new Error("Couldn't find any field of type string in the results");
  }

  // A single field of type string that is neither named "value" nor "text" is considered as "value"
  if (hasNoValueOrText && indices.properties.length === 1) {
    indices.value = indices.properties[0].index;
    indices.properties = [];
  }

  if (hasNoValueOrText && indices.properties.length && !valueProp && !textProp) {
    throw new Error('Properties found in series but missing valueProp and textProp');
  }
}

function getValueOrFallback(
  primary: string | undefined,
  fallback: string | undefined,
  propName: string | undefined,
  fallbackPropName: string | undefined,
  properties?: Record<string, string>
): string {
  return (
    primary ||
    (propName && properties?.[propName]) ||
    fallback ||
    (fallbackPropName && properties?.[fallbackPropName]) ||
    ''
  );
}

type Indices = {
  value: number;
  text: number;
  properties: Array<{ name: string; index: number }>;
  expandable: number;
};

function findFieldsIndices(frames: DataFrame[]): Indices {
  const indices: Indices = {
    value: -1,
    text: -1,
    properties: [],
    expandable: -1,
  };

  for (const frame of getProcessedDataFrames(frames)) {
    for (let index = 0; index < frame.fields.length; index++) {
      const field = frame.fields[index];
      const fieldName = getFieldDisplayName(field, frame, frames).toLowerCase();

      if (field.type === FieldType.string) {
        if (fieldName === 'value' && indices.value === -1) {
          indices.value = index;
        } else if (fieldName === 'text' && indices.text === -1) {
          indices.text = index;
        } else {
          indices.properties.push({ name: fieldName, index });
        }
        continue;
      }

      if (
        fieldName === 'expandable' &&
        (field.type === FieldType.boolean || field.type === FieldType.number) &&
        indices.expandable === -1
      ) {
        indices.expandable = index;
      }
    }
  }

  return indices;
}

function areMetricFindValues(data: any[]): data is MetricFindValue[] {
  if (!data?.length) {
    return data?.length === 0;
  }

  const firstValue: any = data[0];

  if (isDataFrame(firstValue)) {
    return false;
  }

  for (const firstValueKey in firstValue) {
    if (!firstValue.hasOwnProperty(firstValueKey)) {
      continue;
    }

    const value = firstValue[firstValueKey];
    if (
      value !== null &&
      (typeof value === 'string' || typeof value === 'number') &&
      (firstValueKey.toLowerCase() === 'text' || firstValueKey.toLowerCase() === 'value')
    ) {
      return true;
    }
  }

  return false;
}

```

Mainly cleanup, removing repetitions, some common logic extraction.

Feel free to apply these changes

grafakus: Thanks for sharing!
I've applied part of these suggestions, to strike a balance between DRY and readability

packages/scenes/src/variables/variants/query/toMetricFindValues.ts
```diff
@@ -25,74 +33,125 @@ export function toMetricFindValues(): OperatorFunction<PanelData, MetricFindValu
...
+              properties[p.name] = fieldValue(p.index);
             }
 
-            if (fieldName === 'value' && field.type === FieldType.string && valueIndex === -1) {
-              valueIndex = index;
-            }
+            metrics.push({
+              value:
+                value ||
+                (valueProp && properties[valueProp as string]) ||
```
torkelo: valueProp is defiend as string, why the typecast?

grafakus: Good catch, not sure why it's there, removed.

PR #1239, opened 2025-09-05
ScopesVariable: Add flag for emitting on initial load

tskarhed:
Previously, this would cause a change event every time some url reloading would happen.

### DIFF
--- a/packages/scenes/src/variables/variants/ScopesVariable.tsx
+++ b/packages/scenes/src/variables/variants/ScopesVariable.tsx
@@ -30,6 +30,8 @@ export interface ScopesVariableState extends SceneVariableState {
 export class ScopesVariable extends SceneObjectBase<ScopesVariableState> implements SceneVariable<ScopesVariableState> {
   protected _renderBeforeActivation = true;
   protected _context: ScopesContextValue | undefined;
+  // Used to always emit an update event on initial load
+  private _isInitialUpdate = true;
 
   // Special options that enables variables to be hidden but still render to access react contexts
   public UNSAFE_renderAsHidden = true;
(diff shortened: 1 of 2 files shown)


packages/scenes/src/variables/variants/ScopesVariable.tsx
```diff
@@ -100,11 +102,13 @@ export class ScopesVariable extends SceneObjectBase<ScopesVariableState> impleme
     const scopesHaveChanged = !isEqual(oldScopes, newScopes);
 
     // Only update scopes value state when loading is false and the scopes have changed
-    if (!loading && (scopesHaveChanged || newScopes.length === 0)) {
+    // Emit on initial load to ensure that the variable is updated when the scopes are empty
```
torkelo, line 105: > Previously, this would cause a change event every time some url reloading would happen.

what do you mean by 'url reloading' ?  

why would scopesHaveChanged be true and need to be ignored?

torkelo: ah, is it the scenario where newScopes.length === 0? 

hm...  why is state updated with new empty scopes array on every url update?

tskarhed: It was noticed when switching on/off edit mode in dashboards, with empty set of scopes. I would assume that the URL change it causes also makes the scopes update?

Regardless, I think `newScopes.length === 0` is weird check that doesn't really do what it is supposed to. scopesHaveChanged is not true for the initial load.

torkelo: @tskarhed but why is stateObservable emitting a new scope value when going into/out of edit mode?

dprokop: @torkelo apparently this only happens when no scope is selected. @tskarhed is there any special handling for such case with empty scopes? I don't understand why this problem only occurs with scopes being empty.

tskarhed: > @tskarhed but why is stateObservable emitting a new scope value when going into/out of edit mode?

I'll see if I can look further into it next week.

> @torkelo apparently this only happens when no scope is selected. @tskarhed is there any special handling for such case with empty scopes? I don't understand why this problem only occurs with scopes being empty.

Do you mean the emitting of stateObservable when there is no reason to?

torkelo: > Do you mean the emitting of stateObservable when there is no reason to?

yes, exactly

PR #1240, opened 2025-09-05
SceneInteractionProfiler: Measure atomic selectbox interactions on adhocs and groupBy

mdvictor:
This PR creates a new profiler tool for more atomic interactions across scenes components and uses it to measure the duration between when a user opens an adhoc selectbox for example for keys until the keys are actually fetched from the backend.

<details>
  <summary>📦 Published PR as canary version: <code>6.37.0--canary.1240.17829532252.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes-react@6.37.0--canary.1240.17829532252.0
  npm install @grafana/scenes@6.37.0--canary.1240.17829532252.0
  # or 
  yarn add @grafana/scenes-react@6.37.0--canary.1240.17829532252.0
  yarn add @grafana/scenes@6.37.0--canary.1240.17829532252.0
  ```
</details>


dprokop (commented):
@mdvictor  - please have a look at my comment in https://github.com/grafana/grafana/pull/110709 - i don't see a reason for the profile class to be a behavior. What if it simply was a class that's instantiated in  SceneRenderProfiler for scenes access?  This way enabling the profiler would be handled through the same mehanism as scene rendering profiling.

packages/scenes/src/behaviors/SceneQueryController.ts
```diff
@@ -18,18 +19,28 @@ export class SceneQueryController
 
   #tryCompleteProfileFrameId: number | null = null;
 
-  public constructor(state: Partial<SceneQueryStateControllerState> = {}, private profiler?: SceneRenderProfiler) {
+  public constructor(
+    state: Partial<SceneQueryStateControllerState> = {},
+    private renderProfiler?: SceneRenderProfiler,
+    private interactionProfiler?: SceneInteractionProfiler
```
mdvictor: I feel like this does not belong in the queryController... although it makes it easier to have it here

dprokop: @mdvictor - yeah, it doesn't. I was thinking the interaction profiler should be composed in SceneRenderProfiler.

mdvictor: But even if it would be composed, we still interact with the SceneRenderProfiler (and thus the new interaction stuff) through the queryController, right? In both grafana and scenes, due to `onProfileComplete` which lives on the queryController behaviour.

PR #1242, opened 2025-09-08
GroupByVariable: Do not show default value in url

mdvictor:
This brings consistency with how the adhoc filters var does it.

Basically, default values do not get synced in the URL (as they are pulled for other sources e.g.: scopes or dashboards). As soon as there is user input that modifies the variable from its default value we begin syncing with the URL so we can carry this new value across dashboards.

### DIFF
--- a/packages/scenes/src/variables/groupby/GroupByVariableUrlSyncHandler.ts
+++ b/packages/scenes/src/variables/groupby/GroupByVariableUrlSyncHandler.ts
@@ -30,7 +30,10 @@ export class GroupByVariableUrlSyncHandler implements SceneObjectUrlSyncHandler 
     }
 
     return {
-      [this.getKey()]: toUrlValues(this._sceneObject.state.value, this._sceneObject.state.text),
+      [this.getKey()]:
+        this._sceneObject.state.defaultValue && !this._sceneObject.state.restorable
+          ? ['']
+          : toUrlValues(this._sceneObject.state.value, this._sceneObject.state.text),
       [this.getRestorableKey()]: this._sceneObject.state.defaultValue
         ? this._sceneObject.state.restorable
           ? 'true'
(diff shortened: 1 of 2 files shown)


packages/scenes/src/variables/groupby/GroupByVariableUrlSyncHandler.ts
```diff
@@ -30,7 +30,10 @@ export class GroupByVariableUrlSyncHandler implements SceneObjectUrlSyncHandler
     }
 
     return {
-      [this.getKey()]: toUrlValues(this._sceneObject.state.value, this._sceneObject.state.text),
+      [this.getKey()]:
+        this._sceneObject.state.defaultValue && !this._sceneObject.state.restorable
+          ? ['']
+          : toUrlValues(this._sceneObject.state.value, this._sceneObject.state.text),
```
dprokop, line 36: @mdvictor can we, instead of adding an empty string, just do not include this in the url? I think that's a bit more elegant.

mdvictor: IIRC the same happens in filters and there was some reason for this that escapes me now. I'll have a look and see if I can change this!

mdvictor: @dprokop Found the [PR](https://github.com/grafana/scenes/pull/560) related to the AdHocs that made me keep it [there](https://github.com/grafana/scenes/blob/main/packages/scenes/src/variables/adhoc/AdHocFiltersVariableUrlSyncHandler.ts#L55) and do the same in the groupBys

mdvictor: I'd reckon we'd need it here but will doublecheck

PR #1258, opened 2025-10-01
Variables: Introduce a new type called "switch"

leventebalogh:
### What changed?
This PR is the scenes counterpart of the https://github.com/grafana/grafana/pull/111366 core Grafana PR, that introduces a new variable for boolean values, and renders them a switch.

https://github.com/user-attachments/assets/f5251ffe-927b-4b5b-b27a-4890789c8d85

### DIFF
--- a/packages/scenes/src/variables/variants/SwitchVariable.tsx
+++ b/packages/scenes/src/variables/variants/SwitchVariable.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { Observable, of } from 'rxjs';
-import { Switch } from '@grafana/ui';
+import { Switch, useTheme2 } from '@grafana/ui';
+import { css } from '@emotion/css';
 import { SceneObjectBase } from '../../core/SceneObjectBase';
 import { VariableDependencyConfig } from '../VariableDependencyConfig';
 import {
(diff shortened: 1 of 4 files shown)


packages/scenes/src/variables/variants/SwitchVariable.tsx
```diff
@@ -0,0 +1,126 @@
...
+export interface SwitchVariableState extends SceneVariableState {
+  value: boolean;
+}
+
+export class SwitchVariable extends SceneObjectBase<SwitchVariableState> implements SceneVariable<SwitchVariableState> {
+  public static Component = SwitchVariableRenderer;
+
+  protected _variableDependency = new VariableDependencyConfig(this, {
+    statePaths: ['value'],
+  });
```
torkelo: don't think this variable should/need to support variables in it's value definition. Should only be `true` / `false'. and the other variations.

packages/scenes/src/variables/variants/SwitchVariable.tsx
```diff
@@ -0,0 +1,126 @@
...
+      this.setValue(boolValue);
+    }
+  }
+}
+
+function SwitchVariableRenderer({ model }: SceneComponentProps<SwitchVariable>) {
+  const state = model.useState();
+  const theme = useTheme2();
+
+  const containerStyle = css({
```
torkelo: useStyles2

torkelo: Any reason you can't use ControlLabel?

packages/scenes/src/variables/variants/SwitchVariable.tsx
leventebalogh: Removed in https://github.com/grafana/scenes/pull/1258/commits/d1872795a7942be3d45fe3c2a8b110aa3841432d.

packages/scenes/src/variables/variants/SwitchVariable.tsx
leventebalogh: Good catch, thanks, fixed to use `useStyles2()` in https://github.com/grafana/scenes/pull/1258/commits/103c5475aee0e95736b0039f72d6e3fcf44ddad1. 👍

leventebalogh: > Any reason you can't use ControlLabel?

Hmm, isn't it being used in `<VariableValueSelectWrapper>` in scenes and [in `<VariableValueSelectWrapper>` in core Grafana](https://github.com/grafana/grafana/blob/dbadd7a685382a0402de7a276a8b3823695b5c6f/public/app/features/dashboard-scene/scene/VariableControls.tsx#L37)? I was trying to only style the variable itself here, and the label be handled by the wrappers. Or did I miss something?

PR #1266, opened 2025-10-08
Annotations: Pass annotations through when transformations exist but none are annotations transformations

fastfrwrd:
Relates to [https://github.com/grafana/support-escalations/issues/18804](https://github.com/grafana/support-escalations/issues/18804)

If any transformations were set on a panel, and annotations were also set on that panel, but none of the transformations touched the annotations, then the update in https://github.com/grafana/scenes/pull/1207 led to an inadvertent side-effect where annotations were "eaten" by this change. The issue was mainly the missing null coalescence operator here to tack data.annotations on if there were no transformed annotations: https://github.com/grafana/scenes/pull/1207/files#diff-672934780f78622c0c2f2c605ff8848860c289c3e67c924bd1664ceef8d84ae0L206

<details>
  <summary>📦 Published PR as canary version: <code>6.39.5--canary.1266.18407595746.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@6.39.5--canary.1266.18407595746.0
  npm install @grafana/scenes-react@6.39.5--canary.1266.18407595746.0
  # or 
  yarn add @grafana/scenes@6.39.5--canary.1266.18407595746.0
  yarn add @grafana/scenes-react@6.39.5--canary.1266.18407595746.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/querying/SceneDataTransformer.ts
+++ b/packages/scenes/src/querying/SceneDataTransformer.ts
@@ -170,13 +170,21 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
 
     const seriesTransformations = this._filterAndPrepareTransformationsByTopic(
       interpolatedTransformations,
-      (topic) => topic == null || topic === DataTopic.Series,
-      true
+      (transformation) => {
+        if ('options' in transformation || 'topic' in transformation) {
+          return transformation.topic == null || transformation.topic === DataTopic.Series;
+        }
+        return true;
+      }
     );
     const annotationsTransformations = this._filterAndPrepareTransformationsByTopic(
       interpolatedTransformations,
-      DataTopic.Annotations,
-      false
+      (transformation) => {
+        if ('options' in transformation || 'topic' in transformation) {
+          return transformation.topic === DataTopic.Annotations;
+        }
+        return false;
+      }
     );
 
     if (this._transformSub) {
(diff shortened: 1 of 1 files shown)


packages/scenes/src/querying/SceneDataTransformer.test.ts
```diff
@@ -809,6 +837,14 @@ describe('SceneDataTransformer', () => {
...
             meta: { dataTopic: 'annotations' },
           },
+          {
+            fields: [
+              { name: '0', config: {}, values: [400, 500, 600], type: 'number' },
+              { name: '1', config: {}, values: [1, 2, 3], type: 'number' },
+            ],
+            length: 3,
+            meta: { dataTopic: 'annotations' },
+          },
```
fastfrwrd, line 846: @leeoniya I wanted to check one thing - this change changes the outcome of this test in particular, so I want to make sure this is correct. In the original data here, there's a series and an an annotation frame. This new outcome seems correct to me, where they'd both appear as annotations if this transform exists, right? in the previous test, only the transformed series data was coming through.

leeoniya: yeah looks like this wasn't correct before :grimacing:

packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -269,4 +258,21 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
         : t;
     });
   }
+
+  private _filterAndPrepareTransformationsByTopic(
+    interpolatedTransformations: Array<DataTransformerConfig<any> | CustomTransformerDefinition>,
+    topic: DataTopic | ((topic?: DataTopic) => boolean),
+    includeByDefault: boolean
+  ): Array<DataTransformerConfig<any> | CustomTransformOperator> {
+    const topicCheckFn = typeof topic === 'function' ? topic : (t?: DataTopic) => t === topic;
```
mdvictor: Nit: It feels a bit confusing to have both types on topic, how about just having a fn check so it's consistent? Either way, it's nonblocking

fastfrwrd: I cleaned it up a bit, thanks for the feedback!

PR #1272, opened 2025-10-10
Decouple AdHocFiltersComboboxRenderer from AdHocFiltersVariable

adrapereira:
We've seen some requests to use the Combobox filters component in places where the plugin or core Grafana feature isn't using scenes, thus making using the variable not possible/recommended. 

This PR decouples the AdHocFiltersComboboxRenderer React component from the AdHocFiltersVariable by replacing the scene model with a controller class instance. 

This doesn't change anything in the behavior of the AdHocFiltersVariable because we defined a controller for the variable that simply routes the function calls to the variable. See [VariableBackedAdHocFiltersController.ts](https://github.com/grafana/scenes/compare/andre/refactor-adhoc-filters-variable-combobox?expand=1#diff-4fdbe2c7e8e7fcfed29f4fd74c6cfc4dc337076b6e2111861c1f7e7b7fcd2450).

Developers who want to use the combobox filters component without the variable must define the controller to fit their needs. This means that the filters and `wip` filter state must be managed in the controller, the component doesn't keep any state. Same for the keys and values used for the dropdown elements, the developer must provide these since the component doesn't automatically retrieve them from a datasource. 

Additionally it supports setting the input placeholder value, originally defaulted to `Filter by label values`.

This PR in grafana/grafana uses the new controller backed component in the Trace View panel: https://github.com/grafana/grafana/pull/112287

<details>
  <summary>📦 Published PR as canary version: <code>6.42.0--canary.1272.18911329941.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@6.42.0--canary.1272.18911329941.0
  npm install @grafana/scenes-react@6.42.0--canary.1272.18911329941.0
  # or 
  yarn add @grafana/scenes@6.42.0--canary.1272.18911329941.0
  yarn add @grafana/scenes-react@6.42.0--canary.1272.18911329941.0
  ```
</details>

### DIFF
--- a/packages/scenes-app/src/demos/adhocFiltersControllerExample.tsx
+++ b/packages/scenes-app/src/demos/adhocFiltersControllerExample.tsx
@@ -83,8 +83,6 @@ export class SimpleController implements AdHocFiltersController {
   }
 
   updateFilter(filter: AdHocFilterWithLabels, update: Partial<AdHocFilterWithLabels>): void {
-    console.log('updateFilter', filter, update);
-    console.log('wip', this.wip);
     if (filter === this.wip) {
       // If we set value we are done with this "work in progress" filter and we can add it
       if ('value' in update && update['value'] !== '') {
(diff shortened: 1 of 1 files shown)


packages/scenes/src/variables/adhoc/controller/VariableBackedAdHocFiltersController.ts
```diff
@@ -0,0 +1,81 @@
...
+import { AdHocFilterWithLabels, AdHocFiltersVariable } from '../AdHocFiltersVariable';
+import { AdHocFiltersController, AdHocFiltersControllerState } from './AdHocFiltersController';
+import { getQueryController } from '../../../core/sceneGraph/getQueryController';
+import { getInteractionTracker } from '../../../core/sceneGraph/getInteractionTracker';
+
+/**
+ * Adapter that wraps AdHocFiltersVariable to implement the AdHocFiltersController interface.
+ * This allows the combobox UI to work with the variable while keeping the UI decoupled.
+ */
+export class VariableBackedAdHocFiltersController implements AdHocFiltersController {
```
Sergej-Vlasov: small nit: Maybe shorter name like `AdHocFiltersController`?

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -783,8 +784,11 @@ export function AdHocFiltersVariableRenderer({ model }: SceneComponentProps<AdHo
   const { filters, readOnly, addFilterButtonText } = model.useState();
   const styles = useStyles2(getStyles);
 
+  // Create controller adapter for combobox mode
+  const controller = useMemo(() => new VariableBackedAdHocFiltersController(model), [model]);
+
   if (model.state.layout === 'combobox') {
-    return <AdHocFiltersComboboxRenderer model={model} />;
+    return <AdHocFiltersComboboxRenderer controller={controller} />;
```
Sergej-Vlasov, line 794: considering `combobox` UI is GA this is alright but thinking if we should not initiate controller for old UI:
```typescript
  // Create controller adapter for combobox mode
  const controller = useMemo(
    () => (model.state.layout === 'combobox' ? new VariableBackedAdHocFiltersController(model) : undefined),
    [model]
  );

  if (controller) {
    return <AdHocFiltersComboboxRenderer controller={controller} />;
  }
```

packages/scenes/src/variables/adhoc/controller/VariableBackedAdHocFiltersController.ts
mdvictor: Or `AdHocFiltersVariableController` or some derivate, would be nice to have a different name here -- `VariableBacked...` is a bit confusing to me

mdvictor (approved):
LGTM! Awesome work!

PR #1274, opened 2025-10-15
Make getExploreURL() return the explore URL including a possible subpath

### LINKED ISSUE #1273: getExploreURL ignores appSubUrl
When hosting Grafana under a subpath, eg. `example.com/grafana`, `getExploreUrl()` doesn't work as it ignores the path.

https://github.com/grafana/scenes/blob/105080cfa0c6296361c0dd628f80ebf5b7a7ddf0/packages/scenes/src/utils/explore.ts#L62

LinzerToertchen:
Fixes #1273

### DIFF
--- a/packages/scenes/src/utils/explore.ts
+++ b/packages/scenes/src/utils/explore.ts
@@ -3,6 +3,7 @@ import { SceneObject } from '../core/types';
 import { wrapInSafeSerializableSceneObject } from './wrapInSafeSerializableSceneObject';
 import { getDataSourceSrv } from '@grafana/runtime';
 import { DataQuery } from '@grafana/schema';
+import { config } from '@grafana/runtime';
 
 /**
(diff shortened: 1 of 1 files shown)


packages/scenes/src/utils/explore.ts
```diff
@@ -59,7 +60,7 @@ export async function getExploreURL(
       })
     );
 
-    return `/explore?left=${left}`;
+    return `${config.appSubUrl}/explore?left=${left}`;
```
gtk-grafana, line 63: This is a breaking change for anyone that's already worked-around this, it's probably better to find a solution that avoids regression.

Maybe something like adding a new `options` argument to `getExploreURL` that includes a `returnExplorePath` param which is true by default, or maybe `includeSubPath` which is false by default might be cleaner. Documenting the current behavior in the spec of that object would be another great way to keep folks from running into this in the future.

Also we will need test coverage before this can be merged

PR #1277, opened 2025-10-15
SceneTimeRangeCompare: Fix url syncing

drew08t:
This fixes the url syncing for time comparison during page refresh by inputting an option key (to be passed by the panel to include `compareWith-[panel id]`

Before:
Oct-15-2025 16-48-58

After:
Oct-15-2025 16-36-33

Fixes https://github.com/grafana/grafana/issues/112274

### DIFF
--- a/packages/scenes/src/components/SceneTimeRangeCompare.tsx
+++ b/packages/scenes/src/components/SceneTimeRangeCompare.tsx
@@ -17,6 +17,7 @@ interface SceneTimeRangeCompareState extends SceneObjectState {
   compareWith?: string;
   compareOptions: Array<{ label: string; value: string }>;
   hideCheckbox?: boolean;
+  key?: string;
 }
 
 const PREVIOUS_PERIOD_VALUE = '__previousPeriod';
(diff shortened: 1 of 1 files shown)


packages/scenes/src/components/SceneTimeRangeCompare.tsx
```diff
@@ -17,6 +17,7 @@ interface SceneTimeRangeCompareState extends SceneObjectState {
   compareWith?: string;
   compareOptions: Array<{ label: string; value: string }>;
   hideCheckbox?: boolean;
+  key?: string;
```
torkelo, line 20: key is already a used state property that has specific meaning/usage.

packages/scenes/src/components/SceneTimeRangeCompare.tsx
```diff
@@ -156,29 +163,23 @@ export class SceneTimeRangeCompare
   }
 
   public getUrlState(): SceneObjectUrlValues {
-    return {
-      compareWith: this.state.compareWith,
-    };
+    const key = this.state.key || 'compareWith';
+    return { [key]: this.state.compareWith };
```
torkelo, line 167: why this change? the UrlSyncManager and UniqueUrlKeyMapper should already handle making url keys unique for the scene, might be some bug there we can fix?

PR #1281, opened 2025-10-21
UrlSyncManager: Fixes performance issue with unique key logic 

torkelo:
* Fixes performance issue with filtering out orphaned objects. 
* adds new perf test dashboard for time range compare per panel and many panels 

While this fixes the issue there are more issues with url sync and time range compare I think, not really designed for this many panels and per panel compare select with url state. 

Another problem with url state is for repeated panels / layouts, the unique key assigned will be the same as it's only based on depth/location in tree. 

<details>
  <summary>📦 Published PR as canary version: <code>6.40.0--canary.1281.18720650291.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@6.40.0--canary.1281.18720650291.0
  npm install @grafana/scenes-react@6.40.0--canary.1281.18720650291.0
  # or 
  yarn add @grafana/scenes@6.40.0--canary.1281.18720650291.0
  yarn add @grafana/scenes-react@6.40.0--canary.1281.18720650291.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/SceneTimeRangeCompare.tsx
+++ b/packages/scenes/src/components/SceneTimeRangeCompare.tsx
@@ -156,7 +156,6 @@ export class SceneTimeRangeCompare
   }
 
   public getUrlState(): SceneObjectUrlValues {
-    console.log('SceneTimeRangeCompare getUrlState called');
     return {
       compareWith: this.state.compareWith,
     };
--- a/packages/scenes/src/services/UniqueUrlKeyMapper.ts
+++ b/packages/scenes/src/services/UniqueUrlKeyMapper.ts
@@ -75,6 +75,7 @@ function filterOutOrphanedObjects(sceneObjects: SceneObject[], root: SceneObject
 }
 
 function isOrphan(obj: SceneObject, root: SceneObject) {
+  // We don't consider objects without parents to be orphan as they can be top-level scene objects
   if (!obj.parent) {
     return false;
   }
@@ -86,6 +87,7 @@ function isOrphan(obj: SceneObject, root: SceneObject) {
       found = true;
       return false; // stop iteration
     }
+    return;
   });
 
   // If we did not find the object among its parent's children it's an orphan


packages/scenes/src/components/SceneTimeRangeCompare.tsx
```diff
@@ -156,6 +156,7 @@ export class SceneTimeRangeCompare
   }
 
   public getUrlState(): SceneObjectUrlValues {
+    console.log('SceneTimeRangeCompare getUrlState called');
```
bfmatei: I think this needs to be removed

packages/scenes/src/services/UniqueUrlKeyMapper.ts
```diff
@@ -65,22 +64,34 @@ export class UniqueUrlKeyMapper {
...
+      sceneObjects.splice(i, 1);
+      i--; // Decrement i to account for the removed element
     }
   }
 }
 
-function isOrphanOrInActive(obj: SceneObject) {
-  const root = obj.getRoot();
+function isOrphan(obj: SceneObject, root: SceneObject) {
+  if (!obj.parent) {
```
bfmatei, line 79:
```suggestion
  // We don't consider objects without parents to be orphan as they can be top-level scene objects
  if (!obj.parent) {
```

bfmatei: I think a comment can make it a little bit more clearer for long-term maintenance

packages/scenes/src/services/UniqueUrlKeyMapper.ts
```diff
@@ -65,22 +64,34 @@ export class UniqueUrlKeyMapper {
...
+    if (isOrphan(obj, root)) {
+      sceneObjects.splice(i, 1);
+      i--; // Decrement i to account for the removed element
     }
   }
 }
 
-function isOrphanOrInActive(obj: SceneObject) {
-  const root = obj.getRoot();
+function isOrphan(obj: SceneObject, root: SceneObject) {
```
bfmatei, line 77: One thing that bugs me is why the original fn was called `OrInActive`. Should we exclude inactive objects as well?

packages/scenes/src/services/UniqueUrlKeyMapper.ts
```diff
@@ -65,22 +64,34 @@ export class UniqueUrlKeyMapper {
...
+    return false;
+  }
+
+  let found = false;
+
+  obj.parent.forEachChild((child) => {
+    if (child === obj) {
+      found = true;
+      return false; // stop iteration
+    }
```
bfmatei, line 89:
```suggestion
    }
    return;
```

This should prevent the typecheck issue.

bfmatei (approved):
Overall this looks good. I left a couple of nits as well as a question.

packages/scenes/src/services/UniqueUrlKeyMapper.ts
torkelo: that naming was a bit old I think, before I realized that we have to consider in-active objects as well (for say panels inside collapsed rows) , meaning we should only remove orphaned objects

bfmatei: That makes sense.

PR #1297, opened 2025-11-06
AdhocFiltersVariable: Expose `originFilters` in links

mdvictor:
Allows the usage of `${adhoc_variable_name.originFilters}` to retrieve AdHoc `originFilters` in link builders.

<details>
  <summary>📦 Published PR as canary version: <code>6.43.0--canary.1297.19158576084.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@6.43.0--canary.1297.19158576084.0
  npm install @grafana/scenes-react@6.43.0--canary.1297.19158576084.0
  # or 
  yarn add @grafana/scenes@6.43.0--canary.1297.19158576084.0
  yarn add @grafana/scenes-react@6.43.0--canary.1297.19158576084.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.test.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.test.tsx
@@ -1810,7 +1810,7 @@ describe.each(['11.1.2', '11.1.1'])('AdHocFiltersVariable', (v) => {
     expect(value).toEqual(['scopeOriginFilter|=|val#scope']);
   });
 
-  it.only('Returns filters expression through getValue()', () => {
+  it('Returns filters expression through getValue()', () => {
     const { filtersVar } = setup({
       filters: [{ key: 'key1', operator: '=', value: 'val1' }],
     });


packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -55,6 +61,8 @@ export interface AdHocFilterWithLabels<M extends Record<string, any> = {}> exten
   nonApplicableReason?: string;
 }
 
+const ORIGIN_FILTERS_KEY: keyof AdHocFiltersVariableState = 'originFilters';
```
tskarhed, line 64: It is a little bit unclear what the naming means. Maybe because I don't fully understand what originFilters are. Is it all filters or just filters with a specific origin?

mdvictor: just filters with an origin, this is basically a type guard for it's usage below in `getValue`, to not hardcode the string in there, but use this which is certain to be a keyof the state, which contains the said `originFilters` property. 

I think to make it more explicit I could go with `ORIGIN_FILTERS_PROPERTY_NAME` maybe? WDYT?

mdvictor: or `ORIGIN_FILTERS_STATE_PROPERTY_NAME `

tskarhed: I am not too worried about the naming of the constant, more the property it refers to `${adhoc_variable_name.originFilters}`. With some documentation it would probably be okay though.

mdvictor: hmm, yeah. I reckon some docs would help here, my thinking here was to keep the name of the property itself on the variable state as this could be extended to other variable data maybe, like options for example, or maybe baseFilters in adhocs

torkelo (commented):
why would this be needed? originFilters are added automatically from schema

mdvictor:
@torkelo yes, but there is a need for them to be exposed in the url. default values are not persisted in the urlSync so they cannot be used in link building by just calling ${adhoc}

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.test.tsx
```diff
@@ -1794,6 +1793,32 @@ describe.each(['11.1.2', '11.1.1'])('AdHocFiltersVariable', (v) => {
...
+          origin: 'scope',
+        },
+      ],
+    });
+
+    const value = filtersVar.getValue('originFilters');
+    expect(value).toEqual(['scopeOriginFilter|=|val#scope']);
+  });
+
+  it.only('Returns filters expression through getValue()', () => {
```
torkelo:
```suggestion
  it('Returns filters expression through getValue()', () => {
```

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.test.tsx
```diff
@@ -1794,6 +1793,32 @@ describe.each(['11.1.2', '11.1.1'])('AdHocFiltersVariable', (v) => {
...
+          operator: '=',
+          value: 'val',
+          values: ['val'],
+          origin: 'scope',
+        },
+      ],
+    });
+
+    const value = filtersVar.getValue('originFilters');
+    expect(value).toEqual(['scopeOriginFilter|=|val#scope']);
```
torkelo, line 1810: this looks strange, does an external system know about this url param ? and want only the origin filters?

torkelo (approved):
can't they use the label filters associated with the data instead? that should include origin and normal filters? 

the concept of origin filters still seem so complex and messy, can still not really groc it

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.test.tsx
mdvictor: that is the case, yes

mdvictor:
i'm afraid the data approach may be hit or miss due to labels potentially not being on the targets if they are not applicable :(

PR #1303, opened 2025-11-14
Drilldowns: Applicabilty check on drilldown variables

mdvictor:
Adds a way to verify if the drilldown datasource has applicability enabled without having to asynchronously fetch the datasource and verify if `getDrilldownsApplicability` is implemented

<details>
  <summary>📦 Published PR as canary version: <code>6.47.1--canary.1303.19499670138.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@6.47.1--canary.1303.19499670138.0
  npm install @grafana/scenes-react@6.47.1--canary.1303.19499670138.0
  # or 
  yarn add @grafana/scenes@6.47.1--canary.1303.19499670138.0
  yarn add @grafana/scenes-react@6.47.1--canary.1303.19499670138.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/utils.ts
+++ b/packages/scenes/src/variables/utils.ts
@@ -12,6 +12,7 @@ import { sceneGraph } from '../core/sceneGraph';
 import { SceneDataQuery, SceneObject, SceneObjectState } from '../core/types';
 import { SceneQueryRunner } from '../querying/SceneQueryRunner';
 import { DataSourceRef } from '@grafana/schema';
+import { css } from '@emotion/css';
 
 export function isVariableValueEqual(a: VariableValue | null | undefined, b: VariableValue | null | undefined) {
   if (a === b) {


packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -624,9 +630,12 @@ export class AdHocFiltersVariable
     const response = await this.getFiltersApplicabilityForQueries(filters, queries ?? []);
```
dprokop, line 630: @mdvictor - this will still async call the ds service to verify if the method is implemented, so this PR does not add much. Or I am missing sth?

mdvictor: Indeed, this still does the async DS call, but only in the adhocVar - we need to get the ds to actually call the applicability (similar like the getTagKey/Value calls).

The benefit of this is in other places that need to check if the DS supports applicability (e.g. panel subHeader or panel groupBy action). In those places instead of fetching the DS and checking if it supports applicability (per panel for example), we can rely on this flag set on the adhoc/groupby to give us that information

torkelo: I think dominiks point is, if the VizPanel sub header renders before this async completes how will be re-rendered when it does complete?

mdvictor: aah that's a good point, then this can be moved in the state and will trigger a rerender on the subheader when it updates

mdvictor: I've switched to state here

packages/scenes/src/variables/utils.ts
```diff
@@ -275,3 +281,78 @@ export function getNonApplicablePillStyles(theme: GrafanaTheme2) {
...
+  isApplicabilityEnabled?: boolean
+): boolean {
+  const datasourceUid = sceneGraph.interpolate(sourceObject, queriesDataSource?.uid);
+
+  return Boolean(
+    isApplicabilityEnabled && datasourceUid === sceneGraph.interpolate(sourceObject, drilldownDatasource?.uid)
+  );
+}
+
+export async function getDrilldownApplicability(
```
dprokop: @mdvictor I don't see a strong argument for exposing those one-off methods from scenes. WDYT about moving them to core, close to where per-panel applicability code lives? If we encounter a need to make this a reusable util, we can always decide to move it to scenes and make it a public API.

mdvictor: Ok, I can move them there. My only logic was that it is reused in other core components as well - e.g. https://github.com/grafana/grafana/pull/113445, but then again it makes sense to keep them in core

mdvictor:
Moved helpers to core

PR #1306, opened 2025-11-26
Introduce regexApplyTo variable field

kristinademeshchik:
**What is this feature?**

The new field defines whether regex should be applied to the value or the display text, with the default remaining set to the value for consistency.

**Why do we need this feature?**

For multiple data sources, such as Azure, Cloudwatch, or AWS-IOT, when setting up a regex for the variable, it gets applied to the value rather than the text. This can be problematic because, in the UI, customers only see the text, making it difficult to obtain the value.

<details>
  <summary>📦 Published PR as canary version: <code>6.49.0--canary.1306.19943511147.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@6.49.0--canary.1306.19943511147.0
  npm install @grafana/scenes-react@6.49.0--canary.1306.19943511147.0
  # or 
  yarn add @grafana/scenes@6.49.0--canary.1306.19943511147.0
  yarn add @grafana/scenes-react@6.49.0--canary.1306.19943511147.0
  ```
</details>

### DIFF
--- a/packages/scenes/package.json
+++ b/packages/scenes/package.json
@@ -29,7 +29,6 @@
     "clean": "rimraf ./dist ./compiled ./package.tgz",
     "typecheck": "tsc --emitDeclarationOnly false --noEmit",
     "test": "jest",
-    "test:watch": "jest --watch",
     "lint": "eslint --cache .",
     "prepack": "node ./scripts/prepack.js",
     "postpack": "node ./scripts/postpack.js",


packages/scenes/package.json
```diff
@@ -29,6 +29,7 @@
     "clean": "rimraf ./dist ./compiled ./package.tgz",
     "typecheck": "tsc --emitDeclarationOnly false --noEmit",
     "test": "jest",
+    "test:watch": "jest --watch",
```
ivanortegaalba: 👏

packages/scenes/src/variables/variants/query/QueryVariable.test.tsx
```diff
@@ -32,17 +34,30 @@ import { config, setRunRequest } from '@grafana/runtime';
...
 
-const runRequestMock = jest.fn().mockReturnValue(
-  of<PanelData>({
+// TODO: Fallback enum for backward compatibility with older versions of @grafana/data
+const VariableRegexApplyTo =
+  VariableRegexApplyToFromData ||
+  ({
+    value: 'value',
+    text: 'text',
+  } as const);
```
ivanortegaalba: Let's update this if [we finally decide to use string unions](https://github.com/grafana/grafana/pull/114426#discussion_r2588436326) 😄

packages/scenes/src/variables/variants/query/QueryVariable.tsx
```diff
@@ -61,6 +64,8 @@ export class QueryVariable extends MultiValueVariable<QueryVariableState> {
       options: [],
       datasource: null,
       regex: '',
+      // TODO: remove harcoded default value after updating grafana/data
+      regexApplyTo: VariableRegexApplyTo?.value ?? 'value',
```
ivanortegaalba: I'd say it's good to have a default value that doesn't depend on Grafana version to keep the logic working as expected :)

packages/scenes/src/variables/variants/query/QueryVariable.test.tsx
dprokop: +1

packages/scenes/package.json
dprokop: wonder if we need it when you can simply run `yarn test -- --watch`? Matter of preference I think. Anyways adding this won't harm :)

dprokop (approved):

packages/scenes/package.json
kristinademeshchik: Well, that’s the whole point of these scripts, right? You can hide implementation details behind `test:watch`, and if the script changes later (like adding a `--notify` flag), you don’t need to remember or type anything new.

I personally prefer to put `watch` under `test` script, but it looks like that one is used in CI, so I didn’t want to mess anything up 

Alternatively, we could introduce `test:ci` for CI and use `test` for local dev

packages/scenes/src/variables/variants/query/QueryVariable.test.tsx
kristinademeshchik: updated ✅

packages/scenes/package.json
kristinademeshchik: ooh wait a second, I just learned about `test:scenes`
alright, reverting this then

PR #1345, opened 2026-01-30
Fundamentals: No more flickering with render before activation global default option

torkelo:
Problems

* Because scene components are not rendered until after the activation useEffect has triggered this "null" render clears the dom causing flickering when changing scene object instances 
* VizPanel does not render PanelChrome before div measurements are completed, causing minor flickering as PanelChrome is not rendered on first render  
* LazyLoader renders empty div before IntersectionObserver has checked if div is visible or not causing empty div on first render => flickering 

## Changes required when SceneObjectBase.RENDER_BEFORE_ACTIVATION_DEFAULT = true

* Global static flag SceneObjectBase.RENDER_BEFORE_ACTIVATION_DEFAULT can now make all scene objects render before activation 
* This means panels/query runners activate before their parents, required change to isVariableLoadingOrWaitingToUpdate to account for sets that have not been activated yet
* Scenes-react required changes, need to add the scene object to the context in the render path 

## Changes required for LazyLoader 

* We have to render children even on first mount, query execution is still blocked due to LazyLoaderInViewContext state which defaults to false 
* Question do we need an opt-in / option for this change in behavior?

## Changes to VizPanelRendering

* Render PanelChrome even when we have no div measurement (PanelChrome will be rendered with undefined width / height which will switch it to auto size mode). 
* Visualization is still not rendered until we have a measurement 

To try the new mode, a scene app plugin or core Grafana can enable the new behavior by executing 

```ts
  SceneObjectBase.RENDER_BEFORE_ACTIVATION_DEFAULT = true
```

https://github.com/user-attachments/assets/3a34f1d5-4700-401d-bfc0-965e3ea06a61

Related OSS PR: https://github.com/grafana/grafana/pull/118567

<details>
  <summary>📦 Published PR as canary version: <code>7.5.0--canary.1345.24995132997.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@7.5.0--canary.1345.24995132997.0
  npm install @grafana/scenes-react@7.5.0--canary.1345.24995132997.0
  # or 
  yarn add @grafana/scenes@7.5.0--canary.1345.24995132997.0
  yarn add @grafana/scenes-react@7.5.0--canary.1345.24995132997.0
  ```
</details>


packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -390,7 +390,9 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
 
     this._timeSubRange = timeRange;
     this._timeSub = timeRange.subscribeToState(() => {
-      this.runWithTimeRange(timeRange);
+      // setTimeout to let SceneVariableSet also respond to time range change
+      // So that variables that depend on time range have time to switch to loading state
+      setTimeout(() => this.runWithTimeRange(timeRange), 0);
```
torkelo, line 480: This (and same in AnnotationDataLayer) is the only fix that trouble me but cannot figure out a fix without a much large system change. 

The problem here is that SceneQueryRunner activates before any parent SceneVariableSet so SceneQueryRunner is first to subscribe to the time range so get's this event before SceneVariableSet so SceneVariableSet has no time to start loading a variable with refresh on time range change. 

So, without this setTimeout, queries are executed twice if they depend on a variable that has "refresh on time range change". 

to solve this more cleanly, we would need a new way to propagate time range changes from root to leaf (even React would struggle with this as useEffect triggers on leaves before children).

packages/scenes-react/src/contexts/SceneContextObject.tsx
```diff
@@ -87,3 +88,36 @@ export class SceneContextObject extends SceneObjectBase<SceneContextObjectState>
...
+      deactivate();
+    };
+  }, [ctx, obj]);
+
+  // Check if scene contains object instance
+  if (ctx.state.children.includes(obj)) {
+    return;
+  }
+
+  // This is technically a state change during render. Ee have to add it to the state tree right away in order to render the object on the first pass
```
oscarkilhed:
```suggestion
  // This is technically a state change during render. We have to add it to the state tree right away in order to render the object on the first pass
```

packages/scenes/src/core/SceneComponentWrapper.test.tsx
```diff
@@ -31,7 +32,11 @@ export class TestScene extends SceneObjectBase<TestSceneState> {
   };
 }
 
-describe('SceneComponentWrapper', () => {
+describe('SceneComponentWrapper no render before activiation', () => {
```
oscarkilhed:
```suggestion
describe('SceneComponentWrapper no render before activation', () => {
```

packages/scenes/src/core/SceneComponentWrapper.test.tsx
```diff
@@ -58,13 +63,43 @@ describe('SceneComponentWrapper', () => {
     expect(scene.renderCount).toBe(1);
     expect(screen.getByText('isActive: true')).toBeInTheDocument();
   });
+});
 
-  it('should render component before activation whgen renderBeforeActivation is true', () => {
-    const scene = new TestScene({ name: 'nested' });
-    scene.setRenderBeforeActivation(true);
+describe('SceneComponentWrapper render before activiation', () => {
```
oscarkilhed:
```suggestion
describe('SceneComponentWrapper render before activation', () => {
```

oscarkilhed (approved):
I've tested this with Grafana and it seems to work just fine. Left some minor suggestions that were just typos.

PR #1424, opened 2026-04-09
Filters: Report interactions through behaviour

mdvictor:
This PR addresses a shortcoming of https://github.com/grafana/scenes/pull/1422/ by reporting interactions through a behaviour set from core grafana. This is so we do not report all incoming interactions coming from scenes adhoc usage.

<details>
  <summary>📦 Published PR as canary version: <code>8.0.1--canary.1424.25052097964.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@8.0.1--canary.1424.25052097964.0
  npm install @grafana/scenes-react@8.0.1--canary.1424.25052097964.0
  # or 
  yarn add @grafana/scenes@8.0.1--canary.1424.25052097964.0
  yarn add @grafana/scenes-react@8.0.1--canary.1424.25052097964.0
  ```
</details>


packages/scenes/src/variables/adhoc/AdHocFiltersRecommendations.tsx
```diff
@@ -362,7 +362,7 @@ function AdHocFiltersRecommendationsRenderer({ model }: SceneComponentProps<AdHo
       const exists = filters.some((f) => f.key === filter.key && f.value === filter.value);
       if (!exists) {
         model.addFilterToParent(filter);
-        reportInteraction('grafana_unified_drilldown_recent_filter_applied', {
+        getReportInteractionHandler(model)?.reportInteraction('grafana_unified_drilldown_recent_filter_applied', {
```
dprokop: @mdvictor - the behavior approach is fine-ish, i think. My point really is those event names should be up to the consumer, do oyu have any thoughts on that?. And the callback should really just accept payload. 

But this is also interesting approach that you are proposing here, we ccould apply it in other places where we expose interaction tracking (in particular performance profiler).

mdvictor: hmm, I see what you mean, it makes sense. Yeah I think we can continue using the behaviour but extend it so the events come from grafana. will update, thank you!

mdvictor: @dprokop I've modified the code and given control to consumers on implementing this, let me know WDYT

harisrozajac (approved):

PR #1070, opened 2025-03-05
SceneObjectBase: Fixes crash TypeError circular reference when stringifying scene object

torkelo:
We use tests a lot that verify scene object instance is what we expect, when these tests fail jest usually prints this 

```
 Test suite failed to run

    TypeError: Converting circular structure to JSON
        --> starting at object with constructor 'PanelTimeRange'
        |     property '_variableDependency' -> object with constructor 'VariableDependencyConfig'
        --- property '_sceneObject' closes the circle
        at stringify (<anonymous>)

      at messageParent (node_modules/jest-worker/build/workers/messageParent.js:29:19)
```

This PR fixes this to implement a custom toJSON function  

<details>
  <summary>📦 Published PR as canary version: <code>6.2.0--canary.1070.13674378560.0</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install @grafana/scenes-react@6.2.0--canary.1070.13674378560.0
  npm install @grafana/scenes@6.2.0--canary.1070.13674378560.0
  # or 
  yarn add @grafana/scenes-react@6.2.0--canary.1070.13674378560.0
  yarn add @grafana/scenes@6.2.0--canary.1070.13674378560.0
  ```
</details>

PR #1426, opened 2026-04-10
Jest: Upgrade jest to fix issue with logging scene objects

torkelo:
Seems like jest did not always call toJSON on object prototypes (only instances) so we still have issues with jest crashing when logging scene object instances. Related to https://github.com/grafana/scenes/pull/1070 

fixes issues like. 

```
 Test suite failed to run

    TypeError: Converting circular structure to JSON

```

The latest jest fixes this issue. I was unable to create a test case for this as it only happens if an outer expect fails (not possible to replicate with an expect throws, for example). 

We need to upgrade jest in core repo as well to fix it there


packages/scenes-app/package.json
```diff
@@ -29,7 +29,7 @@
     "@swc/jest": "^0.2.36",
```
bfmatei, line 29: (nit) I think we should also update the `@swc/jest` to `0.2.39` to explicitly support it.

```suggestion
    "@swc/jest": "^0.2.39",
```

bfmatei (approved):
🚀

PR #1430, opened 2026-04-20
SceneQueryRunner: Clear incomplete data on deactivation to prevent stuck panels

kristinademeshchik:
**What is this feature?**

when a panel is deactivated while its query is still in flight (e.g. another panel temporarily activated it via the `dashboard` datasource and then released it), the panel's data state remains "Loading" with no active subscription to resolve it.

on reactivation, the panel sees existing data and assumes it doesn't need to re-query and leaving it permanently ostuck in a loading state.

on deactivation, we should discard any data that hasn't reached a terminal state (Done or Error). This ensures that when the panel reactivates, it recognizes there's no usable data and issues a fresh query.

data that already completed successfully is preserved, so panels scrolling in and out of view don't unnecessarily re-fetch.

<details>
  <summary>📦 Published PR as canary version: <code>7.4.1--canary.1430.24681505648.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@7.4.1--canary.1430.24681505648.0
  npm install @grafana/scenes-react@7.4.1--canary.1430.24681505648.0
  # or 
  yarn add @grafana/scenes@7.4.1--canary.1430.24681505648.0
  yarn add @grafana/scenes-react@7.4.1--canary.1430.24681505648.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.ts
@@ -428,6 +428,12 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
     this._timeSub = undefined;
     this._timeSubRange = undefined;
 
+    // Reset data to its initial state unless it reached a terminal state (Done or Error).
+    // Incomplete data is discarded so reactivation starts with a clean slate.
+    if (this.state.data && ![LoadingState.Done, LoadingState.Error].includes(this.state.data.state)) {
+      this.setState({ data: undefined });
+    }
+
     this._drilldownDependenciesManager.cleanup();
   }
 
--- a/packages/scenes/src/querying/SceneQueryRunner.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.ts
@@ -429,7 +429,7 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
     this._timeSubRange = undefined;
 
     // Reset data to its initial state unless it reached a terminal state (Done or Error).
-    // Incomplete data is discarded so reactivation starts with a clean slate.
+    // Incomplete data is discarded so reactivation starts with a clean state.
     if (this.state.data && ![LoadingState.Done, LoadingState.Error].includes(this.state.data.state)) {
       this.setState({ data: undefined });
     }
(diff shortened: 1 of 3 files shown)


torkelo (commented):
Could you add a unit test for this scenario?

kristinademeshchik:
> Could you add a unit test for this scenario?

done ✅

harisrozajac (approved):
looks good!

oscarkilhed (approved):

dprokop:
@kristinademeshchik how can i reproduce this scenario in the UI?

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -428,6 +428,12 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
     this._timeSub = undefined;
     this._timeSubRange = undefined;
 
+    // Reset data to its initial state unless it reached a terminal state (Done or Error).
+    // Incomplete data is discarded so reactivation starts with a clean state.
+    if (this.state.data && ![LoadingState.Done, LoadingState.Error].includes(this.state.data.state)) {
```
ifrost, line 433: Just to double check - could this break something when the state is Streaming?

kristinademeshchik:
> @kristinademeshchik how can i reproduce this scenario in the UI?

It's very edge casey but you need to have a panel that uses query results from another panel (Dashboard datasource), and that source panel is out of view and lazy loaded. When the consuming panel re-runs, it briefly unsubscribes and resubscribes to the dashboard datasource (cancelled requests in the screen record) and causes a quick ref drop on the source panel. The source panel gets deactivated but remains in loading state and when it gets re-activated, the queries dont get re-triggered due to loading state.
I have a dashboard spec on hands with repro steps, let me know if you want to test it yourself.

https://github.com/user-attachments/assets/b9f60f91-7ae3-4e78-b87a-b918f5b4847d

packages/scenes/src/querying/SceneQueryRunner.ts
kristinademeshchik: yeah that's a very good question, it actually fixes similar potential problem for streaming as it does for loading. when `_onDeactivate` kills the subscription without resetting `this.state.data`, `shouldRunQueriesOnActivate()` sees existing data and returns `false` -> the panel never reconnects
we only want to preserve terminal states like `done` or `error` (complete results) that are safe to display without an active subscription
In-progress ones like `Loading` or `Streaming` are not meaningful once the subscription is dead

PR #1433, opened 2026-04-23
Variables: Fix options search to be case-insensitive

mdvictor:
This PR turns variable option search to also be case insensitive when searching with special chars. Case insensitivity was already there so for a var option `test` and a search `TEST` it would correctly match, but this did not work well if a char like `-` appeared, so an option `test-value` would not be matched by a search string `TEST-`. This PR fixes that

<details>
  <summary>📦 Published PR as canary version: <code>7.4.2--canary.1433.24826259640.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@7.4.2--canary.1433.24826259640.0
  npm install @grafana/scenes-react@7.4.2--canary.1433.24826259640.0
  # or 
  yarn add @grafana/scenes@7.4.2--canary.1433.24826259640.0
  yarn add @grafana/scenes-react@7.4.2--canary.1433.24826259640.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/filter.ts
+++ b/packages/scenes/src/variables/filter.ts
@@ -1,4 +1,4 @@
-import { SelectableValue } from '@grafana/data';
+import { escapeRegex, SelectableValue } from '@grafana/data';
 import uFuzzy from '@leeoniya/ufuzzy';
 import { VariableValueOption } from './types';
 
@@ -48,10 +48,12 @@ export function fuzzyFind<T extends SelectableValue | VariableValueOption>(
     // if these differ then special chars exist in the input
     keptChars !== inputChars
   ) {
+    const needleRegex = new RegExp(escapeRegex(needle), 'i');
+
     for (let i = 0; i < haystack.length; i++) {
       let item = haystack[i];
 
-      if (item.includes(needle)) {
+      if (needleRegex.test(item)) {
         matches.push(options[i]);
       }
     }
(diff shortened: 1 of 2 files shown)


Sergej-Vlasov (approved):

packages/scenes/src/variables/filter.ts
```diff
@@ -48,10 +48,12 @@ export function fuzzyFind<T extends SelectableValue | VariableValueOption>(
     // if these differ then special chars exist in the input
     keptChars !== inputChars
   ) {
+    const needleRegex = new RegExp(escapeRegex(needle), 'i');
```
grafakus, line 51: Pretty niche, but: 

```suggestion
    const needleRegex = new RegExp(escapeRegex(needle), 'iu');
```

would enable full Unicode case folding — e.g. so that `ß` matches `SS` or Turkish `İ` matches `i`. Not a regression since this path was case-sensitive before, but it would complete the fix for international users.

grafakus (approved):
👍🏾👍🏾 Just left a minor non-blocking (super niche) suggestion.

packages/scenes/src/variables/filter.ts
mdvictor: That's a good call! I just checked this out though and it doesn't seem to cover folding in a lot of cases, at least not in the JS regex. I've checked both examples you've shared in a test code and it sadly does not match `ß` with `SS` for example, neither for the Turkish `İ`. I'll leave it as it is for now and revisit if there is a larger need.

PR #1453, opened 2026-05-14
chore(yarn): disable lifecycle scripts

jackw:
This an org level requirement to use gh actions in a repo.


ashharrison90 (approved):
lgtm 👍

PR #1457, opened 2026-05-20
Chore: Migrate CI to Github App Token Broker

mdvictor:
Migrates `scenes-repo-bot-access-token` gh app to GATB and away from direct vault secrets access


ivanortegaalba (approved):
Thanks, Victor! I checked grafana/grafana and other repos and I see the pattern is the same! 💯

PR #1458, opened 2026-05-21
Prevent panel dragging in view mode

bfmatei:
Stop passing drag hooks to the PanelChrome when the layout is not in a draggable state.

This fixes an issue where the panel would be draggable between rows in view mode.

Steps to reproduce:
1. Have a dashboard with two rows
2. One of the rows has a custom grid
3. Try to drag the panel inside the grid (works fine)
4. Try to drag the panel to the other row (drag works)

This is not reproducible with the auto grid as auto grid exits early from the drag logic. However, the custom grid emits the drag start event and the parent layout starts the dragging sync through the orchestrator.

Another solution to this would be to check if the layout is in a draggable state before triggering the drag sync in the layout orchestrator. However, I consider this to be a cleaner solution as it prevents unnecessary dragging logic from being executed in view mode.

<details>
  <summary>📦 Published PR as canary version: <code>8.2.6--canary.1458.26394321623.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@8.2.6--canary.1458.26394321623.0
  npm install @grafana/scenes-react@8.2.6--canary.1458.26394321623.0
  # or 
  yarn add @grafana/scenes@8.2.6--canary.1458.26394321623.0
  yarn add @grafana/scenes-react@8.2.6--canary.1458.26394321623.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
@@ -392,7 +392,10 @@ function getDragClasses(panel: VizPanel) {
 
 function getDragHooks(panel: VizPanel) {
   const parentLayout = sceneGraph.getLayout(panel);
-  return (parentLayout?.isDraggable?.() && parentLayout?.getDragHooks?.()) ?? {};
+  if (!parentLayout || !parentLayout.isDraggable() || !parentLayout.getDragHooks) {
+    return {};
+  }
+  return parentLayout.getDragHooks() ?? {};
 }
 
 /**
--- a/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
@@ -392,7 +392,7 @@ function getDragClasses(panel: VizPanel) {
 
 function getDragHooks(panel: VizPanel) {
   const parentLayout = sceneGraph.getLayout(panel);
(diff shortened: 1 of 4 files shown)


grafakus (dismissed):
Neat!

packages/scenes/src/components/VizPanel/VizPanelRenderer.tsx
```diff
@@ -392,7 +392,10 @@ function getDragClasses(panel: VizPanel) {
 
 function getDragHooks(panel: VizPanel) {
   const parentLayout = sceneGraph.getLayout(panel);
-  return parentLayout?.getDragHooks?.() ?? {};
+  if (!parentLayout || !parentLayout.isDraggable() || !parentLayout.getDragHooks) {
```
grafakus: Just a doubt, should we also check for itemDraggingDisabled(panel, parentLayout) (same as L386)?

ifrost: I feel like `SceneLayout.getDragHooks()` should internally check for  `isDraggable` and return no hooks so consumers don't need to check it.

bfmatei: Yeah, that's what I tried initially but I got some types errors. I just tried it again and it seems to work. Fixing now.

ifrost (approved):

PR #1538, opened 2026-06-09
fix(AnnotationsDataLayer): pass sampled field names into `arrayToDataFrame`

### LINKED ISSUE #105257: Annotations: end time null in first row breaks annotation
### What happened?

When an annotation query is configured to retrieve both start and end time of the annotation, resulting in an annotated region, a null end time in the first row returned from the query breaks the annotation.
Returning that same row as the second row of the result works perfectly fine, showing that Grafana can handle a null value for the end timestamp.
Screen recording is zipped as there was some problem uploading it to Github without doing so.

[Grafana annotations end time null bug.mov.zip](https://github.com/user-attachments/files/20160037/Grafana.annotations.end.time.null.bug.mov.zip)

### What did you expect to happen?

The same as when the null value is in the second row of the result: all annotation regions get displayed and the annotation with null end time is a simple annotation instead of a region.

### Did this work before?

I have only tested it with Grafana 12 so far.

### How do we reproduce it?

1. Create a dashboard with a graph panel (e.g. Grafana random walk).
2. Create an annotation query returning both start time and end time (e.g. postgres with select * from values).
3. Make the query return null as end time on the first row of its result.
4. Return to the dashboard to see the annotation not working.

Dashboard used in the screen recording (data replaced by Grafana random walk):
[dashboard.json](https://github.com/user-attachments/files/20160120/dashboard.json)

### Is the bug inside a dashboard panel?

No

### Environment (with versions)?

Grafana: Grafana v12.0.0 (4c0e7045f9)
OS: grafana/grafana-oss:latest Docker image (platform: linux/amd64) on MacOS Sequoia 15.4.1 (24E263)
Browser: Firefox/Chrome (bug appears in both)

### Grafana platform?

Docker

### Datasource(s)?

_No response_

gtk-grafana:
Fixes https://github.com/grafana/grafana/issues/105257

Currently only the first annotation defines the data frame field schema, which results in bugs like in https://github.com/grafana/grafana/issues/105257, where queries that contain valid endTime mappings but contain null values for some of the annotations return invalid annotation frames due to `arrayToDataFrame` only looking at the first annotation event passed in from the query.

This fix looks at a handful (250) of non-contiguous annotations instead of just the first in an attempt to reduce the chance of `arrayToDataFrame` omitting fields, without the overhead of iterating through every key in every annotation event.

### Limitations:
This does not address the bug in [public dashboards](https://github.com/grafana/grafana/blob/main/public/app/features/dashboard-scene/scene/DashboardAnnotationsDataLayer.ts#L39), since that path bypasses the `processEvents` method in the scenes package and calls `arrayToDataFrame` directly. 

We can:
1. Revisit this later
2. export the `getLikelyAnnotationEventFieldNames` from scenes for use in grafana/grafana
3. Move `getLikelyAnnotationEventFieldNames` into grafana/data
4. Make a riskier change in `arrayToDataFrame` by generalizing `getLikelyAnnotationEventFieldNames` and populate the names array with a reasonable sample size.

I would recommend 1 or 2 as a short term fix, and discuss 3/4 as a more systemic fix to `arrayToDataFrame` as a longer term solution.

To test: Pull [this dashboard](https://community.grafana.net/d/gaqw7c7/grafana-issues-105257) and set custom scenes version.

<details>
<summary>Repro dashboard JSON</summary>

```json
{
  "annotations": [
    {
      "kind": "AnnotationQuery",
      "spec": {
        "builtIn": true,
        "enable": true,
        "hide": true,
        "iconColor": "rgba(0, 211, 255, 1)",
        "name": "Annotations & Alerts",
        "query": {
          "datasource": {
            "name": "-- Grafana --"
          },
          "group": "grafana",
          "kind": "DataQuery",
          "spec": {},
          "version": "v0"
        }
      }
    },
    {
      "kind": "AnnotationQuery",
      "spec": {
        "enable": true,
        "hide": false,
        "iconColor": "red",
        "mappings": {
          "time": {
            "source": "field",
            "value": "start_time"
          },
          "timeEnd": {
            "source": "field",
            "value": "stop_time"
          },
          "title": {
            "source": "field",
            "value": "value"
          }
        },
        "name": "Buggy annos",
        "query": {
          "datasource": {
            "name": "psql-employees"
          },
          "group": "grafana-postgresql-datasource",
          "kind": "DataQuery",
          "spec": {
            "dataset": "playdb",
            "editorMode": "code",
            "expr": "",
            "format": "table",
            "interval": "",
            "rawQuery": true,
            "rawSql": "SELECT *\nFROM (\n  VALUES\n    -- ('2026-05-23 07:30:00Z'::timestamp, '2026-05-23 08:00:00Z'::timestamp, '10'),\n    ('2026-05-23 08:30:00Z'::timestamp, null, '10'),\n    ('2026-05-23 09:00:00Z'::timestamp, '2026-05-23 09:30:00Z'::timestamp, '20'),\n    ('2026-05-23 09:30:00Z'::timestamp, '2026-05-23 10:00:00Z'::timestamp, '15'),\n    ('2026-05-23 10:00:00Z'::timestamp, '2026-05-23 10:30:00Z'::timestamp, '30')\n) AS mock_data(start_time, stop_time, value)",
            "refId": "Anno",
            "sql": {
              "columns": [
                {
                  "parameters": [],
                  "type": "function"
                }
              ],
              "groupBy": [
                {
                  "property": {
                    "type": "string"
                  },
                  "type": "groupBy"
                }
              ],
              "limit": 50
            }
          },
          "version": "v0"
        }
      }
    },
    {
      "kind": "AnnotationQuery",
      "spec": {
        "enable": false,
        "hide": false,
        "iconColor": "red",
        "name": "Good annos",
        "query": {
          "datasource": {
            "name": "psql-employees"
          },
          "group": "grafana-postgresql-datasource",
          "kind": "DataQuery",
          "spec": {
            "dataset": "playdb",
            "editorMode": "code",
            "expr": "",
            "format": "table",
            "interval": "",
            "rawQuery": true,
            "rawSql": "SELECT *\nFROM (\n  VALUES\n    ('2026-05-23 07:30:00Z'::timestamp, '2026-05-23 08:00:00Z'::timestamp, '10'),\n    ('2026-05-23 08:30:00Z'::timestamp, null, '10'),\n    ('2026-05-23 09:00:00Z'::timestamp, '2026-05-23 09:30:00Z'::timestamp, '20'),\n    ('2026-05-23 09:30:00Z'::timestamp, '2026-05-23 10:00:00Z'::timestamp, '15'),\n    ('2026-05-23 10:00:00Z'::timestamp, '2026-05-23 10:30:00Z'::timestamp, '30')\n) AS mock_data(start_time, stop_time, value)",
            "refId": "Anno",
            "sql": {
              "columns": [
                {
                  "parameters": [],
                  "type": "function"
                }
              ],
              "groupBy": [
                {
                  "property": {
                    "type": "string"
                  },
                  "type": "groupBy"
                }
              ],
              "limit": 50
            }
          },
          "version": "v0"
        }
      }
    },
    {
      "kind": "AnnotationQuery",
      "spec": {
        "enable": false,
        "hide": false,
        "iconColor": "red",
        "mappings": {
          "text": {
            "source": "field",
            "value": "value"
          },
          "time": {
            "source": "field",
            "value": "start_time"
          },
          "timeEnd": {
            "source": "field",
            "value": "stop_time"
          }
        },
        "name": "Annotation workaround",
        "query": {
          "datasource": {
            "name": "psql-employees"
          },
          "group": "grafana-postgresql-datasource",
          "kind": "DataQuery",
          "spec": {
            "dataset": "playdb",
            "editorMode": "code",
            "expr": "",
            "format": "table",
            "interval": "",
            "rawQuery": true,
            "rawSql": "SELECT start_time, stop_time, value\nFROM (\n  SELECT\n    0 AS _ord,\n    '1999-05-22 07:30:00+00'::timestamptz AS start_time,\n    '1999-05-23 08:00:00+00'::timestamptz AS stop_time,\n    '10'::text AS value\n  UNION ALL\n  SELECT *\nFROM (\n  VALUES\n    (1, '2026-05-23 08:30:00Z'::timestamp, null, '10'),\n    (1, '2026-05-23 09:00:00Z'::timestamp, '2026-05-23 09:30:00Z'::timestamp, '20'),\n    (1, '2026-05-23 09:30:00Z'::timestamp, '2026-05-23 10:00:00Z'::timestamp, '15'),\n    (1, '2026-05-23 10:00:00Z'::timestamp, '2026-05-23 10:30:00Z'::timestamp, '30')\n) AS mock_data(ord, start_time, stop_time, value)\n) AS combined\nORDER BY _ord, start_time;",
            "refId": "Anno",
            "sql": {
              "columns": [
                {
                  "parameters": [],
                  "type": "function"
                }
              ],
              "groupBy": [
                {
                  "property": {
                    "type": "string"
                  },
                  "type": "groupBy"
                }
              ],
              "limit": 50
            }
          },
          "version": "v0"
        }
      }
    }
  ],
  "cursorSync": "Off",
  "editable": true,
  "elements": {
    "panel-1": {
      "kind": "Panel",
      "spec": {
        "data": {
          "kind": "QueryGroup",
          "spec": {
            "queries": [
              {
                "kind": "PanelQuery",
                "spec": {
                  "hidden": false,
                  "query": {
                    "datasource": {
                      "name": "9cY0WtPMz"
                    },
                    "group": "grafana-testdata-datasource",
                    "kind": "DataQuery",
                    "spec": {
                      "scenarioId": "random_walk",
                      "seriesCount": 1
                    },
                    "version": "v0"
                  },
                  "refId": "A"
                }
              }
            ],
            "queryOptions": {
              "maxDataPoints": 10
            },
            "transformations": []
          }
        },
        "description": "",
        "id": 1,
        "links": [],
        "title": "New panel",
        "vizConfig": {
          "group": "timeseries",
          "kind": "VizConfig",
          "spec": {
            "fieldConfig": {
              "defaults": {
                "color": {
                  "mode": "palette-classic"
                },
                "custom": {
                  "axisBorderShow": false,
                  "axisCenteredZero": false,
                  "axisColorMode": "text",
                  "axisLabel": "",
                  "axisPlacement": "auto",
                  "barAlignment": 0,
                  "barWidthFactor": 0.6,
                  "drawStyle": "line",
                  "fillOpacity": 0,
                  "gradientMode": "none",
                  "hideFrom": {
                    "legend": false,
                    "tooltip": false,
                    "viz": false
                  },
                  "insertNulls": false,
                  "lineInterpolation": "linear",
                  "lineWidth": 1,
                  "pointSize": 5,
                  "scaleDistribution": {
                    "type": "linear"
                  },
                  "showPoints": "auto",
                  "showValues": false,
                  "spanNulls": false,
                  "stacking": {
                    "group": "A",
                    "mode": "none"
                  },
                  "thresholdsStyle": {
                    "mode": "off"
                  }
                },
                "thresholds": {
                  "mode": "absolute",
                  "steps": [
                    {
                      "color": "green",
                      "value": 0
                    },
                    {
                      "color": "red",
                      "value": 80
                    }
                  ]
                }
              },
              "overrides": []
            },
            "options": {
              "annotations": {
                "clustering": -1,
                "lines": {
                  "width": 2
                },
                "multiLane": false,
                "regions": {
                  "opacity": 0.1
                }
              },
              "legend": {
                "calcs": [],
                "displayMode": "list",
                "enableFacetedFilter": false,
                "placement": "bottom",
                "showLegend": true
              },
              "tooltip": {
                "hideZeros": false,
                "mode": "single",
                "sort": "none"
              }
            }
          },
          "version": "13.1.0-26978683508"
        }
      }
    }
  },
  "layout": {
    "kind": "GridLayout",
    "spec": {
      "items": [
        {
          "kind": "GridLayoutItem",
          "spec": {
            "element": {
              "kind": "ElementReference",
              "name": "panel-1"
            },
            "height": 8,
            "width": 12,
            "x": 0,
            "y": 0
          }
        }
      ]
    }
  },
  "links": [],
  "liveNow": false,
  "preferences": {
    "layout": {
      "kind": "GridLayout",
      "spec": {
        "items": []
      }
    }
  },
  "preload": false,
  "tags": [],
  "timeSettings": {
    "autoRefresh": "",
    "autoRefreshIntervals": [
      "5s",
      "10s",
      "30s",
      "1m",
      "5m",
      "15m",
      "30m",
      "1h",
      "2h",
      "1d"
    ],
    "fiscalYearStartMonth": 0,
    "from": "2026-05-23T05:22:37.306Z",
    "hideTimepicker": false,
    "timezone": "browser",
    "to": "2026-05-23T17:57:41.125Z"
  },
  "title": "grafana/issues/105257",
  "variables": []
}
```

</details>

https://community.grafana.net/d/gaqw7c7/grafana-issues-105257

<details>
  <summary>📦 Published PR as canary version: <code>8.5.1--canary.1538.28054709060.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@8.5.1--canary.1538.28054709060.0
  npm install scenes-app@8.5.1--canary.1538.28054709060.0
  npm install @grafana/scenes-react@8.5.1--canary.1538.28054709060.0
  # or 
  yarn add @grafana/scenes@8.5.1--canary.1538.28054709060.0
  yarn add scenes-app@8.5.1--canary.1538.28054709060.0
  yarn add @grafana/scenes-react@8.5.1--canary.1538.28054709060.0
  ```
</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -997,7 +997,7 @@ Please migrate to the [`ignore`](https://redirect.github.com/SuperchupuDev/tinyg
 - `docusaurus-plugin-content-blog`
   - [#&#8203;11564](https://redirect.github.com/facebook/docusaurus/pull/11564) test(blog): Add basic tests for blog routes. ([@&#8203;slorber](https://redirect.github.com/slorber))
 
-##### :globe\_with\_meridians: Translations
+##### :globe_with_meridians: Translations
 
 - `docusaurus-theme-translations`
   - [#&#8203;11632](https://redirect.github.com/facebook/docusaurus/pull/11632) feat(i18n): add Urdu (ur) default theme translations ([@&#8203;hammadurrehman2006](https://redirect.github.com/hammadurrehman2006))
@@ -1073,7 +1073,7 @@ Please migrate to the [`ignore`](https://redirect.github.com/SuperchupuDev/tinyg
 - `docusaurus-theme-mermaid`
   - [#&#8203;11438](https://redirect.github.com/facebook/docusaurus/pull/11438) perf(theme-mermaid): lazy load the Mermaid library ([@&#8203;slorber](https://redirect.github.com/slorber))
 
-##### :nail\_care: Polish
+##### :nail_care: Polish
 
 - `docusaurus-theme-classic`
   - [#&#8203;11463](https://redirect.github.com/facebook/docusaurus/pull/11463) fix(theme): remove "Edit this page" button from print view ([@&#8203;richk21](https://redirect.github.com/richk21))
@@ -1085,7 +1085,7 @@ Please migrate to the [`ignore`](https://redirect.github.com/SuperchupuDev/tinyg
 - [#&#8203;11481](https://redirect.github.com/facebook/docusaurus/pull/11481) chore(deps): bump actions/dependency-review-action from 4.8.0 to 4.8.1 ([@&#8203;dependabot\[bot\]](https://redirect.github.com/apps/dependabot))
 - [#&#8203;11446](https://redirect.github.com/facebook/docusaurus/pull/11446) chore(deps): bump actions/dependency-review-action from 4.7.3 to 4.8.0 ([@&#8203;dependabot\[bot\]](https://redirect.github.com/apps/dependabot))
 
-##### :globe\_with\_meridians: Translations
+##### :globe_with_meridians: Translations
 
 - `docusaurus-theme-translations`
   - [#&#8203;11484](https://redirect.github.com/facebook/docusaurus/pull/11484) fix(translations): improve Arabic theme translations ([@&#8203;maysara-elshewehy](https://redirect.github.com/maysara-elshewehy))
@@ -1207,7 +1207,7 @@ Please migrate to the [`ignore`](https://redirect.github.com/SuperchupuDev/tinyg
 - Other
   - [#&#8203;11287](https://redirect.github.com/facebook/docusaurus/pull/11287) chore(website): split changelog per version + adjust changelog plugin implementation ([@&#8203;slorber](https://redirect.github.com/slorber))
 
-##### :globe\_with\_meridians: Translations
+##### :globe_with_meridians: Translations
 
 - `docusaurus-theme-translations`
   - [#&#8203;11315](https://redirect.github.com/facebook/docusaurus/pull/11315) fix(theme-translations): Add missing Portuguese (pt-BR) theme translations and improve some of it. ([@&#8203;marcelocell](https://redirect.github.com/marcelocell))
@@ -1298,9 +1298,8 @@ Please migrate to the [`ignore`](https://redirect.github.com/SuperchupuDev/tinyg
   - [#&#8203;10953](https://redirect.github.com/facebook/docusaurus/pull/10953) feat(plugin-npm2yarn): Add Bun to default tabs conversions ([@&#8203;jakeboone02](https://redirect.github.c
 
 > ✂ **Note**
-> 
+>
 > PR body was truncated to here.
-
 
 </details>
 
@@ -2080,7 +2079,6 @@ A bad version was accidentally published and [it can't be unpublished](https://r
 ##### Allow multiple decorators on same getter/setter ([#&#8203;14584](https://redirect.github.com/prettier/prettier/pull/14584) by [@&#8203;fisker](https://redirect.github.com/fisker))
 
 <!-- prettier-ignore -->
-
 ```ts
 // Input
 class A {
@@ -2119,7 +2117,6 @@ class A {
 ##### Allow decorators on private members and class expressions ([#&#8203;14548](https://redirect.github.com/prettier/prettier/pull/14548) by [@&#8203;fisker](https://redirect.github.com/fisker))
 
 <!-- prettier-ignore -->
-
 ```ts
 // Input
 class A {
@@ -2158,7 +2155,6 @@ TypeScript 5.0 introduces two new syntactic features:
 ##### Add missing parentheses for decorator ([#&#8203;14393](https://redirect.github.com/prettier/prettier/pull/14393) by [@&#8203;fisker](https://redirect.github.com/fisker))
 
 <!-- prettier-ignore -->
-
 ```jsx
 // Input
 class Person {
@@ -2182,7 +2178,6 @@ class Person {
 ##### Add parentheses for `TypeofTypeAnnotation` to improve readability ([#&#8203;14458](https://redirect.github.com/prettier/prettier/pull/14458) by [@&#8203;fisker](https://redirect.github.com/fisker))
 
 <!-- prettier-ignore -->
-
 ```tsx
 // Input
 type A = (typeof node.children)[];
@@ -2201,7 +2196,6 @@ it will be interpreted as a `printWidth` of `Infinity` rather than being ignored
 (which previously resulted in the default `printWidth` of 80 being applied, if not overridden by Prettier-specific configuration).
 
 <!-- prettier-ignore -->
-
 ```html
 <!-- Input -->
 <div className='HelloWorld' title={`You are visitor number ${ num }`} onMouseOver={onMouseOver}/>
@@ -2226,7 +2220,6 @@ it will be interpreted as a `printWidth` of `Infinity` rather than being ignored
 ##### Fix leading comments in mapped types with `readonly` ([#&#8203;13427](https://redirect.github.com/prettier/prettier/pull/13427) by [@&#8203;thorn0](https://redirect.github.com/thorn0), [@&#8203;sosukesuzuki](https://redirect.github.com/sosukesuzuki))
 
 <!-- prettier-ignore -->
-
 ```tsx
 // Input
 type Type = {
@@ -2255,7 +2248,6 @@ unless the length of params exceeds the print width. This also updates the else 
 same as opening blocks.
 
 <!-- prettier-ignore -->
-
 ```hbs
 {{! Input }}
 {{#block param param param param param param param param param param as |blockParam|}}
@@ -2313,7 +2305,6 @@ In [Sapling SCM](https://sapling-scm.com/), `.sl/` is the folder where it stores
 ##### Recognize `@satisfies` in Closure-style type casts ([#&#8203;14262](https://redirect.github.com/prettier/prettier/pull/14262) by [@&#8203;fisker](https://redirect.github.com/fisker))
 
 <!-- prettier-ignore -->
-
 ```jsx
 // Input
 const a = /** @&#8203;satisfies {Record<string, string>} */ ({hello: 1337});
@@ -2331,7 +2322,6 @@ const b = /** @&#8203;type {Record<string, string>} */ ({hello: 1337});
 ##### Fix parens in inferred function return types with `extends` ([#&#8203;14279](https://redirect.github.com/prettier/prettier/pull/14279) by [@&#8203;fisker](https://redirect.github.com/fisker))
 
 <!-- prettier-ignore -->
-
 ```ts
 // Input
 type Foo<T> = T extends ((a) => a is infer R extends string) ? R : never;
@@ -2357,7 +2347,6 @@ type Foo<T> = T extends ((a) => a is infer R extends string) ? R : never;
 See [Angular v15.1.0 release note](https://redirect.github.com/angular/angular/releases/tag/15.1.0) for details.
 
 <!-- prettier-ignore -->
-
 ```html
 // Input
 <app-test/>
@@ -2381,7 +2370,6 @@ SyntaxError: Only void and foreign elements can be self closed "app-test" (1:1)
 ##### Don't lowercase link references ([#&#8203;13155](https://redirect.github.com/prettier/prettier/pull/13155) by [@&#8203;DerekNonGeneric](https://redirect.github.com/DerekNonGeneric) & [@&#8203;fisker](https://redirect.github.com/fisker))
 
 <!-- prettier-ignore -->
-
 ```markdown
 <!-- Input -->
 We now don't strictly follow the release notes format suggested by [Keep a Changelog].
@@ -2403,7 +2391,6 @@ We now don't strictly follow the release notes format suggested by [Keep a Chang
 ##### Preserve self-closing tags ([#&#8203;13691](https://redirect.github.com/prettier/prettier/pull/13691) by [@&#8203;dcyriller](https://redirect.github.com/dcyriller))
 
 <!-- prettier-ignore -->
-
 ```hbs
 {{! Input }}
 <div />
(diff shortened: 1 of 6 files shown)


packages/scenes/CHANGELOG.md
```diff
@@ -666,7 +666,6 @@ A bad version was accidentally published and [it can't be unpublished](https://r
 ##### Allow multiple decorators on same getter/setter ([#&#8203;14584](https://redirect.github.com/prettier/prettier/pull/14584) by [@&#8203;fisker](https://redirect.github.com/fisker))
 
 <!-- prettier-ignore -->
-
```
gtk-grafana: looks like these prettier-ignores are no longer working which is causing CI to fail

CHANGELOG.md
```diff
@@ -997,7 +997,7 @@ Please migrate to the [`ignore`](https://redirect.github.com/SuperchupuDev/tinyg
 - `docusaurus-plugin-content-blog`
   - [#&#8203;11564](https://redirect.github.com/facebook/docusaurus/pull/11564) test(blog): Add basic tests for blog routes. ([@&#8203;slorber](https://redirect.github.com/slorber))
 
-##### :globe\_with\_meridians: Translations
```
gtk-grafana: more prettier fallout 😢

packages/scenes-react/CHANGELOG.md
```diff
@@ -421,15 +421,15 @@ Read the [React 19 Upgrade Guide](https://react.dev/blog/2024/04/25/react-19-upg
 
 ### [`v9.2.0`](https://redirect.github.com/rollup/plugins/blob/HEAD/packages/eslint/CHANGELOG.md#v920)
 
-*2025-10-28*
+_2025-10-28_
```
gtk-grafana, line 424: looks like prettier changed without committing the changes on main, and CI is failing unless I include these changes on this branch

packages/scenes/src/querying/layers/annotations/utils.ts
```diff
@@ -90,3 +90,39 @@ export function dedupAnnotations(annotations: any) {
...
+
+/**
+ * getLikelyAnnotationEventFieldNames tests a few non-contiguous annotation events to get a more complete set of field
+ * names to pass into arrayToDataFrame without iterating through the entire array.
+ * Avoids the first annotation defining the schema of all subsequent annotations. (https://github.com/grafana/grafana/issues/105257)
+ *
+ * @param annotationEvents
+ * @param samples
+ */
+export function getLikelyAnnotationEventFieldNames(annotationEvents: AnnotationEvent[], samples = 20): string[] {
```
gtk-grafana: The runtime complexity of this might not warrant the potential bugginess of sampling, 2000 annotations will run in ~1ms on a 2024 M4 Pro, which is a fraction of the cost of just rendering the annotations.

2000 samples: ~1ms worst case*
1000 samples: ~0.68ms worst case
500 samples: ~0.14ms worst case
100 samples: ~0.05ms worst case
20 samples: ~0.03ms worst case

*Testing behavior validated locally via github datasource and dashboard mapping, I believe alerting and IRM use-cases have more keys so the base cost of Object.keys will be larger.

### Arguments for removing sampling:
We'd probably be fine running this unbounded assuming all of the annotations are rendered in the panel, as that would crash the browser before the overhead here would become noticeable in isolation.

### Arguments for increasing sample size:
This is just one task of many, and annotations can be returned outside of the range of the panel, so out of an abundance of caution we should retain the sampling, but bump up the sample count to 500, which should check every annotation for the vast majority of dashboards, but limit worst case behavior. Using the current sample size of 20 is probably sufficient in most cases, but the runtime difference between 20 and 500 isn't enough to risk needing to come back in here and increase the count when another user runs into this bug.

### Arguments against bumping up the sampling:
Due to the lack of reporting of this bug, we can assume that this is a pretty rare edge-case, and grabbing the first annotation worked well enough until now (citation needed), so 20 is probably fine to catch the vast majority of cases.

Image

gtk-grafana: I decided to bump the sample size from 20 to 250 for PR review

packages/scenes/src/querying/layers/annotations/utils.ts
```diff
@@ -90,3 +90,39 @@ export function dedupAnnotations(annotations: any) {
...
+
+/**
+ * getLikelyAnnotationEventFieldNames tests a few non-contiguous annotation events to get a more complete set of field
+ * names to pass into arrayToDataFrame without iterating through the entire array.
+ * Avoids the first annotation defining the schema of all subsequent annotations. (https://github.com/grafana/grafana/issues/105257)
+ *
+ * @param annotationEvents
+ * @param samples
+ */
+export function getLikelyAnnotationEventFieldNames(annotationEvents: AnnotationEvent[], samples = 250): string[] {
```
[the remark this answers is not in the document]
gtk-grafana: updated pr decription

fastfrwrd, line 102: What motivated 250? in the example given in that ticket, you'd even have gotten away with 2 technically, of course that's a very limited case tho. Overall the only part of the method that's concerning is the `Object.keys` and then looping over the keys of each annotation event, but maybe we should dial it down to like... 50 at first? Or we would want to make this number configurable by grafana, but that seems pretty hard. One other option would be to make this a dynamic relationship with the length of annotationEvents, like `Math.floor(annotationEvents.length / 2)` or something, but that's a bit tortured too. Curious to hear your thoughts!

packages/scenes/src/querying/layers/annotations/utils.ts
```diff
@@ -90,3 +90,39 @@ export function dedupAnnotations(annotations: any) {
...
+
+  const annotationEventNamesSet = new Set<string>();
+  for (let annotationEventIndex = firstIdx; annotationEventIndex < len; annotationEventIndex += stride) {
+    const annotationEvent = annotationEvents[annotationEventIndex];
+    const keys = Object.keys(annotationEvent);
+    for (let i = 0; i < keys.length; i++) {
+      if (!annotationEventNamesSet.has(keys[i])) {
+        annotationEventNamesSet.add(keys[i]);
+      }
+    }
```
fastfrwrd: the Set.prototype.has check is unnecessary, and we probably need to be as efficient as we can in this method:


```suggestion
    for (const key of Object.keys(annotationEvent)) {
      annotationEventNamesSet.add(key);
    }
```

packages/scenes/src/querying/layers/annotations/utils.ts
gtk-grafana: Here was my internal struggle with choosing the length: https://github.com/grafana/scenes/pull/1538#discussion_r3388411504

100 took about 1/20th of a ms
500 took about 1/7th of a ms worst case
I figured 100 or 500 would cover all annotations in most cases, and both had insignificant impact on runtime so I split the difference.
But I doubt it matters, we could check 20 or all of them and have the same impact for 99.999% of cases.

fastfrwrd: that checks out. 1/10th of a ms on your laptop would maybe translate to like 2-4ms on the trashiest laptop in 2026, so that's very reasonable to me. :shipit:

fastfrwrd (approved):

PR #1551, opened 2026-06-18
VizPanel: Increment renderCounter on data-driven re-renders

ivanortegaalba:
## What

`VizPanel` re-renders whenever its data changes (the renderer subscribes to data via `dataObject.useState()`), but it does not bump the `renderCounter` prop on those re-renders. This bumps `renderCounter` when a new data request settles.

## Why

Panel plugins that imperatively mutate their own DOM after render (e.g. text/HTML panels running a post-render script) re-apply those mutations whenever `renderCounter` changes. So when the component re-renders because its data changed, the plugin must receive a new `renderCounter`. Otherwise the re-render resets the plugin's DOM and it is never re-applied, so the scripted content disappears (for example when a query-type variable finishes loading and the panel's query re-runs).

`renderCounter` was hardcoded to `0` until https://github.com/grafana/scenes/pull/934, which introduced `_renderCounter` and bumped it only on options changes (https://github.com/grafana/scenes/pull/954 later added `forceRender` for referenced-variable and time-range changes). Data-driven re-renders were never covered, so a re-render caused purely by new data still passes an unchanged `renderCounter`.

## Fix

Track the data request id in the renderer and add a counter, bumped on each new request, to the value passed to the plugin:

```tsx
renderCounter={_renderCounter + dataRenderCounter}
```

The first request after mount is skipped (initial mount already renders once); keying on `requestId` means streaming frames within the same request do not over-trigger.

## Test

`VizPanel > panel rendering > data plugin > Should increment renderCounter only when a new data request settles`: a same-request state change leaves `renderCounter` unchanged, a new request increments it.

Placement note: this lives in the renderer so it resolves the same data object the renderer already displays and can schedule a re-render via `useState`. Happy to move it model-side (symmetric to the existing `skipDataQuery` time-range `forceRender`) if preferred.

## Affected plugins (rely on `renderCounter`)

Core panels that read the prop today, so they benefit from the contract being restored on data-driven re-renders:

- [`annolist`](https://github.com/grafana/grafana/blob/main/public/app/plugins/panel/annolist/AnnoListPanel.tsx) — re-fetches annotations in `componentDidUpdate` when `renderCounter` changes
- [`dashlist`](https://github.com/grafana/grafana/blob/main/public/app/plugins/panel/dashlist/DashList.tsx) — re-fetches dashboards via `useThrottle(props.renderCounter, 5000)`
- [`stat`](https://github.com/grafana/grafana/blob/main/public/app/plugins/panel/stat/StatPanel.tsx), [`gauge`](https://github.com/grafana/grafana/blob/main/public/app/plugins/panel/gauge/GaugePanel.tsx), [`bargauge`](https://github.com/grafana/grafana/blob/main/public/app/plugins/panel/bargauge/BarGaugePanel.tsx) — forward it to [`VizRepeater`](https://github.com/grafana/grafana/blob/main/packages/grafana-ui/src/components/VizRepeater/VizRepeater.tsx), whose `shouldComponentUpdate` re-renders on `renderCounter` change
- [`logstable`](https://github.com/grafana/grafana/blob/main/public/app/plugins/panel/logstable/LogsTable.tsx) — forwards it to the table

Note on the fetch-triggering panels: `annolist` and `dashlist` run no datasource queries, so the query runner takes the no-queries path and produces no `request`. They never receive a new `requestId`, so this change does not bump their counter and does not introduce extra fetches; only query-running panels (and third-party DOM-mutating panels) get the corrected behavior.

### DIFF
--- a/packages/scenes/src/components/VizPanel/VizPanel.test.tsx
+++ b/packages/scenes/src/components/VizPanel/VizPanel.test.tsx
@@ -15,6 +15,7 @@ import {
   PanelProps,
   toUtc,
   DataTransformerConfig,
+  DataQueryRequest,
 } from '@grafana/data';
 import * as grafanaData from '@grafana/data';
 import { getPanelPlugin } from '../../../utils/test/__mocks__/pluginMocks';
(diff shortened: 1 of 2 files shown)


ivanortegaalba:
Closing this. The rendering issue that motivated the change is better fixed in the affected panel plugin itself: that plugin does not consume `renderCounter`, so bumping it here would not fix the reported behavior. If we later want the data-driven `renderCounter` bump as a standalone improvement for plugins that do rely on the contract (annolist, dashlist, VizRepeater-based panels), we can revisit on its own merits.

PR #1554, opened 2026-06-22
SceneQueryRunner: convert forkJoin to combineLatest 

### LINKED ISSUE #126804: TimeComparison: Loading state
Time comparison currently breaks the LoadingState for any panel where the feature is enabled, so no loading UI is presented to the user while either the primary or compare query is loading.

Time comparison query does not currently set the `PanelData` `LoadingState`, so panels are can be in the "Done" state when the time compare query is still running, and TimeCompare queries that return errors are not presented in the UI.

The current behavior is a poor UX as users do not know if they should expect that the time compare query will return, and do not have any way within the Grafana UI to debug queries that error.

Requirements:
* Loading state is "Loading" until all queries complete successfully.
* Loading state is "Error" if either query returns an error.

Maybe outside the scope of this fix, but as a stretch goal we should check support for displaying multiple query errors with https://github.com/grafana/grafana/pull/126750.

gtk-grafana:
Fixes https://github.com/grafana/grafana/issues/126804

### Problem
Requests with secondary queries do not currently emit loading events. 
Requests with secondary queries that return errors do not show error states in the panel.

### Details
The problem exists as a result of how [forkJoin](https://rxjs.dev/api/index/function/forkJoin) operates. forkJoin takes many observables and only emits when all subscribed observables complete, in Grafana this means that besides first load after the browser loads a panel, there is no loading feedback for any query run in a panel (i.e. on time range or query change), and secondary queries that error out do not display any error state in the panel.

The secondary query functionality was introduced in https://github.com/grafana/scenes/pull/587 as a somewhat general interface, but developed primarily to support query time comparison functionality. 

### Solution
[combineLatest](https://rxjs.dev/api/index/function/combineLatest) takes many observables and returns an observable that emits values from the latest value received in any of the observables, this gives us the ability to merge states across multiple requests which is required in order to utilize loading states and to display errors in secondary queries.

### Concerns
Grafana's `runWithTimeRange` usage of secondary queries are not well documented, and explicit test coverage is thin for `extraQueryProcessingOperator` so there is some regression risk, although it looks as though time range comparison is the only consumer of this functionality in core Grafana so regression risk should be limited to that feature.

#### References:
Comparison of [forkJoin and combineLatest](https://shuji-bonji.github.io/RxJS-with-TypeScript/en/guide/creation-functions/combination/forkJoin-vs-combineLatest)

[rxjs forkJoin doc](https://rxjs.dev/api/index/function/forkJoin)
[rxjs combineLatest doc](https://rxjs.dev/api/index/function/combineLatest)

#### Related PRs:
Original PR that introduced secondary queries for time range comparisons: https://github.com/grafana/scenes/pull/587

#### To test:
Grab the scenes canary info below and update your local grafana package.json and run `yarn install && yarn start`.
You can use the "Flaky query" test data source scenario to verify loading state and errors on the primary or secondary query are displayed in the UI.

<details>
  <summary>📦 Published PR as canary version: <code>8.5.1--canary.1554.27979683361.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@8.5.1--canary.1554.27979683361.0
  npm install scenes-app@8.5.1--canary.1554.27979683361.0
  npm install @grafana/scenes-react@8.5.1--canary.1554.27979683361.0
  # or 
  yarn add @grafana/scenes@8.5.1--canary.1554.27979683361.0
  yarn add scenes-app@8.5.1--canary.1554.27979683361.0
  yarn add @grafana/scenes-react@8.5.1--canary.1554.27979683361.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.ts
@@ -1,5 +1,5 @@
 import { cloneDeep, isEqual } from 'lodash';
-import { forkJoin, ReplaySubject, Unsubscribable } from 'rxjs';
+import { combineLatest, ReplaySubject, Unsubscribable } from 'rxjs';
 
 import { DataQuery, DataSourceRef, LoadingState } from '@grafana/schema';
 
@@ -18,7 +18,7 @@ import {
 
 // TODO: Remove this ignore annotation when the grafana runtime dependency has been updated
 // @ts-ignore
-import { getRunRequest, toDataQueryError, isExpressionReference, config } from '@grafana/runtime';
+import { config, getRunRequest, isExpressionReference, toDataQueryError } from '@grafana/runtime';
 
 import { SceneObjectBase } from '../core/SceneObjectBase';
 import { sceneGraph } from '../core/sceneGraph';
@@ -36,11 +36,11 @@ import { writeSceneLog } from '../utils/writeSceneLog';
 import { VariableValueRecorder } from '../variables/VariableValueRecorder';
 import { emptyPanelData } from '../core/SceneDataNode';
 import { getClosest } from '../core/sceneGraph/utils';
-import { isExtraQueryProvider, ExtraQueryDataProcessor, ExtraQueryProvider } from './ExtraQueryProvider';
-import { passthroughProcessor, extraQueryProcessingOperator } from './extraQueryProcessingOperator';
+import { ExtraQueryDataProcessor, ExtraQueryProvider, isExtraQueryProvider } from './ExtraQueryProvider';
+import { extraQueryProcessingOperator, passthroughProcessor } from './extraQueryProcessingOperator';
 import { filterAnnotations } from './layers/annotations/filterAnnotations';
 import { getEnrichedDataRequest } from './getEnrichedDataRequest';
-import { registerQueryWithController, QueryProfilerLike } from './registerQueryWithController';
+import { QueryProfilerLike, registerQueryWithController } from './registerQueryWithController';
 import { findPanelProfiler } from '../utils/findPanelProfiler';
 import { AdHocFiltersVariable } from '../variables/adhoc/AdHocFiltersVariable';
 import { GroupByVariable } from '../variables/groupby/GroupByVariable';
@@ -574,7 +574,9 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
         // extra request providers.
         const op = extraQueryProcessingOperator(processors);
         // Combine the primary and secondary streams into a single stream, and apply the operator.
-        stream = forkJoin([stream, ...secondaryStreams]).pipe(op);
+        // Use combineLatest so intermediate loading/error states from the primary and secondary requests are emitted as
+        // they arrive, instead of only a single combined emission once every request has completed.
+        stream = combineLatest([stream, ...secondaryStreams]).pipe(op);
       }
 
       const panelProfiler: QueryProfilerLike | undefined = findPanelProfiler(this);
--- a/packages/scenes/src/querying/extraQueryProcessingOperator.ts
+++ b/packages/scenes/src/querying/extraQueryProcessingOperator.ts
@@ -13,7 +13,7 @@ export const passthroughProcessor: ExtraQueryDataProcessor = (_, secondary) => o
  * otherwise return the first/primary response state
  * @param responseState
  */
-function combineLoadingStates(responseState: LoadingState[]): LoadingState {
+export function combineLoadingStates(responseState: LoadingState[]): LoadingState {
   if (responseState.some((s) => s === LoadingState.Error)) {
     return LoadingState.Error;
   }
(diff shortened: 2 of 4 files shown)


packages/scenes/src/querying/extraQueryProcessingOperator.ts
```diff
@@ -1,32 +1,72 @@
...
+export function combineLoadingStates(responseState: LoadingState[]): LoadingState {
+  if (responseState.some((s) => s === LoadingState.Error)) {
+    return LoadingState.Error;
+  }
+  if (responseState.some((s) => s === LoadingState.Loading)) {
+    return LoadingState.Loading;
+  }
+  if (responseState.some((s) => s === LoadingState.Streaming)) {
+    return LoadingState.Streaming;
+  }
```
fastfrwrd, line 25: not sure if this is done in any sort of performance-sensitive spot, and I don't know how many entries could reasonably be in `responseState` but it it pretty simple to optimize this into a single loop instead of three without changing anything about the readability (imo):

```suggestion
  for (const responseState of responseStates) {
    if (responseState === LoadingState.Error) {
      return LoadingState.Error;
    }
    if (responseState === LoadingState.Loading) {
      return LoadingState.Loading;
    }
    if (responseState === LoadingState.Streaming) {
      return LoadingState.Streaming;
    }
  }
```

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -574,7 +574,9 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
         // extra request providers.
         const op = extraQueryProcessingOperator(processors);
         // Combine the primary and secondary streams into a single stream, and apply the operator.
-        stream = forkJoin([stream, ...secondaryStreams]).pipe(op);
+        // Use combineLatest so intermediate loading/error states from the primary and secondary requests are emitted as
+        // they arrive, instead of only a single combined emission once every request has completed.
+        stream = combineLatest([stream, ...secondaryStreams]).pipe(op);
```
fastfrwrd, line 579: makes sense!

packages/scenes/src/querying/extraQueryProcessingOperator.ts
```diff
@@ -1,32 +1,72 @@
...
+ * responses so that intermediate states are reflected while requests are in flight.
+ * @param panelDataProcessors
+ */
+type RequestId = string;
 export const extraQueryProcessingOperator =
-  (processors: Map<string, ExtraQueryDataProcessor>) => (data: Observable<[PanelData, ...PanelData[]]>) => {
+  (panelDataProcessors: Map<RequestId, ExtraQueryDataProcessor>) => (data: Observable<[PanelData, ...PanelData[]]>) => {
+    // combineLatest re-emits whenever any source updates, so the same secondary response object can be seen multiple times.
+    // Processors may mutate their input and are not guaranteed to be idempotent, so we cache the processed output by source identity and only run each processor once per secondary response.
+    const processedCache = new WeakMap<PanelData, PanelData>();
```
fastfrwrd, line 44: traced through scenes to try to figure out the implications of this cache. It should be fine, because the operator only exists for the life of a single call to run query/queries, so the cache will be reconstructed the next time the query is run.

packages/scenes/src/querying/extraQueryProcessingOperator.ts
```diff
@@ -1,32 +1,72 @@
...
+        const allResponseErrors = allResponses.flatMap((d) => d.errors ?? []);
+        return {
+          ...processedPrimary,
+          state: combineLoadingStates(allResponses.map((d) => d.state)),
+          series: [...processedPrimary.series, ...processedSecondaries.flatMap((s) => s.series)],
+          annotations: [
+            ...(processedPrimary.annotations ?? []),
+            ...processedSecondaries.flatMap((s) => s.annotations ?? []),
+          ],
+          ...(allResponseErrors.length > 0 ? { errors: allResponseErrors, error: allResponseErrors[0] } : {}),
```
fastfrwrd, line 68: I mean, the performance implications of a lot of the spread and flatMaps in here scare _me_, given what would probably happen with the amount of object and arrays we'd construct just to throw them away immediately. but it's kind of none of my business, and it matches what was already in here. If it works it works!

fastfrwrd (approved):
This looks good to me and the tests make sense and exercise what you added in my opinion.

packages/scenes/src/querying/extraQueryProcessingOperator.ts
gtk-grafana: @fastfrwrd that's different behavior though, that would return Error if the first request is error, Loading if first request is loading, etc. We could optimize this but we'd need to put all of the loading states into a set and then  check for existence of Error, Loading, Streaming.

gtk-grafana: The size of the array of response loading states is essentially constant, I don't think it can currently be more then 2, and even if it goes as high as 10 (which is higher then the concurrent request limit in grafana frontend) the performance benefit is negligible, so it's mostly a style thing. The set solution is probably just as legible so I'm ambivalent either way 🤷

fastfrwrd: ah of course. if we feel confident in those parameters this is fine.

drew08t (approved):
Tested this thoroughly for time comparison and it is working as expected! Great job. 🚀

PR #1564, opened 2026-07-01
Variables: Block dependent variables and panels when a variable errors or resolves empty

nachopv:
## The problem

When a dashboard template variable fails to update — its datasource query errors, times out, or (very commonly) the datasource swallows the failure and returns an **empty** result — scenes today still cascades as if the variable succeeded:

- chained variables that depend on the failed variable are re-evaluated, and
- panels that reference the failed variable (or its dependents) re-run their queries.

With the failed variable interpolating to an empty string, this produces **unscoped backend queries**. For example, a panel query

```
kube_pod_container_resource_limits{resource="cpu",pod=~"$pod.*"}
```

collapses to

```
kube_pod_container_resource_limits{resource="cpu",pod=~".*"}
```

when `$pod` fails to resolve — matching **every** pod instead of one. Such queries can be extremely expensive and can hammer the datasource at the exact moment it is already unhealthy.

Related issue: grafana/grafana#127689

## What this PR does

Adds two **opt-in, default-off** options to `SceneVariableSet` state:

| Option | Effect |
| --- | --- |
| `blockDependentsOnError` | When a variable is in an error state, its transitive dependent variables and any panels referencing the involved variables are **not** refreshed. They keep their previous (stale) value/data and issue no query. |
| `treatEmptyAsError` | Only meaningful together with the above. A variable that resolves to an empty/null value is also treated as an error. This is required for real-world datasources (e.g. Prometheus `label_values`) that catch their own request failures and return empty results with `LoadingState.Done` rather than propagating an error. |

Both default to `false`, so **there is no behavior change** for existing consumers; the feature is fully opt-in by the embedding application.

### Implementation

- `SceneVariableSet`
  - Tracks errored variables (`_erroredVariables`); a variable that errors, or (with `treatEmptyAsError`) completes empty, is recorded.
  - `_updateNextBatch` guards dependents: if a dependency is in an error state the update is skipped, `loading` is cleared and the stale value is kept (no perpetual spinner).
  - `isVariableInErrorState` / `_isVariableConsideredEmpty` helpers. `_isVariableConsideredEmpty` also treats an "Include All" variable that returned no options as empty (it resolves to the `$__all` token but expands to nothing), and errored variables are re-validated on re-activation so their state is not lost across a deactivate/re-activate cycle.
- `sceneGraph.hasVariableDependencyInErrorState` + `VariableDependencyConfig.hasDependencyInErrorState` walk the dependency graph to detect error state transitively.
- `SceneQueryRunner` skips query execution when a variable dependency is in an error state, preserving the panel's previous data.

## How it was tested

- **Unit tests**
  - `SceneVariableSet.test.tsx` — error and empty-as-error cases over A→B→C chains, asserting dependents are skipped, stale values kept, and recovery works. Includes corner cases: an "Include All" variable that returns no options is treated as empty (no unscoped All-query), and the error/empty state survives a deactivate/re-activate cycle.
  - `SceneQueryRunner.test.ts` — query runner skips execution when a variable dependency is in an error state.
- **Manual end-to-end** (via a Grafana build linked against this branch, behind a feature toggle):
  - 3 variables `t_a` (fails), `t_b` (healthy), `t_c` (chained to `t_a`) and 3 panels `p_a`/`p_b`/`p_c` on a healthy TestData datasource.
  - Result: `t_a` empty→error, `t_c` skipped (`Skipping updateAndValidate, a dependency is in error state`), `t_b` succeeds. Panels: `p_a` and `p_c` issue **no** `/api/ds/query` (`Variable dependency is in error state, skipping query execution`); only `p_b` runs. Confirmed both in the network panel and in the live scene graph (`request: null` for blocked panels, `SQR100` for `p_b`).
  - Verified the feature is a no-op when disabled.

## Notes for the reviewer

- Public API surface: two new optional `SceneVariableSetState` fields + two new `sceneGraph`/`VariableDependencyConfig` helpers. No breaking changes.

## Companion PR

The consuming Grafana wiring (the `dashboardVariablesBlockOnError` feature toggle plus passing `blockDependentsOnError` / `treatEmptyAsError` to the dashboard variable sets) is **grafana/grafana#127691**. This scenes PR is self-contained and can merge independently; the Grafana PR depends on it (must merge + release, then bump `@grafana/scenes`).


packages/scenes/src/variables/types.ts
```diff
@@ -91,6 +91,19 @@ export interface VariableValueOption {
...
+   * failed variable or any of its dependents will not run. This prevents running queries against the
+   * backend with empty/null interpolated values when a variable fails.
+   */
+  blockDependentsOnError?: boolean;
+  /**
+   * Only relevant together with `blockDependentsOnError`. When true a variable that resolves to an
+   * empty value (no options / null) is treated the same as a failed variable for the purpose of
+   * blocking dependent variables and panels.
+   */
+  treatEmptyAsError?: boolean;
```
torkelo, line 106: Not sure why this is needed, if a query fails it should fail properly not return an empty string

torkelo: and if it can't do that this should be an option on the datasource variable query not here

PR #1567, opened 2026-07-07
Chore: Sign commits made by auto

mdvictor:
Adapts the scenes repo release process to sign commits generated by the former `auto shipit` command. That opaque command would do everything for us, and throughout its process it would create 2 commits that would be unsigned (one for version bump and changelog). With our latest enforcements on signed commits, we need to update the release pipeline so those commits are also signed.

Since the initial command didn't support it, this PR breaks `auto shipit` into it's internal commands and deals with the commit signing in this workflow.

<details>
  <summary>📦 Published PR as canary version: <code>8.9.6--canary.1567.28860887668.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@8.9.6--canary.1567.28860887668.0
  npm install scenes-app@8.9.6--canary.1567.28860887668.0
  npm install @grafana/scenes-react@8.9.6--canary.1567.28860887668.0
  # or 
  yarn add @grafana/scenes@8.9.6--canary.1567.28860887668.0
  yarn add scenes-app@8.9.6--canary.1567.28860887668.0
  yarn add @grafana/scenes-react@8.9.6--canary.1567.28860887668.0
  ```
</details>

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -42,11 +42,11 @@ jobs:
       - name: Build library
         run: yarn build
 
-  release:
-    name: Release
+  canary:
+    name: Canary release
     runs-on: ubuntu-latest
     needs: test
-    if: "!contains(github.event.head_commit.message, 'ci skip') && !contains(github.event.head_commit.message, 'skip ci')"
+    if: github.event_name == 'pull_request'
     permissions:
       id-token: write
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/node-ci.yml
```diff
@@ -75,7 +116,104 @@ jobs:
...
+      # `auto shipit` is decomposed into the steps below so that the changelog and
+      # version-bump commits can be created through the GitHub GraphQL API
+      # (createCommitOnBranch) instead of `git push`. Commits created that way are
+      # signed by GitHub ("Verified"), which the branch protection on main requires.
+
+      - name: Calculate version bump
+        id: bump
+        run: |
+          AUTO_OUTPUT="$(yarn auto version)"
+          BUMP="$(echo "$AUTO_OUTPUT" | tail -n1)"
```
Sergej-Vlasov: **Fragile bump parsing.** `tail -n1` assumes the bump keyword is the last stdout line. Any trailing yarn/`auto` log line, warning, or deprecation notice shifts it, and because of the safe-default `case` below an unparsed value silently becomes `''` — which *skips a legitimate release* with no failure signal.

**Fix:** match the keyword anywhere in the output instead of relying on position, e.g.
```bash
BUMP="$(echo "$AUTO_OUTPUT" | grep -oE '^(major|minor|patch)$' | tail -n1)"
```
and `echo "$AUTO_OUTPUT"` when `BUMP` is empty so a mis-parse is debuggable in the logs.

.github/workflows/node-ci.yml
```diff
@@ -75,7 +116,104 @@ jobs:
...
+          GITHUB_TOKEN: ${{ steps.get-github-app-token.outputs.token }}
+
+      - name: Update changelogs
+        if: steps.bump.outputs.bump != ''
+        # `auto changelog` writes the root and per-package changelogs and commits
+        # them locally (unsigned). Drop the commit but keep the file changes so the
+        # next step can re-create it through the API.
+        run: |
+          yarn auto changelog
+          git reset --mixed HEAD~1
```
Sergej-Vlasov: **`git reset --mixed HEAD~1` hard-codes "exactly one commit".** This assumes `auto changelog` created precisely one commit. If it ever creates zero commits (nothing to log) this silently drops a *real* commit; if it creates more than one, leftovers remain and get folded into the next API commit.

**Fix:** capture HEAD before the command and reset back to it:
```bash
BEFORE="$(git rev-parse HEAD)"
yarn auto changelog
git reset --mixed "$BEFORE"
```
(or use a `--no-commit`/`--dry-run` flag if `auto changelog` supports leaving the tree dirty without committing).

.github/workflows/node-ci.yml
```diff
@@ -75,7 +116,104 @@ jobs:
...
+          file_pattern: 'lerna.json packages/*/package.json'
+        env:
+          GITHUB_TOKEN: ${{ steps.get-github-app-token.outputs.token }}
+
+      - name: Check out version bump commit
+        if: steps.bump.outputs.bump != ''
+        run: |
+          git fetch origin main
+          git reset --hard "$VERSION_COMMIT"
+          git tag "v$VERSION" "$VERSION_COMMIT"
```
Sergej-Vlasov, line 202: **Local tag looks dead.** This creates `v$VERSION` locally, but the remote tag is created independently via `gh api` in the "Push release tag" step, and `lerna publish from-package` doesn't need a tag. Unless `auto release` reads the *local* tag, this line has no effect.

**Fix:** confirm whether `auto release` relies on it; if not, drop the `git tag` line to avoid implying the tag is created here.

.github/workflows/node-ci.yml
```diff
@@ -75,7 +116,104 @@ jobs:
...
+        env:
+          VERSION: ${{ steps.version.outputs.version }}
+          VERSION_COMMIT: ${{ steps.version-commit.outputs.commit-hash }}
+
+      - name: Publish to npm
+        if: steps.bump.outputs.bump != ''
+        # Same command the npm plugin runs from `auto shipit`. Publishing
+        # authenticates via npm trusted publishing (OIDC), hence id-token: write.
+        run: yarn lerna publish from-package --yes --no-verify-access
```
Sergej-Vlasov: **Non-atomic release ordering.** Publish happens *before* the tag is pushed ("Push release tag" runs next). npm publishes are irreversible, so if the tag push or GitHub-release step fails, you're left with packages on npm that have **no git tag and no GitHub release**.

**Fix:** push the tag *before* `lerna publish` — the version-bump commit already exists on `main` at this point, so the tag can be created first, and only then publish. That way a failure after publish is limited to the (re-runnable) release-notes step.

.github/workflows/node-ci.yml
```diff
@@ -75,7 +116,104 @@ jobs:
...
+
+      - name: Push release tag
```
Sergej-Vlasov, line 207: **Tag push is not idempotent.** `POST .../git/refs` returns 422 if `refs/tags/v$VERSION` already exists, so a re-run after any mid-pipeline failure fails here.

**Fix:** check for the ref first and `PATCH .../git/refs/tags/v$VERSION` when it exists, or guard with `gh api ... || true` plus an explicit existence check, so re-runs are recoverable.

.github/workflows/node-ci.yml
```diff
@@ -42,11 +42,52 @@ jobs:
...
+
+      # On pull requests `auto shipit` takes its canary path: it publishes a canary
+      # version to npm and comments on the PR. It creates no git commits, so it is
+      # unaffected by the signed-commit requirement and can stay as-is.
+      - name: Run auto shipit (canary)
+        run: yarn auto shipit
+        env:
+          GITHUB_TOKEN: ${{ steps.get-github-app-token.outputs.token }}
+
   release:
```
Sergej-Vlasov, line 86: **Consider a `concurrency` guard on the release job (likely pre-existing).** Two quick pushes to `main` could run overlapping `release` jobs and race on version calc / tag creation.

**Fix:**
```yaml
concurrency:
  group: release-main
  cancel-in-progress: false
```
to serialize releases.

Sergej-Vlasov (commented):
A few correctness/robustness notes on the decomposed release pipeline, inline. Nothing blocking, but I'd address the tag-before-publish and the hard-coded `HEAD~1` before merge. Overall a careful, well-commented change 👍

Sergej-Vlasov (approved):
Re-reviewed after the "address comments" commit — all six points resolved cleanly:

1. **Bump parsing** — `grep -oE '^(major|minor|patch)$'` + raw-output echo on empty. Robust, and it keeps the safe-skip default the old `case` provided. ✅
2. **`HEAD~1`** — now captures `BEFORE_CHANGELOG` and resets back to it. ✅
3. **Local tag** — comment confirms `auto release` resolves the version via `git describe --tags` on HEAD, so it's intentional, not dead code. ✅
4. **Atomicity** — tag push moved ahead of `lerna publish`; a publish failure now leaves git consistent. ✅
5. **Idempotent tag push** — checks the existing ref and create/skip/fail-on-mismatch; nicer than what I suggested since it catches a wrong-SHA tag. ✅
6. **Concurrency** — `release-main` group with `cancel-in-progress: false`. ✅

Reordering looks safe: re-runs skip the already-correct tag and `from-package` only publishes unpublished versions. LGTM 🚀

Sergej-Vlasov (approved):

PR #1572, opened 2026-07-10
SceneQueryRunner: fan out multi-value datasource variable feeding an expression

### LINKED ISSUE #127694: Multi-value datasource variable only queries the first selected datasource when used as an expression input
When a multi-value datasource variable is used as the datasource for a query that feeds a server-side expression (Math, Reduce, etc.), the SceneQueryRunner resolves the variable to a single datasource UID instead of fanning out across all selected datasources. Only the first datasource in the selection list is queried. Direct queries using the same variable (without an expression) fan out correctly and show one result per datasource (mixed-0 / mixed-1 / mixed-2 in the Query Inspector). Reordering the datasource selection changes which datasource is used.

Steps to reproduce:

1. In a Grafana Cloud dashboard, create a datasource template variable: type = Datasource, datasource type = Prometheus, Multi-value enabled. Select 3 Prometheus datasources.
2. Create a panel with datasource set to -- Mixed --.
3. Add query refId A: datasource = ${prom_datasource}, expression = count(up), instant = true, hide = true.
4. Add a server-side Math expression refId B: expression = $A.
5. Observe: panel shows only 1 result (from the first selected datasource) instead of 3.
6. Open Query Inspector → Query tab: confirm single ds_type = expr request with only one datasource UID in the input query.
7. For contrast, create a second panel with the same query refId A but no expression (unhide it). Confirm it shows 3 results (mixed-0 / mixed-1 / mixed-2).
8. To demonstrate order-dependence: reorder the datasource selection. The expression panel value changes to match the new first datasource.

sarahzinger:
## What this PR does

When a dashboard
- has a variable for various datasources including an "all" (example all prometheus datasources)
- and then that dashboard has a query that uses that datasource variable

It is supposed to run that query on all of the datasources when "all" is selected

This generally works, but when you then
- and then you use that query with a **server-side expression** (Math, Reduce, etc.) there is a bug in the`SceneQueryRunner` where it would for some reason just use the first datasource from the list of datasources

This pr tries to properly fan out the server side expression to all of the datasources

For repro steps of the bug look here: https://github.com/grafana/grafana/issues/127694 (set up 3 prometheus datasources instances + import the dashboard in the issue and you'll see)

## Why

Multi-value datasource variables already fan out for direct queries (via the Mixed datasource), but not when the query feeds an expression — the request collapses to a single datasource, so the panel shows one result instead of one per selected datasource.

`prepareRequests` now expands the fan-out root data query and its dependent expression subgraph into one concrete clone per selected datasource, rewriting refIds so each replica is self-contained. Queries and expressions that do not depend on the multi-value variable are left untouched (they run once), so unrelated results are not duplicated.

Fixes [grafana/grafana#127694](https://github.com/grafana/grafana/issues/127694).

## Notes

- I am not very familiar with scenes and the release process. Feel free to take this pr over and do whatever needs to be done to get it fixed thanks! I also tried a different approach [here](https://github.com/grafana/grafana/pull/128125) if we like that one better I'd be happy to re-open. I wasn't really sure tbh. 

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,7 +3,5 @@
   "useNx": false,
   "$schema": "node_modules/lerna/schemas/lerna-schema.json",
   "version": "8.10.2",
-  "packages": [
-    "packages/*"
-  ]
+  "packages": ["packages/*"]
 }
--- a/packages/scenes/src/querying/SceneQueryRunner.test.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.test.ts
@@ -48,6 +48,7 @@ import { SceneQueryStateControllerState } from '../behaviors/types';
 import { config } from '@grafana/runtime';
 import { ScopesVariable } from '../variables/variants/ScopesVariable';
 import { ConstantVariable } from '../variables/variants/ConstantVariable';
+import { CustomVariable } from '../variables/variants/CustomVariable';
 
 const getDataSourceMock = jest.fn().mockImplementation((datasource) => {
   const isMixed = datasource?.uid === '-- Mixed --';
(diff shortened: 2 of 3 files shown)


harisrozajac (commented):
@sarahzinger  thanks for moving this into scenes. I manually verified the fix against three local Prometheus datasources (`gdev-prom-a/b/c`) with a multi-value DS variable + Math expression: before this change only the first DS was queried; with the PR branch linked into Grafana, all selected datasources fan out correctly.

i'll take this PR over 🙂

harisrozajac (approved):

PR #1582, opened 2026-07-15
Filters: propagate scopes and dashboard level filters to section vars

harisrozajac:
## Summary

When a row/tab has its own filter variable, queries under that section previously used only the closest AdHoc (`findClosestAdHocFilterInHierarchy`). Dashboard filters and scope-origin filters were dropped, so panels lost scopes / parent filters (Automon-like “section filters only” broke).

This PR implements a proposal to apply scope + dashboard filters to section filters when **executing queries** only — do not merge them into the section filter UI.

Fixes https://github.com/grafana/hyperion-planning/issues/660

### Demo

https://github.com/user-attachments/assets/6a6b1de1-e7cb-42d7-8a20-4c54b3eba018

### Changes
- Collect all same-DS AdHoc variables up the scene hierarchy and merge filters at query time in `DrilldownDependenciesManager` (ancestor → section), plus scopes from `getScopes()` so scopes still apply with section-only AdHoc.
- Keep scopes out of nested (section) AdHoc `originFilters` so they don’t appear as pills in the section filter bar (dashboard filter bar still shows scopes).
- Continue notifying dependents when a parent AdHoc shares a name with a section AdHoc (`filter0`), so parent filter changes re-run section queries.
- Inject scopes into dashboard-level AdHoc `originFilters` on activation when scopes are already selected.

## Test plan
- [x] Manual: dashboard with scopes + section filter — section UI shows only section filters; query `request.filters` includes scopes + dashboard + section
- [x] Manual: section filter only (no dashboard AdHoc) — scopes still applied to queries
- [x] Manual: section group-by only — scopes still work (regression)

<details>
  <summary>📦 Published PR as canary version: <code>8.13.1--canary.1582.29536064356.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@8.13.1--canary.1582.29536064356.0
  npm install scenes-app@8.13.1--canary.1582.29536064356.0
  npm install @grafana/scenes-react@8.13.1--canary.1582.29536064356.0
  # or 
  yarn add @grafana/scenes@8.13.1--canary.1582.29536064356.0
  yarn add scenes-app@8.13.1--canary.1582.29536064356.0
  yarn add @grafana/scenes-react@8.13.1--canary.1582.29536064356.0
  ```
</details>

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -4,7 +4,6 @@ import {
   GetTagResponse,
   GrafanaTheme2,
   MetricFindValue,
-  // @ts-expect-error (temporary till we update grafana/data)
   DrilldownsApplicability,
   Scope,
   SelectableValue,
@@ -1052,12 +1051,14 @@ export class AdHocFiltersVariable
     groupByKeys?: string[]
   ): Promise<DrilldownsApplicability[] | undefined> {
     const ds = await this._dataSourceSrv.get(this.state.datasource, this._scopedVars);
+    // @ts-expect-error (temporary till we update grafana/data)
     if (!ds || !ds.getDrilldownsApplicability) {
       return;
     }
 
     const timeRange = sceneGraph.getTimeRange(this).state.value;
 
+    // @ts-expect-error (temporary till we update grafana/data)
     return await ds.getDrilldownsApplicability({
       filters,
       queries,
@@ -1229,6 +1230,7 @@ export class AdHocFiltersVariable
     }
 
     const ds = await this._dataSourceSrv.get(this.state.datasource, this._scopedVars);
+    // @ts-expect-error (TODO: remove after upgrading with https://github.com/grafana/grafana/pull/118270)
     if (!ds || !ds.getGroupByKeys) {
       return override ? dataFromResponse(override.values).map(toSelectableValue) : [];
     }
@@ -1239,6 +1241,7 @@ export class AdHocFiltersVariable
     const queriesForKeys =
       queries ?? (this.state.useQueriesAsFilterForOptions ? getQueriesForVariables(this) : undefined);
 
+    // @ts-expect-error (TODO: remove after upgrading with https://github.com/grafana/grafana/pull/118270)
     const response = await ds.getGroupByKeys({
       filters: [],
       queries: queriesForKeys,
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -4,6 +4,7 @@ import {
   GetTagResponse,
   GrafanaTheme2,
   MetricFindValue,
+  // @ts-expect-error (temporary till we update grafana/data)
   DrilldownsApplicability,
   Scope,
   SelectableValue,
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
@@ -361,10 +361,14 @@ export class AdHocFiltersVariable
     // (same as dashboard filters: background merge, not shown in the section UI).
     if (isNestedVariableSetMember(this)) {
       this._clearScopeOriginFilters();
-    } else if (sceneGraph.getScopes(this)?.length) {
-      // If scopes are already selected (activated after scopes), inject them now.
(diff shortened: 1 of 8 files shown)


packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -356,6 +356,17 @@ export class AdHocFiltersVariable
   }
 
   private _activationHandler = () => {
+    // Dashboard-level AdHoc shows scopes as originFilters in the filter bar.
+    // Section AdHoc must not — scopes apply at query time via DrilldownDependenciesManager
+    // (same as dashboard filters: background merge, not shown in the section UI).
+    if (isNestedVariableSetMember(this)) {
+      this._clearScopeOriginFilters();
```
mdvictor, line 363: non blocking - but are we sure we dont want to show them there as well? I can't recall

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -356,6 +356,17 @@ export class AdHocFiltersVariable
...
+    } else if (sceneGraph.getScopes(this)?.length) {
+      // If scopes are already selected (activated after scopes), inject them now.
+      // Skip when scopes are empty so we do not clear pre-seeded / edited scope
+      // originFilters before scopes load.
+      this._updateScopesFilters();
```
mdvictor, line 372: Do we need to update on activation here? I think they should only update on scope variable update, otherwise with this here it will reset the scope filter values, if they have been edited, on deactivation and reactivation -- so if I edit a scope filter value and go to panel edit, going back to the dashboard will update scope filters back to their defaults instead of maintaining the edits

packages/scenes/src/variables/sets/SceneVariableSet.ts
```diff
@@ -340,7 +341,12 @@ export class SceneVariableSet extends SceneObjectBase<SceneVariableSetState> imp
       if (localVar?.isAncestorLoading) {
         variable = localVar;
       } else if (localVar) {
-        return;
+        // AdHoc filters merge across hierarchy at query time (section + dashboard + scopes).
+        // Continue notifying dependents so parent AdHoc changes still re-run section queries
+        // even when the section defines an AdHoc with the same name (e.g. filter0).
+        if (!(variable instanceof AdHocFiltersVariable && localVar instanceof AdHocFiltersVariable)) {
+          return;
```
mdvictor, line 355: Not sure, but I think we would also want to stop traversal if the dashboard adhoc DS differs from the section adhoc DS, then we don't want to merge.

mdvictor (changes requested):
Tested and works very nicely, we need to make sure the scope filters are not reset after going to panel edit. And I'm not entirely sure about the last one regarding traversal, but I think we might want a check there too

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
harisrozajac: yeah we don't want to show them at section level. see [here](https://github.com/grafana/hyperion-planning/issues/660#issuecomment-4840871337)

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
harisrozajac: yes this is a great catch! i've fixed this so the edited value is perserved:

https://github.com/user-attachments/assets/ae3fe881-2c6a-4af1-b7f8-6c169c461fae

packages/scenes/src/variables/sets/SceneVariableSet.ts
harisrozajac: also great catch! i fixed this

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
mdvictor: Cool, cool, makes sense, thank you @harisrozajac !!

mdvictor:
Awesome work! Thank you @harisrozajac for taking a look over those. The initial bug is no longer!

I found another interesting thing


https://github.com/user-attachments/assets/74a937e4-b236-46c5-9e1e-2b74e0764016

Having a default filter on a section causes the same mount/unmount bug, where if I edit a panel and go back it is reset... I wonder if we even want **dashboard** default filters at a section level... I guess I can see benefits to having them at section level.

In any case, this is a bug unrelated to this PR, I can see this in Ops as well, we can tackle in another PR

mdvictor (approved):

PR #1587, opened 2026-07-24
GrafanaCompatability: Make scene demo app work with react 19

torkelo:
This seems to make the demo app work with latest main and react 19 enabled


mdvictor (approved):

jackw (approved):
LGTM! 🚀

PR #1591, opened 2026-07-30
AdHocFiltersVariable: Add All value option for dashboard default filters

samjewell:
## What

Dashboard default (origin) filters with the 'one of' (`=|`) operator now offer an explicit **All** option at the top of the multi-value dropdown, mirroring the All option on query variables.

- Selecting All is exclusive: it clears any other ticked values, and ticking a value clears All.
- Applying it stores the special `$__all` sentinel value (`values: ['$__all']`). The pill stays visible and reads `key =| All`.
- Filters holding `$__all` stop restricting results: they are excluded from query requests (`request.filters`), the rendered filter expression (`${Filters}` interpolation), and `getTagKeys`/`getTagValues` lookups.
- The restore button returns an edited filter to the dashboard default as usual, and a default that *is* All round-trips (`restorable: false`, All shown ticked in the dropdown).
- Display always falls back to "All" from the sentinel, so pills render correctly even when `valueLabels` were not persisted.

## Why

Dashboard authors setting a multi-value default filter currently have to enumerate every value. That snapshot goes stale as new label values appear, and viewers cannot tell "everything selected" from a curated list. An All default keeps the filter pinned and visible while leaving results unrestricted until the viewer narrows it.

This intentionally does not reuse the existing match-all tombstone (`=~ .*`) that filter removal produces: that swaps the author's operator for `=~`, loses the multi-value checkbox editing UX, and cannot round-trip as an authored default in the editor. The `$__all` sentinel keeps the `=|` operator, so viewers can open the pill and pick specific values from the normal checkbox dropdown.

Scope notes:

- Offered only for origin filters - for a regular ad hoc filter, All is equivalent to deleting it.
- Offered only for `=|`, not `!=|`, where "All" would read as excluding everything.

## Relationship to #1596

The exclusive selection here (ticking All clears the other values, and vice versa) matches the behaviour proposed for the query variable picker in #1596. If that lands, both pickers share the same All semantics; if review there settles on different semantics, this PR will follow suit. One nuance specific to the ad hoc path: unlike query variables there is no commit-time normalisation equivalent to `changeValueTo`, so without exclusivity a mixed selection like `['$__all', 'EMEA']` would persist as-is - the pill would read "All, EMEA" while the filter resolves as All, and the viewer's tick would silently do nothing.

This PR deliberately does not add a `Selected (N)` toggle-all header to the ad hoc dropdown (the affordance the query variable picker has for exploding All into individual ticks and collapsing back). The ad hoc combobox has never had a toggle-all row, so that is left as a follow-up - happy to pick it up if full parity is wanted.

## Why this lives in scenes rather than Grafana

Worth spelling out, since much of the default-filters surface area is Grafana's: the **Default filters** editor, its controller (`AdHocOriginFiltersController`), and the dashboard JSON serialisation are all in grafana/grafana. Scenes also exposes hooks that let a host customise ad hoc filter behaviour with no scenes change at all - `getTagKeysProvider`, `getTagValuesProvider` and `expressionBuilder`, all three of which Grafana's alerting scenes already use. With only those, Grafana could have added an All entry to the values dropdown and stripped `$__all` out of the rendered filter expression.

Four parts of the behaviour have no such extension point, which is what puts this change here:

1. **Exclusive selection** (ticking All clears the other values, and ticking a value clears All) is local React state in the combobox, in `handleLocalMultiValueChange`.
2. **`request.filters`** is assembled by `DrilldownDependenciesManager.getFilters()`. That is the path datasources actually read for ad hoc filters, there is no host override for it, and dashboard ad hoc variables run with `applyMode: 'auto'`, so it is the live path.
3. **The `getTagKeys`/`getTagValues` cascade** passes sibling filters as context when computing a given filter's options. An All filter has to be excluded there too, or it constrains every other filter's option list.
4. **The runtime filter bar** renders through `AdHocFiltersVariableController` and `AdHocFilterPill`. Hosts supply their own controller in editor contexts only, not at runtime, so the "All" display fallback has to live here.

Without 2 to 4, an All default would be an editor-only illusion: the dashboard would send `key=|$__all` to the datasource as a literal value, matching nothing.

The upside of this split is that no Grafana-side change is required. `$__all` travels inside the existing `values` array, which both the v1 and v2 dashboard serialisers already persist, so there is no schema work to do. The cost is release cadence - it needs a scenes release and a version bump in Grafana before it can ship.

**Open question for reviewers:** this reuses the `ALL_VARIABLE_VALUE` (`$__all`) sentinel that `MultiValueVariable` already uses, carried as a filter value. If we would rather model "all selected" as an explicit field on the filter instead, that turns into a schema change on the Grafana side, and I am happy to take it that way.

## Testing

- New combobox interaction tests: All shown first for one-of origin filters, not shown for user filters or `!=|`, exclusive selection, commit payload, checked state and restorable flag for an All default, restore back to the dashboard default.
- New unit tests: All-value filters excluded from the filter expression, from `getTagKeys`/`getTagValues` calls, and from `SceneQueryRunner` request filters.

<details>
  <summary>📦 Published PR as canary version: <code>8.14.0--canary.1591.30990876578.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@8.14.0--canary.1591.30990876578.0
  npm install scenes-app@8.14.0--canary.1591.30990876578.0
  npm install @grafana/scenes-react@8.14.0--canary.1591.30990876578.0
  # or 
  yarn add @grafana/scenes@8.14.0--canary.1591.30990876578.0
  yarn add scenes-app@8.14.0--canary.1591.30990876578.0
  yarn add @grafana/scenes-react@8.14.0--canary.1591.30990876578.0
  ```
</details>


samjewell:
Question for @grafana/grafana-dashboards-core before this gets a behaviour review: **which "All" interaction should ad hoc filters converge on?** While comparing this PR against the query variable picker I realised the codebase currently has three All implementations that disagree with each other:

1. **The select-based query variable picker** (`VariableValueSelectMulti` in this repo, what dashboards render today): All is a plain `$__all` option with no exclusivity - ticking All does not clear other values and vice versa, so mixed selections like `[EMEA, APJ, $__all]` are representable. Resolution is `hasAllValue()`, i.e. `value[0] === '$__all'`, which is order-of-selection dependent: depending on tick order, either the All tick or the value ticks silently win. It does have the `Selected (N)` toggle-all header (from `MultiSelect`'s `toggleAllOptions`) with the indeterminate state, which is a nice affordance for exploding All into individual ticks and collapsing back.

   image
   image


3. **The legacy OptionsPicker** (`public/app/features/variables/pickers/OptionsPicker/reducer.ts` in grafana/grafana): fully exclusive - selecting All clears everything else, and `updateAllSelection` strips `$__all` the moment a second value is ticked.

4. **`@grafana/ui` MultiCombobox**: All is not stored at all - it is a select-all / clear-all convenience that materialises the real values. (Not suitable for default filters, since a materialised list goes stale as new label values appear - avoiding that is the point of this PR.)

This PR currently follows the legacy picker's semantics (option 2): ticking All clears other ticks and vice versa, so the committed state is always unambiguous - either `['$__all']` or concrete values, never a mix. The main motivation is avoiding the silent no-op trap: with a default of All, a viewer who ticks a value while All stays ticked would otherwise hit Apply and see nothing change. It does not implement the `Selected (N)` toggle-all header - the ad hoc combobox has never had one.

image
image


Proposal: merge this as is (the behaviour is self-consistent and trap-free), and treat convergence as a follow-up - e.g. adding a toggle-all header to the ad hoc multi-value dropdown, and/or revisiting the query picker's order-dependent resolution. But if you would rather this PR match the select-based picker exactly, or you have a different convergence target in mind, happy to change it - flag it here.

samjewell:
Update on the question above: since writing it we've opened #1596, which fixes the select-based picker's mixed-selection bugs (the order-dependent `value[0]` resolution, and the dropdown showing ticks that get rewritten on blur) and makes its All selection exclusive - selecting All deselects the other values and vice versa. So the convergence question may largely resolve itself: if #1596 lands, this PR's behaviour matches the query variable picker's, and both match what the toggle-all header already does.

The remaining parity delta is the `Selected (N)` toggle-all header itself, which the ad hoc dropdown has never had. We've deliberately left it out of this PR to keep the scope tight, but happy to pick it up as a follow-up if you'd like full parity - it would reuse the `ToggleAllState` enum and mirror the picker's `determineToggleAllState`, rendered as a pinned row in the combobox dropdown.

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -1448,6 +1459,15 @@ export function isMatchAllFilter(filter: AdHocFilterWithLabels): boolean {
   return filter.operator === '=~' && filter.value === '.*';
 }
 
+/**
+ * True when the filter has the special "All" value selected, meaning it should not
+ * restrict queries. Origin (default) filters offer "All" in the multi-value combobox
+ * so dashboard authors can pre-select a key without restricting its values.
+ */
+export function isAllValueFilter(filter: AdHocFilterWithLabels): boolean {
+  return filter.values ? filter.values.includes(ALL_VARIABLE_VALUE) : filter.value === ALL_VARIABLE_VALUE;
```
samjewell: **High — restrict sentinel recognition to the supported operator.** This currently classifies any filter whose value is `$__all` as unrestricted, including `=`, `!=`, and `!=|` filters and ordinary user filters. A legitimate literal filter such as `tenant = "$__all"` will therefore be silently omitted from query requests, expressions, and option lookups, broadening results. Since the All affordance is only defined for `=|`, please require `filter.operator === "=|"` here (and add regression coverage for the other operators).

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -1363,7 +1374,7 @@ function renderExpression(
   filters: AdHocFilterWithLabels[] | undefined
 ) {
   return (builder ?? renderPrometheusLabelFilters)(
-    filters?.filter((f) => isFilterApplicable(f) && !isGroupByFilter(f)) ?? []
+    filters?.filter((f) => isFilterApplicable(f) && !isGroupByFilter(f) && !isAllValueFilter(f)) ?? []
```
samjewell: **Medium — the per-key interpolation path also needs All filtering.** This fixes `${Filters}`, but `getFilterValueByKey()` still includes All filters. As a result, `${Filters["pod"]}` returns the literal `$__all`; if a concrete filter shares that key, it returns `[$__all, concrete]`, even though the query/request paths correctly ignore the All filter. Please apply the same predicate in `getFilterValueByKey()` and test both the All-only and All-plus-concrete cases.

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersCombobox.tsx
```diff
@@ -314,6 +320,17 @@ export const AdHocCombobox = forwardRef(function AdHocCombobox(
     }
   }
 
+  // Origin (default) filters with the "one of" operator offer an explicit "All" option,
+  // like query variables. Selecting it commits the special $__all value, which lifts the
+  // filter's restriction from queries. Not offered for "not one of" where "All" would read
+  // as excluding everything.
+  if (isMultiValueEdit && filter?.origin && filter?.operator === '=|' && !inputValue) {
```
samjewell: **Medium — prepending All invalidates the initial index for grouped options.** `handleFetchOptions()` sets index 1 when `options[0].group` is present, assuming index 0 is the group header. After this `unshift`, index 1 is the group header and the first selectable datasource value is index 2. Opening a grouped dropdown and immediately pressing Enter can therefore pass the header (with an undefined value) to `handleLocalMultiValueChange()` and later commit malformed state. The initial active index should account for this inserted row; please add a grouped-options keyboard test.

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
```diff
@@ -24,7 +25,13 @@ export function AdHocFilterPill({ filter, controller, readOnly, focusOnWipInputR
     useEditablePill({ filter, controller, readOnly, focusOnWipInputRef, isFilterEmpty: isAdhocFilterEmpty });
 
   const keyLabel = filter.keyLabel ?? filter.key;
-  const valueLabel = filter.valueLabels?.join(', ') || filter.values?.join(', ') || filter.value;
+  // The special $__all value always displays as "All", even when no valueLabels were stored
+  const valueLabel =
+    (filter.valueLabels?.length ? filter.valueLabels : filter.values ?? [filter.value])
```
samjewell: **Low — derive the All label from the sentinel, not persisted `valueLabels`.** Selecting All stores the current translation in `valueLabels`, and URL sync persists it. On reload or locale change, this branch prefers that stale label and never translates the `$__all` value. Pair labels with `filter.values` and force the current All translation whenever the corresponding value is the sentinel.

samjewell (commented):
**AI review disclosure:** This review was produced by **GPT 5.6 Sol — High Thinking**.

I would request changes, but GitHub does not allow authors to request changes on their own pull requests. The overall approach is sound and CI is green, but I found three correctness regressions that should be addressed before merge, plus one localization issue. The inline comments contain concrete reproduction paths and suggested fixes.

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
samjewell: Fixed in c6a39f66 - `isAllValueFilter` now requires `operator === '=|'`, so a literal `$__all` on `=`, `!=`, `!=|` or regex operators keeps filtering. Regression test added covering `=` and `!=|`.

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
samjewell: Fixed in c6a39f66 - `getFilterValueByKey()` applies the same predicate, so `${Filters["pod"]}` returns `''` when the filter is All, and only the concrete values when an All filter shares the key. Both cases tested.

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFiltersCombobox.tsx
samjewell: Fixed in c6a39f66 - when the All row is prepended, it takes the initial highlight (index 0), which is also the query-variable-picker convention of All-first. The same rule is applied in `onChange` for the empty-input case, since clearing typed text re-prepends the row and would have re-introduced the same off-by-one. Grouped-options keyboard test added (immediate Enter toggles All, not the group header).

packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
samjewell: Fixed in c6a39f66 - labels are now paired with `filter.values` by index and the sentinel always renders with the current translation, so a persisted label (e.g. from another locale) never wins. Test added with a stale `valueLabels: ['Alle']` rendering as `All`.

mdvictor:
I like this! I think the UI is going in a nice direction with the All value in multiselect. I'm not as keen on the query exclusion of these filters based on `isAllFilterValue`. We already have support for that under the `isMatchAllFilter` functionality, which is basically the same, but for the regex structure we talked about in our meeting. What if we update the matchAll logic to include this All branch as well, instead of creating a new `isAllFilterValue`.

In this case, we should also update `updateToMatchAll` to turn a filter to matchall depending on it's operator (so if the operator is `=|` we use the new `$_all`, otherwise default to regex `=* .*` as before. This would fall more in line with the logic default filters follows -- they always have the X icon associated with basically nullifying the filter, but keeping the pill there so it can be restored. so if we set the filter to be matchAll from the get-go, aka being `=| All`, the UI would pick it up in the filters bar and now show the X, for example. This video explains it a bit better I hope:


https://github.com/user-attachments/assets/84f0e773-e55d-4755-b05d-7bddbe78b686

Basically, as soon we set the default filter to be  `=| All`, it should be set as matchAll so the UI doesnt show the X icon anymore.

And in this way we don't need `isAllFilterValue` and the extra logic as it will fall under `isMatchAllFilter`.

@samjewell WDYT?  Let me know if it makes sense.

I really wanna thank you again for putting the time in this, but also realize that you might be busy with other stuff as well, so please let me know if you want me to take over this at any point and I can prioritize as best I can to help unblock RevOps! I really think this should land!

mdvictor (approved):

kristinademeshchik (approved):

PR #1593, opened 2026-07-30
feat(SceneQueryController): add registerPendingWork to hold query readiness for non-query work (like repeat clone mounting)

grafakus:
Support for https://github.com/grafana/grafana/pull/129758

### DIFF
--- a/packages/scenes/src/behaviors/SceneQueryController.ts
+++ b/packages/scenes/src/behaviors/SceneQueryController.ts
@@ -2,7 +2,12 @@ import { SceneObjectBase } from '../core/SceneObjectBase';
 import { SceneObject, SceneStatelessBehavior } from '../core/types';
 import { writeSceneLog } from '../utils/writeSceneLog';
 import { SceneRenderProfiler } from '../performance/SceneRenderProfiler';
-import { SceneQueryControllerEntry, SceneQueryControllerLike, SceneQueryStateControllerState } from './types';
+import {
+  SceneQueryControllerEntry,
+  SceneQueryControllerEntryType,
+  SceneQueryControllerLike,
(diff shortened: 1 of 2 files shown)


packages/scenes/src/behaviors/SceneQueryController.ts
```diff
@@ -77,6 +82,17 @@ export class SceneQueryController
     }
   }
 
+  /**
+   * Registers non-query pending work (e.g. repeat clone mounting) so that profile completion
+   * and readiness checks based on runningQueriesCount cannot fire while the work is pending.
+   * Returns a release function. Releasing more than once is a no-op.
+   */
+  public registerPendingWork(type: SceneQueryControllerEntryType, origin: SceneObject): () => void {
```
ifrost, line 90: Related to https://github.com/grafana/grafana/pull/129758/changes#r3713343432. It just looks like a hack to hook into query controller, so it's a bit dangerous to expose it publicly like that. Maybe there could be a separate profiler/tracker where components could register pending async work?

grafakus: Agreed, exposing this on the query controller feels unnecessary/wrong.

`registerPendingWork` is also redundant with what we already have: It wraps `queryStarted` / `queryCompleted`, which are already public on `SceneQueryControllerLike` and which   [registerQueryWithController](https://github.com/grafana/scenes/blob/main/packages/scenes/src/querying/registerQueryWithController.ts) already uses.

grafakus: I've closed the PR, let's continue the conversation in https://github.com/grafana/grafana/pull/129758

PR #1583, opened 2026-07-20
SceneTimeRangeCompare: make timeShiftAlignmentProcessor non-mutating and export it

dprokop:
Fixes duplicate comparison series when time compare runs against streaming/split-chunk queries (e.g. Loki). `timeShiftAlignmentProcessor` mutated `refId`/`meta` in place on frames owned by the datasource's response accumulator, so every re-emitted chunk registered as a brand-new series. Exposed by #1554.

The processor now returns new frame objects - per the contract `ExtraQueryProvider` already documents - and is exported so grafana core can delete its private fork of the same function: [grafana/grafana#128796](https://github.com/grafana/grafana/pull/128796).

**Test:** scenes-app → Demos → *Time compare + streaming duplicate series (regression demo)*, plus regression tests in `SceneTimeRangeCompare.test.tsx`.

<details>
  <summary>📦 Published PR as canary version: <code>8.14.0--canary.1583.29911997371.0</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install @grafana/scenes@8.14.0--canary.1583.29911997371.0
  npm install scenes-app@8.14.0--canary.1583.29911997371.0
  npm install @grafana/scenes-react@8.14.0--canary.1583.29911997371.0
  # or 
  yarn add @grafana/scenes@8.14.0--canary.1583.29911997371.0
  yarn add scenes-app@8.14.0--canary.1583.29911997371.0
  yarn add @grafana/scenes-react@8.14.0--canary.1583.29911997371.0
  ```
</details>

PR #1550, opened 2026-06-18
Dashboards: Push error and notice status items into panel header

mdvictor:
This PR gathers query errors and notices so they can be rendered into a single UI tooltip that links to the inspector to show a unified view of these per panel

For https://github.com/grafana/grafana/pull/126750

<details>
  <summary>📦 Published PR as canary version: <code>8.5.1--canary.1550.28094466812.0</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install @grafana/scenes@8.5.1--canary.1550.28094466812.0
  npm install scenes-app@8.5.1--canary.1550.28094466812.0
  npm install @grafana/scenes-react@8.5.1--canary.1550.28094466812.0
  # or 
  yarn add @grafana/scenes@8.5.1--canary.1550.28094466812.0
  yarn add scenes-app@8.5.1--canary.1550.28094466812.0
  yarn add @grafana/scenes-react@8.5.1--canary.1550.28094466812.0
  ```
</details>

PR #1595, opened 2026-07-31
TimeComparison: Warning when comparison is empty

### LINKED ISSUE #126825: TimeComparison: UX - Empty query feedback
Currently if the time compare query returns no results, but the primary query does, there's nothing in the UI that lets the user know that nothing being returned was expected (there's no no-data view)

drew08t:
I went ahead and moved the awesome contribution from @macayu17 in this [PR](https://github.com/grafana/grafana/pull/126918) to scenes because the `timeShiftAlignmentProcessor` was moved out of grafana and into scenes with https://github.com/grafana/grafana/pull/128796 and https://github.com/grafana/scenes/pull/1583.

This PR adds an informational notice when a panel time comparison query returns no data while the primary query still has data.

Before:
image

After:
image

Fixes https://github.com/grafana/grafana/issues/126825

### DIFF
--- a/packages/scenes/src/components/SceneTimeRangeCompare.test.tsx
+++ b/packages/scenes/src/components/SceneTimeRangeCompare.test.tsx
@@ -550,5 +550,105 @@ describe('SceneTimeRangeCompare', () => {
 
       expect(result.series[0].refId).toBe('A-compare');
     });
+
+    describe('empty comparison notice', () => {
+      // When the primary query returns data but the comparison query does not, the processor surfaces
+      // an informational notice so users understand the empty comparison was expected.
+      const NO_DATA_NOTICE = { severity: 'info', text: 'No data returned for time comparison' };
+
+      const primaryWithData = () =>
+        makePanelData('2024-01-10T00:00:00.000Z', '2024-01-10T01:00:00.000Z', [
+          toDataFrame({ refId: 'A', fields: [{ name: 'value', values: [1] }] }),
+        ]);
+
+      const getEmptyComparisonProcessor = () => {
+        const timeRange = new SceneTimeRange({ from: 'now-1h', to: 'now' });
+        const comparer = new SceneTimeRangeCompare({ compareWith: '24h' });
+        return getProcessor(comparer, timeRange);
+      };
+
+      it('should emit a notice frame from the request targets when the compare query returns no frames', async () => {
+        const processor = getEmptyComparisonProcessor();
+        const secondary: PanelData = {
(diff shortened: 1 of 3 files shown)


fastfrwrd:
I wonder if this should somehow use the new Warnings/Errors/Notices UI worked on by @mdvictor to avoid adding another icon and tooltip in the already crowded title bar. There was just a discussion occurring that it's already getting confusing to distinguish between the "description" icon and the icon for that UI, so I feel like this adds another competitor to the mix. WDYT Dashboards folks?

gelicia:
> I wonder if this should somehow use the new Warnings/Errors/Notices UI worked on by @mdvictor to avoid adding another icon and tooltip in the already crowded title bar. There was just a discussion occurring that it's already getting confusing to distinguish between the "description" icon and the icon for that UI, so I feel like this adds another competitor to the mix. WDYT Dashboards folks?

I looked into this, the `notices` property that we add this notice onto is a long standing field that displays this data. I do think the UI Victor is adding should incorporate that data, but I also think we are doing the right thing in this ticket by using existing fields.

Edit: Confirmed that the new work brings in this data, so it is already using the paved path. https://github.com/grafana/scenes/pull/1550/changes#diff-b4e762e26aeb297a8f36c1ada3213ecd5a9bcd972799a753018dccf96fc75a30R471

gelicia:
@drew08t what's your testing setup like for this? I usually have just used random walk, which will have data no matter what. 

Edit: Nvm, CSV File > `ohlc_dogecoin.csv` does the trick!

gelicia (changes requested):
Having trouble replicating this.

Created a CSV file with the following (requires converting the timestamp field)

```
"timestamp","value"
10/08/2026 09:02:24,1
10/08/2026 09:08:30,2
10/08/2026 09:14:36,3
10/08/2026 09:20:42,2
10/08/2026 09:26:48,1
10/08/2026 09:32:54,2
10/08/2026 09:39:00,3
10/08/2026 09:45:06,2
10/08/2026 09:51:12,1
10/08/2026 09:57:18,2
10/08/2026 10:03:24,3
10/08/2026 10:09:30,2
10/08/2026 10:15:36,1
10/08/2026 10:21:42,2
10/08/2026 10:27:48,3
10/08/2026 10:33:54,2
10/08/2026 10:40:00,1
10/08/2026 10:46:06,2
10/08/2026 10:52:12,3
10/08/2026 10:58:18,2
```

This should give data for a small amount of time and return nothing for a week prior.  The series doesn't toggle like it should, and I don't see the notification icon. 

Let me know what I should do instead. Thank you!

gelicia (approved):
Used a prometheus query on a server I only started today. Saw the notice as expected.

PR #876, opened 2024-08-27
Add toggle all option to variable value select

oscarkilhed:
Needs https://github.com/grafana/grafana/pull/92483 to be merged before this will work.

PR #1596, opened 2026-07-31
VariableValueSelect: Make All value selection exclusive in the multi picker

samjewell:
## What

Selecting the **All** option in the multi-value variable picker now deselects the other values in the open dropdown, and selecting another value deselects All. This mirrors the normalisation `changeValueTo` already applies on commit, so the dropdown no longer shows mixed selections that get silently rewritten on blur.

It also closes two gaps in that existing normalisation:

- `changeValueTo` removed the All value from a mixed selection only when it sat **first** in the array. It now removes it wherever it sits. (URL values are covered by this path too: `updateFromUrl()` passes them through `changeValueTo()`, so a mixed URL value is normalised without interpolating a literal `$__all`.)
- `hasAllValue()` only detected the All value in **first** position (`value[0] === '$__all'`). It now detects it at any position, so a mixed selection that bypasses normalisation entirely (e.g. direct state updates) resolves as All instead of interpolating a literal `$__all` into queries.
- `formatRegistry.isAllValue()` - an unused, unexported duplicate of the same order-dependent check - is removed rather than patched.

## Why

Today the dropdown lets All and concrete values appear ticked together, but that state is not what gets committed - `changeValueTo` rewrites it on blur using positional heuristics. Two concrete problems:

1. **The UI lies during interaction.** With `All` selected, ticking `EMEA` and `APJ` shows all three ticked; on blur the committed value is actually `[EMEA, APJ]`. The user never sees the real selection until they reopen the dropdown.
2. **Mixed selections that reach state produce broken queries.** When the All value ends up mid-array (e.g. `[EMEA, $__all, APJ]`, reachable today by ticking All between two value ticks), no normalisation applies and `value[0]` checks miss it - so queries interpolate a literal `$__all` (e.g. `env=~"EMEA|\$__all|APJ"`).

The exclusive behaviour also matches the legacy `OptionsPicker` (`public/app/features/variables/pickers/OptionsPicker/reducer.ts` in grafana/grafana), which selects only All when All is clicked and strips All when a second value is ticked - so this restores the semantics that didn't survive the port to the select-based picker. The toggle-all header from #876 is untouched and already behaves exclusively (it explodes All into the concrete values, or collapses the selection back to All).

Builds on the toggle-all work in #876 / grafana/grafana#92483 (cc @oscarkilhed).

Related: this came out of the discussion on #1591 about which "All" interaction ad hoc filter default values should converge on - see https://github.com/grafana/scenes/pull/1591#issuecomment-5134231043.

## Testing

- New interaction tests: selecting All deselects the other values in the open dropdown; selecting a value deselects All; committed values on blur are `['$__all']` / the concrete value respectively.
- New interaction test for the toggle-all header (which calls `onChange` with a synthetic `{ option: {} }` action): explodes All into the concrete values and collapses back to All.
- New unit tests: `changeValueTo` with the All value mid-array; `getValueText`/`getValue` when the All value is not first.
- Full suite, typecheck, lint and prettier pass.

### DIFF
--- a/packages/scenes/src/variables/components/VariableValueSelect.tsx
+++ b/packages/scenes/src/variables/components/VariableValueSelect.tsx
@@ -1,6 +1,7 @@
 import { t } from '@grafana/i18n';
 import { isArray } from 'lodash';
 import React, { RefCallback, useEffect, useMemo, useState } from 'react';
+import { type ActionMeta } from 'react-select';
 import {
   Checkbox,
   InputActionMeta,
@@ -13,6 +14,7 @@ import {
 } from '@grafana/ui';
 
 import { MultiValueVariable, MultiValueVariableState } from '../variants/MultiValueVariable';
+import { ALL_VARIABLE_VALUE } from '../constants';
 import { VariableValue, VariableValueSingle } from '../types';
 import { selectors } from '@grafana/e2e-selectors';
 import { GrafanaTheme2, SelectableValue } from '@grafana/data';
@@ -24,7 +26,28 @@ import { getVariableControlId } from '../utils';
 
 const filterNoOp = () => true;
 
-const filterAll = (v: SelectableValue<VariableValueSingle>) => v.value !== '$__all';
+const filterAll = (v: SelectableValue<VariableValueSingle>) => v.value !== ALL_VARIABLE_VALUE;
+
+/**
+ * Mirrors changeValueTo's commit-time normalisation so the open dropdown reflects what will
+ * actually be committed: selecting All deselects the other values, and selecting another
+ * value deselects All. Without this the dropdown can show mixed selections that are
+ * silently rewritten on blur.
+ */
+const enforceAllExclusivity = (
+  values: VariableValueSingle[],
+  action: ActionMeta<SelectableValue<VariableValueSingle>>
+): VariableValueSingle[] => {
+  if (action.action === 'select-option' && action.option?.value === ALL_VARIABLE_VALUE) {
+    return [ALL_VARIABLE_VALUE];
+  }
+
+  if (values.length > 1) {
+    return values.filter((v) => v !== ALL_VARIABLE_VALUE);
+  }
+
+  return values;
+};
 
 const determineToggleAllState = (
   selectedValues: Array<SelectableValue<VariableValueSingle>>,
@@ -34,7 +57,7 @@ const determineToggleAllState = (
     return ToggleAllState.allSelected;
   } else if (
     selectedValues.length === 0 ||
-    (selectedValues.length === 1 && selectedValues[0] && selectedValues[0].value === '$__all')
+    (selectedValues.length === 1 && selectedValues[0] && selectedValues[0].value === ALL_VARIABLE_VALUE)
   ) {
     return ToggleAllState.noneSelected;
   } else {
@@ -198,7 +221,12 @@ export function VariableValueSelectMulti({
           model.changeValueTo([], undefined, true);
         }
 
-        setUncommittedValue(newValue.map((x) => x.value!));
+        setUncommittedValue(
+          enforceAllExclusivity(
+            newValue.map((x) => x.value!),
+            action
+          )
+        );
       }}
     />
   );
--- a/packages/scenes/src/variables/interpolation/formatRegistry.ts
+++ b/packages/scenes/src/variables/interpolation/formatRegistry.ts
@@ -415,7 +415,9 @@ function customFormatQueryParameter(name: string, value: VariableValueSingle, va
 }
 
 export function isAllValue(value: VariableValueSingle) {
-  return value === ALL_VARIABLE_VALUE || (Array.isArray(value) && value[0] === ALL_VARIABLE_VALUE);
+  // Match hasAllValue: the All value counts wherever it sits, so a mixed selection is
+  // never interpolated as a literal $__all
+  return value === ALL_VARIABLE_VALUE || (Array.isArray(value) && value.includes(ALL_VARIABLE_VALUE));
 }
 
 const SQL_ESCAPE_MAP: Record<string, string> = {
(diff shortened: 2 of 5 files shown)


packages/scenes/src/variables/components/VariableValueSelect.tsx
```diff
@@ -198,7 +221,12 @@ export function VariableValueSelectMulti({
           model.changeValueTo([], undefined, true);
         }
 
-        setUncommittedValue(newValue.map((x) => x.value!));
+        setUncommittedValue(
+          enforceAllExclusivity(
+            newValue.map((x) => x.value!),
+            action
```
samjewell, line 227: **Low:** Could we add an interaction test for the toggle-all header path? `SelectBase` invokes `onChange` with the synthetic action `{ action: "select-option", option: {} }`; this currently relies on the length/filter fallback in `enforceAllExclusivity`, but the new tests only cover clicking ordinary options.

packages/scenes/src/variables/interpolation/formatRegistry.ts
```diff
@@ -415,7 +415,9 @@ function customFormatQueryParameter(name: string, value: VariableValueSingle, va
 }
 
 export function isAllValue(value: VariableValueSingle) {
-  return value === ALL_VARIABLE_VALUE || (Array.isArray(value) && value[0] === ALL_VARIABLE_VALUE);
+  // Match hasAllValue: the All value counts wherever it sits, so a mixed selection is
+  // never interpolated as a literal $__all
+  return value === ALL_VARIABLE_VALUE || (Array.isArray(value) && value.includes(ALL_VARIABLE_VALUE));
```
samjewell: **Low:** Could we add a direct test for `isAllValue()` with `$__all` in a non-first position? This behavior changes here but is not covered directly. The helper also has no in-repo caller and is not exported from the package index, so if it is not intended for a follow-up call site, dropping this hunk may be clearer.

samjewell (commented):
Review from GPT-5.6 Sol, High.

two code comments and one nit:

**Nit:** The PR description says edge-case URLs resolve as All, but `updateFromUrl()` passes URL values through `changeValueTo()`, so a mixed URL value can instead have All stripped depending on its position. Could we adjust that wording to say the URL value is normalized without interpolating a literal `$__all`?

packages/scenes/src/variables/components/VariableValueSelect.tsx
samjewell: Added in c28b6d20 - the new test drives the toggle-all header through both directions (explode All into the concrete values, collapse back to All). The header fires the synthetic `{ action: 'select-option', option: {} }`, so both clicks go through `enforceAllExclusivity`'s length/filter fallback rather than the option branch.

packages/scenes/src/variables/interpolation/formatRegistry.ts
samjewell: Agreed - it has no callers and is not exported from the package index, so I've removed `isAllValue` entirely in c28b6d20 rather than reverting the hunk: leaving an order-dependent duplicate of `hasAllValue` lying around is how this class of bug breeds. Happy to reinstate it (with the `.includes()` fix and a direct test) if a follow-up call site is planned.

samjewell:
Good catch on the nit - `updateFromUrl()` does route through `changeValueTo()`, so a mixed URL value is normalised there (position decides whether All wins or is stripped) rather than resolving as All via `hasAllValue()`. I've reworded the description: the `.includes()` change matters for values that bypass `changeValueTo` entirely (e.g. direct state updates), and the URL case is now described as normalised-without-literal-$__all.

Both inline comments addressed in c28b6d20: added a toggle-all header interaction test, and removed the dead `isAllValue` helper rather than patching it.

packages/scenes/src/variables/variants/MultiValueVariable.ts
```diff
@@ -329,16 +331,18 @@ export abstract class MultiValueVariable<TState extends MultiValueVariableState
...
       if (value[value.length - 1] === ALL_VARIABLE_VALUE) {
         value = [ALL_VARIABLE_VALUE];
         text = [ALL_VARIABLE_TEXT];
       }
-      // If the first value is the ALL value and we have other values, then remove the All value
-      else if (value[0] === ALL_VARIABLE_VALUE && value.length > 1) {
-        value.shift();
+      // Otherwise a value was added after the All value, so the All value is
+      // deselected wherever it sits in the selection
+      else if (value.length > 1 && value.includes(ALL_VARIABLE_VALUE)) {
```
mdvictor, line 341: I don't think $__all could ever be set among other vars, it would always normalize to either an array with one element that is $__all, or everything else? But this might be a good safeguard?

packages/scenes/src/variables/variants/MultiValueVariable.ts
```diff
@@ -281,7 +281,9 @@ export abstract class MultiValueVariable<TState extends MultiValueVariableState
 
   public hasAllValue() {
     const value = this.state.value;
-    return value === ALL_VARIABLE_VALUE || (Array.isArray(value) && value[0] === ALL_VARIABLE_VALUE);
+    // Values that bypass changeValueTo's normalisation (e.g. direct state updates) can hold
+    // the All value at any position; treating it as All avoids interpolating a literal $__all
+    return value === ALL_VARIABLE_VALUE || (Array.isArray(value) && value.includes(ALL_VARIABLE_VALUE));
```
mdvictor, line 286: changing this to verify across the array generalizes an existing bug potentially, I'm not saying this should be changed but more pointing it out as an existing bug that surfaced even more potentially.

On main -- If I change the JSON schema from the JSON editor and add something like `[ "$__all", "1", "2" ]` in the values of some custom variable for example, and apply those changes, it'll show both `All, 1 ,2` in there. I think this is not intended, but also not a big deal maybe -- on refresh it goes away.

With this change we can add `$__all` anywhere in the array and it will be kept as such, while on main the All value would get dropped. Again, would be dropped on refresh. So I don't think it's that big of a deal, but something for us to keep an eye on.

The only way I could repro this also was through directly manipulating the schema, since the UI is stricter

PR #1597, opened 2026-07-31
fix(SceneQueryRunner): secondary query behavior fixes

### LINKED ISSUE #126186: TimeComparison: Show primary query results before compare query completes
When time comparison is enabled, the panel should show current-period results as soon as the primary query completes. Compare series should appear when the compare query finishes — not block the whole panel until both are done.

Today both queries must finish before anything renders, which makes every compare panel feel twice as slow.

Fix the query runner / panel data flow so primary and compare can resolve independently.

gtk-grafana:
**What**
Avoid outside range notice when secondary returns before primary. 
Avoids dropping into error/no-data state when secondary returns error, but primary returns valid data.
Drops into no-data/error state when primary returns error, even if secondary returns valid data.
Prevents panel from rendering without visible data and no y-axis with "Data outside range" when primary is still loading.

Fixes https://github.com/grafana/grafana/issues/126186

**How**
Secondary queries should not set hasFetchedData in SceneQueryRunner.

**Why**
Secondary time compare queries should not visualize without something to compare against. 
But if the secondary fails, the primary should still be visualized.

### DIFF
--- a/packages/scenes/src/querying/extraQueryProcessingOperator.ts
+++ b/packages/scenes/src/querying/extraQueryProcessingOperator.ts
@@ -6,15 +6,18 @@ import { ExtraQueryDataProcessor } from './ExtraQueryProvider';
 export const passthroughProcessor: ExtraQueryDataProcessor = (_, secondary) => of(secondary);
 
 /**
- * Combines loading responseState across multiple responses
- * If any response contains an error => Error
- * Otherwise, if any response is loading => Loading
- * Otherwise, if any response is streaming => Streaming
- * otherwise return the first/primary response state
+ * Combines loading state across the primary response and any secondary (extra query) responses.
+ *
+ * If primary response is error => Error (set series warning meta instead?)
+ * Else if any response is loading => Loading
+ * Else if any response is streaming => Streaming
+ * Else => primary state
  * @param responseState
  */
 export function combineLoadingStates(responseState: LoadingState[]): LoadingState {
-  if (responseState.some((s) => s === LoadingState.Error)) {
+  const primaryState = responseState[0];
+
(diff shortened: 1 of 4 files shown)


drew08t (dismissed):
Works as expected for time comparison using this dashboard to test locally.

[time-comparison-primary-first.json](https://github.com/user-attachments/files/31125087/time-comparison-primary-first.json)

Would love a second set of eyes on this from @dprokop or @ivanortegaalba before merging!

packages/scenes/src/querying/extraQueryProcessingOperator.ts
```diff
@@ -6,15 +6,18 @@ import { ExtraQueryDataProcessor } from './ExtraQueryProvider';
...
 
 /**
- * Combines loading responseState across multiple responses
- * If any response contains an error => Error
- * Otherwise, if any response is loading => Loading
- * Otherwise, if any response is streaming => Streaming
- * otherwise return the first/primary response state
+ * Combines loading state across the primary response and any secondary (extra query) responses.
+ *
+ * If primary response is error => Error (set series warning meta instead?)
```
harisrozajac: i don't understand this question about series warning meta. This helper only chooses merged `data.state` from the primary (`responseState[0]`). It does not decide where errors are stored and secondary failures still get flattened onto `PanelData.errors / error` in the operator below, they just no longer force `state: Error`.

harisrozajac (dismissed):
LGMT - i left a note about one line that i was unclear about

drew08t (approved):

drew08t:
Hey @harisrozajac please re-review, thanks!

dprokop (approved):
looks good!

PR #1604, opened 2026-08-10
Adhoc/GroupBy: don't apply variables to queries when applyMode is manual

gillesdemey:
Setting `applyMode: 'manual'` on an `AdHocFiltersVariable` or `GroupByVariable` means "don't apply this to queries automatically, I'll do it myself." Scenes applies them anyway, so they end up applied twice.

When a query runs, scenes walks up the scene graph and collects every adhoc/groupby variable whose datasource matches the query's. That walk only compared datasource UIDs — it never checked `applyMode`.

This PR adds a gate to the dependency manager to make sure no variable with `applyMode === 'manual'` will be included in the variable inheritance graph.

**Note for reviewers**

Anyone who set `applyMode: 'manual'` after #1351 and came to rely on the accidental auto-application will see a behaviour change. That's the intended direction — they were getting filters applied twice — but worth a look.

Several workaround are already in place in `grafana/grafana` for Alert Notification History, Alert Activity and in the Drilldown app.

Supersedes https://github.com/grafana/scenes/pull/1429/

### DIFF
--- a/packages/scenes/src/querying/SceneQueryRunner.test.ts
+++ b/packages/scenes/src/querying/SceneQueryRunner.test.ts
@@ -650,6 +650,41 @@ describe.each(['11.1.2', '11.1.1'])('SceneQueryRunner', (v) => {
       expect(runRequestCall2[1].filters).toEqual(filtersVar.state.filters);
     });
 
+    it('should not pass adhoc filters via request object when applyMode is manual', async () => {
+      const queryRunner = new SceneQueryRunner({
+        datasource: { uid: 'test-uid' },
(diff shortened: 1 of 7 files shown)


edvard-falkskar:
Looks like this will fix my newly created issue https://github.com/grafana/scenes/issues/1607

torkelo (approved):

edvard-falkskar (commented):
Code LGTM, great with the test also for the `patchGetAdhocFilters` path. I have tested it together with App O11y plugin and it fixes our problem.

PR #1606, opened 2026-08-13
Variables: Add xxhash and xxh3 template variable format functions

justin-thurston:
## Summary

Adds two new template variable format functions:

- **`${var:xxhash}`** — hashes with **xxhash64**, zero-padded 16-char lowercase hex, seed-configurable. Pure-JS synchronous implementation.
- **`${var:xxh3}`** — hashes with **xxh3-64**, zero-padded 16-char lowercase hex, seed fixed at 0. WebAssembly implementation with a lazy-init warmup at module load.

### Motivation

Some data sources store hashed identifiers (a common pattern for high-cardinality labels in Loki/Prometheus, where the emitter replaces a serial number, tenant id, or account id with a hash at ingest for faster label matching). Users querying those data sources today have no way to type the original unhashed value in a variable input — they have to compute the hash themselves, out-of-band, and paste it. `${var:xxhash}` and `${var:xxh3}` close that gap: users type the raw value, Grafana hashes it client-side before the query is sent to the data source.

Two formatters are shipped because different producers pick different xxHash variants: `xxhash64` is the older widely-adopted default; `xxh3-64` is the newer default that many recent tools have moved to.

### Behavior

```
serial = 'SN-42'
${serial:xxhash}         -> 16-char lowercase hex xxhash64 digest, seed=0
${serial:xxhash:1234}    -> same, but with decimal seed 1234
${serial:xxhash:0x1234}  -> same, but with hex seed 0x1234
${serial:xxh3}           -> 16-char lowercase hex xxh3-64 digest, seed=0

serials = ['SN-1', 'SN-2']
${serials:xxhash}        -> '<digest-1>,<digest-2>'
${serials:xxh3}          -> '<digest-1>,<digest-2>'
```

Non-string values are stringified via `String(value)` before hashing. Invalid seed args for `xxhash` fall back to 0.

### `${var:xxhash}` — synchronous via `xxhashjs`

`xxhashjs` (pierrec/js-xxhash, ~298 stars, MIT, updated Jul 2026) is a pure-JS xxhash implementation with a synchronous API. Fits Grafana's synchronous `FormatRegistryItem.formatter(value, args, variable): string` contract without any redesign. Matches Cyan4973/xxHash reference output.

### `${var:xxh3}` — WebAssembly via `hash-wasm` with lazy init

xxh3 in pure sync JS in the browser is a puzzle: there is no widely-adopted, actively-maintained pure-JS xxh3 implementation. The realistic options were:

| Library | xxh3 variants | Sync? | Stars |
|---|---|---|---|
| `xxh3-ts` | XXH3-128 only | ✅ sync | 11 ★ |
| `hash-wasm` | XXH3-64 + XXH3-128 | ❌ async init | 1,152 ★ |
| `xxhash-wasm` | none (h32/h64 only) | — | 184 ★ |

`xxh3-ts` was tempting for its sync API but only ships XXH3-128 and has a low supply-chain score. `hash-wasm` is the mature choice, and it provides both XXH3-64 and XXH3-128 with a factory-hasher pattern whose per-call `update`/`digest` are synchronous once the WebAssembly module has initialized.

To fit `hash-wasm` into the synchronous formatter contract:

- `createXXHash3(0, 0)` is fired at module load and its resolved `IHasher` is cached in a module-level variable.
- The `${var:xxh3}` formatter checks whether the hasher is ready and, if so, calls `init()`/`update(input)`/`digest('hex')` synchronously.
- If a formatter call arrives before init has resolved (typically only on the very first render after a hard reload), the formatter returns the raw input value with a `console.warn`. A dashboard refresh interpolates the correct hash. This is the least-surprising graceful degradation given the contract mismatch.
- The initialization promise is exported as an internal `_xxh3Ready` helper for tests only. This is not intended as a public API surface.

**Seeds are intentionally not supported for `${var:xxh3}`.** `hash-wasm` binds the seed at `createXXHash3()` time (not per-hash), so supporting arbitrary per-call seeds would require a hasher-per-seed cache with an async init on each unseen seed. Users who need a seeded hash can use `${var:xxhash}`.

### Enum entries

Both formatters are registered with string-literal `id`s (`'xxhash'` and `'xxh3'`) with the standard `// not yet available in depended @grafana/schema version` comment, matching the `join` and `customqueryparam` precedent. A companion PR against `grafana/grafana` adds `XXHash = 'xxhash'` and `XXH3 = 'xxh3'` to `VariableFormatID`.

### Sync + tree-shaking

`sideEffects: false` is preserved for the `xxhash` formatter (`xxhashH64` is only invoked inside the formatter closure). The `xxh3` formatter has one intentional top-level side effect: `createXXHash3(0, 0).then(...)`. Because `formatRegistry` is imported and used by grafana's `formatVariableValue`, the module always evaluates and the init always fires; `sideEffects: false` remains truthful at the package level for consumers that never touch the registry.

## Companion PR

- grafana/grafana#130648 — adds the `VariableFormatID.XXHash` and `VariableFormatID.XXH3` enum entries and the user-facing docs.

## Test plan

- [x] `formatRegistry.test.ts` — new `describe('xxhash')` block with 7 tests (empty-string reference vector, output shape, determinism, seed variance, non-string coercion, multi-value arrays, invalid seed fallback).
- [x] `formatRegistry.test.ts` — new `describe('xxh3')` block with 6 tests, gated on `_xxh3Ready()` in `beforeAll` so tests wait for WASM init (empty-string reference vector `2d06800538d394c2`, output shape, determinism, non-string coercion, multi-value arrays, cross-formatter divergence vs xxhash).
- [ ] Reviewer: run `yarn test packages/scenes/src/variables/interpolation/formatRegistry.test.ts` after `yarn install` picks up the new `xxhashjs` and `hash-wasm` deps. `hash-wasm` uses WASM; if the jest jsdom env cannot instantiate the WebAssembly module, the `xxh3` tests may need a jest-environment tweak — please flag if that happens.

### DIFF
--- a/packages/scenes/package.json
+++ b/packages/scenes/package.json
@@ -58,7 +58,8 @@
     "react-select": "^5.10.2",
     "react-use": "^17.5.0",
     "react-virtualized-auto-sizer": "^1.0.24",
-    "uuid": "^11.0.0"
+    "uuid": "^11.0.0",
+    "xxhashjs": "^0.2.2"
   },
   "peerDependencies": {
     "@grafana/data": ">=11.6",
@@ -84,6 +85,7 @@
     "@types/react-grid-layout": "1.3.6",
     "@types/react-virtualized-auto-sizer": "1.0.8",
     "@types/systemjs": "^6.15.1",
+    "@types/xxhashjs": "^0.2.4",
     "eslint": "^9.28.0",
     "i18next-parser": "9.4.0",
     "jest": "30.4.2",
--- a/packages/scenes/package.json
+++ b/packages/scenes/package.json
@@ -52,6 +52,7 @@
     "@floating-ui/react": "^0.27.0",
     "@leeoniya/ufuzzy": "^1.0.16",
     "@tanstack/react-virtual": "^3.9.0",
+    "hash-wasm": "^4.12.0",
     "history": "^4.9.0",
     "lodash": "^4.17.21",
     "react-grid-layout": "^1.3.4",
--- a/packages/scenes/src/variables/interpolation/formatRegistry.test.ts
+++ b/packages/scenes/src/variables/interpolation/formatRegistry.test.ts
@@ -159,6 +159,66 @@ describe('formatRegistry', () => {
     });
   });
 
+  describe('xxhash', () => {
+    // Well-known xxhash64 reference vector for the empty string with seed=0.
+    it('hashes the empty string to the reference xxhash64 vector (seed=0)', () => {
+      // @ts-expect-error xxhash not in depended @grafana/schema yet
+      expect(formatValue('xxhash', '')).toBe('ef46db3751d8e999');
+    });
+
+    it('produces zero-padded 16-char lowercase hex', () => {
+      // @ts-expect-error xxhash not in depended @grafana/schema yet
+      const out = formatValue('xxhash', 'serial-12345');
+      expect(out).toMatch(/^[0-9a-f]{16}$/);
+    });
+
+    it('is deterministic for the same input and seed', () => {
+      // @ts-expect-error xxhash not in depended @grafana/schema yet
+      const first = formatValue('xxhash', 'SN-42');
+      // @ts-expect-error xxhash not in depended @grafana/schema yet
+      const second = formatValue('xxhash', 'SN-42');
+      expect(first).toBe(second);
+    });
+
+    it('changes when the seed changes', () => {
+      // @ts-expect-error xxhash not in depended @grafana/schema yet
+      const withoutSeed = formatValue('xxhash', 'SN-42');
+      // @ts-expect-error xxhash not in depended @grafana/schema yet
+      const withDecimalSeed = formatValue('xxhash', 'SN-42', 'text', ['1']);
+      // @ts-expect-error xxhash not in depended @grafana/schema yet
+      const withHexSeed = formatValue('xxhash', 'SN-42', 'text', ['0x1']);
+      expect(withoutSeed).not.toBe(withDecimalSeed);
+      expect(withDecimalSeed).toBe(withHexSeed);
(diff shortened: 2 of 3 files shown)


justin-thurston:
Cursor Bugbot flagged three items worth calling out; leaving these here so any human reviewer coming in sees my read on each:

1. **[Medium] `xxhash64` seeds truncated to 32 bits.** Real limitation — `parseXxhashSeed` returns a JS `number` and xxhashjs constructs its `UINT64` via `| 0`, so seeds ≥ 2^32 silently truncate. The follow-ups I'm considering, in order of increasing scope:
   - Document the 32-bit seed limit in the description/docs (most producers use ≤ 32-bit seeds anyway).
   - Parse seeds as BigInt and construct a proper UINT64 via `cuint` for values that don't fit in 32 bits.
   - Explicitly throw / warn when a user supplies a > 32-bit seed rather than silently truncating.
   Which of these is preferable is a design call — happy to take direction.

2. **[High] `xxh3` interpolates unhashed values before WASM init completes.** Acknowledged in the PR description as intentional graceful degradation. The formatter contract is synchronous and hash-wasm's init is async; the alternatives (throwing on early call, blocking the render, switching to `xxh3-ts` at 11★, or bumping the format function contract to async) each have tradeoffs I called out in the PR body. Genuinely open to a reviewer preference here.

3. **[Medium] `xxh3` init is a load-side side effect vs `sideEffects: false`.** Fair point that under aggressive tree-shaking the init could be dropped. Easiest fix: move `createXXHash3(0, 0)` inside a lazy initializer that fires on the first `xxh3Hex` call. Slightly worse first-hash race but simpler to reason about. Can push that as a follow-up commit if it looks good.

Two blockers on my end: I still need to sign the CLA and re-push signed commits — will handle both.

PR #1610, opened 2026-08-17
AdhocFilters: Fixes focus state when visual refresh is enabled

torkelo:
Noticed that adhoc filters combobox focus state was white when visual refresh was enabled


ashharrison90 (approved):
lgtm 👍

PR #1614, opened 2026-08-18
SceneDataTransformer: support system and URL provided transformations

dprokop:
Adds support for system transformations in scenes. 

System transformations are currently:
* Provided by `SceneDataTransformer` parent (e.g. `VizPanel`, gated with `applyPluginTransformations` state flag)
* Not saved in `state.transformations`
* * read with `getResolvedSystemTransformations`
* Discovered on `SceneDataTransformer` activation
* Prepended before or appended after user transformations
* Tagged with `origin` (i.e. `plugin`) and `position` (`prepend`, `append`)
* Noop when system transformation provider throw

Also fixes variable interpolation dropping the `operator` function from object-form custom transformers.

<details>
  <summary>📦 Published PR as canary version: <code>9.0.0--canary.1614.35114143706.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@9.0.0--canary.1614.35114143706.0
  npm install scenes-app@9.0.0--canary.1614.35114143706.0
  npm install @grafana/scenes-react@9.0.0--canary.1614.35114143706.0
  # or 
  yarn add @grafana/scenes@9.0.0--canary.1614.35114143706.0
  yarn add scenes-app@9.0.0--canary.1614.35114143706.0
  yarn add @grafana/scenes-react@9.0.0--canary.1614.35114143706.0
  ```
</details>


packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -105,6 +142,69 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
...
+
+    const transformations = [
+      ...groups.system.prepend,
+      ...groups.url.prepend,
+      ...userTransformations,
+      ...groups.url.append,
+      ...groups.system.append,
+    ];
+
+    if (isEqual(transformations, this.state.transformations)) {
```
gtk-grafana: custom transforms of the form `() => (source) => source` always fail this check since [isEqual](https://lodash.info/doc/isEqual) compares functions by reference 
```
(() => (source) => source) === (() => (source) => source)
/// returns false
```

I think we'll need a `key` param added to the `CustomTransformerDefinition` so we can compare custom transforms

gtk-grafana: Added in https://github.com/grafana/scenes/pull/1614/changes/78eb43738bb7d1e1a15c77086e1c2b220aa7e89b

packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -105,6 +142,69 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
...
+   * configured ones. Prepended transformations run before the user transformations, appended ones after.
+   * Each provided transformation is tagged with the origin (default 'system'). Previous transformations
+   * with the same origin are replaced, so repeated calls (e.g. on every data render from a panel) are
+   * idempotent; transformations from other origins are preserved.
+   *
+   * Resulting pipeline order: system prepend, url prepend, user, url append, system append -
+   * panel provided (system) transformations wrap everything, url provided ones sit closest
+   * to the user configured transformations.
+   */
+  public setSystemTransformations({
```
gtk-grafana: should we call these `setRuntimeTransformations` and `isRuntimeTransformation` instead since they set `url` and `system` transforms?

gtk-grafana: Nah I think `system` and `user` transformations work fine

gtk-grafana:
@dprokop I made a few minor changes while working on the grafana implementation and PoCs
1. Fix: custom transforms calling setState on every call of `setSystemTransformations`: https://github.com/grafana/scenes/pull/1614/commits/78eb43738bb7d1e1a15c77086e1c2b220aa7e89b
2. Fix: useDataTransformer wiping out system transforms (note adds new public `setUserTransformations` method): https://github.com/grafana/scenes/pull/1614/commits/fdb7af0e955d3f4a17320b95c35e455176b5fce0
3. Test: add coverage for interpolation https://github.com/grafana/scenes/pull/1614/commits/10c9c4316803e87fe81b807e1e0bb1a875f33417
4. Chore: `setSystemTransformations` types https://github.com/grafana/scenes/pull/1614/commits/d177c345f464fb84257234e304ff036ad831db9f
5. Fix: bug introduced in setUserTransformations refactor https://github.com/grafana/scenes/commit/3074fbb24ad810388d6b67305d427c89308c7892 + regression test
6. [Address question from Dom](https://github.com/grafana/scenes/pull/1614#discussion_r3821465062) asking if we should remove `url` and rename `system` to `plugin` https://github.com/grafana/scenes/commit/00e5b54e18ce167ef3aa72fe779fabc0f1b11dfc
7. Regression with null transformations in v1 dashboards: https://github.com/grafana/scenes/pull/1614/commits/456fa65c3558ea278738c0d417c80f0e2d525923 : https://github.com/grafana/scenes/pull/1614#discussion_r3832868551

Other then that everything looks good, marking as ready for review

packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -105,6 +169,100 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
...
+   * Replaces the user configured transformations, keeping any system transformations in place around them.
+   * Use this instead of setState({ transformations }) from a transformations editor or a React binding:
+   * writing the array directly would drop the runtime transformations added via setSystemTransformations,
+   * since they live in the same array.
+   */
+  public setUserTransformations(transformations: SceneDataTransformation[]) {
+    const { system } = this._partitionTransformations();
+
+    this._applyTransformations(this._combineTransformations(system, transformations));
+  }
```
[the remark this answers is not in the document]
gtk-grafana: Good bot, fixed in https://github.com/grafana/scenes/pull/1614/commits/3074fbb24ad810388d6b67305d427c89308c7892

packages/scenes/src/core/types.ts
```diff
@@ -225,8 +225,33 @@ export type SceneObjectUrlValue = string | string[] | undefined | null;
...
 
 export type CustomTransformOperator = (context: DataTransformContext) => MonoTypeOperatorFunction<DataFrame[]>;
+
+/**
+ * Identifies where a transformation comes from. Transformations without an origin are user configured.
+ * Transformations with an origin are runtime transformations added programmatically via
+ * SceneDataTransformer.setSystemTransformations - 'system' for panel provided ones, 'url' for ones
+ * provided through the URL. They are not meant to be persisted or edited in the transformations editor.
+ */
+export type TransformationOrigin = 'system' | 'url';
```
dprokop: @gtk-grafana / @leeoniya   - do you think we need this explicit differentiation? Maybe instead of using 'system' we should use 'plugin'

gtk-grafana: It's tough to know before the ad-hoc part is implemented, I can foresee needing to tag saved transformations as coming from a url/ad-hoc transform, but system transforms will never really apply for that use-case, so the differentiation might be necessary down the road but not strictly needed today.

To be defensive I'd agree that we should rename as plugin, remove `url`, and follow up with additional options to the discriminated union when ad-hoc or something else needs the differentiation

dprokop: 💯

gtk-grafana: Updated in https://github.com/grafana/scenes/pull/1614/commits/00e5b54e18ce167ef3aa72fe779fabc0f1b11dfc

packages/scenes/src/core/types.ts
```diff
@@ -225,8 +225,37 @@ export type SceneObjectUrlValue = string | string[] | undefined | null;
...
+ * origin would delete it - which is why provenance that outlives promotion needs its own persisted field.
+ */
+export type TransformationOrigin = 'plugin';
+
+/**
+ * Whether a system transformation runs before ('prepend') or after ('append') the user configured
+ * transformations. Stored on the transformation because the placement can not be derived from the
+ * array position when there are no user transformations.
+ */
+export type SystemTransformationPosition = 'prepend' | 'append';
```
mdvictor: what would be the point of appending transformations? I might miss something here, I'm not sure what use-case this would serve. For prepends it feels natural because the panel would expect the data in some form and the system transforms would make sure the user sees something at the end of the panel, but for appends I guess I can see very little transformation usage idk like naming/organizing the fields in some specific way, but still would be very dependent on what the user adds as transformations too? 

What I'm envisioning is a scenario where the user adds some of his transformations, which together with the appends crash and the user has no idea what happened because they are not aware of the post appended transformations. Can that be a problem? Or would we mitigate this with some explanations on the panel?

mdvictor (dismissed):
this LGTM! I like the approach on the sceneDataTransformer, left a comment related to the system transformations appending functionality, but otherwise I think this is great!

packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -13,16 +13,91 @@ import { toDataQueryError } from '@grafana/runtime';
...
+
+/**
+ * Returns true for transformations added via SceneDataTransformer.setSystemTransformations, whichever origin
+ * injected them. That is the question persisting and editing want to ask; use isTransformationFrom when a
+ * provider needs to reason about only the entries it owns.
+ */
+export function isSystemTransformation(
+  transformation: SceneDataTransformation
+): transformation is SystemTransformation {
+  return typeof transformation === 'object' && 'origin' in transformation && transformation.origin != null;
```
gtk-grafana:
```suggestion
  return typeof transformation === 'object' && transformation !== null && 'origin' in transformation && transformation.origin != null;
```

If a v1 dashboard has `[null]` provisioned for the transformations (which can only happen in externally authored JSON) the dashboard will throw an error with the flag off.


Repro:
```bash
curl -s -u admin:admin -H 'Content-Type: application/json' \
  http://localhost:3000/api/dashboards/db -d '{
  "dashboard": {
    "title": "null-transformation-repro",
    "schemaVersion": 39,
    "panels": [{
      "type": "timeseries", "title": "P1", "id": 1,
      "gridPos": {"h": 8, "w": 12, "x": 0, "y": 0},
      "targets": [{"refId": "A"}],
      "transformations": [null]
    }]
  },
  "overwrite": true
}'
```

Custom.ini
```
dashboardNewLayouts = false
grafana.panelPluginTransformations=false
```

packages/scenes/src/core/types.ts
gtk-grafana: We're purposely keeping the scenes implementation wide so panel developers can address a variety of different problems. Appending is really more for upcoming functionality then it is for the currently defined prepend system transform use-cases:

The pre-requisite "adaptor" transformations are always going prepend, but viewer added "ad-hoc" transformations (e.g. a non-editor user hiding or reordering columns in a table) also fall in the "system" transformation bucket, and they will always append after the user transforms are applied, as they originate from the rendered panel UI (e.g. table column drag-and-drop can't apply to fields that might not exist after user transforms).

gtk-grafana: But @mdvictor your comment might support the argument to rename "system" to "runtime" (especially if ad-hoc decides to not differentiate transforms generated by UI interactions in the panel against transforms added in the classic transformation UI), since they fundamentally are just transforms that don't show up in the dashboard spec.

But I don't have strong feelings either way.

mdvictor (dismissed):

packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -102,9 +228,252 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
...
+   *
+   * Defaults to the frames the pipeline is working on, which is what a caller asking "what is running
+   * right now" wants, and shares the pipeline's memo instead of resolving the suppliers a second time.
+   * Do not reach for this object's own `state.data.series` to fill the argument: that is pipeline output,
+   * and resolving a supplier against it asks the question the supplier exists to avoid. Pass frames
+   * explicitly only to ask about a set this transformer is not currently running.
+   */
+  public getResolvedSystemTransformations(
+    series: DataFrame[] = this.getSourceData().state.data?.series ?? []
+  ): ResolvedSystemTransformations {
```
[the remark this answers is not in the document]
gtk-grafana: Fixed in https://github.com/grafana/scenes/pull/1614/commits/ed73af9a23736c5e020395c4e3ca5f9ce53d70b0

gtk-grafana:
After @dprokop's suggestion [here](https://github.com/grafana/grafana/pull/131174#discussion_r3842667788), we elected to add a new supplier in the system transforms to solve the additional complexity from the new subclass in core, which still allows for the supplier needing pre-transform frames, and opens up usage in explore and scenes apps instead of only applying to dashboards. 

### Changes since last review:

New supplier: https://github.com/grafana/scenes/pull/1614/commits/34b01c842fb3fadcf309b4f7333aec2968bab14a
Clone fix: https://github.com/grafana/scenes/pull/1614/commits/05090da52e54fdb1c63a4019309b4a2fcfa4d031

#### A few other minor tweaks:
Types clean up: https://github.com/grafana/scenes/pull/1614/commits/df8275d94fe53fa21f1b1a89d2b5c63efb65503d
Set default series in getResolvedSystemTransforms: https://github.com/grafana/scenes/pull/1614/commits/f01074cd6eb79613b3117bf86d0a1508d1d28fa1
Co-pilot review comment fix: https://github.com/grafana/scenes/pull/1614/commits/ed73af9a23736c5e020395c4e3ca5f9ce53d70b0

Sorry @mdvictor for pulling the rug out from under you yet again, but we're getting there! It looks like the additional supplier makes the grafana side changes considerably easier to review, so should be net-win with the latest change here.

gtk-grafana:
Aaaand in the core implementation we ran into needing 3 more small changes:
1. One-line fix: https://github.com/grafana/scenes/pull/1614/commits/fc5d4745e7fb555fbb22f3e14ef6eb048d0aad56 - `transform()`'s empty-pipeline early return leaks the previous pass
2. Skip the forced re-run when a registration resolves to the same thing - https://github.com/grafana/scenes/pull/1614/commits/d7ff11c86137a4f472df96666109807e41b4b01e - slightly bigger change, not strictly necessary, but the fix can remove some unnecessary reprocessing
3. don't call setState when data hasn't changed - also not necessary, but prevents subscribers from getting events on noop https://github.com/grafana/scenes/pull/1614/commits/40da41a226287e9cb725489a09ca5a05a3453f94

harisrozajac (dismissed):

packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -102,9 +235,296 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
...
+   * builds them inline would re-run the pipeline and emit new data on every call, which loops if it
+   * re-applies on data change. Give such operators a stable `key` to declare identity instead - matching
+   * keys make the operator reference irrelevant, and changing the key signals a real change.
+   *
+   * Resulting pipeline order: prepended system transformations, the user configured ones, then appended
+   * system transformations. Origins are replaced independently, so once a second provider exists its
+   * entries slot into the same prepend and append tiers without disturbing this one's, ordered by which
+   * origin called this first.
+   */
+  public setSystemTransformations({
```
torkelo: Why do we need to store system transformations in the SceneDataTransformer state? 

can't it just ask the "scene" (it walks up the scene tree, until it finds a scene object that implements SystemTransformationSupplier), this VizPanel can implement and get the transformations from the PanelPlugin. 

That way the PanelPlugin API could be something like 

```ts
getPanelTransformation(userTransformations) 
```

and the panel can choose to prepend or append (or both)

torkelo: all the logic to group and filter transfomations would go away

packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -102,9 +235,296 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
...
+   * no transformations resolves empty both before and after, and forcing a pass for that costs a redundant
+   * emission per panel on every dashboard load, since `$data` activates before whatever registers the
+   * supplier and the registration therefore always lands after the first pass.
+   *
+   * Answering it calls both suppliers once for the current frames, which is why the contract asks for them
+   * to be cheap and free of side effects. Comparing the two directly rather than against the last pass
+   * keeps the question narrow: does this swap change anything? A supplier whose own output has drifted for
+   * unrelated reasons is what reprocessTransformations is for.
+   */
+  private _supplierSwapChangesResolution(
```
torkelo: not sure I understand this function/comment

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -351,6 +360,116 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
...
+
+        const next = this._pluginForTransformations();
+
+        if (next === this._pluginForTransformationsResolved?.plugin) {
+          return;
+        }
+
+        this._pluginForTransformationsResolved = { plugin: next };
+        callback();
+      });
```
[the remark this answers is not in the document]
gtk-grafana: Outside the scope of this PR, this is pre-existing bug on main and not impacted by the changes on this branch.

packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -71,6 +101,35 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
...
+   */
+  private _discoverProvider() {
+    const provider = this.parent;
+
+    if (!isSystemTransformationsProvider(provider)) {
+      // This only ever runs from the activation handler, where the parent is already final, so a parent
+      // that is not a provider means anything discovered on a previous activation belongs to a tree this
+      // transformer is no longer part of.
+      this._provider = undefined;
+      this._resolvedSystem = undefined;
```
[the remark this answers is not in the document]
gtk-grafana: Good catch, addressed in  [`ac9f862`](https://github.com/grafana/scenes/pull/1614/commits/ac9f862326a8934011ce7760d4fa78b000f947d8)

gtk-grafana:
Rewritten to check for parents of `SceneDataTransformer` that implement `SystemTransformationsProvider`. PTAL!

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -351,6 +365,113 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
...
+   * This method "warms" the import cache and re-reads it synchronously.
+   *
+   * A data provider can stay active while its panel never renders (e.g. a dashboard datasource panel)
+   * In which case _onActivate never runs and the plugin is never imported.
+   * It does not go through _loadPlugin, which runs option migrations and writes state on an inactive panel.
+   *
+   * Done from a subscription rather than getSystemTransformations because a failed import is
+   * evicted from the cache and would be retried on every pass.
+   */
+  private _warmPluginForTransformations(transformer: SceneDataTransformer, callback: () => void) {
```
torkelo: I think this really warrants a refactor of plugin loading in VizPanel so it can be async and awaited, something we have needed for a very long time and hacked around in core.

packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -102,9 +173,91 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
...
+    // is the reason callers reach for this - so the memo cannot survive a forced re-run.
+    this._resolvedSystem = undefined;
     this.transform(this.getSourceData().state.data, true);
   }
 
+  /**
+   * The system transformations for the given source frames.
+   * This is the source of truth to know what the pipeline is running since provider output never reaches state.
+   */
+  public getResolvedSystemTransformations(series?: DataFrame[]): ResolvedSystemTransformations {
```
torkelo, line 213: does this need to be public? where is it called from?

packages/scenes/src/querying/SceneQueryRunner.ts
```diff
@@ -804,7 +804,7 @@ export class SceneQueryRunner extends SceneObjectBase<QueryRunnerState> implemen
     request.interval = norm.interval;
     request.intervalMs = norm.intervalMs;
 
-    // If there are any extra request providers, we need to add a new request for each
+    // If there are any extra request systemTransformations, we need to add a new request for each
```
torkelo: ?

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
```diff
@@ -552,7 +552,7 @@ export class AdHocFiltersVariable
   /**
    * Updates the variable's `filters` and `filterExpression` state.
    * If `skipPublish` option is true, this will not emit the `SceneVariableValueChangedEvent`,
-   * allowing consumers to update the filters without triggering dependent data providers.
+   * allowing consumers to update the filters without triggering dependent data systemTransformations.
```
torkelo: ?

packages/scenes/src/variables/DrilldownDependenciesManager.ts
```diff
@@ -22,7 +22,7 @@ import { getScopes } from '../core/sceneGraph/sceneGraph';
 import { SCOPES_VARIABLE_NAME } from '../variables/constants';
 
 /**
- * Manages ad-hoc filters and group-by variables for data providers.
+ * Manages ad-hoc filters and group-by variables for data systemTransformations.
```
torkelo: ?

packages/scenes/src/querying/SceneQueryRunner.test.ts
```diff
@@ -2202,7 +2202,7 @@ describe.each(['11.1.2', '11.1.1'])('SceneQueryRunner', (v) => {
       expect(queryRunner.state.data?.series.map((s) => s.refId)).toEqual(['A', 'secA', 'secB']);
     });
 
-    it('merges secondaries contributed by multiple providers', async () => {
+    it('merges secondaries contributed by multiple systemTransformations', async () => {
```
torkelo: think claude is conflating the word provider here to with system transformation provider

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -351,6 +365,113 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
...
+    const plugin = this._pluginForTransformations();
+
+    return plugin ? getPluginSystemTransformations(plugin, { series }) : {};
+  }
+
+  /**
+   * @internal
+   * The plugin resolves asynchronously, so the pipeline is told to reprocess whenever the plugin changes.
+   */
+  public subscribeToSystemTransformationsChanged(
```
torkelo: a little bit tempted to instead of this subscription call reprocessTransformation from VizPanel changePlugin (if $data is SceneDataTransformer). feels like it might be less complex?

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -351,6 +365,113 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
...
+  /**
+   * @internal
+   * SystemTransformationsProvider. Identifies this panel's tier in its data pipeline; not for app code.
+   */
+  public origin: TransformationOrigin = 'plugin';
+
+  /**
+   * The plugin the data pipeline last resolved against.
+   */
+  private _pluginForTransformationsResolved?: { plugin: PanelPlugin | undefined };
```
torkelo: why do we need this and `private _plugin?: PanelPlugin;` on line 119? 

I think we need a async plugin-loading refactoring, ie 

```ts
  public async getPluginAsync(): Promise<PanelPlugin | undefined> {
    ....
  }
```

so we can use this from core when we need the plugin, or here when we need the transformations. So we only have one plugin state (cache), and one place we load the plugin

torkelo (commented):
Looks good except the duplicate plugin loading in VizPanel and some I think faulty comment changes

packages/scenes/src/querying/SceneQueryRunner.ts
gtk-grafana: Bad rename, sorry about that fixed in [refactor: fix bad rename](https://github.com/grafana/scenes/pull/1614/commits/0bf93ccf66f85a3d51f18ba2c168581f0fccf8d7)

packages/scenes/src/variables/adhoc/AdHocFiltersVariable.tsx
gtk-grafana: Fixed in [refactor: fix bad rename](https://github.com/grafana/scenes/pull/1614/commits/0bf93ccf66f85a3d51f18ba2c168581f0fccf8d7)

packages/scenes/src/variables/DrilldownDependenciesManager.ts
gtk-grafana: [refactor: fix bad rename](https://github.com/grafana/scenes/pull/1614/commits/0bf93ccf66f85a3d51f18ba2c168581f0fccf8d7)

packages/scenes/src/querying/SceneQueryRunner.test.ts
gtk-grafana: [refactor: fix bad rename](https://github.com/grafana/scenes/pull/1614/commits/0bf93ccf66f85a3d51f18ba2c168581f0fccf8d7)

packages/scenes/src/querying/SceneDataTransformer.ts
gtk-grafana: called by [public/app/features/dashboard-scene/scene/systemTransformations.ts:31](https://github.com/grafana/grafana/pull/132111/changes#diff-d13d0b23b9001ae35c6a7d9c79d9589e262e3214720e8b2bfad65533df38b909R31) and [public/app/features/dashboard-scene/panel-edit/PanelDataPane/PanelDataTransformationsTab.tsx:118](https://github.com/grafana/grafana/pull/132111/changes#diff-14cff1b56d064c986970c8a9a17223deb5ecbb5f0c21fedd3a806702464baaddR118), since the provider output isn't added to state.transformations this is the mechanism for grafana to display the UI.

Clarified the comment in [`141d0ce`](https://github.com/grafana/scenes/pull/1614/commits/141d0ce95a65cfcfc9f9906fbf4c1fd60d56a604)

packages/scenes/src/components/VizPanel/VizPanel.tsx
gtk-grafana: Cleaned up in [`39c9828`](https://github.com/grafana/scenes/pull/1614/commits/39c9828bb9f00d27ffdf3ae99aa417f41b582e16)

packages/scenes/src/components/VizPanel/VizPanel.tsx
gtk-grafana: `reprocessTransformations` from `changePluginType` misses the plugin arriving async after the transformer's first pass, and panels that never activate (dashboard datasources loading data from non-rendered panels), but [`39c9828`](https://github.com/grafana/scenes/pull/1614/commits/39c9828bb9f00d27ffdf3ae99aa417f41b582e16) was able to clean up some of the additional complexity here

packages/scenes/src/components/VizPanel/VizPanel.tsx
gtk-grafana: cleaned up in [`39c9828`](https://github.com/grafana/scenes/pull/1614/commits/39c9828bb9f00d27ffdf3ae99aa417f41b582e16)

packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -60,17 +91,57 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
 
     this._subs.add(sourceData.subscribeToState((state) => this.transform(state.data)));
 
+    // Before the first transform below, so a discovered provider is part of that first pass rather than
+    // needing a second, corrective one.
+    const providerChanged = this._discoverProvider();
+
     if (sourceData.state.data) {
-      this.transform(sourceData.state.data);
+      this.transform(sourceData.state.data, providerChanged);
```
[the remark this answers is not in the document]
gtk-grafana: Good bot. Fixed & added coverage in [`1c55fd9`](https://github.com/grafana/scenes/pull/1614/commits/1c55fd96ce28969576f9b916c177f64e41763e28)

packages/scenes/src/querying/SceneDataTransformer.ts
```diff
@@ -256,13 +455,19 @@ export class SceneDataTransformer extends SceneObjectBase<SceneDataTransformerSt
...
       // Calculate transformation complexity metrics
-      const metrics = this._calculateTransformationMetrics(data, this.state.transformations);
+      const metrics = this._calculateTransformationMetrics(data, transformations);
 
       // Start the DataProcessing phase with centralized logging - get end callback
       endTransformCallback = profiler.onDataTransformStart(timestamp, transformationId, metrics);
     }
 
-    const interpolatedTransformations = this._interpolateVariablesInTransformationConfigs(data);
+    this._lastPassSystem = system;
```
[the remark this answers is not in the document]
gtk-grafana: Fixed in [`05aad61`](https://github.com/grafana/scenes/pull/1614/commits/05aad610405b9060b180f0889666a1bffaba0090)

packages/scenes/src/components/VizPanel/VizPanel.tsx
torkelo: not sure I understand, how would changePluginType be called from those scenarios? 

I thought this was only to handle plugin type changes from panel edit?

gtk-grafana: Yeah apologies, that was not clear. `changePluginType` is only for panel edit changes, I had meant to say in those scenarios `changePluginType` isn't called, so a reprocess in `changePluginType` wouldn't cover those cases.

We are currently use the subscription:
1. First load, since the plugin is loaded async we need the subscription to run the transform after the plugin has resolved. 
2. On `applyPluginTransformations` state change
3. Inactive panels, where we might not even have the plugin loaded
4. VizType change, which `changePluginType` should be able to replace

But after digging in a bit more I think we can delete this sub if we reprocess in `_pluginLoaded`:

1. First load: trivially covered
2. `applyPluginTransformations` not covered, but since the feature flag drives this the chance of it getting flipped mid-transform doesn't warrant the effort
3. Inactive panels: not covered, but since the dashboard datasource is the only case, and I think it calls `activateSceneObjectAndParentTree`, so it looks like we added the subscription for this case only to cover a theoretical API shape instead of solving an actual problem
4. VizType change: covered with `changePluginType` calling `_pluginLoaded`

We still might want a subscription for ad-hoc/url transforms, but we can cross that bridge when we come to it. 

I'm just wrapping up the subscription removal locally and should have it pushed up in a few. Great catch!

gtk-grafana: Removed `subscribeToSystemTransformationsChanged` in [`7170dbf`](https://github.com/grafana/scenes/pull/1614/commits/7170dbf42dcc3ad44a6a16d21e3f76562fa422a2)

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -336,6 +367,11 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
       pluginId: plugin.meta.id,
     });
 
+    // The plugin arrives async so the pipeline may have already run a pass without its system transformations.
+    if (this.state.applyPluginTransformations && $data instanceof SceneDataTransformer && $data.isActive) {
+      $data.reprocessTransformations();
+    }
```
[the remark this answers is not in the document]
gtk-grafana: Yup, and we decided that this was an acceptable tradeoff for removing the subscription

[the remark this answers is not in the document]
gtk-grafana: I triple checked and the assumption that a dashboard datasource doesn't activate source panels is untrue, on both canaries before and after removing the subscription we see the exact same behavior where the source panel is activated independently of the transformations being applied in the destination dashboard datasource panel before the data is ever rendered. 

This appears to be the expected behavior on main due to Grafana's `[datasource.ts](https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/dashboard/datasource.ts#L108)`calls `activateSceneObjectAndParentTree`, which activates the source panel, so the destination panel only receives frames after `_pluginLoaded` has processed on the source and system transformations are already avialable.

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -106,6 +113,11 @@ export interface VizPanelState<TOptions = {}, TFieldConfig = {}> extends SceneOb
    **/
   _UNSAFE_clearPreviousFieldValues?: boolean;
 
+  /**
+   * Whether this panel provides system transformations
+   */
+  applyPluginTransformations?: boolean;
```
torkelo, line 119: why is this an option on VizPanel, is it not just a PanelPlugin option? or is this to control it with a feature toggle?

packages/scenes/src/components/VizPanel/VizPanel.tsx
torkelo: should this not check if the plugin provides transformations before calling reprocessTransformations?

packages/scenes/src/querying/systemTransformations/systemTransformationTypes.ts
```diff
@@ -0,0 +1,59 @@
...
+  ): {
+    prepend?: Array<DataTransformerConfig | CustomTransformerDefinition>;
+    append?: Array<DataTransformerConfig | CustomTransformerDefinition>;
+  };
+
+  /**
+   * Optional. Asks to be told when what getSystemTransformations resolves to may have changed without new
+   * data arriving.
+   */
+  subscribeToSystemTransformationsChanged?(transformer: SceneDataTransformer, callback: () => void): Unsubscribable;
```
torkelo: do we really need this? what scenario is this for?

packages/scenes/src/components/VizPanel/VizPanel.tsx
gtk-grafana: Yup, this is to control with the `grafana.panelPluginTransformations` feature flag (see https://github.com/grafana/grafana/pull/132109)

gtk-grafana: Made this a bit more clear in [`77c6cec`](https://github.com/grafana/scenes/pull/1614/commits/77c6cec6381c8ff0476ecf3be7c011748ddcee95)

packages/scenes/src/components/VizPanel/VizPanel.tsx
gtk-grafana: Yes it should, otherwise enabling the flag will add unnecessary reprocess for panels without provider. Fixed in [`77c6cec`](https://github.com/grafana/scenes/pull/1614/commits/77c6cec6381c8ff0476ecf3be7c011748ddcee95)

packages/scenes/src/querying/systemTransformations/systemTransformationTypes.ts
gtk-grafana: No, this was added to rerun pipeline when the plugin loaded after the transform pass had already ran, but after the refactor to remove the VizPanel subscription this was dead code that should have been cleaned up.

Removed in [`77c6cec`](https://github.com/grafana/scenes/pull/1614/commits/77c6cec6381c8ff0476ecf3be7c011748ddcee95)

gtk-grafana: `transformer` param in `getSystemTransformations` was also unused, removed in [`7d30c21`](https://github.com/grafana/scenes/pull/1614/commits/7d30c214b7fbe41ecd4d4d878629856f6f976261)

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -336,6 +367,17 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
...
     });
 
+    // The plugin arrives async so the pipeline may have already run a pass without its system
+    // transformations.
+    if (
+      this.state.applyPluginTransformations && // does panel support system transformations?
+      pluginProvidesSystemTransformations(plugin) && // does the plugin supply provider method?
+      $data instanceof SceneDataTransformer &&
+      $data.isActive
+    ) {
```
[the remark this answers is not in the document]
gtk-grafana: Fixed in [`900c38e`](https://github.com/grafana/scenes/pull/1614/commits/900c38e470fa1155f4b3468c60c9d3b30f26948f)

gtk-grafana:
I think we're finally pretty close here, if anyone has a few minutes for final review I would greatly appreciate it!

packages/scenes/src/components/VizPanel/VizPanel.tsx
torkelo: Would it not make more sense to have SceneDataTransformer check the feature flag? 

Feels messy to add this to VizPanel just for feature flag

torkelo (dismissed):

packages/scenes/src/querying/systemTransformations/systemTransformationProvider.ts
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Narrows SystemTransformationsProvider
+ */
+export function isSystemTransformationsProvider(o: unknown): o is SystemTransformationsProvider {
+  return (
+    typeof o === 'object' &&
+    o !== null &&
+    'origin' in o &&
+    typeof (o as SystemTransformationsProvider).getSystemTransformations === 'function'
+  );
```
mdvictor: This is a bit nitty, but I think it would be easier to have some marker flag on the interface like `isSystemTransformationProvider = true` and just check here that `return 'isSystemTransformationProvider' in o`

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -347,10 +390,67 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}> extends Scene
...
+    const { pluginId } = this.state;
+
+    return this._plugin?.meta.id === pluginId ? this._plugin : loadPanelPluginSync(pluginId);
+  }
+
+  /**
+   * @internal
+   * SystemTransformationsProvider. Identifies this panel's tier in its data pipeline; not for app code.
+   */
+  public origin: TransformationOrigin = 'plugin';
```
mdvictor: This could maybe pose some problems in the future with it being public and tagged internal and 'not for app code'. Should we enforce this better for the future as it can be set externally? Maybe have it readonly?

mdvictor (dismissed):
LGTM, left a couple of nits!

packages/scenes/src/querying/systemTransformations/systemTransformationProvider.ts
gtk-grafana: Good call, cleaned up in [`846df9f`](https://github.com/grafana/scenes/pull/1614/commits/846df9f7f350d2b39937e61e548b0386a51dcbc9) with new `isSystemTransformationsProvider` flag

packages/scenes/src/components/VizPanel/VizPanel.tsx
gtk-grafana: Also a good call. Added in [`c710747`](https://github.com/grafana/scenes/pull/1614/commits/c7107476eaa3125c1a78028b7ad96b11d49e7e50)

mdvictor (approved):
LGTM!

PR #1620, opened 2026-08-20
Deps: Migrate to React 19.2.8

Sergej-Vlasov:
Migrates the repo to React **19.2.8**, matching the React 19 bump in `grafana/grafana`. No Grafana dependency changes — peers stay at `@grafana/* >=11.6`.

## 1. React 19 update

`react`/`react-dom` → `19.2.8`, `@types/react` → `19.2.18`, `@types/react-dom` → `19.2.4`.

**Dependencies**

- Pin `react`/`react-dom` in `scenes-app` and the docs website, and add root `resolutions` for both so every workspace resolves a single copy. The website previously declared `^18.2.0` and pulled its own nested React `18.3.1`, so the monorepo was running two Reacts.
- `@testing-library/react` `14.3.1` → `16.3.0` (v14 doesn't support React 19), and `@testing-library/dom` `9.3.4` → `10.4.1` for its peer range.
- Drop `@testing-library/react-hooks` (unused, peers `react@^16 || ^17`) and the unused `@testing-library/react@^12` in `scenes-app`.

**Code** — four breaking changes, 11 type errors across 6 files:

| Breaking change | Fix |
|---|---|
| `act` removed from `react-dom/test-utils` | Import from `@testing-library/react` |
| `useRef` requires an initial argument | Lazy refs take `undefined`; the two `useImperativeHandle` refs take `null` to stay assignable to `Ref<T>` |
| Global `JSX` namespace removed | `JSX.*` → `React.JSX.*` |
| `ReactElement` props default `any` → `unknown` | `isValidElement` takes an explicit props type |

**Peer ranges** widen to `^18.0.0 || ^19.0.0` — strictly more permissive, so no consumer loses support. They deliberately don't move to `>=19`: scenes peers `@grafana/* >=11.6`, and Grafana only requires React 19 from `@grafana/ui@13.2.0` onward, so that would break everyone below Grafana 13.2.

Published types stay compatible with both majors. `prepack.js` ships `dist/index.d.ts`, where all 46 JSX references emit as `React__default.JSX.Element` — and `React.JSX` exists in `@types/react@18.3.29` as well as 19.

## 2. CI: React 18 lane

The widened peer range promises React 18 support, but the workspace pins 19 — so CI would only ever test half of it, and React-19-only APIs could reach consumers below Grafana 13.2 as a runtime break.

New `Run tests (React 18)` job: `scripts/use-react-18.js` downgrades the workspace pins, then `typecheck` + `test` run against React 18. Typecheck is the valuable half — `ref` as a plain prop, `<Context value>`, `use()` and `useActionState` are all *type* errors under `@types/react` 18, so they're caught at review time. `@testing-library/react` 16 peers both majors, so no test-tooling changes were needed.

Worth review attention:

- Both release jobs `needs:` this lane, since the peer range is a published promise. Say the word if you'd rather it be advisory at first.
- The job disables the node-modules, install-state and turbo caches. `yarn.lock` is unchanged at cache-restore time so the keys collide with the React 19 job — and a turbo hit could let the lane report success without running anything.
- `yarn-nm-install` gains an `immutable` input. It passes `--no-immutable` explicitly rather than just omitting `--immutable`, because yarn defaults `enableImmutableInstalls` to true whenever `CI` is set.

## Verification

Both lanes green in CI. React 18 and React 19 give identical results: typecheck 3/3, `@grafana/scenes` 84 suites / 1812 passed / 103 snapshots, `@grafana/scenes-react` 18 suites / 42 passed. The React 18 run log confirms `react@npm:18.3.1` resolved and `cache bypass` on all turbo tasks, so it isn't replaying the React 19 results.

Two things reviewers might otherwise wonder about:

- The suite logs 48 `not wrapped in act(...)` warnings from `AdHocFiltersRecommendationsRenderer`. Pre-existing — unmodified `main` on React 18.3.1 emits the same 48. Left alone to keep this PR focused.
- `react-grid-layout@1.5.3` and `react-resizable` both pass `nodeRef` to `DraggableCore`, so React 19's removal of `findDOMNode` is never hit on the drag/resize path that jsdom tests don't cover.

**Suggested release: minor.** Peers only widen, no public API changes, runtime behaviour unchanged, and the published `.d.ts` still compiles against `@types/react` 18.

Published canary version: 8.17.0--canary.1620.34205354089.0


mdvictor:
I've tested this around and seems to work well, alongside local development and the scenes demo app itself

packages/scenes-react/package.json
```diff
@@ -54,24 +54,25 @@
...
-    "@grafana/runtime": ">=11.6",
-    "@grafana/schema": ">=11.6",
-    "@grafana/ui": ">=11.6",
-    "react": "^18.0.0",
-    "react-dom": "^18.0.0",
+    "@grafana/data": ">=12.4",
+    "@grafana/e2e-selectors": ">=12.4",
+    "@grafana/runtime": ">=12.4",
+    "@grafana/schema": ">=12.4",
+    "@grafana/ui": ">=12.4",
```
MattIPv4: Are these correct, does every 13.x release support React 19?

packages/scenes/package.json
```diff
@@ -61,26 +61,27 @@
...
-    "@grafana/schema": ">=11.6",
-    "@grafana/ui": ">=11.6",
-    "react": "^18.0.0",
-    "react-dom": "^18.0.0",
+    "@grafana/data": ">=12.4",
+    "@grafana/e2e-selectors": ">=12.4",
+    "@grafana/i18n": ">=12.4",
+    "@grafana/runtime": ">=12.4",
+    "@grafana/schema": ">=12.4",
+    "@grafana/ui": ">=12.4",
```
MattIPv4: Ditto, is every 13.x release compatible?

harisrozajac (dismissed):

packages/scenes-react/package.json
Sergej-Vlasov: we have 18 or 19 as peer deps, react 19 is only forced in dev deps. And its quite unlikely that major scenes version will be backported to older versions

package.json
```diff
@@ -44,7 +44,10 @@
     "i18n-extract": "yarn workspace @grafana/scenes run i18n-extract"
   },
   "resolutions": {
-    "@types/react": "18.3.29",
+    "react": "19.2.8",
+    "react-dom": "19.2.8",
```
ivanortegaalba, line 48: This pins react and react-dom to exactly 19.2.8 across the whole repo, so nothing in CI ever resolves React 18, while the published peers still claim `^18.0.0 || ^19.0.0`. That leaves the 18 half of the range untested rather than wrong. Would a small `@types/react@18` plus `react@18` typecheck lane be worth adding? It would make the claim checkable, and I think it also gives a concrete answer to the two open questions on this PR. Fine as a follow-up.

ivanortegaalba (dismissed):
Approving! 💯

mdvictor (approved):
Tested again and LGTM!

PR #1622, opened 2026-08-21
AdHocFiltersVariable: Stop rendering match all pills as disabled

### LINKED ISSUE #1621: AdHocFiltersVariable: `=| All` pills render with the disabled treatment and read as non-interactive
Image

## Problem

Since #1591, `isMatchAllFilter()` also returns true for `=|` filters carrying the `$__all` sentinel, not just the `=~ .*` form. That is right for query semantics - both should be stripped from the rendered expression, from `request.filters` and from key/value option lookups.

But `isMatchAllFilter` also drives presentation, so All pills were silently opted into the **disabled pill treatment**:

Image

An All pill loses its text colour, its border **and** its hover state - three independent "not interactive" signals - on a pill that is fully interactive.

## Root cause

`isMatchAllFilter` is a **query-semantics** predicate - "this filter does not restrict results, so strip it". It is also wired into three **presentation** decisions:

- `AdHocFilterPill.tsx:77` - `disabled={isMatchAllFilter(filter) || filter.nonApplicable}` applies `disabledPill` (`text.disabled`, `border: 0`, hover removed)
- `AdHocFilterPill.tsx:121` - `matchAllPillIcon` paints the history / "Restore the value set by this dashboard" **button** in `text.disabled` and removes its hover, while leaving `cursor: pointer`
- `AdHocFilterPill.tsx:70` - `showRemove` hides the `x`

#1591 widened the predicate for correct query behaviour, and all three presentation behaviours moved with it.

**The vocabulary is the problem. `disabled` means "you cannot interact with this". It is being used to mean "this is not currently narrowing anything". Only the second statement is true.**

The restore icon is the clearest illustration: it is a live button that still sets `cursor: pointer`, so we are telling the user it is clickable and painting it as though it is not, with no hover feedback to resolve the contradiction.

Image

## Why it matters

The use case is dashboards that pin a set of dimensions as origin filters so viewers can **see at a glance which dimensions are filterable and click straight into one**. An authored-as-All default is exactly the "pinned but not yet narrowed" case - and it is the one rendered as disabled. Viewers read those pills as unavailable and do not click them, which defeats the point of pinning them.

## Proposal

Stop using disabled styling to express "not narrowing".

1. `disabled={filter.nonApplicable}` - take `isMatchAllFilter` out of the expression entirely, so `disabled` means only what it says.
2. **Express "not narrowing" on the value instead**: the value keeps its current muted colour, while the key, operator, border and hover return to normal. This should apply to **both** match-all forms, not just the sentinel. Needs a small `BasePill` change - it currently takes a flat `label` string (`label={pillTextContent}`), so the value has to be passed as a separate node or via a `mutedValue` prop.
3. Drop `matchAllPillIcon` so the restore button renders as the button it is.

`showRemove` can stay as it is - hiding a control whose action would be a no-op is a different statement from greying out an active one, and a reasonable one to make.

## Alternatives considered

- **Exempt only the `$__all` sentinel from `disabled`.** Narrows the misuse rather than ending it, and leaves `=~ .*` pills mislabelled.
- **Drive the grey off `filter.restorable`.** That flag means "differs from the authored default", so it would grey out precisely the pills a viewer has just narrowed.

## Related

- #1591 - introduced the `$__all` sentinel and widened `isMatchAllFilter`
- #1602 - would make `=| $__all` the standard cleared state on multi-value-capable datasources
- #1152 - introduced the `disabled` to `disabledPill` wiring for non-applicable filters

samjewell:
Closes #1621

image

## What

`isMatchAllFilter` is a query-semantics predicate - "this filter does not restrict results, so strip it". It was also driving three presentation decisions, so when #1591 widened it to cover the `=| $__all` sentinel, All pills silently inherited the **disabled pill treatment**: `text.disabled`, `border: 0`, hover removed.

That asserts "you cannot interact with this" about a pill that is fully interactive - and on dashboards that pin dimensions as origin defaults, it is the pill users most need to click. They read it as unavailable and leave it alone.

## Changes

1. **`disabled={filter.nonApplicable}`** - `isMatchAllFilter` comes out of the expression, so `disabled` means only what it says.
2. **"Not narrowing" moves onto the value.** `BasePill` gains an optional `mutedValue`: the value keeps its current muted colour while the key, operator, border and hover return to normal. Applies to **both** match all forms, not just the sentinel.
3. **The restore button drops `matchAllPillIcon`**, so a live button (it still sets `cursor: pointer`) no longer renders as though it were disabled.

`showRemove` is deliberately unchanged - hiding a control whose action would be a no-op is a different statement from greying out an active one.

Visually the only difference on a match all pill is that the key, operator and border come back. The value grey is unchanged, so the "this is not filtering" cue survives.

## Testing

- `packages/scenes` adhoc suites: 487 passed, 0 failed.
- Two new tests: match-all value in its own node, pill still a button, click enters edit mode — for both `=| All` and `=~ .*`.
- Full `packages/scenes` suite: 1785 passed. The 15 failures in `VizPanelBuilder` / `SceneDataTransformer` / `calculateField` / `filterByName` reproduce identically on an unmodified tree and are unrelated to this change.
- Match-all pill text is now split across two nodes, so RTL `getByText('key =| All')` no longer matches. Tests here query by `Edit filter with key ...` instead. Grepped grafana/grafana: no in-tree tests assert that concatenated string.

### DIFF
--- a/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
+++ b/packages/scenes/src/variables/adhoc/AdHocFiltersCombobox/AdHocFilterPill.tsx
@@ -24,8 +24,9 @@ export function AdHocFilterPill({ filter, controller, readOnly, focusOnWipInputR
     useEditablePill({ filter, controller, readOnly, focusOnWipInputRef, isFilterEmpty: isAdhocFilterEmpty });
 
   const keyLabel = filter.keyLabel ?? filter.key;
+  const isMatchAll = isMatchAllFilter(filter);
   // A saved All default can arrive without valueLabels, so the sentinel always renders as All
-  const isAllValueSelected = filter.operator === ONE_OF_OPERATOR && isMatchAllFilter(filter);
+  const isAllValueSelected = filter.operator === ONE_OF_OPERATOR && isMatchAll;
   const valueLabel = isAllValueSelected
     ? t('grafana-scenes.components.adhoc-filter-pill.all-values', 'All')
     : filter.valueLabels?.join(', ') || filter.values?.join(', ') || filter.value;
@@ -52,7 +53,11 @@ export function AdHocFilterPill({ filter, controller, readOnly, focusOnWipInputR
   const cleanFilter = !filter.restorable && !filter.readOnly && !filter.nonApplicable;
 
   if (viewMode) {
-    const pillTextContent = `${keyLabel} ${filter.operator} ${valueLabel}`;
+    // A match all filter is not narrowing anything, so its value is muted. The key, operator,
+    // border and hover stay normal - the pill is still fully interactive.
+    const pillTextContent = isMatchAll
+      ? `${keyLabel} ${filter.operator}`
+      : `${keyLabel} ${filter.operator} ${valueLabel}`;
 
     const handleRemove = () => {
       if (filter.origin && filter.origin === 'dashboard') {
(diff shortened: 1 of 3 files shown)


mdvictor (approved):
LGTM!

PR #1630, opened 2026-08-26
Variables: Fix multi value picker staying empty after clearing the selection

ivanortegaalba:
## What

Fixes a multi-value variable picker getting stuck on the `Select value` placeholder after the user clears the selection, instead of falling back to the variable's default (`All` when `includeAll` is on, otherwise the first option).

Reported against a Cloud dashboard where a `Resource Group` variable would not go back to `All`.

## Why

`VariableValueSelectMulti` held a *copy* of the variable's value in local state, so that selecting options does not re-run queries on every click, and committed it on blur. It healed that copy through a `useEffect` on `state.value`, so only when a state change was actually published.

`MultiValueVariable.changeValueTo` rewrites an empty selection back to the default via `getDefaultMultiState`, and then early-returns when that rewrite equals the value the variable already had:

```ts
if (isEqual(value, this.state.value) && isEqual(text, this.state.text)) {
  return;
}
```

Clearing a variable that already sits on its default therefore produced: local copy `[]` -> commit -> variable rewrites to `['$__all']` -> equals the current value -> nothing published -> no re-render -> the picker kept rendering the empty copy. The variable itself was unchanged, so panels kept returning the right data and only the picker looked wrong.

One detail that made this look intermittent: on activation, validation normalises `text` from `['All']` to `'All'`, and that mismatch let the first commit slip past the `isEqual` guard. So the first clear recovered and every clear after it stuck.

## Fix

The local state stops being a copy of the value and becomes "there is an edit in flight". It is `undefined` unless the user is mid-edit, and the picker otherwise renders the variable's value directly, so the two cannot drift apart and the variable is free to reject or rewrite what gets committed without having to announce it.

Alternative considered and rejected: making `changeValueTo` always publish a state change. That would re-run dependent queries on a value change that is a no-op.

Behaviour kept as it was: an external value change still wins over an in-flight edit, and queries still only run on commit.

## Tests

`should fall back to the All value every time the selection is cleared` reproduces the bug and fails on `main`. `should commit the selection made while the menu was open` passes on `main` and is there to guard the primary path against this rework.

### DIFF
--- a/packages/scenes/src/variables/components/VariableValueSelect.test.tsx
+++ b/packages/scenes/src/variables/components/VariableValueSelect.test.tsx
@@ -1,6 +1,6 @@
 import { MultiOrSingleValueSelect } from './VariableValueSelect';
 import React from 'react';
-import { render, screen } from '@testing-library/react';
+import { render, screen, within } from '@testing-library/react';
 import { SceneVariableSet } from '../sets/SceneVariableSet';
 import { TestScene } from '../TestScene';
 import { selectors } from '@grafana/e2e-selectors';
@@ -205,6 +205,25 @@ describe('VariableValueSelectMulti', () => {
 
     return { model, ...result };
   }
+
+  function getRemoveButtonFor(label: string) {
+    return within(screen.getByText(label).parentElement!).getByRole('button', { name: 'Remove' });
+  }
+
+  it('should only fall back to the All value once the last selected option is removed', async () => {
+    const { model } = setupMultiVariable({ value: ['A', 'B'], text: ['A', 'B'] });
+
+    await userEvent.click(getRemoveButtonFor('A'));
+    await userEvent.click(screen.getByTestId('outside'));
+
+    expect(model.state.value).toEqual(['B']);
+
+    await userEvent.click(getRemoveButtonFor('B'));
+    await userEvent.click(screen.getByTestId('outside'));
+
+    expect(model.state.value).toEqual([ALL_VARIABLE_VALUE]);
+    expect(screen.getByText(ALL_VARIABLE_TEXT)).toBeInTheDocument();
+  });
 
   it('should fall back to the All value every time the selection is cleared', async () => {
     const { model, container } = setupMultiVariable({ value: [ALL_VARIABLE_VALUE], text: [ALL_VARIABLE_TEXT] });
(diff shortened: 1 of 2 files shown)


packages/scenes/src/variables/components/VariableValueSelect.test.tsx
```diff
@@ -169,3 +170,62 @@ describe('VariableValueSelect', () => {
...
+      <div>
+        <MultiOrSingleValueSelect model={model} />
+        <button data-testid="outside">outside</button>
+      </div>
+    );
+
+    return { model, ...result };
+  }
+
+  it('should fall back to the All value every time the selection is cleared', async () => {
```
grafakus, line 228: I haven't reviewed the production code so this test feels slightly weird... I'd expect checking a scenario similar to:

0. setup: having with 2 options selected
1. clearing one of the two option does not falls back to "All"
2. clearing the other option falls back to "All"
3. clicking on the clear icon falls back to "All" as well?

packages/scenes/src/variables/components/VariableValueSelect.test.tsx
```diff
@@ -169,3 +170,62 @@ describe('VariableValueSelect', () => {
...
 });
+
+describe('VariableValueSelectMulti', () => {
+  function setupMultiVariable(state: Partial<MultiValueVariableState>) {
+    const model = new CustomVariable({
+      name: 'test',
+      query: 'A,B,C',
+      isMulti: true,
+      includeAll: true,
+      defaultToAll: true,
```
grafakus, line 181: [non-blocking suggestion] Should we also test `defaultToAll: false`?

packages/scenes/src/variables/components/VariableValueSelect.test.tsx
```diff
@@ -169,3 +170,62 @@ describe('VariableValueSelect', () => {
...
+  });
+
+  it('should commit the selection made while the menu was open', async () => {
+    const { model } = setupMultiVariable({ value: [ALL_VARIABLE_VALUE], text: [ALL_VARIABLE_TEXT] });
+
+    await userEvent.click(screen.getByRole('combobox'));
+    await userEvent.click(screen.getByText('B'));
+    await userEvent.click(screen.getByTestId('outside'));
+
+    expect(model.state.value).toEqual(['B']);
```
grafakus, line 258:
```suggestion
expect(model.state.value).toEqual(['B']);
expect(
screen.getByTestId(selectors.pages.Dashboard.SubMenu.submenuItemValueDropDownValueLinkTexts('B'))
).toBeInTheDocument();
```
?

grafakus (dismissed):
LGTM, just a few suggestion for the unit tests

grafakus (approved):

PR #1636, opened 2026-09-08
QueryVariable: interpolate legacy variable queries against their own scene

tskarhed:
## Summary
- Legacy datasources (Graphite, Snowflake, and others that predate the `variables` API) interpolate via `templateSrv.replace()` and ignore the `scopedVars` `LegacyQueryRunner` passes them. They read `window.__grafanaSceneContext` instead.
- When more than one scene is active (for example a reporting form with one scene per dashboard), only the last scene to activate owns that global, so chained queries on the others run uninterpolated (`$env.*`).
- Point the global at the variable’s own `SceneVariableSet` for the synchronous part of `metricFindQuery`, then restore it in `finally` before awaiting the promise so concurrent scenes cannot overwrite each other.

## Steps to reproduce (report drawer)

Needs Grafana Enterprise. The redesigned reporting UI is behind the `newShareReportDrawer` feature toggle.

1. Enable `newShareReportDrawer` (`[feature_toggles]` in `grafana.ini`, or Admin → Feature toggles) and restart if you used the ini file.
2. Create two dashboards that each have a chained query variable on a **legacy** datasource (Graphite, Snowflake, or any plugin without the `variables` API). Example:
   - Custom variable `env` with values `prod` on dashboard A and `dev` on dashboard B
   - Query variable `hosts` with query `$env.*`
3. Open **Reporting** (`/reports`) and click **Create report** (or from a dashboard: Share → Schedule report). The form should open as a drawer, not the old full-page wizard.
4. In **Dashboards**, pick dashboard A, then **Add dashboard** and pick dashboard B. Both scenes stay mounted at once.
5. Expand the variable pickers on each dashboard row.

**Before this change:** the picker for `hosts` is empty on both rows. The report form's dashboard rows never own `window.__grafanaSceneContext`, so `$env` resolves against no variable and both the scene interpolator and `templateSrv` return the token unchanged — the datasource is queried for a literal `$env.*`, which matches no metrics. If a scene that does define `env` happens to own the global (for example the drawer opened from a dashboard that has its own `env`), the picker shows that scene's values instead of the row's.
**After this change:** A's picker shows `prod.*` and B's shows `dev.*`.

## Test plan
- [ ] `yarn test:scenes packages/scenes/src/variables/variants/query/createQueryVariableRunner.test.ts --watch=false`
- [ ] Confirm two scenes with different `$env` values both interpolate correctly when activated together
- [ ] Confirm the previous `__grafanaSceneContext` is restored after the query
- [ ] Spot-check a Graphite/Snowflake query variable that depends on another variable
- [ ] Reproduce via the report drawer steps above against a linked scenes build

### DIFF
--- a/packages/scenes/src/variables/variants/query/createQueryVariableRunner.test.ts
+++ b/packages/scenes/src/variables/variants/query/createQueryVariableRunner.test.ts
@@ -1,6 +1,7 @@
 import { lastValueFrom } from 'rxjs';
 
 import { DataSourceApi, MetricFindValue } from '@grafana/data';
+import { waitFor } from '@testing-library/react';
 
 import { EmbeddedScene } from '../../../components/EmbeddedScene';
 import { SceneCanvasText } from '../../../components/SceneCanvasText';
@@ -129,9 +130,11 @@ describe('LegacyQueryRunner', () => {
   it('interpolates through the global scene context when the datasource ignores scopedVars', async () => {
     const { scene, queryVariable } = buildScene('prod', 'hosts');
 
+    // Activation runs the query on its own, so nothing else may trigger a refresh here: the
+    // assertion below pins the exact calls the datasource saw.
     activateFullSceneTree(scene);
 
-    await lastValueFrom(queryVariable.validateAndUpdate());
+    await waitFor(() => expect(queryVariable.state.options).toEqual([{ label: 'prod.*', value: 'prod.*' }]));
 
     expect(metricFindQueryCalls).toEqual([{ variableName: 'hosts', interpolated: 'prod.*' }]);
   });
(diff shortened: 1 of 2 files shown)


ifrost:
Could we update data source plugins to use scopedVars instead? 🤔

tskarhed:
I am doing that too. I added this as a fallback - feel free to deny it

torkelo:
I feel like updating the plugin is best

tskarhed:
This may be in issue in a lot of datasources. Just looking at the Enterprise ones, I've identified at least 8 possible plugins.

[The example in the docs](https://grafana.com/developers/plugin-tools/how-to-guides/data-source-plugins/add-support-for-variables#implement-metricfindquery) will also cause this bug to appear. So this is a public functionality that we support.

If we don't intend to support this functionality we should:

1. Inform developers that they are using it incorrectly (deprecatedWarning in code maybe?)
2. Update the docs accordingly

mdvictor:
I'm also a bit torn on this... I understand the need but agree this would be better done in the plugins. My main concern is that adding this might further mean plugins never get updated and even new ones might continue building on the threaded path of not using `scopedVars`. We definitely need to update docs and inform devs.

PR #1640, opened 2026-09-09
CI: Release through a pull request instead of committing to main

Sergej-Vlasov:
Releases have failed since 2026-08-25 with `Changes must be made through a pull request` — the org ruleset **Repo required PRs** forbids committing to `main`, which is what the release job did. No stable release since 8.16.1. Platform Productivity declined a bypass and asked for a release-PR model.

## Change

The `release` job becomes two:

| Job | Trigger | Does |
|---|---|---|
| `release-pr` | push to `main` | Writes changelog + version bump to `release/vX.Y.Z`, opens a PR, enables auto-merge |
| `publish` | that PR merging | Tag, npm publish, GitHub release — existing logic, unchanged |

This works because `Repo required PRs` targets only the default branch, so a release branch is writable. `Signed Commits` covers *all* branches, so commits still go through `ghcommit-action`.

The jobs discriminate on commit subject: squash is the only merge method enabled here, so a merged release PR lands as `Release vX.Y.Z`. `publish` additionally checks whether the version is already on npm, so a retitled PR or a re-run is a no-op rather than a failure or double publish.

Also: canary is skipped on `release/*` (it would otherwise publish a canary of the release PR itself), and `release-pr` drops the build step since it publishes nothing.

## Note on releases

Releases are no longer fully unattended — the ruleset requires 1 approving review, and auto-merge waits for it rather than bypassing it. So a release is now: merge a labelled PR → approve the release PR it opens → publish happens automatically.

## Review findings (second commit)

A six-lens adversarial review raised 25 findings; 13 were refuted, 6 distinct real defects fixed:

| Severity | Defect |
|---|---|
| Critical | `release-pr` set `contents: read`, but `create-github-app-token` uses OIDC and needs `id-token: write` — the job would have failed on its first step |
| High | `release-pr` and `publish` shared one concurrency group. GitHub keeps a single pending entry per group, so a queued `publish` could be cancelled by a later `release-pr` — version bumped on main, nothing tagged or published, and reported as *cancelled* rather than *failed*, so nothing alerts |
| High | The publish gate probed only `@grafana/scenes`. lerna publishes topologically, so a partial failure left `scenes-react` unpublished, and the retry went green having skipped everything |
| Medium | The GitHub release was gated on the same npm probe, so a failed `auto release` could never be retried |
| Medium | The superseded-PR loop matched any `release/*` branch and closed it with `--delete-branch` — a human's `release/my-refactor` would have been deleted |
| Medium | No recovery path if the trigger ever missed; added `workflow_dispatch` |

Verified locally: YAML parses; version bumps don't dirty `yarn.lock` (workspaces record as `0.0.0-use.local`), so the release PR's own `yarn install --immutable` stays green and auto-merge isn't blocked; `npm view` exit codes and `lerna list --json --no-private` behave as the gate assumes; the branch matcher accepts `release/v8.17.0` and rejects `release/my-refactor`.

## Reviewer feedback (third commit)

All three comments applied:

- **Tag commit on re-drive.** On a `workflow_dispatch` re-drive `GITHUB_SHA` is the tip of the chosen branch, and dispatch cannot target a commit, so the tag would have landed on whatever merged since. Now resolved from the `Release v` commit subject. (`auto release` still describes from HEAD, so notes can widen on a late dispatch — noted in the step.)
- **npm probe error handling.** `2>&1` conflated E404 with registry/network errors and both set `PUBLISH=true`; harmless for `from-package` itself, but it also unblocked the tag push, so an npm outage would tag a release that never published. Now only E404 counts as "not published".
- **Approval visibility.** The PR body and run summary now say an approval is required, since `--auto` waits for a review rather than bypassing it.

**One finding I could not confirm** and did not act on: a claim that an unlabelled merge landing while a release PR is open drops that commit from the changelog permanently. My reading is that `auto version` still sees the earlier release-labelled PR (untagged until the release PR merges), so the job proceeds and regenerates the PR including the new commit. Worth watching on the first real release.

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -5,6 +5,9 @@ on:
     branches: [main]
   pull_request:
     types: [opened, synchronize, reopened]
+  # Manual re-drive for `publish`. Pushing to main is forbidden by the ruleset, so
+  # without this a publish that never fired could not be recovered.
+  workflow_dispatch:
 
 permissions:
   contents: read
@@ -127,8 +130,9 @@ jobs:
   # forbids committing to main directly: this job opens a release PR, and the
   # `publish` job below reacts to that PR being merged. The marker is the commit
   # subject, which for a squash merge (the only method enabled here) is the PR
-  # title. `publish` also has a version-based fallback so a retitled or manually
-  # landed release still ships.
+  # title. `publish` re-checks npm before acting, so a re-run is a no-op rather
+  # than a double publish, and workflow_dispatch is the manual re-drive if the
+  # subject-based trigger ever misses.
   release-pr:
     name: Open release PR
     runs-on: ubuntu-latest
@@ -138,7 +142,8 @@ jobs:
       group: release-main
       cancel-in-progress: false
     permissions:
-      contents: read
+      # create-github-app-token fetches its credentials over OIDC.
+      id-token: write
     steps:
       - name: Get GitHub App Token
         id: get-github-app-token
@@ -206,7 +211,7 @@ jobs:
         # recomputes cumulatively, so closing the old one and opening a fresh PR
         # always yields the correct version.
         run: |
-          for NUM in $(gh pr list --state open --json number,headRefName --jq '.[] | select(.headRefName | startswith("release/")) | .number'); do
+          for NUM in $(gh pr list --state open --json number,headRefName --jq '.[] | select(.headRefName | test("^release/v[0-9]+\\.[0-9]+\\.[0-9]+$")) | .number'); do
             echo "Closing superseded release PR #${NUM}"
             gh pr close "$NUM" --delete-branch --comment "Superseded by the release PR for v${VERSION}."
           done
@@ -263,9 +268,12 @@ jobs:
     name: Publish release
     runs-on: ubuntu-latest
     needs: [test, test-react-18]
-    if: "github.event_name == 'push' && startsWith(github.event.head_commit.message, 'Release v')"
+    if: "github.event_name == 'workflow_dispatch' || (github.event_name == 'push' && startsWith(github.event.head_commit.message, 'Release v'))"
     concurrency:
-      group: release-main
+      # Deliberately NOT release-main: GitHub keeps a single pending entry per
+      # group, so sharing one let a queued publish be cancelled by a later
+      # release-pr, silently dropping an already-merged release.
+      group: release-publish
       cancel-in-progress: false
     permissions:
       id-token: write
@@ -295,24 +303,39 @@ jobs:
       - name: Build library
         run: yarn run lerna run build --no-private
 
-      - name: Resolve version to publish
+      - name: Resolve what still needs releasing
         id: version
-        # Fallback for the commit-subject trigger: if this version is already on
-        # npm there is nothing to do, so a re-run (or a push whose subject merely
-        # looks like a release) is a no-op rather than a failure.
+        # Probe every publishable package, not just one. lerna publishes
+        # topologically, so a failure partway through leaves some packages on npm
+        # and some not; gating on a single package would make the retry a silent
+        # no-op and strand the rest. The GitHub release is tracked separately so a
+        # failed `auto release` can be retried after a successful npm publish.
         run: |
           VERSION="$(node -p "require('./lerna.json').version")"
           echo "version=${VERSION}" >> "$GITHUB_OUTPUT"
-          if npm view "@grafana/scenes@${VERSION}" version >/dev/null 2>&1; then
-            echo "publish=false" >> "$GITHUB_OUTPUT"
-            echo "v${VERSION} is already published; nothing to do."
+          PUBLISH=false
+          SPECS="$(yarn lerna list --json --no-private --loglevel silent \
+            | node -e "let s='';process.stdin.on('data',d=>s+=d).on('end',()=>console.log(JSON.parse(s).map(p=>p.name+'@'+p.version).join('\n')))")"
+          for SPEC in $SPECS; do
+            if npm view "$SPEC" version >/dev/null 2>&1; then
+              echo "${SPEC} is already published."
+            else
+              echo "${SPEC} is not on npm."
+              PUBLISH=true
+            fi
+          done
+          echo "publish=${PUBLISH}" >> "$GITHUB_OUTPUT"
+          if gh release view "v${VERSION}" >/dev/null 2>&1; then
+            echo "release=false" >> "$GITHUB_OUTPUT"
           else
-            echo "publish=true" >> "$GITHUB_OUTPUT"
-            echo "Publishing v${VERSION}."
+            echo "release=true" >> "$GITHUB_OUTPUT"
           fi
+          echo "Needs npm publish: ${PUBLISH}"
+        env:
+          GH_TOKEN: ${{ steps.get-github-app-token.outputs.token }}
 
       - name: Push release tag
-        if: steps.version.outputs.publish == 'true'
+        if: steps.version.outputs.publish == 'true' || steps.version.outputs.release == 'true'
         # Tag before publishing to npm: a publish failure then leaves git fully
         # consistent, and this step is skipped on a re-run if the tag already
         # points at the right commit.
@@ -331,7 +354,7 @@ jobs:
           VERSION: ${{ steps.version.outputs.version }}
 
       - name: Create local tag
-        if: steps.version.outputs.publish == 'true'
+        if: steps.version.outputs.publish == 'true' || steps.version.outputs.release == 'true'
         # `auto release` resolves the release version from `git describe --tags`
         # on HEAD, so the tag has to exist locally as well as on the remote.
         run: git tag "v${VERSION}" "${GITHUB_SHA}" || true
@@ -346,7 +369,7 @@ jobs:
         run: yarn lerna publish from-package --yes --no-verify-access
 
       - name: Create GitHub release
-        if: steps.version.outputs.publish == 'true'
+        if: steps.version.outputs.release == 'true'
         run: yarn auto release
         env:
           GITHUB_TOKEN: ${{ steps.get-github-app-token.outputs.token }}
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -86,7 +86,7 @@ jobs:
     name: Canary release
     runs-on: ubuntu-latest
     needs: [test, test-react-18]
-    if: github.event_name == 'pull_request'
+    if: "github.event_name == 'pull_request' && !startsWith(github.head_ref, 'release/')"
     permissions:
       id-token: write
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/node-ci.yml
```diff
@@ -182,96 +187,189 @@ jobs:
...
-          git tag "v$VERSION" "$VERSION_COMMIT"
+          PR_URL="$(gh pr create \
+            --base main \
+            --head "release/v${VERSION}" \
+            --title "Release v${VERSION}" \
+            --label no-changelog \
+            --body "Automated release of v${VERSION}. Merging publishes to npm and creates the git tag and GitHub release.")"
+          echo "Opened ${PR_URL}"
+          gh pr merge "$PR_URL" --auto --squash
         env:
```
ivanortegaalba: I went and read the live repo config for this one, and I think auto-merge will park the PR here rather than merge it.

`main`'s branch protection does list `scenes-repo-bot-access-token` under `bypass_pull_request_allowances.apps`, so on that rule alone the app could merge without an approval. But org ruleset 15595097 "Repo required PRs" (scoped to `~DEFAULT_BRANCH`) has `bypass_actors: null` together with `required_approving_review_count: 1`, `require_last_push_approval: true` and `require_extra_approval_for_unattributed_changes: true`, and where the two overlap the more restrictive one wins. If that reading is right, `--auto` enables auto-merge and then waits indefinitely for a human approval on every release.

That may well be the intent, and it does match what Platform Productivity asked for. If so, would it be worth making it explicit rather than implicit? Two small things that would help: say it in the `--body` text (it currently reads "Merging publishes to npm", which sounds hands-off), and echo the PR URL into `$GITHUB_STEP_SUMMARY` so whoever has to approve can find it without going digging.

One thing that is not a problem, in case it looked like one: there are no required status checks anywhere on this repo (`required_status_checks.contexts` is `[]` and neither ruleset carries a `required_status_checks` rule), so `Canary release` being skipped on `release/*` heads cannot hold the merge up.

Also worth checking, and I could not settle it from the API: does `require_extra_approval_for_unattributed_changes` treat a `createCommitOnBranch` commit authored by the app as attributed? If it does not, the release PR would need two approvals rather than one.

.github/workflows/node-ci.yml
```diff
@@ -182,96 +187,189 @@ jobs:
...
         # consistent, and this step is skipped on a re-run if the tag already
         # points at the right commit.
         run: |
           EXISTING="$(gh api "repos/${GITHUB_REPOSITORY}/git/ref/tags/v${VERSION}" --jq .object.sha 2>/dev/null)" || EXISTING=""
           if [ -z "$EXISTING" ]; then
-            gh api "repos/${GITHUB_REPOSITORY}/git/refs" -f "ref=refs/tags/v${VERSION}" -f "sha=${VERSION_COMMIT}"
-          elif [ "$EXISTING" = "$VERSION_COMMIT" ]; then
-            echo "Tag v${VERSION} already exists at ${VERSION_COMMIT}, skipping"
+            gh api "repos/${GITHUB_REPOSITORY}/git/refs" -f "ref=refs/tags/v${VERSION}" -f "sha=${GITHUB_SHA}"
+          elif [ "$EXISTING" = "$GITHUB_SHA" ]; then
```
ivanortegaalba: Small thing about the `workflow_dispatch` re-drive specifically. Here `GITHUB_SHA` is the tip of whatever ref was selected in the dispatch, not the release commit, and since `workflow_dispatch` only takes a branch or tag you cannot point it at the release commit directly.

On the happy path the two are the same, because squash-merge plus `delete_branch_on_merge` leaves the release commit at the tip of `main`. But the whole reason this input exists is to recover a publish that never fired, and by then something else has usually merged. In that case `v${VERSION}` gets tagged on a later commit, and since `auto release` resolves the version from `git describe --tags` on HEAD, the GitHub release notes widen to cover the extra commits too.

Would resolving the sha from the commit that last touched `lerna.json` be sturdier here? Or is accepting it and noting it in the comment good enough, given it is the manual escape hatch?

.github/workflows/node-ci.yml
```diff
@@ -182,96 +187,189 @@ jobs:
...
+          yarn lerna version "${{ steps.bump.outputs.bump }}" --force-publish --no-commit-hooks --yes --no-push --no-git-tag-version
+          echo "version=$(node -p "require('./lerna.json').version")" >> "$GITHUB_OUTPUT"
 
-      - name: Sync with remote
+      - name: Replace any superseded release PR
         if: steps.bump.outputs.bump != ''
+        # A release PR left open while another labelled PR merges is stale: its
+        # version no longer reflects everything since the last tag. `auto version`
+        # recomputes cumulatively, so closing the old one and opening a fresh PR
+        # always yields the correct version.
```
ivanortegaalba: This claim is right about `auto version` being cumulative, but it is cumulative *since the last git tag*, and the tag is only pushed by `publish` after the release PR merges. That leaves a window where the two disagree.

Concretely: the release PR for v8.17.0 merges, `publish` starts, and before it reaches `Push release tag` another PR carrying the `release` label merges. The `release-pr` job that fires for it runs `auto version`, still sees `v8.16.1` as the last tag, and so recomputes over commits that were just released. `lerna.json` is already at 8.17.0 by then, so the version itself comes out fine (8.18.0), but the changelog it generates re-includes the entries that already shipped in 8.17.0.

Narrow window, and the direct-to-main flow had the same shape, so I do not think it is worth blocking on. Worth a line in this comment so the next person does not have to re-derive it?

.github/workflows/node-ci.yml
```diff
@@ -182,96 +187,189 @@ jobs:
...
+        # failed `auto release` can be retried after a successful npm publish.
+        run: |
+          VERSION="$(node -p "require('./lerna.json').version")"
+          echo "version=${VERSION}" >> "$GITHUB_OUTPUT"
+          PUBLISH=false
+          SPECS="$(yarn lerna list --json --no-private --loglevel silent \
+            | node -e "let s='';process.stdin.on('data',d=>s+=d).on('end',()=>console.log(JSON.parse(s).map(p=>p.name+'@'+p.version).join('\n')))")"
+          for SPEC in $SPECS; do
+            if npm view "$SPEC" version >/dev/null 2>&1; then
+              echo "${SPEC} is already published."
```
ivanortegaalba: A transient `npm view` failure reads the same as "not published here", since the `2>&1` swallows the difference between E404 and a network or registry error, and either way it sets `PUBLISH=true`.

The publish itself is fine with that, `from-package` skips whatever is already on npm exactly as the comment above says. What made me look twice is that this also feeds `Push release tag`, which runs when `publish == 'true' || release == 'true'`. That step is careful about an existing tag pointing elsewhere, so I think the worst case is still just a no-op re-run.

Is it worth separating the E404 case from the rest here, or is fail-open-and-let-lerna-decide the deliberate choice?

.github/workflows/node-ci.yml
```diff
@@ -182,96 +187,189 @@ jobs:
...
-          git reset --hard "$VERSION_COMMIT"
-          git tag "v$VERSION" "$VERSION_COMMIT"
+          PR_URL="$(gh pr create \
+            --base main \
+            --head "release/v${VERSION}" \
+            --title "Release v${VERSION}" \
+            --label no-changelog \
+            --body "Automated release of v${VERSION}. Merging publishes to npm and creates the git tag and GitHub release.")"
+          echo "Opened ${PR_URL}"
+          gh pr merge "$PR_URL" --auto --squash
```
ivanortegaalba: Super small nit: `--auto` enables auto-merge and then waits for a human approval on every release but the body text reads as if merging is automatic. Making the approval step visible:

```suggestion
            --body "Automated release of v${VERSION}. Needs one approval; merging then publishes to npm and creates the git tag and GitHub release.")"
          echo "Opened ${PR_URL}"
          echo "Release PR for v${VERSION} is open and needs an approval before it can merge: ${PR_URL}" >> "$GITHUB_STEP_SUMMARY"
          gh pr merge "$PR_URL" --auto --squash
```

.github/workflows/node-ci.yml
```diff
@@ -182,96 +187,189 @@ jobs:
...
           VERSION: ${{ steps.version.outputs.version }}
-          VERSION_COMMIT: ${{ steps.version-commit.outputs.commit-hash }}
+
+      - name: Create local tag
+        if: steps.version.outputs.publish == 'true' || steps.version.outputs.release == 'true'
+        # `auto release` resolves the release version from `git describe --tags`
+        # on HEAD, so the tag has to exist locally as well as on the remote.
+        run: git tag "v${VERSION}" "${GITHUB_SHA}" || true
+        env:
+          VERSION: ${{ steps.version.outputs.version }}
```
ivanortegaalba, line 380: This is a good find from the agent review

-----

On the `workflow_dispatch` re-drive `GITHUB_SHA` is the tip of the chosen branch, not the release commit, and since dispatch only accepts a branch or tag you cannot point it at the commit. That is the one path where main has usually moved on, so the tag lands on a later commit. Resolving it from the `Release v` commit instead, and reusing it for the local tag:

```suggestion
      - name: Push release tag
        id: tag
        if: steps.version.outputs.publish == 'true' || steps.version.outputs.release == 'true'
        # Tag before publishing to npm: a publish failure then leaves git fully
        # consistent, and this step is skipped on a re-run if the tag already
        # points at the right commit. The sha is resolved from the `Release v`
        # commit rather than GITHUB_SHA, which on a workflow_dispatch re-drive is
        # whatever sits at the tip of the chosen branch.
        run: |
          RELEASE_SHA="$(git rev-list -1 --grep "^Release v${VERSION}" HEAD)"
          if [ -z "$RELEASE_SHA" ]; then
            echo "No 'Release v${VERSION}' commit reachable from HEAD" >&2
            exit 1
          fi
          echo "sha=${RELEASE_SHA}" >> "$GITHUB_OUTPUT"
          EXISTING="$(gh api "repos/${GITHUB_REPOSITORY}/git/ref/tags/v${VERSION}" --jq .object.sha 2>/dev/null)" || EXISTING=""
          if [ -z "$EXISTING" ]; then
            gh api "repos/${GITHUB_REPOSITORY}/git/refs" -f "ref=refs/tags/v${VERSION}" -f "sha=${RELEASE_SHA}"
          elif [ "$EXISTING" = "$RELEASE_SHA" ]; then
            echo "Tag v${VERSION} already exists at ${RELEASE_SHA}, skipping"
          else
            echo "Tag v${VERSION} already exists but points at ${EXISTING} instead of ${RELEASE_SHA}" >&2
            exit 1
          fi
        env:
          GH_TOKEN: ${{ steps.get-github-app-token.outputs.token }}
          VERSION: ${{ steps.version.outputs.version }}

      - name: Create local tag
        if: steps.version.outputs.publish == 'true' || steps.version.outputs.release == 'true'
        # `auto release` resolves the release version from `git describe --tags`
        # on HEAD, so the tag has to exist locally as well as on the remote.
        run: git tag "v${VERSION}" "${RELEASE_SHA}" || true
        env:
          VERSION: ${{ steps.version.outputs.version }}
          RELEASE_SHA: ${{ steps.tag.outputs.sha }}
```

Prefix match rather than `^Release v${VERSION}$` on purpose: a squash subject is the PR title plus ` (#N)`. Worth knowing this only fixes the tag. `auto release` still describes from HEAD, so on a late dispatch the notes can still widen.

.github/workflows/node-ci.yml
```diff
@@ -182,96 +187,189 @@ jobs:
...
+          SPECS="$(yarn lerna list --json --no-private --loglevel silent \
+            | node -e "let s='';process.stdin.on('data',d=>s+=d).on('end',()=>console.log(JSON.parse(s).map(p=>p.name+'@'+p.version).join('\n')))")"
+          for SPEC in $SPECS; do
+            if npm view "$SPEC" version >/dev/null 2>&1; then
+              echo "${SPEC} is already published."
+            else
+              echo "${SPEC} is not on npm."
+              PUBLISH=true
+            fi
+          done
```
ivanortegaalba, line 334: `2>&1` here hides the difference between 404 and a registry or network error, and both set `PUBLISH=true`. `from-package` makes that harmless for the publish itself, but it also unblocks `Push release tag`, so a npm down will do the tag push. Splitting the two:

```suggestion
          for SPEC in $SPECS; do
            if NPM_OUT="$(npm view "$SPEC" version 2>&1)"; then
              echo "${SPEC} is already published."
            elif printf '%s' "$NPM_OUT" | grep -q 'E404'; then
              echo "${SPEC} is not on npm."
              PUBLISH=true
            else
              echo "npm view ${SPEC} failed: ${NPM_OUT}" >&2
              exit 1
            fi
          done
```

.github/workflows/node-ci.yml
Sergej-Vlasov: Applied. Verified the mechanics first: `git rev-list -1 --grep "^Release v${VERSION}" HEAD` returns empty with **exit 0** on no match rather than failing, so the explicit `-z` guard is doing real work — `set -e` would not have caught it.

Kept your note about `auto release` still describing from HEAD as a comment in the step, since it's a real remaining limitation on a late dispatch rather than something this fixes.

.github/workflows/node-ci.yml
Sergej-Vlasov: Applied. Confirmed the discrimination holds before taking it: a missing version gives `npm error code E404`, while pointing npm at an unreachable registry gives `ENOTFOUND` and correctly does not match, so a registry outage now fails the job instead of silently pushing the tag.

.github/workflows/node-ci.yml
Sergej-Vlasov: Applied, both the body wording and the step-summary line. The approval requirement is inherent here — the ruleset requires one approving review and `--auto` waits for it rather than bypassing it — so making it visible at the point someone reads the PR is worth more than a nit.

.github/workflows/node-ci.yml
```diff
@@ -182,96 +187,204 @@ jobs:
...
         env:
+          GH_TOKEN: ${{ steps.get-github-app-token.outputs.token }}
           VERSION: ${{ steps.version.outputs.version }}
-          VERSION_COMMIT: ${{ steps.version-commit.outputs.commit-hash }}
+
+  publish:
+    name: Publish release
+    runs-on: ubuntu-latest
+    needs: [test, test-react-18]
+    if: "github.event_name == 'workflow_dispatch' || (github.event_name == 'push' && startsWith(github.event.head_commit.message, 'Release v'))"
```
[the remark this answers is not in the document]
Sergej-Vlasov: Confirmed and fixed in 9872f9df.

Traced it: `publish`'s `if` accepted any dispatch with no ref check, `ghcommit-action` writes the commit subject as exactly `Release v<version>` on the release branch, and the tag step's `git rev-list --grep "^Release v${VERSION}"` would match that branch's own commit. So a dispatch from `release/v8.17.0` would publish to npm and tag a commit that is not on main — and `delete_branch_on_merge` is enabled here, so that commit becomes unreachable once the release PR merges.

The dispatch path is now gated on `github.ref == 'refs/heads/main'`. The push path was already main-only via the workflow trigger.

ivanortegaalba (approved):
I tested locally and it works! 💯

PR #1642, opened 2026-09-10
CI: Harden the release pipeline after the v8.17.0 release

Sergej-Vlasov:
Hardening for the release pipeline, driven by what actually happened releasing **v8.17.0** — the first live run of the release-PR flow. Every change below traces to a real failure in that run, not a hypothetical.

Recap of the run: the release PR opened, was approved and auto-merged correctly, and `release-pr`/`canary` skipped exactly as designed. Then `publish` failed twice — once half-published, once fully released but still red — and needed a manual re-run.

## 1. Pin the publish tree to the release commit

On the `workflow_dispatch` path `actions/checkout` takes **main's tip**, not the release commit, and everything downstream reads that tree: the build, `lerna.json`, the `gitHead` lerna stamps into npm metadata, and the commit range `auto release` turns into notes.

A re-drive after later merges would publish a tree the tag doesn't describe. `Push release tag` already resolved the `Release v` commit; the working tree now follows it. **No-op on the push path**, where HEAD is already that commit.

This was a gap in the earlier `workflow_dispatch` fix — that pinned the tag's SHA but left the checkout on the tip.

## 2. Retry the npm publish

v8.17.0 aborted partway with a Rekor 409 (`TLOG_CREATE_ENTRY_ERROR`), leaving `@grafana/scenes` published and `@grafana/scenes-react` not. The *identical* command succeeded minutes later on a manual re-run — so the retry is precisely what a human was doing by hand.

The registry, not lerna's exit code, decides completion: after a non-zero exit the step re-probes npm, treats "everything is on npm" as success, and only fails once packages are still missing after three attempts. Safe because `from-package` recomputes the publish set from the registry each call.

Deliberately **not** applied to `Create GitHub release` — `auto release` calls `repos.createRelease` unconditionally, so an in-job retry would 422 once the release exists and turn a good release permanently red. Its re-entrancy correctly comes from the `release` gate across runs.

## 3. Assert the GitHub release exists, don't trust `auto`'s exit code

`auto release` created the release and commented on 10 PRs, then hit issue #1621 with `403 Resource not accessible by integration` and failed the step — after the release was already correct.

This is not a one-off. In `@auto-it/released`, the PR-comment path is wrapped in `try/catch` but **the issue-comment path is not**, so any release containing a `Closes #N` PR fails the same way. A post-release failure is now a `::warning`; a genuinely missing release is still an error.

Granting the app `issues: write` is still worth doing, but it's now a quality-of-life fix rather than a blocker.

## 4. Report the end state to the step summary

Both v8.17.0 attempts surfaced only a bare red X — one half-published, one fully released. Working out which took reading ~2,000 log lines. The summary now lists each package, the tag (and whether it points at the right commit) and the release, then says plainly whether a re-run would change anything.

Also: `Resolve` now fails if `lerna list` returns no publishable packages, which would otherwise leave `PUBLISH=false` and skip the publish **silently green**.

## Verification

The release path can't be exercised by CI, so this is verified by targeted local testing rather than a live run:

- YAML parses; `publish` has the expected 13 steps; `canary` untouched
- Release-commit resolution returns `f52184de` on main — the correct v8.17.0 commit
- Retry loop tested across four cases: success first try → 0; non-zero exit but registry complete → 0; transient then success → 0; genuinely broken → **1** (failure detection preserved)
- `specs` output format verified as space-separated

Honest caveat: **the real test is the next release.** These paths only execute during a publish. The design intent is that every change either fails closed or is a no-op on the happy path.

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -297,6 +297,22 @@ jobs:
       - name: Prepare repository
         run: git fetch --unshallow --tags
 
+      # On the workflow_dispatch path the checkout is main's tip, not the release
+      # commit, and everything below reads the working tree: the build, lerna.json,
+      # the gitHead lerna stamps into the npm metadata, and the commit range
+      # `auto release` turns into release notes. Pin the tree to the same commit
+      # `Push release tag` tags, so the tarballs, the tag and the notes all describe
+      # one commit. On the push path HEAD is already that commit, so this is a no-op.
+      - name: Check out the release commit
+        run: |
+          RELEASE_SHA="$(git rev-list -1 --grep "^Release v[0-9]" HEAD)"
+          if [ -z "$RELEASE_SHA" ]; then
+            echo "No 'Release v' commit reachable from HEAD" >&2
+            exit 1
+          fi
+          echo "Publishing from ${RELEASE_SHA}"
+          git checkout --detach "$RELEASE_SHA"
+
       - name: Test using Node.js
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
         with:
@@ -321,6 +337,12 @@ jobs:
           PUBLISH=false
           SPECS="$(yarn lerna list --json --no-private --loglevel silent \
             | node -e "let s='';process.stdin.on('data',d=>s+=d).on('end',()=>console.log(JSON.parse(s).map(p=>p.name+'@'+p.version).join('\n')))")"
+          # An empty list would leave PUBLISH=false and skip the publish silently green.
+          if [ -z "$SPECS" ]; then
+            echo "lerna list returned no publishable packages" >&2
+            exit 1
+          fi
+          echo "specs=$(printf '%s' "$SPECS" | tr '\n' ' ')" >> "$GITHUB_OUTPUT"
           for SPEC in $SPECS; do
             if NPM_OUT="$(npm view "$SPEC" version 2>&1)"; then
               echo "${SPEC} is already published."
@@ -383,12 +405,112 @@ jobs:
       - name: Publish to npm
         if: steps.version.outputs.publish == 'true'
         # Publishing authenticates via npm trusted publishing (OIDC), hence
-        # id-token: write. from-package skips anything already on npm, so a
-        # partially completed publish can be re-run safely.
-        run: yarn lerna publish from-package --yes --no-verify-access
+        # id-token: write. from-package recomputes the publish set from the registry
+        # on every call, so repeating it is safe and is exactly what a human does by
+        # hand: provenance signing intermittently aborts partway through a
+        # multi-package publish with a Rekor 409 (TLOG_CREATE_ENTRY_ERROR), which is
+        # how v8.17.0 left one package published and one not. The registry, not
+        # lerna's exit code, decides when we are done.
+        run: |
+          for ATTEMPT in 1 2 3; do
+            if yarn lerna publish from-package --yes --no-verify-access; then
+              exit 0
+            fi
+            echo "lerna publish attempt ${ATTEMPT} exited non-zero"
+            MISSING=""
+            for SPEC in $SPECS; do
+              npm view "$SPEC" version >/dev/null 2>&1 || MISSING="${MISSING} ${SPEC}"
+            done
+            if [ -z "$MISSING" ]; then
+              echo "Every package is on npm; treating the publish as complete."
+              exit 0
+            fi
+            echo "Still missing:${MISSING}"
+            if [ "$ATTEMPT" -eq 3 ]; then break; fi
+            sleep $((ATTEMPT * 30))
+          done
+          echo "npm publish still incomplete after 3 attempts:${MISSING}" >&2
+          exit 1
+        env:
+          SPECS: ${{ steps.version.outputs.specs }}
 
       - name: Create GitHub release
         if: steps.version.outputs.release == 'true'
-        run: yarn auto release
+        # `auto release` creates the release, then the `released` plugin comments on
+        # every PR and on every issue referenced by a "Closes #N". Upstream guards the
+        # PR path with a try/catch but not the issue path, so a failure there fails
+        # this step after the release is already correct — which is what happened on
+        # v8.17.0 (403, the app has pull_requests:write but not issues:write). Assert
+        # on the release itself and keep a post-release failure visible as a warning.
+        run: |
+          AUTO_RC=0
+          yarn auto release || AUTO_RC=$?
+          if ! gh release view "v${VERSION}" >/dev/null 2>&1; then
+            echo "::error::auto release exited ${AUTO_RC} and no GitHub release exists for v${VERSION}"
+            exit 1
+          fi
+          if [ "$AUTO_RC" -ne 0 ]; then
+            echo "::warning::GitHub release v${VERSION} was created, but auto exited ${AUTO_RC} afterwards; a post-release comment or label failed"
+          fi
         env:
           GITHUB_TOKEN: ${{ steps.get-github-app-token.outputs.token }}
+          GH_TOKEN: ${{ steps.get-github-app-token.outputs.token }}
+          VERSION: ${{ steps.version.outputs.version }}
+
+      - name: Report release state
+        # Always runs, because the end state is least obvious exactly when a step
+        # above failed: v8.17.0 attempt 1 was half-published and attempt 2 was fully
+        # released but still red, and both looked like a bare red X.
+        if: always() && steps.version.outputs.version != ''
+        run: |
+          INCOMPLETE=false
+          {
+            echo "## Release v${VERSION}"
+            echo
+            echo "| Artifact | State |"
+            echo "| --- | --- |"
+          } >> "$GITHUB_STEP_SUMMARY"
+          for SPEC in $SPECS; do
+            # Retried: this can run seconds after publishing, and a packument read can
+            # lag. A false MISSING would red a release that actually landed.
+            STATE=""
+            for _ in 1 2 3; do
+              if NPM_OUT="$(npm view "$SPEC" version 2>&1)"; then
+                STATE="published"
+                break
+              fi
+              sleep 5
+            done
+            if [ -z "$STATE" ]; then
+              if printf '%s' "$NPM_OUT" | grep -q 'E404'; then
+                STATE="**MISSING from npm**"
+              else
+                STATE="unknown (npm view failed)"
+              fi
+              INCOMPLETE=true
+            fi
+            echo "| npm \`${SPEC}\` | ${STATE} |" >> "$GITHUB_STEP_SUMMARY"
+          done
+          EXPECTED_SHA="$(git rev-list -1 --grep "^Release v${VERSION}" HEAD || true)"
+          TAG_SHA="$(gh api "repos/${GITHUB_REPOSITORY}/git/ref/tags/v${VERSION}" --jq .object.sha 2>/dev/null)" || TAG_SHA=""
+          if [ -n "$TAG_SHA" ] && [ "$TAG_SHA" = "$EXPECTED_SHA" ]; then
+            echo "| tag \`v${VERSION}\` | \`${TAG_SHA}\` |" >> "$GITHUB_STEP_SUMMARY"
+          else
+            echo "| tag \`v${VERSION}\` | **MISSING or wrong commit** (\`${TAG_SHA:-none}\`) |" >> "$GITHUB_STEP_SUMMARY"
+            INCOMPLETE=true
+          fi
+          if gh release view "v${VERSION}" >/dev/null 2>&1; then
+            echo "| GitHub release | created |" >> "$GITHUB_STEP_SUMMARY"
+          else
+            echo "| GitHub release | **MISSING** |" >> "$GITHUB_STEP_SUMMARY"
+            INCOMPLETE=true
+          fi
+          if [ "$INCOMPLETE" = true ]; then
+            printf '\n**v%s is INCOMPLETE - re-run this job.**\n' "$VERSION" >> "$GITHUB_STEP_SUMMARY"
+            exit 1
+          fi
+          printf '\nv%s is fully released on npm, git and GitHub. Any failure above is post-release; re-running will change nothing.\n' "$VERSION" >> "$GITHUB_STEP_SUMMARY"
+        env:
+          GH_TOKEN: ${{ steps.get-github-app-token.outputs.token }}
+          VERSION: ${{ steps.version.outputs.version }}
+          SPECS: ${{ steps.version.outputs.specs }}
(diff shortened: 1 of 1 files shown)


.github/workflows/node-ci.yml
```diff
@@ -297,6 +297,22 @@ jobs:
...
+      # one commit. On the push path HEAD is already that commit, so this is a no-op.
+      - name: Check out the release commit
+        run: |
+          RELEASE_SHA="$(git rev-list -1 --grep "^Release v[0-9]" HEAD)"
+          if [ -z "$RELEASE_SHA" ]; then
+            echo "No 'Release v' commit reachable from HEAD" >&2
+            exit 1
+          fi
+          echo "Publishing from ${RELEASE_SHA}"
+          git checkout --detach "$RELEASE_SHA"
```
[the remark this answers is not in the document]
mdvictor: I think this makes sense, otherwise this looks good!

.github/workflows/node-ci.yml
```diff
@@ -297,6 +297,22 @@ jobs:
...
+      # `auto release` turns into release notes. Pin the tree to the same commit
+      # `Push release tag` tags, so the tarballs, the tag and the notes all describe
+      # one commit. On the push path HEAD is already that commit, so this is a no-op.
+      - name: Check out the release commit
+        run: |
+          RELEASE_SHA="$(git rev-list -1 --grep "^Release v[0-9]" HEAD)"
+          if [ -z "$RELEASE_SHA" ]; then
+            echo "No 'Release v' commit reachable from HEAD" >&2
+            exit 1
+          fi
```
ivanortegaalba: `--grep` matches any line of a commit message, not just the subject, because git compiles the pattern with `REG_NEWLINE`. So a commit whose *body* contains a line starting with `Release v` matches too, and `rev-list -1` returns the newest match. GitHub's default squash body is the concatenated commit messages of the PR, so this can happen without anyone doing anything unusual.

Reproduced locally:

```console
$ git log --format='%h %s'
b2c3d4e Docs: mention the release     # body: "Release v8.17.0 shipped the thing"
a1b2c3d Release v8.17.0 (#1640)

$ git rev-list -1 --grep "^Release v[0-9]" HEAD
b2c3d4e
```

On the `workflow_dispatch` path that would detach to the wrong tree, and I think the run would still go green: `Push release tag` and `Report release state` each re-resolve `^Release v${VERSION}` from the new HEAD, both find the real release commit as an ancestor, and the summary prints that the tag matches. The push path looks unaffected, since HEAD is already the release commit.

Would asserting on the subject after resolving be enough here?

```suggestion
          RELEASE_SHA="$(git rev-list -1 --grep "^Release v[0-9]" HEAD)"
          if [ -z "$RELEASE_SHA" ]; then
            echo "No 'Release v' commit reachable from HEAD" >&2
            exit 1
          fi
          case "$(git log -1 --format=%s "$RELEASE_SHA")" in
            "Release v"[0-9]*) ;;
            *) echo "Resolved ${RELEASE_SHA} is not a release commit" >&2; exit 1 ;;
          esac
```

.github/workflows/node-ci.yml
```diff
@@ -383,12 +405,112 @@ jobs:
...
-        run: yarn lerna publish from-package --yes --no-verify-access
+        # id-token: write. from-package recomputes the publish set from the registry
+        # on every call, so repeating it is safe and is exactly what a human does by
+        # hand: provenance signing intermittently aborts partway through a
+        # multi-package publish with a Rekor 409 (TLOG_CREATE_ENTRY_ERROR), which is
+        # how v8.17.0 left one package published and one not. The registry, not
+        # lerna's exit code, decides when we are done.
+        run: |
+          for ATTEMPT in 1 2 3; do
+            if yarn lerna publish from-package --yes --no-verify-access; then
```
ivanortegaalba: The step comment above says the registry, not lerna's exit code, decides when we are done, but a zero exit returns here without probing the registry. The probe only runs on the failure path.

Low impact in practice since `Report release state` catches it, but would it be worth either probing on success too, or relaxing the comment?

.github/workflows/node-ci.yml
```diff
@@ -383,12 +405,112 @@ jobs:
...
+              exit 0
+            fi
+            echo "lerna publish attempt ${ATTEMPT} exited non-zero"
+            MISSING=""
+            for SPEC in $SPECS; do
+              npm view "$SPEC" version >/dev/null 2>&1 || MISSING="${MISSING} ${SPEC}"
+            done
```
ivanortegaalba: This probe runs seconds after a failed publish with no delay and no retry, while the equivalent probe in `Report release state` retries three times because a packument read that soon can lag.

The blast radius here is only one extra `lerna publish` (which `from-package` self-corrects), so this is minor, but should the two probes treat lag the same way?

.github/workflows/node-ci.yml
```diff
@@ -383,12 +405,112 @@ jobs:
...
         env:
           GITHUB_TOKEN: ${{ steps.get-github-app-token.outputs.token }}
+          GH_TOKEN: ${{ steps.get-github-app-token.outputs.token }}
+          VERSION: ${{ steps.version.outputs.version }}
+
+      - name: Report release state
+        # Always runs, because the end state is least obvious exactly when a step
+        # above failed: v8.17.0 attempt 1 was half-published and attempt 2 was fully
+        # released but still red, and both looked like a bare red X.
+        if: always() && steps.version.outputs.version != ''
```
ivanortegaalba: `always()` also runs on cancellation, so a cancelled run would write a summary and exit 1 for a release nobody wanted judged. Was `!cancelled()` the intent?

```suggestion
        if: ${{ !cancelled() && steps.version.outputs.version != '' }}
```

.github/workflows/node-ci.yml
```diff
@@ -383,12 +405,112 @@ jobs:
...
+              sleep 5
+            done
+            if [ -z "$STATE" ]; then
+              if printf '%s' "$NPM_OUT" | grep -q 'E404'; then
+                STATE="**MISSING from npm**"
+              else
+                STATE="unknown (npm view failed)"
+              fi
+              INCOMPLETE=true
+            fi
```
ivanortegaalba: A transient `npm view` failure that is not an `E404` sets `INCOMPLETE`, which reds the job and prints "re-run this job" for a release that may be entirely complete. That looks like the same false red this step was written to remove.

Would it make sense to keep the hard failure for a confirmed `E404` and warn on the unknown case?

```suggestion
            if [ -z "$STATE" ]; then
              if printf '%s' "$NPM_OUT" | grep -q 'E404'; then
                STATE="**MISSING from npm**"
                INCOMPLETE=true
              else
                STATE="unknown (npm view failed)"
                echo "::warning::npm view failed for ${SPEC}, could not confirm it is published"
              fi
            fi
```

.github/workflows/node-ci.yml
```diff
@@ -383,12 +405,112 @@ jobs:
...
+            echo "| npm \`${SPEC}\` | ${STATE} |" >> "$GITHUB_STEP_SUMMARY"
+          done
+          EXPECTED_SHA="$(git rev-list -1 --grep "^Release v${VERSION}" HEAD || true)"
```
ivanortegaalba: The release commit is now resolved three separate times: here, in `Check out the release commit`, and in `Push release tag`. Two of them use `^Release v${VERSION}` and one uses `^Release v[0-9]`, so they can disagree, and when they do this summary would report a tag match against a different commit than the one that was actually built.

Would it make sense to resolve it once in `Check out the release commit`, write it to `$GITHUB_OUTPUT`, and have this step and `Push release tag` read that output? That would also make the second point here moot: `${VERSION}` goes into a regex unescaped, so `8.17.0` also matches `8x17y0`.

.github/workflows/node-ci.yml
```diff
@@ -383,12 +405,112 @@ jobs:
...
+            INCOMPLETE=true
+          fi
+          if gh release view "v${VERSION}" >/dev/null 2>&1; then
+            echo "| GitHub release | created |" >> "$GITHUB_STEP_SUMMARY"
+          else
+            echo "| GitHub release | **MISSING** |" >> "$GITHUB_STEP_SUMMARY"
+            INCOMPLETE=true
+          fi
+          if [ "$INCOMPLETE" = true ]; then
+            printf '\n**v%s is INCOMPLETE - re-run this job.**\n' "$VERSION" >> "$GITHUB_STEP_SUMMARY"
```
ivanortegaalba: One state reaches this line where a re-run cannot help: npm and the GitHub release both complete, only the tag missing. `Push release tag` is gated on `publish == 'true' || release == 'true'`, so on the re-run there is nothing left to publish or release and the tag step is skipped again.

The gate is pre-existing and the state needs a manually deleted tag, so this is narrow. Worth a follow-up rather than a change here?

.github/workflows/node-ci.yml
ivanortegaalba: `--grep` matches any line of a commit message, not just the subject, so a commit whose *body* contains a line starting with `Release v` matches too

Asserting on the subject after resolving should be enough

```suggestion
          RELEASE_SHA="$(git rev-list -1 --grep "^Release v[0-9]" HEAD)"
          if [ -z "$RELEASE_SHA" ]; then
            echo "No 'Release v' commit reachable from HEAD" >&2
            exit 1
          fi
          case "$(git log -1 --format=%s "$RELEASE_SHA")" in
            "Release v"[0-9]*) ;;
            *) echo "Resolved ${RELEASE_SHA} is not a release commit" >&2; exit 1 ;;
          esac
```

.github/workflows/node-ci.yml
ivanortegaalba: `always()` also runs on cancellation, so a cancelled run would write a summary and exit 1 

```suggestion
        if: ${{ !cancelled() && steps.version.outputs.version != '' }}
```

.github/workflows/node-ci.yml
ivanortegaalba, line 509: `npm view` failure that is not an `E404` sets `INCOMPLETE`, which reds the job and prints "re-run this job" for a release that may be entirely complete. That looks like the same false red this step was written to remove.

Would it make sense to keep the hard failure for a confirmed `E404` and warn on the unknown case?

```suggestion
            if [ -z "$STATE" ]; then
              if printf '%s' "$NPM_OUT" | grep -q 'E404'; then
                STATE="**MISSING from npm**"
                INCOMPLETE=true
              else
                STATE="unknown (npm view failed)"
                echo "::warning::npm view failed for ${SPEC}, could not confirm it is published"
              fi
            fi
```

.github/workflows/node-ci.yml
ivanortegaalba: The release commit is resolved three times, independently: here, in `Check out the release commit`, and in `Push release tag`. This passes even when that value is wrong

Would it make sense to resolve it once in `Check out the release commit`, write it to `$GITHUB_OUTPUT`, and have this step and `Push release tag` read that output? A wrong resolution would then stay a single point of failure instead of being confirmed twice

Also `${VERSION}` goes into a regex unescaped, so `8.17.0` also matches `8x17y0`

.github/workflows/node-ci.yml
ivanortegaalba, line 526: If only the tag is missing here, a re-run cannot help

`Push release tag` is gated on `publish == 'true' || release == 'true'`, so on the re-run there is nothing left to publish or release and the tag step is skipped again.

Wondering if we have any way to solve it 🤔

ivanortegaalba (dismissed):
I tested with the agent, and we found some small things. Can you have a look?

.github/workflows/node-ci.yml
Sergej-Vlasov: Fixed, and taken a step further than the suggestion.

Reproduced it first in a scratch repo: a commit whose *body* contained `Release v8.17.0 was wrong` resolved AHEAD of the real release commit, so `--grep` returned entirely the wrong sha. Confirmed the subject assertion catches it.

Rather than assert-and-fail, candidates are now filtered on the subject:

    git log --grep "^Release v[0-9]" --format='%H%x09%s' HEAD | awk -F'\t' '$2 ~ /^Release v[0-9]/ {print $1; exit}'

Same fail-closed guarantee, but it *skips* the decoy instead of blocking a legitimate release. Verified by extracting the rendered step out of the YAML and running it against that scratch repo — it picks the real commit.

Worth flagging: this bug is already live on main, since the merged `Push release tag` greps the same way. This PR fixes it there too.

.github/workflows/node-ci.yml
Sergej-Vlasov: Both fixed. The sha is now resolved once in `Check out the release commit`, written to `$GITHUB_OUTPUT`, and read by `Push release tag`, `Create local tag` and the report. `rev-list --grep` appears zero times now.

That also removes the unescaped `${VERSION}` regex by deletion rather than by escaping — you were right that `8.17.0` matched `8x17y0`, and now there is no second pattern to get wrong.

.github/workflows/node-ci.yml
Sergej-Vlasov: Good catch — that was genuinely unrecoverable. `Resolve` now computes a third gate:

    TAG_SHA="$(gh api .../git/ref/tags/v${VERSION} --jq .object.sha)" || TAG_SHA=""
    [ "$TAG_SHA" = "$RELEASE_SHA" ] && echo tag=false || echo tag=true

and both tag steps are gated on `publish || release || tag`. So a fully published release whose tag is missing or on the wrong commit now re-tags on a re-run instead of skipping forever.

.github/workflows/node-ci.yml
Sergej-Vlasov: Fixed as suggested. Only a confirmed E404 sets `INCOMPLETE`; an inconclusive probe emits `::warning::` and leaves the job's own result alone.

You identified the actual flaw in it — reddening a complete release because npm was briefly unreachable is exactly the false red this step was written to eliminate.

.github/workflows/node-ci.yml
Sergej-Vlasov: Fixed, now `!cancelled() && steps.version.outputs.version != ''`. A cancelled run would have written a summary claiming the release was incomplete and then exited 1.

.github/workflows/node-ci.yml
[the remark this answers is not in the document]
Sergej-Vlasov: Fixed in adb54d5a. Confirmed the mechanism in a scratch repo — a commit whose body quoted a `Release v...` line resolved ahead of the real release commit, so the two steps could disagree.

Rather than aligning the two patterns, the commit is now resolved once, filtered on the subject, and shared via a step output; both tag steps and the report read it. The divergence is gone because there is only one resolution left.

Also live on main via the merged `Push release tag`, so this fixes it there too.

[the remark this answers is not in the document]
Sergej-Vlasov: Agreed and fixed — see the thread above. Reproduced it in a scratch repo: a commit whose body quoted a `Release v...` line resolved ahead of the real release commit. The sha is now resolved once, filtered on the subject, and shared via a step output.

mdvictor (approved):

PR #1650, opened 2026-09-17
PoC: runtime transformation layer provider

gtk-grafana:
## Summary

This change adds runtime transformation layers to `SceneDataTransformer`. A runtime layer is a temporary transformation owner. Independent owners can add, update, and remove transformations without changing saved scene state. Existing `SystemTransformationsProvider` behavior remains compatible. This PR does not change renderers, serializers, editors, or panel state.

## Problem

Grafana integrations currently wrap `VizPanel.getSystemTransformations` to add temporary transformations. Each owner must read, combine, and replace the complete provider result. One owner can overwrite changes from another owner. Consumers also create custom operator identities and call `reprocessTransformations()` directly.

This API gives each owner an independent registration. Each owner updates or removes only its own layer.

## Execution order

The transformation pipeline uses this fixed order:

1. Run provider prepend transformations.
2. Run registered `beforeUser` layers in registration order.
3. Run saved transformations from `state.transformations`.
4. Run provider append transformations.
5. Run registered `afterUser` layers in registration order.
6. Run registered `final` layers in registration order.

Registration time cannot move a `final` layer before provider append transformations.

## Consumer guide

Use the `SceneDataTransformer` that supplies data to the panel. Store the temporary transformations in the object that owns their lifecycle.

Register the layer once:

```ts
import type {
  RuntimeTransformation,
  RuntimeTransformationRegistration,
} from '@grafana/scenes';
import { SceneDataTransformer } from '@grafana/scenes';

class AdHocTransformationOwner {
  private transformations: RuntimeTransformation[] = [];
  private readonly registration: RuntimeTransformationRegistration;

  public constructor(transformer: SceneDataTransformer) {
    this.registration = transformer.registerRuntimeTransformationLayer({
      id: 'grafana-table-ad-hoc',
      phase: 'final',
      getTransformations: ({ series }) => {
        return this.transformations;
      },
    });
  }

  public update(transformations: RuntimeTransformation[]) {
    this.transformations = transformations;
    this.registration.changed();
  }

  public dispose() {
    this.registration.dispose();
  }
}
```

Use a stable and unique `id` for each active owner. A duplicate active ID causes an error. The ID also becomes the `origin` value in the resolved snapshot.

Choose the phase that matches the required behavior:

1. Use `beforeUser` when the layer must run before saved user transformations.
2. Use `afterUser` when the layer must run after saved transformations and provider append transformations.
3. Use `final` when no later runtime layer phase can change the result.

`getTransformations` receives the source series. Source series are query results before any transformation runs. Return `DataTransformerConfig` values, custom transformer definitions, or both.

Do not rely on scene variable interpolation in a runtime layer. Scenes interpolates only saved user transformations. If a runtime layer uses variables, resolve the variables before returning its transformations.

Replace the stored transformations before you call `changed()`. The call clears only that owner memo. Scenes then runs one transformation pass from the existing source frames. It does not run a new data source query.

If the transformer is inactive, `changed()` records the update. Scenes applies the update once when the transformer becomes active.

Call `dispose()` when the owner lifecycle ends. The call removes only that layer and preserves all other layers. The call is idempotent, which means that it is safe to call twice.

## Introspection

Use the following method to inspect the layers that produced the current output:

```ts
const layers = transformer.getResolvedRuntimeTransformationLayers();
```

The method returns a frozen snapshot. Repeated reads return the same snapshot until another pass completes. A read does not call providers or runtime suppliers.

Each resolved layer contains:

- The layer `id`.
- The execution `phase`.
- The layer `origin`.
- The frozen transformation array from the last completed pass.

Registered layers use their ID as the origin. Provider layers keep the origin from `SystemTransformationsProvider`.

## Error behavior

If one supplier throws an error, Scenes logs the error with the layer ID. Scenes treats only that layer as empty. Saved transformations and other runtime layers continue to run.

## Persistence and cloning

Runtime layers stay outside `state.transformations`. They do not enter scene serialization or persisted dashboard state. They do not survive scene reconstruction or a browser reload.

Clones do not copy runtime registrations. When a clone becomes active, it removes output that came from runtime layers on the original transformer.

## Current limits

Stage input is the data entering one layer. This API does not expose stage input. It also does not retain values that a layer removes.

The table visualization must keep its current hidden-field restoration path. A later change can add input access and selective value retention.

## Migration guide

Consumers can replace the old integration with these steps:

1. Remove the wrapper around `VizPanel.getSystemTransformations`.
2. Remove custom identities that exist only to manage temporary stages.
3. Register one runtime layer for each independent owner.
4. Update the owner value and call its `changed()` method.
5. Dispose the registration when the owner lifecycle ends.
6. Keep existing restoration logic that needs pre-layer field values.

Consumers no longer need whole-provider read and write updates. They also do not need to call `reprocessTransformations()` directly.

## Test coverage

The tests cover:

- Exact phase order.
- Stable registration order.
- Independent owner updates.
- Duplicate ID rejection.
- Active and inactive changes.
- Idempotent disposal.
- Data source query avoidance.
- Supplier error isolation.
- Clone isolation.
- Frozen introspection snapshots.
- One supplier resolution for each required pass.
- Existing provider behavior.


gtk-grafana:
we can do better

PR #1651, opened 2026-09-17
VizPanel: Runtime transformations API

gtk-grafana:
## What is this change?

This PR adds a panel-owned runtime transformation controller to `VizPanel`.

A runtime transformation changes current panel data without changing saved scene state, so Grafana can use it for temporary viewer state such as hidden or reordered table columns.

## API

```ts
const runtimeTransformations = panel.getRuntimeTransformations();

runtimeTransformations.set('table:column-state', transformations);
runtimeTransformations.get('table:column-state');
runtimeTransformations.getSourceSeries('table:column-state');

const unsubscribe = runtimeTransformations.subscribe('table:column-state', listener);

runtimeTransformations.set('table:column-state', []);
unsubscribe();
```

Each owner has independent transformations and listeners, runs in registration order, and keeps its position when updated.

## Behavior

- Runtime owners run after plugin and saved transformations
- Empty transformation lists remove the owner
- Removed owners move to the end when added again
- Active changes reprocess current data without a new data source request
- Inactive changes run when the transformer becomes active
- Source capture includes the output from earlier runtime owners
- Plugin changes clear all runtime owners
- Runtime transformations stay out of scene state and serialization

## Safety

A cloned panel gets an empty controller and does not inherit runtime output, while deep-cloned and frozen snapshots prevent changes outside `set()`.

## Tests

Tests cover ordering, updates, removal, source capture, listeners, inactive changes, data replacement, plugin changes, cloning, snapshot immutability, cleanup restoration, and query reuse.

<details>
  <summary>📦 Published PR as canary version: <code>8.19.0--canary.1651.35915442795.0</code></summary>
  <br />

  ```bash
  npm install @grafana/scenes@8.19.0--canary.1651.35915442795.0
  npm install scenes-app@8.19.0--canary.1651.35915442795.0
  npm install @grafana/scenes-react@8.19.0--canary.1651.35915442795.0
  # or 
  yarn add @grafana/scenes@8.19.0--canary.1651.35915442795.0
  yarn add scenes-app@8.19.0--canary.1651.35915442795.0
  yarn add @grafana/scenes-react@8.19.0--canary.1651.35915442795.0
  ```
</details>


packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -778,7 +792,12 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}>
 
   public clone(withState?: Partial<VizPanelState>) {
     // Clear _pluginInstanceState and _pluginLoadError as it's not safe to clone
-    return super.clone({ _pluginInstanceState: undefined, _pluginLoadError: undefined, ...withState });
+    return super.clone({
+      _pluginInstanceState: undefined,
+      _pluginLoadError: undefined,
+      _UNSAFE_clearPreviousFieldValues: this._runtimeTransformations.getClearPreviousFieldValuesForClone(),
```
gtk-grafana, line 802: 🤔

fastfrwrd: YEAH. yeah.

gtk-grafana: 🤖 🤮 

When _UNSAFE_clearPreviousFieldValues is enabled, VizPanelRenderer clears value arrays from old data frames to reduce memory use. Runtime transformations retain references to intermediate source frames, so they temporarily disable this cleanup to preserve values needed for later reprocessing or restoring hidden fields.

A normal clone would copy the temporary value false, but runtime owners are not cloned. The clone would therefore disable cleanup permanently for no reason. getClearPreviousFieldValuesForClone() restores the value that existed before runtime transformations changed it.

CC @leeoniya

gtk-grafana: So my understanding is that this is to support Leon's memory optimization to keep a second copy of the frames from getting stored in memory, something we should explicitly test here

fastfrwrd: https://github.com/grafana/scenes/pull/1655 solves it, at least as far as I have been able to test against grafana. gotta do more testing on this before we move forward but I believe this will work. I think originally we needed this because we _weren't_ hooking into scenes more directly to do this, so all the more reason to not do the Grafana-side version 😓

docusaurus/docs/transformations.md
```diff
@@ -175,6 +175,48 @@ The resulting table will look similar to the one that follows:
...
+const runtimeTransformations = panel.getRuntimeTransformations();
+
+runtimeTransformations.set('table:column-filter', [
+  {
+    id: 'filterFieldsByName',
+    options: {
+      include: {
+        names: ['handler', 'Value'],
+      },
+    },
```
dprokop, line 194: Nit, not sure how much it matters to you at this stage, but what happens if topic is set here? should it be respected? Seems like this test proves it's not

```ts
it.each([false, true])('routes annotation transformations through runtime: %s', (runtime) => {
    const { panel, source, transformer } = setup();
    source.setState({
      data: {
        ...source.state.data,
        annotations: [
          toDataFrame({
            meta: { dataTopic: DataTopic.Annotations },
            fields: [{ name: 'value', type: FieldType.number, values: [10] }],
          }),
        ],
      },
    });
    const transformations = [
      { id: 'runtimeMath', topic: DataTopic.Annotations, options: { operation: 'add', value: 1 } },
    ];

    if (runtime) {
      panel.getRuntimeTransformations().set('annotations', transformations);
    } else {
      transformer.setState({ transformations });
      transformer.reprocessTransformations();
    }

    expect({
      series: transformer.state.data?.series[0].fields[0].values,
      annotations: transformer.state.data?.annotations?.[0].fields[0].values,
    }).toEqual({ series: [1], annotations: [11] });
  });
```

docusaurus/docs/transformations.md
```diff
@@ -175,6 +175,48 @@ The resulting table will look similar to the one that follows:
...
+
+A runtime transformation changes a panel's current output without adding configuration to scene state. Use it for temporary visualization behavior that must run after the panel's saved transformations.
+
+Get the panel-owned controller and set one owner's transformations:
+
+```ts
+const runtimeTransformations = panel.getRuntimeTransformations();
+
+runtimeTransformations.set('table:column-filter', [
+  {
```
dprokop, line 188: can you add real life example, or a scenes demo for this new api?

packages/scenes/src/components/VizPanel/VizPanelRuntimeTransformations.ts
```diff
@@ -0,0 +1,250 @@
...
+      }
+    } else {
+      const snapshot = freezeDeep(cloneDeep(transformations));
+      const group = this._groups.get(owner);
+      if (group) {
+        group.transformations = snapshot;
+        group.operator = this._createOperator(group, snapshot);
+      } else {
+        this._groups.set(owner, this._createGroup(snapshot));
+      }
```
dprokop: I think we can achieve this without branching and only through _createGroup code path. just set the group  regardless if it already existed or now. I'm not convinced the complexity of replacing is needed given that map.set should preserve inserion order (if that's the concern)

dprokop (commented):
LGTM, left some minor things to consider

docusaurus/docs/transformations.md
gtk-grafana: Good catch, it was treated as series only, updated to match saved transformations to use one operator per topic. Also added an annotations example to the (new) demo.
Fixed in [`4f9df0f`](https://github.com/grafana/scenes/pull/1651/commits/4f9df0f88b1e32f2bc2357e76b997c06abb6e77f)

docusaurus/docs/transformations.md
gtk-grafana: Yeah we can!

image
image
image

Added in [`4f9df0f`](https://github.com/grafana/scenes/pull/1651/commits/4f9df0f88b1e32f2bc2357e76b997c06abb6e77f)

packages/scenes/src/components/VizPanel/VizPanelRuntimeTransformations.ts
gtk-grafana: Another good catch, simplified to a single branch in [`4f9df0f`](https://github.com/grafana/scenes/pull/1651/commits/4f9df0f88b1e32f2bc2357e76b997c06abb6e77f)

docusaurus/docs/transformations.md
dprokop: thanks, this is great

packages/scenes/src/components/VizPanel/VizPanel.tsx
gtk-grafana: Yeah let's follow up with that clean-up and refactor after this is merged in, it doesn't look like there's any regression introduced here

packages/scenes/src/components/VizPanel/VizPanel.tsx
```diff
@@ -777,8 +790,19 @@ export class VizPanel<TOptions = {}, TFieldConfig extends {} = {}>
   };
 
   public clone(withState?: Partial<VizPanelState>) {
+    const dataWithoutRuntimeOutput =
+      this._runtimeTransformations.getOperators().length > 0 && this.state.$data instanceof SceneDataTransformer
+        ? { $data: this.state.$data.clone({ data: undefined }) }
+        : {};
```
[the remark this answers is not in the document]
fastfrwrd: Commit 179be3c9 is intended to fix this issue. SceneDataTransformer.clone() now avoids inheriting the previous-pass cache when the clone has no output data, so activation reprocesses the shared source. Added a regression test covering a saved transformation, a runtime owner added while inactive, and a clone using the same parent data; it fails before the fix and passes afterward.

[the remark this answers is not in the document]
gtk-grafana: Aight, Paul fixed it, I was gonna argue that this use-case is a tad too esoteric (https://github.com/grafana/scenes/pull/1659) and document it for potential follow-up if someone can explain how a panel can add a transformation to itself before it was deactivated, and then get cloned. 

Maybe possible in a scene app, but I guess it's a one-line fix so 🤷

dprokop (approved):

PR #1655, opened 2026-09-22
Preserve runtime transformation source values during panel cleanup

fastfrwrd:
Runtime transformations should not need to disable `_UNSAFE_clearPreviousFieldValues`. Hiding a column can remove its values from the rendered output while the query source or a runtime stage still needs those same arrays. Clearing them at that point corrupts subsequent reprocessing or unhiding.

This PR is stacked on #1651, targeting `gtk-grafana/dataviz/runtime-transforms-minimal`. It tries the alternative discussed in https://github.com/grafana/scenes/pull/1651#discussion_r4043840825.

The renderer now preserves arrays referenced by current upstream transformer sources and runtime source captures. Previously rendered arrays remain cleanup candidates while protected, then are emptied when those references disappear, for example after a query refresh. Series and annotations are checked together so sharing across topics does not truncate live values. Upstream arrays that were never rendered are not adopted as cleanup candidates; legacy streaming buffers remain excluded.

This removes the controller's cleanup-flag mutation/restoration, previous-setting bookkeeping, and `getClearPreviousFieldValuesForClone()`. Clones still discard runtime output and get an empty controller. The owner-based runtime API is unchanged.

<details>
  <summary>way-too-long explanation of how this was tested</summary>

Testing:

- On the submitted cleanup commit: 214 Scenes tests and 46 snapshots passed across the VizPanel and SceneDataTransformer suites; one existing test remained skipped. Coverage includes hide/unhide without refetching, filter/reset, deferred cleanup after refresh, intermediate runtime captures, nested/inherited sources, shared series/annotation arrays, streaming buffers, cleanup opt-out, clone settings, and a full VizPanel renderer regression.
- Built the Scenes CJS, ESM, and declaration outputs with Rollup and installed that exact local distribution into the Grafana table experiment. Verified the installed bundle matched the build byte-for-byte.
- For Grafana compatibility, the local testing branch also cherry-picked upstream #1645, which supplies the undo/redo events already present in the published canary. That separate upstream change is deliberately excluded from this PR. On that combined testing branch, 233 Scenes tests and 46 snapshots passed, including SceneGridLayout coverage; one existing test remained skipped.
- Grafana exposes the Scenes controller directly through `PanelContext.adHocTransformations`, typed as `VizPanelRuntimeTransformations`. Table consumers select `grafana:table-view`; no adapter or separate API interface is required. These Grafana consumer changes are not part of this Scenes PR.
- Grafana integration verification: 99 tests passed across six affected table/context suites with the final direct wiring, including an owner-isolation/subscription regression. Grafana `tsc --noEmit`, changed-file ESLint, and formatting checks passed.
- Grafana commit `5f3f361c13d` on `dataviz/table-adhoc-transformations` checks in the tested distribution as a Yarn patch over stable Scenes 8.18.0, with stable scenes-react 8.18.0. It no longer needs a canary publication or local file dependency. The installed distribution matches the tested build byte-for-byte, and `yarn install --immutable` passes.
  - in plain non-slop: **if you check out `dataviz/table-adhoc-transformations` you can test this integrated with grafana on the Table panel.** 
- Final Playwright run against the rebuilt local Grafana instance: 12 checks passed (authentication plus 11 table scenarios). Covered column hide/unhide, filters on every column, refresh retention, reload reset, frame isolation, saved transformations, sidebar initialization/accessibility, numeric range filtering through refresh/hiding/reset, keyboard range editing, and absolute date bounds.
- Two existing browser specs were excluded after confirming stale assertions: the override picker expects an accessible name without its description; the pinning spec expects pinning to be absent even though the Grafana branch implements it. They were not reported as passing.

The memory behavior is tested by checking preserved arrays retain their values and obsolete rendered arrays become empty. No heap-size benchmark was performed. This remains an experiment within the existing unsafe cleanup mechanism; it does not establish global ownership of arrays shared by unrelated consumers.
</details>


packages/scenes/src/components/VizPanel/VizPanelRuntimeTypes.ts
```diff
@@ -26,7 +26,6 @@ export interface RuntimeTransformationGroup {
 interface RuntimeTransformationsPanelState {
   pluginId: string;
   $data?: SceneDataProvider;
-  _UNSAFE_clearPreviousFieldValues?: boolean;
```
fastfrwrd, line 29: this state remains on [VizPanelState](https://github.com/grafana/scenes/blob/d45bbac1ca7ded48806503d99ae543fd60c07020/packages/scenes/src/components/VizPanel/VizPanel.tsx#L104-L114) itself but is no longer on RuntimeTransformationsPanelState because of this change (yay)

PR #1659, opened 2026-09-23
fix(runtime-transforms): Clone can skip transform and stay empty

gtk-grafana:
See https://github.com/grafana/scenes/pull/1651#discussion_r4086515325.

So, this seems like a non-problem but I wanted to document the possible bug and fix without blocking https://github.com/grafana/scenes/pull/1651.

Bug:
If a runtime transform is added to an inactive panel that removes its SceneDataTransformer, the panel clone breaks.
But ad-hoc transform's primary use-case is to add transformations within the panel UI, so it seems unlikely that this would ever happen.

See the "Runtime transformations clone bug" demo added in this branch for an example repro


gtk-grafana:
@fastfrwrd fixed with https://github.com/grafana/scenes/commit/179be3c91d40567beb74a97c674e3754a6127928

